PR #34, opened 2024-05-14
SRM fixes

kassibuss:
Fire rate and damage should be the same as base game


AdamTadeusz:
Would it not be better to balance all weapons in one go?

kassibuss:
> Would it not be better to balance all weapons in one go?

Probably, i just did this fix so I would not forget, probably better to compare all of the weapons

nullsystem:
This PR will be replaced by a more general fire-rate/dmg parity PR so closing it.

Issue #146, opened 2024-05-14
Spectators able to see player outlines


blaberry:
**Background**
When spectating the game, players currently have a team symbol centered on their torso to let people see where they are.

**Issue**
The team symbol provides very little detailed information about the living players, such as which animation state they're in or which direction they are looking in.

**Solution**
Draw an outline around players, allowing spectators and streamers to easily see which direction players are facing and which animation state they are in. This would let spectators and commentators during competetive games more easily ascertain what is going on.

kassibuss:
fixed by https://github.com/NeotokyoRebuild/neo/pull/187 or?

PR #187, opened 2024-05-15
Ghost and Frindly Markers with Networking

VantoNortim:
Network friendlies and people in ghost range if holding the ghost
Friendly markers now based on client's position
Added base class for fading UI elements
Moved ghost beacons to UI from weapon ghost

#146 #153


VantoNortim:
Tested this but there were lots of conflicts please check before merging

mp/src/game/server/neo/neo_player.cpp
```diff
@@ -2505,6 +2505,29 @@ float CNEO_Player::GetSprintSpeed(void) const
 	}
 }
 
+int CNEO_Player::ShouldTransmit(const CCheckTransmitInfo* pInfo)
+{
+	auto player = Instance(pInfo->m_pClientEnt);
+	
+	if(player)
+	{
+		auto neoPlayer = dynamic_cast<CNEO_Player*>(player);
```
nullsystem, line 2514: Could `neoPlayer` be NULL here? If `neoPlayer` is expected in this scope, assign and check `neoPlayer` as the scope's if condition, otherwise check `neoPlayer` when getting ghost.

AdamTadeusz (commented):
See no lowercase overrides, and no changes to code I personally wrote, and the code compiles so that's an approval from me. But do check with @nullsystem

mp/src/game/client/neo/ui/neo_hud_elements.cpp
```diff
@@ -238,6 +245,13 @@ void CNeoHudElements::InitHud()
 	InitHTA();
 	InitRoundState();
 	InitTargetID();
+	InitGhostBeacons();
+}
+
+void CNeoHudElements::InitGhostBeacons()
+{
+	Assert(!m_pGhostBeacons);
+	m_pGhostBeacons = new CNEOHud_GhostBeacons(UI_ELEMENT_NAME_GHOST_BEACONS, this);
```
nullsystem, line 254: Memory leak: This needs to be followed up with a `delete` on deconstructor.

mp/src/game/client/neo/ui/neo_hud_friendly_marker.cpp
```diff
@@ -66,87 +69,103 @@ CNEOHud_FriendlyMarker::CNEOHud_FriendlyMarker(const char* pElemName, vgui::Pane
...
+			memberCount = team->GetNumPlayers();
+			for (int i = 0; i < memberCount; ++i)
+			{
+				auto player = team->GetPlayer(i);
+				if(player && player->IsAlive())
+				{
+					DrawPlayer(GetTeamColour(team->GetTeamNumber()), player);
+				}		
+			}
 		}
```
nullsystem, line 102: Just loop over the clients and check the team per player, offsetting by spec is a bit odd.
Either that or put an assert to make sure team given are Jinrai or NSF.

mp/src/game/client/neo/ui/neo_hud_friendly_marker.cpp
```diff
@@ -66,87 +69,103 @@ CNEOHud_FriendlyMarker::CNEOHud_FriendlyMarker(const char* pElemName, vgui::Pane
...
-	surface()->DrawSetTextColor(teamColor);
-
-	surface()->DrawSetColor(teamColor);
-	surface()->DrawSetTexture(m_hTex);
-
-	auto localPlayer = C_NEO_Player::GetLocalNEOPlayer();
-
-	for (int i = 0; i < MAX_PLAYERS; ++i)
+	int x, y;
+	constexpr float heightOffset = 48.0f;
```
nullsystem, line 119: Until we've moved off the very old steam-runtime on Linux, I'm not sure if constexpr will compile on GCC 4.6. If not, change this to `static const float`. Also applies to: `maxNameLenght` constexpr usage (and a typo)

nullsystem (changes requested):

VantoNortim:
> Some more problems with this, there is no wind up time when taking the ghost into your hands and enemy markers appearing, and there are no beeps when new enemies come into view.
> 
> Also when you take the ghost into your hands sometimes the program breaks in neo_hud_ghost_beacons around line 105, might also need a check if the C_NEO_Player cast is successful. image

I will open a new issue for this. As @nullsystem has already added the delay so I can use that. Rest of the comments are resolved!

AdamTadeusz (approved):

PR #232, opened 2024-05-17
Configurable FOV/Field of view, new convar: neo_fov

nullsystem:
* Most importantly, weapon switching shouldn't use pre-set fov default, because somehow convar isn't updated (client variable) even though it's updated (server and console recognizes)?????
* New convar: `neo_fov`, default 90, range 60-110

### DIFF
--- a/mp/src/game/shared/shareddefs.h
+++ b/mp/src/game/shared/shareddefs.h
@@ -262,7 +262,12 @@ enum CastVote
 
 #define MAX_PLACE_NAME_LENGTH		18
 
+#ifdef NEO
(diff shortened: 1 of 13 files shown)


mp/src/game/shared/shareddefs.h
```diff
@@ -262,7 +262,12 @@ enum CastVote
 
 #define MAX_PLACE_NAME_LENGTH		18
 
+#ifdef NEO
+#define MAX_FOV						110
```
AdamTadeusz, line 266: 120 bros...

PR #252, opened 2024-05-20
Handle gibbed player corpse once, but not necessarily on first tick

AdamTadeusz:


AdamTadeusz:
Also fixes the amount of cloak when switching from recon to assault

nullsystem (approved):

PR #253, opened 2024-05-20
Allow the tachi to switch fire mode

nullsystem:
* Should also prevent grenade from throwing as it uses the same button


AdamTadeusz:
Should also prevent grenade from throwing? Could you elaborate?

nullsystem:
@AdamTadeusz The grenade also uses the same button as tachi's alt-fire toggle to throw the grenade at a different way.

AdamTadeusz (approved):

PR #255, opened 2024-05-20
Supa7 - Fix animation usage between slug and primary

nullsystem:
* Yes, the slug uses primary animation while the main one uses secondary animation


AdamTadeusz:
Primary attack is the slug attack?

AdamTadeusz (approved):

PR #261, opened 2024-05-23
Clear decals on round start

AdamTadeusz:
Execute client command for clients to clear all decals, is there a better way to do this?


AdamTadeusz:
Also tried calling clearAllDecalsFromAllModels from extern modelRender, didn't do anything though.

nullsystem:
@AdamTadeusz So the decal type for EX: blood and bullets is a mod_brush, yet RemoveAllDecals only deals with mod_studio.
The closest I've found relating to this is: https://github.com/Facepunch/garrysmod-issues/issues/3838#issuecomment-489805888 however there doesn't really seem to be related source code within here/the given SDK that allows for this.
This solution is probably the best it could be done cause they exposed easily to create but not delete those decal traces.

nullsystem (approved):

PR #262, opened 2024-05-23
Merge into a single README.md, update CONTRIBUTING.md, new cmake + qtcreator instructions

nullsystem:
* Replace some links with our ones
* Add in additional cmake instructions
* Merge some markdowns into single README.md
* Add credits
* Add qtcreator (linux) instructions


Rainyan (approved):
LGTM, thanks for adding more detailed instructions!

I tested with the following, and both compiled without too much trouble:
* Visual Studio 2022 and CMake for Windows 10 (no Ninja)
* QT Creator 6* and CMake for Linux Mint

*Linux Mint default package repos at current time provides QT Creator 6 instead of 13, but it seemed to work fine as well:
```bash
sudo apt-get update
sudo apt-get install cmake qtcreator gcc-multilib g++-multilib
```

Also for `LD_LIBRARY_PATH`, I had to use `find / -type d -name 'steam-runtime' 2>&1 /dev/null` to find the script, which ended up being in `~/.steam/debian-installation/ubuntu12_32/steam-runtime` instead of the  `~/.local/share(...)` path mentioned in the docs.

I'd imagine this also holds true for Ubuntu & other Debian based distros.

Finally, on QT Creator I had to omit the `--parallel` flag for my ancient CPU, as including it ended up deadlocking the system until the build would finish, with mouse & keyboard going completely unresponsive. Probably solvable with some nice level fiddling, but I don't mind the sequentiality.

AdamTadeusz:
Couldn't figure out how to add -neopath so its picked up, but using the default neotokyo installation path works

nullsystem:
I have updated the README after Rain's feedback, also expanded symlink especially since Linux's `sudo mount --bind` wasn't persistent.

README.md
```diff
@@ -2,10 +2,221 @@
...
+* [Toolbx](https://containertoolbx.org/): `toolbox create -i registry.gitlab.steamos.cloud/steamrt/sniper/sdk sniper; toolbox enter sniper`
+
+Depending on the terminal, you may need to install an additional terminfo in the container just to make it usable.
+The container is based on Debian 11 "bullseye", just look for and install the relevant terminfo package if needed.
+
+##### CLI Building steps
+Using with the ninja build system, to build NT;RE using the CLI can be done with:
+
+```
+$ cd /PATH_TO_REPO/neo/mp/src
```
Masterkatze, line 86: Missing step to run docker container (only for [Docker] and [Podman]?)
For [Docker] it would be:
```
docker run -v /PATH_TO_REPO/neo/mp/src:/root/neo/mp/src --rm -it --entrypoint /bin/bash registry.gitlab.steamos.cloud/steamrt/sniper/sdk
cd /root/neo/mp/src/
```

nullsystem: I think podman maps practically 1:1 to docker, just without need of special privilage. Nevertheless tried your example with podman and it worked so.

nullsystem:
@Masterkatze I've updated the docs with based on your feedback

Masterkatze (approved):

PR #263, opened 2024-05-24
Squad HUD First iteration

AdamTadeusz:
First iteration of the new team hud. Doesn't make much sense that all enemy players show on the hud but only people in the same squad show on the friendly side but oh well, can be easily changed later. Might want to add class Icons and names, will probably have to move the friendly healthbar down and make it move side to side like in siege, or add a red overlay instead that moves up the more damage a friendly has taken


nullsystem:
@AdamTadeusz 
squadHUD
Found two things:
* Team player count in 2 digits seems cut off
* The "(Warmup) " indicator makes the timer cut off. I think the indicator could probably simply be moved below the score/round count. Also there'll be more indicators like match point/sudden death set in another PR.

AdamTadeusz:
I was thinking we replace the "(Warmup)" text with like a "T-", just to keep that middle section from getting too big, but if we're gonna have more indicators then yeah I need to come up with another solution, maybe just making the section 3 rows high and placing the indicator on its own row like you suggested

AdamTadeusz:
@nullsystem mind if I just leave the warmup text as is, we merge in both this and your different status changes, and then I make another commit after to move the status onto a different line? Just to prevent merge conflicts
image

AdamTadeusz:
I pushed two commits, first one has changes to the icons so you can see team mate's class selection, will replace with nicer looking ones once someone talented makes them

nullsystem:
> @nullsystem mind if I just leave the warmup text as is, we merge in both this and your different status changes, and then I make another commit after to move the status onto a different line? Just to prevent merge conflicts image

That's fine

nullsystem (approved):

PR #268, opened 2024-05-26
Fix vgui button issue

Masterkatze:
* Set default optimization level to -O2 and disable two optimizations to fix vgui button issue.

* Create NEO_OUTPUT_LIBRARY_PATH directory when configuring.


nullsystem (approved):
Can confirm both: Release build from SteamRT 3.0/GCC 10 has its vgui button issue fixed, and the `bin` directory created by cmake.

Issue #301, opened 2024-05-27
Support doesn't carry a knife


blaberry:
Support still has a knife - Wakfu

PR #306, opened 2024-05-28
Prevent supports from being given a knife

nullsystem:
Fixes https://github.com/NeotokyoRebuild/neo/issues/301


AdamTadeusz (approved):

AdamTadeusz:
Approving, but not sure if we resolved this conversation, might want to put a bit about this in the design document or something.

PR #308, opened 2024-05-28
Do not create symlinks on Windows (fix #307)

### LINKED ISSUE #307: Windows build regression
Commit e4489aff6f8698ec982b6aad66750089ed1c3054 adds build commands that are incompatible with Windows. This should probably be Linux only, or possibly Linux & MacOS(?), if we're supporting MacOS as a target.

Windows build fails with:
```
Severity	Code	Description	Project	File	Line	Suppression State	Details
Error		failed to create symbolic link 'E:/code/rebuild/neo/mp/src/../game/neo/bin/client_srv.so': A required privilege is not held by the client.	E:\code\rebuild\neo\mp\src\out\build\x86-Debug\src	E:\code\rebuild\neo\mp\src\out\build\x86-Debug\EXEC	1		
```

Masterkatze:


Rainyan (commented):
Thanks, this fixes the Windows build.

However, looking at the generated symlinks on Linux (`ls -la`):
```
client_srv.so -> ~/code/rebuild/neo/mp/src/../game/neo/bin/client.so
server_srv.so -> ~/code/rebuild/neo/mp/src/../game/neo/bin/server.so
```
Presumably, the user of a pre-built release would not have the `src` folder available to relatively walk through, so wouldn't  this symlink fail for them? And also, wouldn't this hard-code the build machine's absolute `NEO_OUTPUT_LIBRARY_PATH` to the symlinks(?)

I.e. shouldn't we rather have the symlinks defined as something like:
```
client_srv.so -> ./client.so
server_srv.so -> ./server.so
```
such that they are independent of any surrounding folder structure? Unless I'm misunderstanding something with how these symlinks are used.

I guess this is somewhat unrelated to this PR, so I could approve this as-is, but perhaps you could take a look, as you know more about the CMakeLists structure we're using for this.

Masterkatze:
Thanks for noticing, fixed

Rainyan (approved):
Looks good to me :+1:

PR #311, opened 2024-05-29
Reduce compiler warnings

nullsystem:
* Use 0/'\0' instead of NULL if assigning to int and not pointer
* Misleading indentation warnings
* `constexpr int/float` to replace enum usages being depreciated
* Disable some warnings by ifdef blocks
* Missing virtual deconstructor
* Hide a lot of MSVC warnings which are often just about namings in general
* Fixed `CNEOHud_RoundState::DrawFriend` is alive if block


Masterkatze (approved):

mp/src/game/client/neo/ui/neo_hud_round_state.h
```diff
@@ -7,8 +7,7 @@
 #include "neo_hud_childelement.h"
 #include "hudelement.h"
 #include <vgui_controls/Panel.h>
-
-class vgui::ImagePanel;
+#include <vgui_controls/ImagePanel.h>
```
Rainyan, line 10: Couldn't we include in the cpp file, and leave the forward declaration here?

mp/src/game/shared/neo/neo_ghost_cap_point.cpp
Rainyan, line 1: Not strictly related to changes in this PR, but I'm seeing a warning here:
```
/mp/src/game/shared/neo/neo_ghost_cap_point.cpp:255: warning: embedded ‘\0’ in format [-Wformat-contains-nul]
/mp/src/game/shared/neo/neo_ghost_cap_point.cpp: In member function ‘void CNEOGhostCapturePoint::Think_CheckMyRadius()’:
/mp/src/game/shared/neo/neo_ghost_cap_point.cpp:255:37: warning: embedded ‘\0’ in format [-Wformat-contains-nul]
  255 |                 V_sprintf_safe(msg, "%s captured the ghost\0",
      |                                     ^~~~~~~~~~~~~~~~~~~~~~~~~
```
I don't think we need an explicit terminator here - might as well also clean this for this PR?

mp/src/CMakeLists.txt
Rainyan, line 1: MSVC Nitpick:

```
Severity	Code	Description	Project	File	Line	Suppression State	Details
Warning	D9025	overriding '/EHs-' with '/EHs'	E:\code\REBUILD\neo\mp\src\out\build\x86-Debug\src	E:\code\REBUILD\neo\mp\src\out\build\x86-Debug\cl	1		


Severity	Code	Description	Project	File	Line	Suppression State	Details
Warning	D9025	overriding '/EHc-' with '/EHc'	E:\code\REBUILD\neo\mp\src\out\build\x86-Debug\src	E:\code\REBUILD\neo\mp\src\out\build\x86-Debug\cl	1		
```

Possibly related to line 128?

mp/src/game/shared/neo/neo_ghost_cap_point.cpp
nullsystem: I think it's relevant also, resolved it now

mp/src/game/server/ai_memory.cpp
```diff
@@ -546,7 +546,11 @@ bool CAI_Enemies::HasFreeKnowledgeOf( CBaseEntity *pEnemy )
 
 	if ( pEnemy != AI_UNKNOWN_ENEMY )
 		DevWarning( 2,"Asking HasFreeKnowledgeOf for enemy that's not in my memory!!\n");
+#ifdef NEO
+	return false;
```
Rainyan, line 550: I think we should return `true` here, because the base engine behaviour is casting `AI_INVALID_TIME` to bool, and it evaluates to a nonzero value. And any nonzero boolean has the truth value of `true`, not `false`. Unless there's some specific reason for us to differ from the base engine implementation here?

mp/src/CMakeLists.txt
nullsystem: I think I'll leave this up to @Masterkatze 
It seems `tier1\CMakeLists.txt` `processor_detect.cpp` re-enables exception handling for just that

mp/src/game/server/ai_memory.cpp
nullsystem: My rational was it seems like the usage of `AI_INVALID_TIME` was an accidental mistake and was intend to be false. Although you are right it was basically a true originally.

nullsystem: I've opted to do the static_cast route and put a comment on it regarding this line.

Rainyan: Cool, yeah I think semantically this PR should probably be a pure refactor. If this particular behaviour ever becomes an issue, we should probably handle it separately.

mp/src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -1333,8 +1333,8 @@ bool CNEORules::RoundIsMatchPoint() const
 		{
 			const int roundDiff = neo_round_limit.GetInt() - (m_iRoundNumber - 1);
 			return (!RoundIsInSuddenDeath() &&
-				((teamJinrai->GetRoundsWon() + 1) > (teamNSF->GetRoundsWon() + roundDiff)) ||
-				((teamNSF->GetRoundsWon() + 1) > (teamJinrai->GetRoundsWon() + roundDiff)));
+				(((teamJinrai->GetRoundsWon() + 1) > (teamNSF->GetRoundsWon() + roundDiff)) ||
```
Rainyan, line 1336: This seems like an implementation change unrelated to the refactor. But I guess we can merge it here if you think this should be changed?

Rainyan: Upon further pondering, I guess this doesn't change the truthness after all? It is a bit confusing with all the braces though.

Rainyan: Perhaps we could open this up as something like:
```cpp
			if (RoundIsInSuddenDeath())
			{
				return false;
			}

			if (teamJinrai->GetRoundsWon() + 1 > teamNSF->GetRoundsWon() + roundDiff)
			{
				return true;
			}

			if (teamNSF->GetRoundsWon() + 1 > teamJinrai->GetRoundsWon() + roundDiff)
			{
				return true;
			}

			return false;
```
so it's a bit more readable?

nullsystem: Did it a bit different style but done nevertheless.

Rainyan (approved):
LGTM, thanks

mp/src/CMakeLists.txt
Masterkatze: Noted, I'll eliminate this warning later.

PR #314, opened 2024-05-30
Change ghost cap distance calculation to match OG NT.

### LINKED ISSUE #269: Ghost capzones match NT06 in size
Ghost cap zones seem to be smaller in NT;Re than in Legacy, which leads to ghost capping not being possible on some maps or delayed, for example Rise heli cap. - StellaNova

Making ghost capture region the same. - Wakfu

Agiel:
For parity the distance is calculated in 3D instead of 2D, and the cap zone radius is not halved. Also, a distance less than or equal to the radius should be a cap, not just less than (this caused some annoying off by a frame issues in one of my old plugins 😅).

Fixes #269.

### DIFF
--- a/mp/src/game/shared/neo/neo_ghost_cap_point.cpp
+++ b/mp/src/game/shared/neo/neo_ghost_cap_point.cpp
@@ -215,13 +215,13 @@ void CNEOGhostCapturePoint::Think_CheckMyRadius(void)
 		}
 
 		const Vector dir = player->GetAbsOrigin() - GetAbsOrigin();
-		const float distance = dir.Length2D();
+		const float distance = dir.Length();
 
 		Assert(distance >= 0);
 
 		// Has the ghost carrier reached inside our radius?
 		// NEO TODO (Rain): newbie UI helpers for approaching wrong team cap
-		if (distance >= (m_flCapzoneRadius / 2.0f))
+		if (distance > m_flCapzoneRadius)
 		{
 			continue;
 		}


Rainyan:
This is clearly better than what we had, but I think it's still slightly off from the original.

For example, comparing the van cap in nt_dawn:
```
map nt_dawn_ctg

// spawn a bot to the other team
sv_cheats 1; bot_add

// steps:
//   1. make sure you spawn in the outside, so your capzone is inside
//   2. go grab the ghost from the world

cl_showpos 1

// This teleports you just barely outside the capzone range.
// Might still not be quite 100% accurately the actual limit, but it's pretty close:
setpos -350 179.750344 -135; setang 0 -90 0

// This triggers the van cap for original NT:
setpos -350 179.750343 -135; setang 0 -90 0
```

For the remake, this limit is somewhere closer to `setpos -350 178.7 -135; setang 0 -90 0` before the cap actually goes in. So I think we're most likely off by 1 unit (~179.7 vs ~178.7) somewhere. Peeking at the disassembly, I think the original implementation calls `CEntitySphereQuery` for this, but I'm guessing we could just offset the calculation accordingly to get the same behaviour.

AdamTadeusz (dismissed):

Rainyan (changes requested):
As per https://github.com/NeotokyoRebuild/neo/pull/314#issuecomment-2141845010

I'd be willing to merge, but in that case we shouldn't close #269 as this isn't full parity currently.

Agiel:
The off-by-one unit discrepancy is certainly weird, but I based nt-wincond off the disassembly and there's nothing weird going on there as far as I could tell https://github.com/Agiel/nt-wincond/blob/main/addons/sourcemod/scripting/nt_wincond.sp#L271.

Hopefully I'll have some time to dig into it this weekend, but if it's not something obvious I think it's fine to merge this and call it close enough.

Rainyan (approved):

Rainyan:
> The off-by-one unit discrepancy is certainly weird, but I based nt-wincond off the disassembly and there's nothing weird going on there as far as I could tell https://github.com/Agiel/nt-wincond/blob/main/addons/sourcemod/scripting/nt_wincond.sp#L271.
> 
> Hopefully I'll have some time to dig into it this weekend, but if it's not something obvious I think it's fine to merge this and call it close enough.

Alright. I think this is good to merge then, but we should re-open #269 until the possible distance discrepancy has been investigated.

nullsystem (approved):

Issue #302, opened 2024-05-27
Sprinting with knife has a sprinting animation cycle


blaberry:
No sprinting animation when using a knife. -Wakfu

PR #317, opened 2024-05-31
Knife - Refactor to use CNEOBaseCombatWeapon, use sprint animation when sprinting

nullsystem:
* Refactored the knife to use `CNEOBaseCombatWeapon` as its based class, pulled over some melee specific stuff over
* Refactored on functions that differentiates the knife when it wasn't a Neo-weapon
* When sprinting, use the sprint animation. Fixes https://github.com/NeotokyoRebuild/neo/issues/302


Agiel (approved):
LGTM, but we should move the damage to the script in the future.

PR #320, opened 2024-06-01
Compass HUD visual closer to OG:NT Styling + HudLayout.res customization

nullsystem:
* Styling closer to OG:NT, rounded corners, faded text, compass length, and NSWE direction adjusted to OG:NT's NSWE directions
* Changed hud customisation, now can be done via `HudLayout.res` (currently OG:NT's HudLayout.res takes precedent)
* Fixes: https://github.com/NeotokyoRebuild/neo/issues/50

### DIFF
--- a/mp/game/neo/scripts/HudLayout.res
+++ b/mp/game/neo/scripts/HudLayout.res
@@ -502,4 +502,12 @@
 		"visible"		"0"
 		"enabled"		"0"
 	}
+	NHudCompass
+	{
+		"fieldName"		"NHudCompass"
+		"visible"		"1"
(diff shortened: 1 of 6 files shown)


Agiel:
In OG NT the fade is a lot stronger, to the point where the opacity goes to 0 near the edges of the compass.
Also, maybe outside the scope for this issue but the text moves outside the background at 1280x720.
image

nullsystem:
@Agiel Applied a stronger fade (full opacity color) and fixed text moved outside by just simply applying the text width with padding to the width.

Agiel (approved):
Much better 💯

PR #321, opened 2024-06-01
NT;RE assets take precedence over OG NT assets.

Agiel:
Not the prettiest solution but it seems to be working.


nullsystem (approved):
Tested on Linux and Windows, looks fine by me.

mp/game/neo/cfg/autoexec.cfg
Rainyan, line 1: Should we also drop the `neo_config.cfg` and `neo_config_custom.cfg` files from `mp/game/neo/cfg`, as we aren't using these anymore?

Rainyan (approved):
LGTM

mp/game/neo/cfg/autoexec.cfg
Agiel: You're right. I'll remove them.

PR #322, opened 2024-06-01
Fix cloak sound not giving sounds server side

nullsystem:
* Fixes: https://github.com/NeotokyoRebuild/neo/issues/129


Agiel (approved):
LGTM

PR #326, opened 2024-06-02
Fix exit crash around m_pGameEvent in gHud situation

nullsystem:
* Just let neo_hud_elements handle it like the rest of the stuff there

### DIFF
--- a/mp/src/game/client/neo/c_neo_player.cpp
+++ b/mp/src/game/client/neo/c_neo_player.cpp
@@ -915,7 +915,7 @@ void C_NEO_Player::PreThink( void )
 			Warning("Couldn't find ghostMarker\n");
 		}
 
-		auto indicator = m_pNeoPanel->GetGameEventIndicator();
+		auto indicator = GET_HUDELEMENT(CNEOHud_GameEvent);
 
 		if (indicator)
 		{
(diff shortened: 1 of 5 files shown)


mp/src/game/client/hud.cpp
```diff
@@ -520,10 +524,25 @@ void CHud::Shutdown( void )
 
 	// Deleting hudlist items can result in them being removed from the same hudlist (m_bNeedsRemove).
 	//	So go through and kill the last item until the array is empty.
+#ifdef NEO
+	// NEO JANK (nullsystem): The original while/delete loop didn't even make sense as the size didn't really come
```
Rainyan, line 528: I think the original code's `delete` calls the `~CHudElement` desctructor, which in turn calls `CHud::RemoveHudElement` for any `m_bNeedsRemove` elements, and that's what eventually decrements the `while` counter, and shifts the array left.

Perhaps to play nice with the `CHudElementHelper::CreateAllElements` system, we could declare our HUD with `DECLARE_HUDELEMENT` in the implementation, so we didn't need to manually manage this memory?

nullsystem: Looks like it works out. For now I'll apply this `DECLARE_HUDELEMENT` to this, then the rest of the stuff migrating to that as a separate PR.

PR #328, opened 2024-06-02
Colourful healthbars

AdamTadeusz:
also picture outline and combined score with center box


Agiel (approved):
Tested in game and it looks good

Rainyan (approved):

PR #331, opened 2024-06-03
Show equipped weapons on model

AdamTadeusz:
Instead of hiding weapons on holster - stop them from being attached to the hand bone, and set their default attachment point on equip.


mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
```diff
@@ -117,7 +117,7 @@ void CNEOBaseCombatWeapon::Equip(CBaseCombatCharacter* pOwner)
 	BaseClass::Equip(pOwner);
 #ifndef CLIENT_DLL
 	NEO_WEP_BITS_UNDERLYING_TYPE weapon = GetNeoWepBits();
-	if (weapon & NEO_WEP_KYLA || weapon & NEO_WEP_MILSO || weapon & NEO_WEP_TACHI)
+	if (weapon & (NEO_WEP_KYLA + NEO_WEP_MILSO + NEO_WEP_TACHI))
```
Rainyan, line 120: While this happens to work here, I'd prefer a bitwise OR; `a+b+c` does not always equal `a|b|c` which could be an issue for composite flags etc.

nullsystem (commented):
So I can confirm it does show equipped weapons on model, both using single player + third person, and also two players/computers and viewing them both. Now, there's seems to be some flicker on the primary weapons when it's on the player's back, other weapons like secondary and grenades looks fine/doesn't flicker at least. Nevertheless I'm ok with letting this in, and the flicker figured out later up to you.
I should also note I tested with master merged in and two computers both running Linux, not sure if that affects things or not.

mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
Rainyan, line 1: If sees that if you:
* Spawn as assault
* Drop your primary gun
* Equip  the pistol

then the grenade attachment will move smoothly.

But, if you switch to your knife (so now you have both the grenade and the pistol attachments simultaneously), the previously smooth grenades start jittering when running around. I'm not really sure why this happens, but maybe the attachments are interfering with each other somehow?

AdamTadeusz: They are attached to separate attachment points, and also I'm not able to replicate this specific type of jittering on a dedicated server with other players, or locally with bots. Seems to only be present when viewing own character model in thirdperson

Rainyan (approved):

nullsystem (approved):

PR #334, opened 2024-06-05
Third person muzzle flashes

AdamTadeusz:
Adds Destroy's third person muzzle flash to all un-suppressed weapons. Also /ZI on Windows for hot-reloading


mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.h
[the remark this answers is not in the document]
AdamTadeusz: image
Are compile time asserts supposed to stop compilation? Because I'm not having any issues.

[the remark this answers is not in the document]
AdamTadeusz: image
This is the problem probably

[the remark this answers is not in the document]
Rainyan: > Are compile time asserts supposed to stop compilation? Because I'm not having any issues.

Yes, it should trigger a `static_assert` error, but it's behind some evil C macro that might be misbehaving.

mp/src/game/shared/neo/neo_player_shared.h
```diff
@@ -16,13 +16,7 @@
 
 #include "neo_predicted_viewmodel.h"
 
-#ifdef INCLUDE_WEP_PBK
```
Rainyan, line 19: Seems unnecessary to require 64 bits here, since we only use 32 (without the PBK).

AdamTadeusz: so previously were we using 32 with the PBK? Why the long long int in the first place?

Rainyan: Because previously we were using C++11, and so we needed 1 extra bit to avoid the UB of flipping the sign bit.

AdamTadeusz: 0x0 is still invalid yes?

Rainyan: > 0x0 is still invalid yes?

Yes, although we might not need to have it if it's not being referred to anywhere except for that old assertion.

mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
```diff
@@ -94,6 +103,7 @@ CNEOBaseCombatWeapon::CNEOBaseCombatWeapon( void )
 
 void CNEOBaseCombatWeapon::Spawn()
 {
+	PrecacheParticleSystem("ntr_muzzle_source");
 	// If this fires, either the enum bit mask has overflowed,
 	// this derived gun has no valid NeoBitFlags set,
 	// or we are spawning an instance of this base class for some reason.
```
Rainyan, line 109: This assertion is also outdated for the same reasons as the previous one, and is triggering incorrectly. Should rewrite as `Assert(GetNeoWepBits() != NEO_WEP_INVALID)`.

mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
```diff
@@ -617,6 +630,51 @@ bool CNEOBaseCombatWeapon::CanBePickedUpByClass(int classId)
 	return true;
 }
 
+void CNEOBaseCombatWeapon::DispatchMuzzleParticleEffect() {
+	const char* particleName = "ntr_muzzle_source";
+	const bool reselAllParticlesOnEntity = false;
```
Rainyan, line 635: Nit: typo on var name & this could be made constexpr

mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
```diff
@@ -617,6 +630,51 @@ bool CNEOBaseCombatWeapon::CanBePickedUpByClass(int classId)
...
+	const ParticleAttachment_t iAttachType = ParticleAttachment_t::PATTACH_POINT_FOLLOW;
+
+	int iAttachment = -1;
+	if (GetBaseAnimating()) {
+		// Find the attachment point index
+		iAttachment = GetBaseAnimating()->LookupAttachment("muzzle");
+		if (iAttachment <= 0)
+		{
+			Warning("Model '%s' doesn't have attachment '%s' to attach particle system '%s' to.\n", GetPrintName(), "muzzle", particleName);
```
Rainyan, line 644: This warning seems to be triggering for the local player on client.dll. Should this be a `<0` check, instead?

Rainyan (changes requested):

mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
AdamTadeusz: This is taken from one of the first DispatchParticleEffect functions in particle_parse.cpp so I think it should be fine
image
It might be trying to attach the flash effect to a weapon that doesn't have a muzzle, should probably add a check that weapon is of type firearm or something

Rainyan: I'm seeing the muzzle correctly if I silence this error with `==-1` check though, so I'm assuming there's some kind of discrepancy between client & server implementation here. Haven't looked at the details though.

AdamTadeusz: the whole reason for this function instead of calling dispatchparticleeffect is that i want to modify the filter thats created at one point so the owner of the gun being fired doesn't see their own 3rd person muzzle flash effect

Rainyan: Ok, this actually only triggers if you switch to `thirdperson`. But even after returning back to `firstperson`, it keeps spamming the warning:

> ] thirdperson 
] firstperson 
Model 'TACHI' doesn't have attachment 'muzzle' to attach particle system 'ntr_muzzle_source' to.
Model 'TACHI' doesn't have attachment 'muzzle' to attach particle system 'ntr_muzzle_source' to.
Model 'TACHI' doesn't have attachment 'muzzle' to attach particle system 'ntr_muzzle_source' to.
Model 'TACHI' doesn't have attachment 'muzzle' to attach particle system 'ntr_muzzle_source' to.
Model 'TACHI' doesn't have attachment 'muzzle' to attach particle system 'ntr_muzzle_source' to.
Model 'TACHI' doesn't have attachment 'muzzle' to attach particle system 'ntr_muzzle_source' to.
Model 'TACHI' doesn't have attachment 'muzzle' to attach particle system 'ntr_muzzle_source' to.

AdamTadeusz: If I remember correctly you can also no longer see any muzzle flashes at all when back in first person?

Rainyan: > If I remember correctly you can also no longer see any muzzle flashes at all when back in first person?

I can still see a bot's muzzle fine after the `thirdperson` swap. I think the error is only happening because the local client is attempting to draw their own muzzle, and failing (even in `firstperson`, after you've entered `thirdperson` previously).

AdamTadeusz: I think I resolved the issue, muzzle flashes in thirdperson are disabled for the owner, need to figure out a way to check if a player is in thirdperson mode or not to re-enable them

mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
```diff
@@ -617,6 +632,40 @@ bool CNEOBaseCombatWeapon::CanBePickedUpByClass(int classId)
 	return true;
 }
 
+#ifdef GAME_DLL
```
Rainyan, line 635: Seems like there's no need for the server to calculate the flashes & send them over the wire here - couldn't the clients themselves do this work for the events they are locally aware of?

mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.h
```diff
@@ -148,6 +148,9 @@ class CNEOBaseCombatWeapon : public CBaseHL2MPCombatWeapon
 	virtual void ItemPostFrame(void);
 
 	virtual void PrimaryAttack(void);
+#ifdef GAME_DLL
+	virtual void DispatchMuzzleParticleEffect(void);
```
Rainyan, line 152: Probably doesn't need to be virtual if we're rolling our own function. Alternatively could explore overriding `C_BaseAnimating::DispatchMuzzleEffect` and/or `C_BaseAnimating::ProcessMuzzleFlashEvent`.

mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
AdamTadeusz: I was following this tutorial https://developer.valvesoftware.com/wiki/Particle_Muzzle_Flash which is why its done the way it is, but I can probably see what function is fired on the clientside when the particleeffect is dispatched from the server, and call it in this function itself only on the client side

AdamTadeusz: Figured it out

mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.h
AdamTadeusz: Ok im overriding processmuzzleflashevent now, this allows us to also hide the ent_light when using suppressed weapons, and also now a muzzle flash is created by the client when the muzzleflashevent is triggered.

Rainyan (changes requested):
Thanks, the client side muzzle flashes seem good! Wrote some notes though.

There's also this strange behaviour where the tachi pistol fires a muzzle in the environment - but this seems unrelated to changes in this PR, as this bug is also present in the current master.

But in case you want to investigate - note the second muzzle flash near the floor:

[tachi.webm](https://github.com/NeotokyoRebuild/neo/assets/6595066/b35d054a-d829-4c41-a9d9-be3ccbef12b2)

I'm guessing this second muzzle might be the original Tachi first person muzzle flash, wherever it's being called from, but I'm not sure.

mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
```diff
@@ -617,6 +625,68 @@ bool CNEOBaseCombatWeapon::CanBePickedUpByClass(int classId)
...
+		{
+			Warning("Model '%s' doesn't have attachment '%s'\n", GetPrintName(), "muzzle_flash");
+			return;
+		}
+	}
+
+	// Muzzle flash light
+	Vector vAttachment;
+	QAngle dummyAngles;
+	GetAttachment(iAttachment, vAttachment, dummyAngles);
```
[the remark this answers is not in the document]
AdamTadeusz: image
How about we instead return early if GetBaseAnimating doesn't return anything? Saves us an operation, assuming GetAttachment will always return an attachment given that GetBaseAnimating()->LookupAttachment() found an attachment

mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
```diff
@@ -617,6 +625,68 @@ bool CNEOBaseCombatWeapon::CanBePickedUpByClass(int classId)
...
+	el->color.r = 255;
+	el->color.g = 192;
+	el->color.b = 64;
+	el->color.exponent = 5;
+
+	// Muzzle flash particle
+	DispatchMuzzleParticleEffect(iAttachment);
+}
+
+void CNEOBaseCombatWeapon::DispatchMuzzleParticleEffect(int iAttachment) {
```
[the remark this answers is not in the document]
AdamTadeusz: image
Good find, was supposed to pass this to DispatchEffect via data. Since most 3rd person weapon models only have one attachment point (the muzzle_flash), it still worked before

mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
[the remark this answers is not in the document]
Rainyan: Sure, if you think it's a safe assertion that `GetAttachment` can never fail in this context. At which point one might also question whether `GetBaseAnimating` can ever fail, although if it does, it'd trigger a null pointer dereference, so I guess the `GetAttachment` failure state of "just" rendering incorrectly positioned muzzle flashes is less bad.

AdamTadeusz:
I have also changed the dlight_t elements to a light that draws both on entities and the world, and changed it's size and lifetime to correspond with the base games flash size and lifetime (https://discord.com/channels/1235346473827434517/1235355938458701874/1248940735911432235)

Rainyan (approved):
Looks good, thanks!

nullsystem (approved):
Tested single-thirdperson and 2-clients viewing muzzle flash from both clients and it does show.

PR #335, opened 2024-06-05
Scoped weapon zoom at parity

AdamTadeusz:
Scoped weapon zoom now decoupled from neo_fov and should always show the same amount of stuff as og tokyo at 4:3.

### DIFF
--- a/mp/src/game/client/neo/c_neo_player.cpp
+++ b/mp/src/game/client/neo/c_neo_player.cpp
@@ -1414,9 +1414,15 @@ void C_NEO_Player::Weapon_SetZoom(const bool bZoomIn)
 	if (bZoomIn)
 	{
 		m_Local.m_iHideHUD &= ~HIDEHUD_CROSSHAIR;
-
-		const int zoomAmount = 30;
-		SetFOV((CBaseEntity*)this, fov - zoomAmount, zoomSpeedSecs);
+		auto neoWep = dynamic_cast<CNEOBaseCombatWeapon*>(GetActiveWeapon());
+		if (neoWep && neoWep->GetNeoWepBits() & NEO_WEP_SCOPEDWEAPON)
+		{
+			SetFOV((CBaseEntity*)this, 16, zoomSpeedSecs);
+		}
+		else {
+			const int zoomAmount = 30;
+			SetFOV((CBaseEntity*)this, fov - zoomAmount, zoomSpeedSecs);
+		}
 	}
 	else
 	{
--- a/mp/src/game/client/neo/c_neo_player.cpp
+++ b/mp/src/game/client/neo/c_neo_player.cpp
@@ -1414,6 +1414,7 @@ void C_NEO_Player::Weapon_SetZoom(const bool bZoomIn)
 	if (bZoomIn)
 	{
 		m_Local.m_iHideHUD &= ~HIDEHUD_CROSSHAIR;
+		const int zoomAmount = 30;
 		auto neoWep = dynamic_cast<CNEOBaseCombatWeapon*>(GetActiveWeapon());
 		if (neoWep)
 		{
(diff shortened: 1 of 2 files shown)


mp/src/game/client/neo/c_neo_player.cpp
```diff
@@ -1414,9 +1414,15 @@ void C_NEO_Player::Weapon_SetZoom(const bool bZoomIn)
 	if (bZoomIn)
 	{
 		m_Local.m_iHideHUD &= ~HIDEHUD_CROSSHAIR;
-
 		const int zoomAmount = 30;
-		SetFOV((CBaseEntity*)this, fov - zoomAmount, zoomSpeedSecs);
+		auto neoWep = dynamic_cast<CNEOBaseCombatWeapon*>(GetActiveWeapon());
```
nullsystem, line 1418: But this affected non-scoped weapons?

nullsystem (changes requested):

mp/src/game/client/neo/c_neo_player.cpp
AdamTadeusz: yes, but now they too are at parity no? Or do we want non scoped weapons fov when aiming to change with neo_fov?

nullsystem: I think non-scoped weapons aiming should keep with being scaled with neo_fov. With this change aiming non-scoped weapons with low or high neo_fov looks odd.

Agiel: I'm going to nip this one in the bud right here. I'm fine with having hip fire fov be adjustable because it was making people queasy, but adjustable aim fov would majorly affect gameplay. If you want it for WCD you take it through the proper channels.

AdamTadeusz: Nip all you want, but this branch currently doesn't change unscoped weapons fov behaviour from master, so unless there are any other issues I think we can merge and close this issue, maybe open another better describing the issue you mentioned

Agiel: It did before the last commit...

AdamTadeusz: And as @nullsystem rightfully pointed out this is a larger issue that probably deserves its own issue and pull request, given the different ways we could go about solving it, and given the gameplay impact that you pointed out

Agiel: I don't see the problem. You had an if-statement that made all guns parity, but then you deliberately changed it to only some guns. We're either working towards parity or we're not. This isn't going to work if people pull in different directions.

AdamTadeusz: I'm using the same argument as you to justify leaving this change that is outside the scope of this pull request outside the scope of this pull request. If you want we can open up another issue and discuss this problem there, I'm not going to force a controversial change to a mechanic in a pr with a name that doesn't address it. And don't worry about the if-statement that once existed, we're squashing changes into a single commit so it shouldn't appear .

AdamTadeusz: What we had before the last commit was plugin parity

mp/src/game/server/neo/neo_player.cpp
```diff
@@ -1251,7 +1251,14 @@ void CNEO_Player::Weapon_SetZoom(const bool bZoomIn)
 	if (bZoomIn)
 	{
 		const int zoomAmount = 30;
-		SetFOV((CBaseEntity*)this, fov - zoomAmount, NEO_ZOOM_SPEED);
+		auto neoWep = dynamic_cast<CNEOBaseCombatWeapon*>(GetActiveWeapon());
+		if (neoWep)
```
nullsystem, line 1255: Think this need to be altered also.

PR #337, opened 2024-06-05
Ghost Marker animated

AdamTadeusz:
in-code ghost marker animation

### DIFF
--- a/mp/src/CMakeLists.txt
+++ b/mp/src/CMakeLists.txt
@@ -131,6 +131,7 @@ if(OS_WINDOWS)
         /errorReport:prompt # ErrorReporting - Prompt Immediately
         /W4
         /Gw # Optimize Global Data
+        /ZI
 
         /wd4456
         /wd4457
--- a/mp/src/CMakeLists.txt
+++ b/mp/src/CMakeLists.txt
@@ -131,7 +131,6 @@ if(OS_WINDOWS)
         /errorReport:prompt # ErrorReporting - Prompt Immediately
         /W4
         /Gw # Optimize Global Data
-        /ZI
 
         /wd4456
         /wd4457
--- a/mp/src/game/client/neo/ui/neo_hud_ghost_marker.cpp
+++ b/mp/src/game/client/neo/ui/neo_hud_ghost_marker.cpp
@@ -130,18 +130,18 @@ void CNEOHud_GhostMarker::DrawNeoHudElement()
 		if (m_fMarkerScalesCurrent[i] > 0.5)
 			alpha *= ((0.5 - (m_fMarkerScalesCurrent[i] - 0.5)) * 2);
 
-		Color ghostColor = Color(204, 204, 204, alpha);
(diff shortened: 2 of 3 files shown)


mp/src/game/client/neo/ui/neo_hud_ghost_marker.cpp
```diff
@@ -111,31 +111,49 @@ void CNEOHud_GhostMarker::DrawNeoHudElement()
...
+			{
+				// If viewing from playing player, but opposite of ghosting team, show red
+				ghostColor = COLOR_RED;
+			}
+			else
+			{
+				// Otherwise show ghosting team color (if friendly or spec)
+				ghostColor = (m_iGhostingTeam == TEAM_JINRAI) ? COLOR_JINRAI : COLOR_NSF;
+			}
 		}
```
nullsystem, line 146: These will override the alpha values. Set the initial colors then edit the alpha value on the color.

nullsystem (changes requested):

mp/src/game/client/neo/ui/neo_hud_ghost_marker.cpp
AdamTadeusz: yeah thats why i marked it as a draft but thanks xd

nullsystem: I was thinking more like `ghostColor[3] = alpha` so the colors can still be used but it's the same result.

nullsystem (approved):

mp/src/CMakeLists.txt
```diff
@@ -131,6 +131,7 @@ if(OS_WINDOWS)
         /errorReport:prompt # ErrorReporting - Prompt Immediately
         /W4
         /Gw # Optimize Global Data
+        /ZI
```
nullsystem, line 134: Actually, don't commit this change in. MSVC is just blasting with warnings.

nullsystem (changes requested):

mp/src/game/client/neo/ui/neo_hud_ghost_marker.cpp
```diff
@@ -129,13 +124,38 @@ void CNEOHud_GhostMarker::DrawNeoHudElement()
...
 	}
 
-	surface()->DrawSetColor(ghostColor);
-	surface()->DrawSetTexture(m_hTex);
-	surface()->DrawTexturedRect(
-		offset_X,
-		offset_Y,
-		offset_X + (m_iMarkerTexWidth * scale),
-		offset_Y + (m_iMarkerTexHeight * scale));
+	const float scale = neo_ghost_marker_hud_scale_factor.GetFloat();
```
nullsystem, line 127: Also, change the default of the scale factor. It looked very small with 0.125 default.

nullsystem (changes requested):

nullsystem (approved):

PR #339, opened 2024-06-06
Fix SUB_Remove deferred thinks

### LINKED ISSUE #330: Fix deferred SUB_Remove not working for Neo guns
Currently, we're overriding virtual `SUB_Remove` for `INEOBaseWeapon` as a no-op, which was presumably a hack to prevent weapons disappearing mid-round. This breaks stuff like weapon attachment rendering upon death. We should revert this hack, and properly set the deferred removes to only occur at appropriate times for NT guns.

Also, technically, guns disappearing mid-round at inopportune moments *is* parity, so perhaps we should guard this behind a dev cvar, with the sane default of not removing the guns(?) Typically, on vanilla NT servers, this behaviour would be overridden by plugins to prevent the guns from disappearing.

Rainyan:
Fix #330 

Previously, we were overriding `SUB_Remove` for NT guns as a hack to prevent them from disappearing mid-round. This fix re-enables the correct `SUB_Remove` behaviour, and instead handles the deferred removals at the `SetPickupTouch` override, based on `mp_weaponstay` cvar's boolean value - with the exception of the ghost never being deleted from the world, regardless of the cvar's value.

We're defaulting to the more sensible `mp_weaponstay 1` behaviour to keep dropped guns in the world over parity, but the original NT behaviour can be restored by toggling this cvar to 0 instead.

### DIFF
--- a/mp/src/game/server/subs.cpp
+++ b/mp/src/game/server/subs.cpp
@@ -86,20 +86,6 @@ void CBaseEntity::SUB_Remove( void )
 		DevWarning( 2, "SUB_Remove called on entity with health > 0\n");
 	}
 
-#if(0)
-#ifdef NEO
-	// NEO HACK (Rain): Hacky way to ensure Neo weapons don't remove themselves.
-	// There's probably some flag we could raise to prevent it from happening
-	// in a nicer way.
-	auto neoWep = dynamic_cast<CNEOBaseCombatWeapon*>(this);
-
-	if (neoWep)
-	{
-		return;
-	}
-#endif
-#endif
-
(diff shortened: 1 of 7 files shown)


AdamTadeusz (approved):
Haven't tested behavior on a server with the convar but otherwise lgtm

nullsystem (approved):
Tested, LGTM:
* `mp_weaponstay 1` - Weapons stays throughout round, ghost stays
* `mp_weaponstay 0` - Weapons eventually disappears, ghost stays

Rainyan:
Geh, should've squashed it. Oh well.

PR #340, opened 2024-06-06
Convars in Advanced multiplayer options

AdamTadeusz:
Moves the advanced options button so it can be displayed, and adds a list of helpful convars to the advanced options


nullsystem:
Add in `cl_fakenick` checkbox toggle
Other than that looks good

AdamTadeusz:
> Add in `cl_fakenick` checkbox toggle Other than that looks good

Changed to cl_onlysteamnick and added. Also we're now archiving all convars in that menu so they don't have to be set each time

mp/src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -31,7 +31,7 @@ ConVar mp_neo_latespawn_max_time("mp_neo_latespawn_max_time", "15", FCVAR_REPLIC
 ConVar sv_neo_wep_dmg_modifier("sv_neo_wep_dmg_modifier", "0.5", FCVAR_REPLICATED, "Temp global weapon damage modifier.", true, 0.0, true, 100.0);
 
 ConVar neo_name("neo_name", "", FCVAR_USERINFO | FCVAR_ARCHIVE, "The nickname to set instead of the steam profile name.");
-ConVar cl_fakenick("cl_fakenick", "1", FCVAR_USERINFO | FCVAR_ARCHIVE, "Show players set neo_name, otherwise only show Steam names.", true, 0.0f, true, 1.0f);
+ConVar cl_fakenick("cl_onlysteamnick", "1", FCVAR_USERINFO | FCVAR_ARCHIVE, "Only show players Steam names, otherwise show player set names.", true, 0.0f, true, 1.0f);
```
nullsystem, line 34: All `cl_fakenick` in code both variable and in comments should change to `cl_onlysteamnick` also. Also change comment a bit EX: L1375 neo_gamerules.cpp `(only used if cl_fakenick=1` to `(only used if cl_onlysteamnick=0`

Also I think `cl_onlysteamnick` should be off/0 like how cl_fakenick=1 was before.

mp/game/neo/cfg/user_default.scr
```diff
@@ -0,0 +1,116 @@
...
+		"Fake Name"
+		{ STRING }
+		{ "Neotokyo Noob" }
+	}
+	"cl_onlysteamnick"
+	{
+		"Only use Steam names"
+		{ BOOL }
+		{ 1 }
+	}
```
nullsystem, line 55: Default `neo_name` change to "". Player's Steam name will be used when empty even if `cl_onlysteamnick=0`

Likewise as the other suggestion, flip `cl_onlysteamnick` default here also.

nullsystem (changes requested):

mp/game/neo/cfg/user_default.scr
```diff
@@ -0,0 +1,116 @@
...
+
+VERSION 1.0
+
+DESCRIPTION INFO_OPTIONS
+{
+	"neo_name"
+	{
+		"Fake Name"
+		{ STRING }
+		{ "Neotokyo Noob" }
```
nullsystem, line 48: We're going with NeotokyoNoob rather than steam name default?

nullsystem (approved):

Rainyan (dismissed):

mp/game/neo/cfg/user_default.scr
```diff
@@ -0,0 +1,119 @@
...
+	{
+		"Only use Steam names"
+		{ BOOL }
+		{ "0" }
+	}
+	"neo_fov"
+	{
+		"FOV (60 - 110)"
+		{ NUMBER 60 110 } // Alternatively SLIDER, but can't have both?
+		{ "90" }
```
Rainyan, line 58: One more thing I have to point out: as per the currently open PR #390 , we're likely going to change both the min FOV and default FOV values to 75. So that should probably be reflected here. Otherwise looks good.

Rainyan (changes requested):

Rainyan (approved):
:+1:

nullsystem (approved):

PR #341, opened 2024-06-06
Add Discord link to main menu

### LINKED ISSUE #312: Main menu has a button for the games discord (like how Dystopia has it)

Rainyan:
Fix #312 

discord_example

This is currently pointing to the main ANPA Discord server.

Suggestions welcome as to what the invite link should say, or which position in the menu it should inhabit.

I'm linking via the `https://` scheme; Discord does have a `discord://` scheme, but it's currently undocumented, so we're probably better off not using it directly lest the API suddenly changes.


blaberry:
Let's hold off on this for now, I think we should have a discussion on where this should link going forward :)

Rainyan:
> is there any way to make a menu item not display when in-game? Might want to hide this when connected to a server

This should be doable, iirc there's some kind of magic keywords for whether the user is in the main menu or not. I'll look into it.

AdamTadeusz:
> > is there any way to make a menu item not display when in-game? Might want to hide this when connected to a server
> 
> This should be doable, iirc there's some kind of magic keywords for whether the user is in the main menu or not. I'll look into it.

There was an idea to use some stuff that tf2 introduced to make the menu spanking brand new, need to find that tutorial

AdamTadeusz:
> > is there any way to make a menu item not display when in-game? Might want to hide this when connected to a server
> 
> This should be doable, iirc there's some kind of magic keywords for whether the user is in the main menu or not. I'll look into it.
> 
> Update: So, there's `OnlyInGame` but I can't seem to find the reverse of that. And as it's implemented in the binary-only GameUI library, it gets a bit tricky to override.

image
Could probably just overwrite this element after its initialized from somewhere else in the code, a bit like they do in this repo https://github.com/HL2RP/HL2RP/blob/b6f9e3a76088f29260d561ae4ad92b5569f78d24/mp/src/game/client/hl2rp/hl2rp_configuration.cpp#L139

But that might be a lot of hassle for very little gain

PR #342, opened 2024-06-06
Weapon innacuracy same between own shots and shots of others

AdamTadeusz:
Use NEOShotManipulator both when shooting and when displaying other people's shots. When using shotguns, opt for the default ShotManipulator, as NEOShotManipulator does not match the expected behaviour of shotguns very well


AdamTadeusz:
actually it just occured to me I only change the shotgun c_te file, should go back and check what the other weapons are using

Rainyan (changes requested):

kassibuss:
Shouldnt we get this merged for the next playtest?

mp/src/game/shared/neo/weapons/weapon_aa13.h
```diff
@@ -41,6 +41,8 @@ class CWeaponAA13 : public CNEOBaseCombatWeapon
 	Activity GetPrimaryAttackActivity(void) OVERRIDE;
 
 	virtual float GetFireRate() OVERRIDE { return 0.333f; }
+	
+	virtual const Vector& GetBulletSpread(void) OVERRIDE { return VECTOR_CONE_7DEGREES; };
```
Rainyan, line 45: This is dangling, because the referenced Vector goes out of scope when this returns. Should be declared as a static variable inside the function.

Rainyan (approved):

nullsystem (approved):

PR #344, opened 2024-06-06
Fix missing HudLayout.res defines (HudMenu/HudRadio) - Fixes attacker info missing

nullsystem:
The HudLayout.res wasn't updated for the asset change, causing HudMenu to have empty texts which caused the attacker info to look like nothing.


AdamTadeusz (approved):
If you want to center the HudRadio regardless of its size, use cs-0.5 for the ypos

Rainyan (approved):

Issue #333, opened 2024-06-04
Crash bug on opening scoreboard/spectator GUI after changing resolution


nullsystem:
* To reproduce:
    1. Open NT;RE, Create Server, enter some map
    2. Spawn in, then press tab, opening the scoreboard + spectator GUI
    3. Esc, go to settings and change the resolution, can be any different resolution (EX: from 900p to 720p)
    4. After it changed, go back to the game and press tab, but this time it crashes the game
* Crash happens in SpectatorGUI when exiting the function
    * Error message: "Run-Time Check Failure #0 - The value of ESP was not properly saved across a function call.  This is usually a result of calling a function declared with one calling convention with a function pointer declared with a different calling convention."
* This is replicated in v5, some commits in between that and 13e08c16d277a059bf8a364b2b179319e9d33e82
* This is not replicable using the VPC + vcproj build system
* EDIT: This also happens on Linux, not really cmake's fault, and seems to be a dangling pointer issue as rain's suggestion

Rainyan:
Can verify this also occurs at commit 17ef464

Rainyan:
Potentially dangling pointers at https://github.com/NeotokyoRebuild/neo/blob/13e08c16d277a059bf8a364b2b179319e9d33e82/mp/src/game/client/game_controls/SpectatorGUI.cpp#L562-L563

PR #348, opened 2024-06-08
Fix SpectatorGUI memory corruption crash

nullsystem:
* Changed from static variable to class's member variable
* Reset pointer if new scheme changed
* Also checked to make sure the score still goes up with team wins switching back and forth resolutions
* Tested on Windows + Linux, loopback + dedicated
* Fixes: https://github.com/NeotokyoRebuild/neo/issues/333

### DIFF
--- a/mp/src/game/client/game_controls/SpectatorGUI.cpp
+++ b/mp/src/game/client/game_controls/SpectatorGUI.cpp
@@ -85,6 +85,10 @@ ConVar cl_spec_mode(
 	FCVAR_ARCHIVE | FCVAR_USERINFO | FCVAR_SERVER_CAN_EXECUTE,
 	"spectator mode" );
 
+namespace {
+constexpr char LABEL_JINRAI[] = "CTScoreValue";
+constexpr char LABEL_NSF[] = "TERScoreValue";
+}
 
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Rainyan (approved):

mp/src/game/client/game_controls/SpectatorGUI.cpp
```diff
@@ -559,30 +568,29 @@ void CSpectatorGUI::OnThink()
...
-		if (!nsfScoreValueLabel)
+		if (!m_scoreValueLabelNSF)
 		{
-			nsfScoreValueLabel = dynamic_cast<Label*>(FindChildByName("TERScoreValue"));
+			m_scoreValueLabelNSF = static_cast<Label*>(FindChildByName(LABEL_NSF));
 		}
-		Assert(jinScoreValueLabel);
-		Assert(nsfScoreValueLabel);
+		Assert(m_scoreValueLabelJinrai);
+		Assert(m_scoreValueLabelNSF);
```
AdamTadeusz, line 581: Would it be more efficient to move these Asserts into the if statements above?

nullsystem: Yes, but they're asserts so I wouldn't worry too much about it.

AdamTadeusz (approved):

mp/src/game/client/game_controls/SpectatorGUI.cpp
Rainyan: The assert macros will be defined as essentially no-ops for release builds, so there's no runtime cost outside debug

```cpp
#define  Assert(_exp)    ((void)0)
```
etc.

PR #351, opened 2024-06-08
GitHub issues + pull request templates

### LINKED ISSUE #350: GitHub issues and pull requests templates
To encourage issues writers to write better and more detailed issues, utilizing GitHub's templates would be a good idea. Documentation in here: https://docs.github.com/en/enterprise-cloud@latest/communities/using-templates-to-encourage-useful-issues-and-pull-requests

Mostly, for bug report to be filled with steps to reproduce, what's expected vs actual, and some extra details. Feature request to give a description and maybe even pictures of what's wanted.

On top of that, Rain suggested (https://github.com/NeotokyoRebuild/neo/pull/349#pullrequestreview-2105932216) it can be linked from the game also, using GitHub issue creation URL feature documented in here: https://docs.github.com/en/enterprise-cloud@latest/issues/tracking-your-work-with-issues/creating-an-issue#creating-an-issue-from-a-url-query

nullsystem:
* **NOTE: This was first done via GitHub's settings and some template to create the issues template. That's why the branch is within the main repo and why the issues markdown formatting is different to the PR's one (which doesn't have a template to create a template).**
* Add issues template giving an option between bug report or feature request
* Structured especially for bug report so it can be followed through
* Altered slightly from GitHub's given default to fit a bit more for NT;RE
* resolves #350
* Issues reporting are both YAML files to leverage form feature. PR will be kept as markdown since it'd rather it be more free-formed.


kassibuss:
LGTM

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -7,15 +7,28 @@ assignees: ''
...
+and [GIT HASH] with the Git short hash the build was built on.
 The main menu will show the build's date and Git hash.
 -->
 Build: [DATE]_[GIT HASH]
 
 **Description**
+<!--
+Build, Description, "To Reproduce", "Expected/Actual behavior", and
+"Machine info" ALL MUST be completed in. DO NOT put in chatlogs
+and instead forumlate them into detailed description and
```
Rainyan, line 29: Typo: formulate

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,33 @@
...
+## Linked Issues
+<!--
+Applying issues here will auto-link the PR to its related issues if starting with "resolves".
+If there's a related PR but don't want to resolve/close the issue, mark them with "related".
+
+See: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword
+-->
+
+* resolves #
+* related #
```
Rainyan, line 32: Perhaps default to "fixes" or "closes" keyword, so we don't have 2 similar "re<...>" keywords right next to each other, for readability.

Rainyan (approved):

nullsystem:
Actually, should the bug reports be turned into using the "issue forms" yaml format instead? It does seem to give more controls, probably an advantage with the URL linking and put in build info + OS at specific places. Not a fan of the yaml format though.

Rainyan:
> Actually, should the bug reports be turned into using the "issue forms" yaml format instead? It does seem to give more controls, probably an advantage with the URL linking and put in build info + OS at specific places. Not a fan of the yaml format though.

YAML comes up more prominently when looking for GitHub docs as to how to do this stuff. Might be a good idea.

blaberry:
I mentioned it earlier, but one of the issues with issues not created
through the projects view is that we lose the bounty tracking by default,
though we could probably fix this relatively easily with some auto-tagging.

Apologies for the surface level insight, I might have some time later today
to give a more in depth review.

nullsystem:
> I mentioned it earlier, but one of the issues with issues not created through the projects view is that we lose the bounty tracking by default, though we could probably fix this relatively easily with some auto-tagging. Apologies for the surface level insight, I might have some time later today to give a more in depth review.

@blaberry It seems like creating an issue through projects is able to use the templates anyway. Clicking on "+" -> Create a new issue -> Above "Add a title" currently it says "Blank issue in NeotokyoRebuild/neo" but with templates put in it'll be some template instead.
Even then if an issue is created through the standard issue, it can be added to the projects stuff anyway.

blaberry:
Yeah using the templates for project-created issues is a great thing, I was only concerned about the opposite flow, ie how do we make sure issues don't get lost by not being assigned a project/milestone. But if we can add them to the bounty project by default then all good by me.

nullsystem:
So regarding the URL, I'm wondering if it's a GitHub's bug but the URL GET parameters seemed to overridden the defaults set in the template. However these parameters can be set back anyway. Ending URL with: `/issues/new?template=bug_report.yml&build=20240609_abc123&labels=bug&title=[BUG]&projects=NeoTokyoRebuild/1` for example will fill the build input box with `20240609_abc123`, put in "bug" label, prefix title with "[BUG]", and auto-add it to the project board/list.

Rainyan (approved):

Rainyan:
One thing worth clarifying for the title wording is whether we want:

* State of things currently:
  * `"Recon class spawns with smoke grenades"`
* The desired state of things:
  * `"Recon class doesn't spawn with smoke grenades"`

as having many issues using both styles could get confusing.

It seems most of our existing issues follow the latter, although I could see "Thing X is broken" being what people reporting bugs would naturally gravitate towards.

nullsystem:
@Rainyan I kind of think bugs titles as "State of things currently" and features/changes titles as "The desired state of things". Either not a big deal or give a guideline for bugs title to be consistent.

Rainyan (dismissed):

blaberry (dismissed):

nullsystem:
@blaberry 
https://github.com/NeotokyoRebuild/neo/blob/0466cdce4b24f7133ba538ad94a473113dfcb0d0/.github/ISSUE_TEMPLATE/feature_request.yml

> There is a problem with this template
> 
> YAML syntax error: (): could not find expected ':' while scanning a simple key at line 40 column 7. [Learn more about this error.](https://docs.github.com/communities/using-templates-to-encourage-useful-issues-and-pull-requests/common-validation-errors-when-creating-issue-forms)

blaberry (dismissed):

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,56 @@
...
+      placeholder: Add a football physics object in each spawn during warmup
+    validations:
+      required: true
+  - type: textarea
+    id: proposal
+    attributes:
+      label: Proposed Solution
+      description: |
+        A clear and concise description of what you want to happen.
+      placeholder: At the RoundStart event trigger during warmup, spawn a football physics object to mess around with
```
Rainyan, line 39: Might be worth baking in the [user story](https://en.wikipedia.org/wiki/User_story#Common_templates) style "who", "what", "when", "why" of the request in the example text.

* "As a **player**, I often **get bored** during the **warmup**, so it **would be fun** to **have a football** physics object to mess around with."

* "As a **plugin developer**, I would really like **feature ABC** **so that** I can **XYZ**."

* "As a **mapper**, I want to have **bullet proof textures** so I **can make my bullet proof windows less thick**.

etc.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,120 @@
+name: Bug report
+description: General bugs, program crashes, broken files, wrong values, etc...
+title: "[BUG]"
```
Rainyan, line 3: Nit: Add trailing space, so the user can start typing right away. Since we probably want `[BUG] Foobar` (spaced) and not `[BUG]Foobar` formatting.

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,56 @@
+name: Feature request
+description: Suggest an idea for NT;RE
+title: "[FEATURE]"
```
Rainyan, line 3: Ditto on trailing space char.

Rainyan (approved):

Rainyan (approved):

blaberry (approved):

PR #352, opened 2024-06-09
Change ADS toggle from on-release to on-press and sprint related bug fixes

### LINKED ISSUE #276: Sprinting takes precedence over ADS, ADS never prevents you from sprinting
sometimes sprint doesn't come out (I think related to ADS?) - Moffin_dono

### LINKED ISSUE #285: Toggling ADS happens onclick instead of onrelease
Toggle ADS should enable on press of the button, not on release of it (this is how it's like in OGNT) - Wakfu

### LINKED ISSUE #355: Sprint key interrupts aiming for supports
As the support class, whilst being aimed in, pressing the sprint key cancels the aim zoom - even though supports cannot enter into sprint, and hence pressing it should not disrupt their aim.

Likewise, holding the sprint key whilst aiming in causes a buggy jitter where the sprint key is fighting against the zooming motion.

nullsystem:
* Also made sure `neo_aim_hold 1` is not affected negatively by this change
* fixes #285 fixes #276 fixes #355

### DIFF
--- a/mp/src/game/server/neo/neo_player.cpp
+++ b/mp/src/game/server/neo/neo_player.cpp
@@ -1080,6 +1080,7 @@ void CNEO_Player::PostThink(void)
 
 	if (pWep)
 	{
+		const bool clientAimHold = ClientWantsAimHold(this);
 		if (pWep->m_bInReload && !m_bPreviouslyReloading)
 		{
 			Weapon_SetZoom(false);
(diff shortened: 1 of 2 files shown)


Rainyan (dismissed):

mp/src/game/server/neo/neo_player.cpp
```diff
@@ -1096,14 +1097,14 @@ void CNEO_Player::PostThink(void)
 			{
 				static_cast<CWeaponGrenade*>(pNeoWep)->SecondaryAttack();
 			}
-			else if (ClientWantsAimHold(this))
+			else
```
Agiel, line 1100: There's a mismatch between the client and server here. We're not checking if sprint is being held, which means it's possible to start aiming while sprinting.

I'd probably take the time to fix #276 and #355 in the same PR.

Agiel (changes requested):

mp/src/game/server/neo/neo_player.cpp
nullsystem: Added sprint fixes in PR, although I'm not going to change shared code stuff for the time being as that's more of a wider look at c_neo_player/neo_player split than just this.

Rainyan (approved):

Agiel (approved):

PR #354, opened 2024-06-09
Reduce ADS change viewmodel FOV/offsets issue on dedicated servers

### LINKED ISSUE #284: Viewmodels don't move around weirdly on linux clients connected to dedicated servers when ADSing
https://files.catbox.moe/4lf3dq.mp4

*(embedding the video link here into github, in case that link ever goes down, should this issue become relevant at some later point for something -Rain)*

https://github.com/NeotokyoRebuild/neo/assets/6595066/698028e8-7153-4051-93cd-a0d2bbe72ce1

-Wakfu

nullsystem:
* I don't consider this an absolute fix as the problem haven't fully gone away, but rather the issue of it happening reduced.
* Contrary to the reported issue, this happens on both Windows and Linux
* Offsets changes are basically refactors, but FOV changes is greatly simplified and should generally be more robust since it's changed from timing range-based to delta-based
* Speed up the timing on switching ADS by almost half, which would actually make it more similar to OG:NT ADS switching time. This would also make the ADS viewmodel change issue a bit less obvious if it occurs.
* fixes #284


nullsystem:
I'm pulling this to draft for now, going to split the player (not viewmodel) FOV timings to a separate PR as it's not relevant to this specific bug, and also relook at a different strategy as it seemed to make the viewmodel change looked non-linear.

nullsystem:
Im going to close this in favor of a better solution im gonna put as a new PR instead

nullsystem:
This PR is replaced by: https://github.com/NeotokyoRebuild/neo/pull/360

PR #358, opened 2024-06-10
Make zoom speed close to OG:NT

### LINKED ISSUE #357: Changing ADS speed should be close to OG:NT
Currently, the ADS change speed is (proven) slower than in OG:NT. The speed value should be adjusted to be as close to OG:NT, which takes more than half the time less than NT;RE.

nullsystem:
* This equivalates to very nearly 1 second when using host_timescale 0.1
* fixes #357


Rainyan (approved):

Rainyan:
Perhaps add a
```cpp
static_assert(NEO_ZOOM_SPEED != 0, "div by zero");
```
or similar under the var declaration, since we're dividing by the value in many places, so if for whatever reason someone were to zero the value, the bug gets caught at compile time.

Rainyan (approved):

AdamTadeusz (approved):

PR #359, opened 2024-06-10
Fix dead player observing & bot team join logic parity

### LINKED ISSUE #336: Players are not able to spectate a players spectating camera
If a player dies while you are spectating them, you can stay with their POV through to them as an observer. This means you basically follow their POV as they spectate.

Rainyan:
Fix #336 

If spectating a player who has died, and it's been `>=DEATH_ANIMATION_TIME` since their death, look up the next valid target and switch to them.

---

Also sneaking in a `bot_next_team` server cvar, because it was useful for testing this. Set `bot_next_team` to a team index to force the next bot created by `bot_add` to join that specific team. Default value is `-1`, which instead makes the bot join the player team with less players. There's also `-2`, which makes bot join a player team completely at random. Anecdotally, this actually fixes a parity issue with bots - before this change, they were joining player teams completely randomly, but OG NT bots follow the `-1` value behaviour (which is the default now).

---

This PR duplicates some of the base `CBasePlayer` methods for client side access, to be called from the `C_NEO_Player::Calc<...>View` overrides. The client's also now receiving the `m_flDeathTime` prop for calculating the death deltatime client side for observing purposes.

---

### Steps to repro:
* Launch a CTG map
* Spawn some bots in both teams
* Join a team
* End the warmup with `mp_restartgame 1` if needed
* Set `mp_forcecamera 0`
* Suicide and start spectating other players
* Use `ent_fire !picker sethealth 0` to kill the current spec target and see what happens
  * If it targets the wrong client whilst in 1st person spec, go to chase cam for a second to update the picker target

### What happens before this patch:
* Can observe a dead player observing some third player

### What happens after this patch:
* Observing a dead client will automatically find the next valid obs target after `DEATH_ANIMATION_TIME` (3 seconds) since their death

### DIFF
--- a/mp/src/game/client/c_baseplayer.cpp
+++ b/mp/src/game/client/c_baseplayer.cpp
@@ -620,6 +620,109 @@ void C_BasePlayer::SetObserverMode ( int iNewMode )
 	}
 }
 
+int C_BasePlayer::GetNextObserverSearchStartPoint(bool bReverse)
+{
+	int iDir = bReverse ? -1 : 1;
+
+	int startIndex;
+
+	if (m_hObserverTarget)
+	{
+		// start using last followed player
+		startIndex = m_hObserverTarget->entindex();
+	}
+	else
+	{
+		// start using own player index
+		startIndex = this->entindex();
+	}
+
+	startIndex += iDir;
+	if (startIndex > gpGlobals->maxClients)
+		startIndex = 1;
+	else if (startIndex < 1)
+		startIndex = gpGlobals->maxClients;
+
+	return startIndex;
+}
+
+CBaseEntity* C_BasePlayer::FindNextObserverTarget(bool bReverse)
+{
+	int startIndex = GetNextObserverSearchStartPoint(bReverse);
+
+	int	currentIndex = startIndex;
+	int iDir = bReverse ? -1 : 1;
+
+	do
+	{
+		CBaseEntity* nextTarget = UTIL_PlayerByIndex(currentIndex);
+
+		if (IsValidObserverTarget(nextTarget))
+		{
+			return nextTarget;	// found next valid player
+		}
+
+		currentIndex += iDir;
+
+		// Loop through the clients
+		if (currentIndex > gpGlobals->maxClients)
+			currentIndex = 1;
+		else if (currentIndex < 1)
+			currentIndex = gpGlobals->maxClients;
+
+	} while (currentIndex != startIndex);
+
+	return nullptr;
+}
+
+bool C_BasePlayer::IsValidObserverTarget(CBaseEntity* target)
+{
+	if (target == NULL)
+		return false;
+
+	// MOD AUTHORS: Add checks on target here or in derived method
+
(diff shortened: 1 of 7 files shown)


mp/src/game/client/c_baseplayer.cpp
```diff
@@ -620,6 +620,111 @@ void C_BasePlayer::SetObserverMode ( int iNewMode )
...
+
+	// check forcecamera settings for active players
+	if (GetTeamNumber() != TEAM_SPECTATOR)
+	{
+		switch (mp_forcecamera.GetInt())
+		{
+		case OBS_ALLOW_ALL:	break;
+		case OBS_ALLOW_TEAM:	if (GetTeamNumber() != target->GetTeamNumber())
+			return false;
+			break;
```
nullsystem, line 720: Styling nitpick: if and its internal return line looks disjointed

nullsystem (commented):
Tested the PR:

Dead player spec fix, can confirm: "Observing a dead client will automatically find the next valid obs target after DEATH_ANIMATION_TIME (3 seconds) since their death" after going through the steps.

bot_next_team tests report:

bot_next_team -1: Bot joins team with lesser player, stacking up bots is quite equal
bot_next_team -2: Bot joins randomly like previous behavior
bot_next_team 0: Bot joins unassigned
bot_next_team 1: Bot joins spec
bot_next_team 2: Bot joins Jinrai
bot_next_team 3: Bot joins NSF
bot_next_team 4+: Bot joins unknown/unassigned

mp/src/game/client/c_baseplayer.cpp
Rainyan: This function is pulled as-is from Valve's `CBasePlayer::IsValidObserverTarget` server implementation so that we have client side access, but I can fix the formatting for the switch.

nullsystem (dismissed):

mp/src/game/server/neo/neo_player.cpp
```diff
@@ -2331,14 +2333,44 @@ bool CNEO_Player::ProcessTeamSwitchRequest(int iTeam)
...
+			else if (numNsf < numJin) {
+				iTeam = TEAM_NSF;
+				break;
+			}
+			[[fallthrough]];
+		case JoinMode::Random:
+			iTeam = RandomInt(TEAM_JINRAI, TEAM_NSF);
+			break;
+		default:
+			iTeam = joinMode.GetInt();
```
AdamTadeusz, line 2372: Would it be worth defining an enum TOTAL_PLAYABLE_TEAMS or something and defaulting to the random behaviour if iTeam >= TOTAL_PLAYABLE_TEAMS ? Otherwise we hit a few asserts when passing a value greater than TEAM_NSF and the bots remain unassigned.

Rainyan: Right, I didn't want to hard code a team limit here, however unlikely it is we'd ever have more. I've clamped the value between playable teams now.

nullsystem (approved):

PR #361, opened 2024-06-11
Update .gitignore

Rainyan:
Ignore Qt Creator build files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
 # Files to ignore when considering what GIT should commit.
 
+# Qt Creator
+mp/build-src-*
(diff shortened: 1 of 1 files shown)


nullsystem:
I think that's quite dependent on setup, like mine Qt-Creator is set on: `mp/src/build/debug`

Rainyan:
This is what my old Qt Creator creates by default, and I'd just like to get it the gitignore so I don't accidentally commit those files one of these days.

nullsystem (approved):

AdamTadeusz (approved):

PR #368, opened 2024-06-15
Replace hud_reloadscheme hack with EditablePanel + LoadControlSettings, indirectly fixes MOTD title + layout

### LINKED ISSUE #367: neo_hud health/compass/ammo need to change to EditablePanel
Currently, on initialization, `hud_reloadscheme` is a forced to call by both the health and ammo huds. The main intention of these is so the customization from `scripts/HudLayout.res` gets applied, however these panels should really instead be using `EditablePanel` as the inherited panel class instead of just `Panel` as `EditablePanel` provides the utility to apply those res settings instead. There are visible issue that came around from this, mostly the MOTD panel resetting the data and making look like the MOTD didn't get applied when its really just resetted after it applied.

nullsystem:
* Both ammo and health-aux huds had a hack which forces hud_reloadscheme to be called on initialization. However this has visibly caused an issue with the MOTD which in Valve's code resets the data and look a bit odd. This also enforced reload on all huds which is unnecessary.
* The proper way to load those HudLayout.res settings is instead to have those inherit EditablePanel which most importantly provides a method `LoadControlSettings` to properly load those settings instead.
* This also refactor the font applying to ApplySchemeSettings
* Minor edit in HudLayout.res chat box is intentional otherwise there's a permanent background with using this EditablePanel change.
* fixes #367

## Testing
* What to change around:
    * Change/add around values in: `scripts/HudLayout.res`. Minimal example to differentiate from defaults. Compass +20 instead of +3, Ammo read bg, Health bar RGB, 255-0-255:
        ```
        NHudCompass
        {
                "fieldName"             "NHudCompass"
                "y_bottom_pos"          "20"
        }
        NHudWeapon
        {
                "fieldName"             "NHudWeapon"
                "box_color_r"           "254"
        }
        NHudHealth
        {
                "fieldName"             "NHudHealth"
                "health_color_g"        "0"
        }
        ```
* What needed to be checked:
    * First time loading in, make sure HudLayout.res setting is applied
    * `hud_reloadscheme` works as usual
    * N-times loading in, HudLayout.res setting is applied
    * Without PR #366 the MOTD retains its layout and server title. It doesn't display something in the central area but that's part of that PR's change. With PR #366 it should look functional.

### DIFF
--- a/mp/src/game/client/neo/ui/neo_hud_ammo.cpp
+++ b/mp/src/game/client/neo/ui/neo_hud_ammo.cpp
@@ -43,7 +43,7 @@ ConVar neo_cl_hud_debug_ammo_color_a("neo_cl_hud_debug_ammo_color_a", "255", FCV
 	"Alpha color value of the ammo, in range 0 - 255.", true, 0.0f, true, 255.0f);
 
 CNEOHud_Ammo::CNEOHud_Ammo(const char* pElementName, vgui::Panel* parent)
-	: CHudElement(pElementName), Panel(parent, pElementName)
+	: CHudElement(pElementName), EditablePanel(parent, pElementName)
 {
 	SetAutoDelete(true);
 
@@ -60,23 +60,9 @@ CNEOHud_Ammo::CNEOHud_Ammo(const char* pElementName, vgui::Panel* parent)
 		SetParent(g_pClientMode->GetViewport());
 	}
 
-	// NEO HACK (Rain): this is kind of awkward, we should get the handle on ApplySchemeSettings
-	vgui::IScheme* scheme = vgui::scheme()->GetIScheme(neoscheme);
-	if (!scheme) {
-		Assert(scheme);
-		Error("CNEOHud_Ammo: Failed to load neoscheme\n");
-	}
-
-	m_hSmallTextFont = scheme->GetFont("NHudOCRSmall");
-	m_hBulletFont = scheme->GetFont("NHudBullets");
-	m_hTextFont = scheme->GetFont("NHudOCR");
-
-	InvalidateLayout();
-
 	SetVisible(neo_cl_hud_ammo_enabled.GetBool());
 
 	SetHiddenBits(HIDEHUD_HEALTH | HIDEHUD_PLAYERDEAD | HIDEHUD_NEEDSUIT | HIDEHUD_WEAPONSELECTION);
-	engine->ClientCmd("hud_reloadscheme"); //NEO FIXME (Adam) this reloads the scheme of all elements, is there a way to do it just for this one?
 }
 
 void CNEOHud_Ammo::Paint()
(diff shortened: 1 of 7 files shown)


mp/src/game/client/neo/ui/neo_hud_ammo.cpp
```diff
@@ -101,6 +87,14 @@ void CNEOHud_Ammo::ApplySchemeSettings(vgui::IScheme* pScheme)
 {
 	BaseClass::ApplySchemeSettings(pScheme);
 
+	LoadControlSettings("scripts/HudLayout.res");
+
+	m_hSmallTextFont = pScheme->GetFont("NHudOCRSmall");
+	m_hBulletFont = pScheme->GetFont("NHudBullets");
+	m_hTextFont = pScheme->GetFont("NHudOCR");
+
+	InvalidateLayout();
```
Rainyan, line 96: Nit: `LoadControlSettings` already calls `InvalidateLayout`, so we can probably omit this.

mp/src/game/client/neo/ui/neo_hud_health_thermoptic_aux.cpp
```diff
@@ -89,6 +76,13 @@ void CNEOHud_HTA::ApplySchemeSettings(vgui::IScheme* pScheme)
 {
 	BaseClass::ApplySchemeSettings(pScheme);
 
+	LoadControlSettings("scripts/HudLayout.res");
+
+	m_hFont = pScheme->GetFont("NHudOCRSmall");
+	m_hFontBuildInfo = pScheme->GetFont("Default");
+
+	InvalidateLayout();
```
Rainyan, line 84: Ditto

Rainyan (commented):
So, the MOTD layout indeed seems fixed by this PR:



However, after using `hud_reloadscheme`, the layout is broken again upon next server join:



Is this MOTD business outside the scope of this PR?

Rainyan:
Also seems we've a merge conflict to be resolved.

nullsystem:
@Rainyan Actually, only the server title/contents being intact being the main fix of this regarding MOTD. So I shouldn't really have mentioned the layout stuff. Technically though, the 2nd picture's "broken" layout is the "actual" res layout of the MOTD straight from OG:NT's TextWindow.res file I think. The fix is probably more like define our own one that doesn't look so bar/closer to HL2's one.

Rainyan (commented):
One more thing I noticed - we're getting a new console error:
```
Resize parent (panel(HudChat) -> parent(NHudWeapon)) not sized yet!!!
```
when joining the server.

This wasn't the case in master branch (d90900fa9c0e22799de3749509b0bce228fd579c as of writing this).

`NHudWeapon` is:

```cpp
#define UI_ELEMENT_NAME_AMMO "NHudWeapon"
```
so I'm guessing this is related to changes in `CNEOHud_Ammo`.

To reproduce:
* Join server
* Press escape when loading in the game completes
* See error:

huderr

nullsystem:
@Rainyan Try out with the latest commit in, I think it's just moving around SetScreenSize/SetBounds lines. Don't think I see the warning after the commit but have a check.

Rainyan (approved):

AdamTadeusz (approved):

PR #370, opened 2024-06-15
Motion Vision Working

VantoNortim:
Now overrides cloak
Change material shader to ignore lighting and stop it from flickering

Still missing: 
Shader with gradient effect
Items/props having MV applied

### DIFF
--- a/mp/src/game/client/neo/c_neo_player.cpp
+++ b/mp/src/game/client/neo/c_neo_player.cpp
@@ -540,33 +540,15 @@ int C_NEO_Player::GetAttackerHits(const int attackerIdx) const
 	return m_rfAttackersHits.Get(attackerIdx);
 }
 
-int C_NEO_Player::DrawModel( int flags )
+int C_NEO_Player::DrawModel(int flags)
 {
-	// Do cloak if cloaked
-	if (IsCloaked())
-	{
-		IMaterial *pass = materials->FindMaterial("dev/toc_cloakpass", TEXTURE_GROUP_CLIENT_EFFECTS);
-		Assert(pass && !pass->IsErrorMaterial());
-
-		if (pass && !pass->IsErrorMaterial())
-		{
-			//const int extraFlags = STUDIO_RENDER | STUDIO_TRANSPARENCY | STUDIO_NOSHADOWS | STUDIO_DRAWTRANSLUCENTSUBMODELS;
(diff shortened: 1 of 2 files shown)


mp/src/game/client/neo/c_neo_player.cpp
```diff
@@ -540,33 +540,15 @@ int C_NEO_Player::GetAttackerHits(const int attackerIdx) const
...
-			//const int extraFlags = STUDIO_RENDER | STUDIO_TRANSPARENCY | STUDIO_NOSHADOWS | STUDIO_DRAWTRANSLUCENTSUBMODELS;
-			modelrender->ForcedMaterialOverride(pass);
-			const int ret = BaseClass::DrawModel(flags /*| extraFlags*/);
-			modelrender->ForcedMaterialOverride(NULL);
-
-			return ret;
-		}
-	}
-
 	int ret = BaseClass::DrawModel(flags);
```
Rainyan, line 545: If BaseClass is not ready to draw, we should probably return early here:
```cpp
if (!ret)
{
	return ret;
}
```
etc

mp/src/game/client/neo/c_neo_player.cpp
```diff
@@ -540,33 +540,15 @@ int C_NEO_Player::GetAttackerHits(const int attackerIdx) const
...
 
 	auto pLocalPlayer = C_NEO_Player::GetLocalNEOPlayer();
 	if (pLocalPlayer && pLocalPlayer->IsInVision())
 	{
-#define SPEED_BETWEEN_WALK_AND_RUN ((NEO_ASSAULT_WALK_SPEED + NEO_ASSAULT_NORM_SPEED) / 2.0)
-		if ((pLocalPlayer->GetClass() == NEO_CLASS_ASSAULT) &&
-			(GetAbsVelocity().Length() >= SPEED_BETWEEN_WALK_AND_RUN))
+		auto vel = GetAbsVelocity().Length2D();
```
Rainyan, line 550: Use `Length` instead of `Length2D`, so we also catch vertical motion for parity.

Rainyan (changes requested):

Rainyan:
> Still missing:
> Shader with gradient effect

I could look into this, and replace the current experimental MV with something more in line with parity.

Rainyan (approved):

nullsystem (approved):

PR #380, opened 2024-06-17
Fadeout after death + first person head ragdoll death cam

### LINKED ISSUE #143: fadeout plays after death

nullsystem:
* 8s -> fade 2s stays -> 2s unfade
* Effect local client/player after death 1st tick only
* Death goes to death cam, first person by the head ragdoll
* Some stuff to figure out when to force unfade and un-deathcam
* fixes #143

### DIFF
--- a/mp/src/game/client/neo/c_neo_player.cpp
+++ b/mp/src/game/client/neo/c_neo_player.cpp
@@ -997,12 +997,15 @@ void C_NEO_Player::PostThink(void)
 			Weapon_SetZoom(false);
 			m_bInVision = false;
 
-			if (this == GetLocalNEOPlayer())
+			if (IsLocalPlayer() && (GetTeamNumber() == TEAM_JINRAI || GetTeamNumber() == TEAM_NSF))
 			{
+				m_iSavedObserverMode = GetObserverMode();
+				SetObserverMode(OBS_MODE_DEATHCAM);
+				// Fade out 8s to blackout + 2s full blackout
 				ScreenFade_t sfade{
-					.duration = static_cast<unsigned short>(static_cast<float>(1<<SCREENFADE_FRACBITS) * 3.0f),
-					.holdTime = static_cast<unsigned short>(static_cast<float>(1<<SCREENFADE_FRACBITS) * 1.0f),
-					.fadeFlags = FFADE_OUT,
+					.duration = static_cast<unsigned short>(static_cast<float>(1<<SCREENFADE_FRACBITS) * 8.0f),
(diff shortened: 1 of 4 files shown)


mp/src/game/client/neo/c_neo_player.cpp
Rainyan, line 1: There's a bug with this fade triggering on initial team join.

We'll also need an un-fade using `FFADE_IN|FFADE_PURGE`, for parity.

Here's a very rough annotated estimation of OG NT fade timings:

[fade_annotated.webm](https://github.com/NeotokyoRebuild/neo/assets/6595066/9a00a95d-a496-4e48-9d6e-fba6fddc7481)

There does seem to be distinct long fade-out, and some sustain (probably the `duration` and `holdTime`), followed by a fade-in that purges the fade-out.

Perhaps we could do something like `SetObserverMode(OBS_MODE_DEATHCAM)` at `m_bFirstDeathTick` or wherever, and then override `C_BasePlayer::CalcDeathCamView` for handling that death camera & fade, using deltatime from `GetDeathTime()`.

We'd also wanna keep the camera fixed to the ragdoll(?) head position for parity, but that's kind of unrelated to this PR.

And once the fade sequence is completed, `SetObserverMode` back to whatever the user had selected before the deathcam sequence, so they can start observing as usual (and we stop calling `::CalcDeathCamView`).

Rainyan (changes requested):

mp/src/game/client/neo/c_neo_player.cpp
nullsystem: @Rainyan Made the fade like OG:NT now and set the death cam at head position

Rainyan (approved):
Looks good.

One more thing I realised is that in OG NT, you can't start speccing your teammates until after the fadeout+hold periods (8+2 secs) - until then, pressing `+jump` or `+attack` does nothing. But in the rebuild, you can press `+jump` after dying and instantly jump to spectating. This might theoretically have some gameplay implications in competitive matches context etc.

Up to you whether you're interested in implementing it here or leave it for another day (in which case we probably want to open a bug for tracking it).

nullsystem:
@Rainyan I've implemented it in here. The button blocking is done similar to how freeze time does it, so there's server-side section this time. If statement made fairly restricted to only round live as: a) doesn't affect implication outside of live rounds, and b) would need more loaded to ensure it doesn't affect other situation if not simply restricted to live.

Rainyan (approved):

AdamTadeusz (approved):

PR #381, opened 2024-06-19
Client side MOTD control

Rainyan:
## Description
This is a complementary PR for #366 and provides client-side control for MOTD rendering.

It also defaults the MOTD to text-only mode, so we're not by default vulnerable to attacks against the HTML renderer, and adds client side control of dismissed MOTD unloading.

Changes:
* Add `cl_disablehtmlmotd` value `2` for completely disabling the MOTD client side.
* Set default of `cl_disablehtmlmotd` to `1` instead of `0`.
* Add `cl_motd_unload_on_dismissal` which allows clients to override a `sv_motd_unload_on_dismissal 0` to force the unload.
* Set default of `sv_motd_unload_on_dismissal ` to `1` instead of `0`.

## Toolchain

- Windows MSVC VS2022

## Linked Issues

- related #366

### DIFF
--- a/mp/src/game/client/clientmode_shared.h
+++ b/mp/src/game/client/clientmode_shared.h
@@ -16,6 +16,12 @@
 #include "GameEventListener.h"
 #include <baseviewport.h>
 
+#ifdef NEO
+#include "vguitextwindow.h"
+
+extern ConVar cl_disablehtmlmotd;
+#endif
+
 class CBaseHudChat;
 class CBaseHudWeaponSelection;
 class CViewSetup;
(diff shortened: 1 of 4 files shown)


nullsystem (approved):
Tested with merger of other 2 branches (368 + 366), server enabled MOTD.
`cl_disablehtmlmotd 0` - Shows HTML/Website
`cl_disablehtmlmotd 1` - Shows text-only MOTD
`cl_disablehtmlmotd 2` - Doesn't show MOTD

AdamTadeusz (approved):

PR #390, opened 2024-06-22
Default to not relay player's neo_fov to 1st person specator, add ADS FOV transition for spec, use script AimFov

### LINKED ISSUE #389: [BUG] neo_fov by spectating player is relayed to spectator by default
### Build Info

20240622_f670f82 (master-f670f828687a22886a4d6b89fc09fcaea9b23e70)

### Description

Currently, `neo_fov` value is done per-player with their FOV values broadcast to other spectators and not only local client. This means if the player has a `neo_fov` value of 110, the person spectating the player will also see that player's `neo_fov` of 110, and not their own `neo_fov` value.

EDIT: It been decided instead to keep the behavior, but have this off by default

### To Reproduce

1. Have two clients/computers running, one run the server and join the server having two clients in it
2. One player join spec, the other join a team
3. The player in a team change `neo_fov` values while the spectator is spectating the player

### Expected behavior

The local client's/spectator's `neo_fov` value is used for FOV.

### Actual behavior

The player's it's spectating `neo_fov` value is used for FOV instead.

### Operating System

- [X] Windows
- [X] Linux

### Version/Distro

Arch Linux; Windows 10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

nullsystem:
## Description

* `neo_fov` ConVar fixed, now set to only define client-side and also used as such instead of the weird server->client routing before. However the server-side will still need to utilize it for server-player and for relay-mode `neo_fov_relay_spec 1`
* `neo_fov_relay_spec` new ConVar, defaults to disabled which uses client's `neo_fov` for 1st person spec. If enabled, relays spectated player's `neo_fov` to whoever spectating the player in 1st person
* Spectator looking at a player switching aim/unaim will now get a gradual rather than instant FOV transition. This transition is done differently by percentage of time verses delta used on player's main FOV, however it works well enough to show the spectator a gradual effect.
* Use AimFov for ADS FOV, fix some missing AimFov in scripts
* `neo_fov` min FOV changed from 60 to 75, default to 75

## Toolchain

## Linked Issues

- fixes #389


Rainyan (changes requested):
After this PR, it seems `neo_fov` no longer works reliably for changing your own FOV.

If you try changing `neo_fov` value between 90 and 110 a few times whilst alive, you'll notice it sometimes doesn't update. Tapping the sprint key seems to make it update again however.

nullsystem:
@Rainyan I don't think it was consistent before either, but anyway it should be now be immediate when neo_fov setting gets applied.

Rainyan:
> @Rainyan I don't think it was consistent before either, but anyway it should be now be immediate when neo_fov setting gets applied.

This fixes the FOV change not being applied. But it seems our aim FOV is now somehow coupled to the base FOV. For example, if you set `neo_fov 75` and try aiming, it pulls the viewmodel in but your actual FOV doesn't change at all, i.e. there is no "zooming" occurring. I'm assuming we'd want the zoom FOV change to be the same ratio as it is in original NT, decoupled from actual FOV.

nullsystem:
@Rainyan Whoops look like `GetMinFOV` shouldn't been altered. Try it now.

Rainyan (dismissed):

Rainyan (approved):

AdamTadeusz (approved):

PR #391, opened 2024-06-22
Improve lean prediction

Rainyan:
## Description
* Fix `m_bInLean` not being client side predicted
  * You can use `cl_pdump 1` to verify this - `m_bInLean` should no longer flash red whilst leaning
* Improve the lean lerp prediction

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/mp/src/game/client/neo/c_neo_player.cpp
+++ b/mp/src/game/client/neo/c_neo_player.cpp
@@ -512,33 +512,34 @@ void C_NEO_Player::CheckVisionButtons()
 
 void C_NEO_Player::CheckLeanButtons()
 {
-	if (IsAlive())
+	if (!IsAlive())
 	{
-		if (ClientWantsLeanToggle(this))
+		return;
+	}
+
+	if (ClientWantsLeanToggle(this))
+	{
+		if (m_afButtonPressed & IN_LEAN_LEFT)
 		{
-			if (m_afButtonPressed & IN_LEAN_LEFT)
-			{
-				if (m_bInLean == NEO_LEAN_LEFT) m_bInLean = NEO_LEAN_NONE;
-				else m_bInLean = NEO_LEAN_LEFT;
-			}
-			if (m_afButtonPressed & IN_LEAN_RIGHT)
-			{
-				if (m_bInLean == NEO_LEAN_RIGHT) m_bInLean = NEO_LEAN_NONE;
-				else m_bInLean = NEO_LEAN_RIGHT;
-			}
+			if (m_bInLean == NEO_LEAN_LEFT) m_bInLean = NEO_LEAN_NONE;
+			else m_bInLean = NEO_LEAN_LEFT;
 		}
-		else
+		if (m_afButtonPressed & IN_LEAN_RIGHT)
 		{
-			m_bInLean = NEO_LEAN_NONE;
-			if ((m_nButtons & IN_LEAN_LEFT) && !(m_nButtons & IN_LEAN_RIGHT))
-			{
-				m_bInLean = NEO_LEAN_LEFT;
-			}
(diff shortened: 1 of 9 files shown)


mp/src/game/client/neo/c_neo_player.cpp
```diff
@@ -508,6 +508,38 @@ void C_NEO_Player::CheckVisionButtons()
...
+			{
+				m_bInLean = NEO_LEAN_LEFT;
+			}
+			else if ((m_nButtons & IN_LEAN_RIGHT) && !(m_nButtons & IN_LEAN_LEFT))
+			{
+				m_bInLean = NEO_LEAN_RIGHT;
+			}
+		}
+	}
+}
```
nullsystem, line 541: Adding this to client player however, for dedicated server, caused violent shaking when leaning using with `neo_lean_toggle 1` and more drastic shaking during changing leaning for hold toggle also.

nullsystem (changes requested):

mp/src/game/client/neo/c_neo_player.cpp
Rainyan: @nullsystem I think it was my changes in the predicted viewmodel that broke it actually. Could you try again now? We've still got some jitter in the listen server loopback lean, but at least the lean inputs should be predicted properly for online/SRCDS now.

nullsystem: @Rainyan Still have the same shaking especially with `neo_lean_toggle 1` on dedicated, and now loopback doesn't even lean properly anymore?

Rainyan: I've converted the toggle lean into binds, so we don't have to try and predict based on it, or some previous buttons state.

> now loopback doesn't even lean properly anymore (both neo_lean_toggle modes)?

Hmm, unable to reproduce this on Windows at least.

nullsystem: Loopback still broken for me, reproduced it both on Linux and Windows

nullsystem: @Rainyan https://streamable.com/s2chfe
It looks to me that part that differentiates between loopback vs dedicated seems to be needed?
But then reverting that ends up making this PR more like changing the toggle lean from convar to keybind instead.

Rainyan: @nullsystem Looks like it's dependent on your rates, I was testing with the default `cl_cmdrate 30` and `cl_updaterate 20`. But I can indeed reproduce that bug with both at 66. Back to the drawing board.

But considering this is what the current master branch looks like to me on listen server, I think it needs some fixing in any case:

[master_8d02776.webm](https://github.com/NeotokyoRebuild/neo/assets/6595066/1aad834d-6f8a-453d-99d3-ff77ab9679f8)

nullsystem: Yeah the jittery lean is certainly an issue would want to fix. For me it's much more noticeable on dedicated but still there on loopback.

Rainyan: @nullsystem Pushed some changes, could you test again when you get the chance

nullsystem: @Rainyan Yep, this is definitely an improvement, the lean looks fine now + even better.
Tested: Linux loopback, Windows loopback, Windows dedicated
Both hold and toggle lean confirmed working and no more jittery motions at all even in dedicated server.

nullsystem (approved):

Rainyan:
Had to rebase to fix a merge conflict.

nullsystem (approved):

AdamTadeusz (approved):

PR #402, opened 2024-06-26
Scoreboard - Get_Score -> GetRoundsWon

nullsystem:
## Description

* Otherwise you'll get a "score" that's based on kills, not team round wins
* Long term solution could to change how that score works but for now this is fine

## Toolchain

### DIFF
--- a/mp/src/game/client/neo/ui/neo_scoreboard.cpp
+++ b/mp/src/game/client/neo/ui/neo_scoreboard.cpp
@@ -408,7 +408,7 @@ void CNEOScoreBoard::UpdateTeamInfo()
 
 				if (i != TEAM_SPECTATOR)
 				{
-					V_snwprintf(val, ARRAYSIZE(val), L"%ls: %d", teamName, team->Get_Score());
+					V_snwprintf(val, ARRAYSIZE(val), L"%ls: %d", teamName, team->GetRoundsWon());
 					pPlayerList->ModifyColumn(sectionID, "ping", val);
 					V_snwprintf(string1, ARRAYSIZE(string1), L"Players: %ls", wNumPlayers);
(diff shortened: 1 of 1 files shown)


Rainyan (approved):

Agiel:
Isn't it better to change how Get_Score works? I kinda assumed it already did the right thing. I mean if we're going to have a switch on game mode and everything I think it's the wrong place to put it in the scoreboard class.

Rainyan:
@Agiel As these are declared together in the base class:
```cpp
	virtual int		Get_Score( void );
	virtual int		Get_Deaths( void );
	virtual int		Get_Ping( void );
```
I'm assuming "score" here is basically a misnomer for "frags"/"XP"/whatever yardstick of individual player performance, tallied up for that whole team.

But I guess we could override the base team, and check NEO rules for whether it's CTG, etc etc, if we wanted to do this "proper".

Although then we'd have both `GetRoundsWon` and a virtual overridden `Get_Score` doing the same thing for that class.

AdamTadeusz:
The server side Team class doesn't seem to offer that much help either,
```
//------------------------------------------------------------------------------------------------------------------
// SCORING
// Purpose: Add / Remove score for this team
void CTeam::AddScore( int iScore )
{
	m_iScore += iScore;
}

void CTeam::SetScore( int iScore )
{
	m_iScore = iScore;
}
```
Maybe just include a NEO HACK comment so we can find this easily later?

nullsystem:
I think rain's suggestion of something like a `CNEOTeam` that can deal with more gamemodes would be the way to go. But for now `GetRoundsWon` will do, which is already utilized in multiple places anyway.

AdamTadeusz (approved):

nullsystem:
Superseded by #410

PR #407, opened 2024-06-28
Return music and background to original

AdamTadeusz:
There is an issue with the intro being the default toilet image at 4:3 aspect ratios, can be fixed by including the files background01.vtf and background01.vmt in this repository instead of relying on them being mounted from the original game. Notice how this behaviour is also present in the original where at 4:3 aspect ratio the intro is the background image, but at 16:9 the intro is black.


AdamTadeusz:
step one of using a tool like bfg to clean git history of a file is making sure that file is not on the head of the main branch

Rainyan (approved):

nullsystem (approved):

PR #409, opened 2024-06-28
Update the bug report template

blaberry:
rephrased some stuff and changed the label used for reports to an original one.

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #
- related #

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,7 +1,7 @@
 name: Bug report
 description: General bugs, program crashes, broken files, wrong values, etc...
-title: "[BUG] "
-labels: bug
+title: ""
+labels: "Bug Report"
 projects: ["NeotokyoRebuild/1"]
 body:
   - type: markdown
@@ -14,8 +14,10 @@ body:
         **DO NOT** paste chatlogs. Formulate the bug into a detailed
         description and reproducible steps using the form below.
 
-        Fill in the title after the **[BUG]** tag and make sure it's
-        in the **"current state of things"** wording.
+        Make sure the title describes the ideal state of the world
+        (ie. if the bug is that recons can see through walls when
+        using their NVGs, the title should be "recons can't see
+        through walls with NVGs").
   - type: input
     id: build
     attributes:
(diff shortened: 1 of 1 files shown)


nullsystem:
I still object with the requirement of "desired state" in a bug report title. From the perspective of an outsider who just comes here and files a bug report, it's typical for the title to be written as a "the current state" like how other projects allows/typical done (can just pick out chromium, firefox, etc...) and this "Make sure..." text could even easily be glossed over.

Rainyan (dismissed):

blaberry:
I'm not sure I understand the second approach you recommend, does "use either should or should not phrasing" not fulfill "have a guideline for both style of title"? @nullsystem

nullsystem:
@blaberry It's about "desired" vs "current" state of title phrasing

blaberry:
Right, then we run into the issue we discussed at length over chat, where some items are desired state, some are not, and the mishmash of both will most likely turn out much harder to read than what's on offer in this PR

nullsystem:
@blaberry There's an important distinction between a bug, change request (opinionated, current state isn't a bug), and feature request. There shouldn't be a blanket all of those have same "desired" state titles, and what on offer currently also gives a clearer distinction with the "[BUG] " title prefix even. The "desired" state titles are natural gravitation for change/feature requests, but as I've put it, the "current" state titles are natural gravitation for bug reports. It's not really a case of random some are/some are not thing, it's that there are distinction that issues titles written for bugs are done differently to change/feature requests.

blaberry:
Alright so two different topics:

using a tag in the title: labels in Github are meant to categorize issues for easier filtering and searching. "[BUG]" in the title achieves a similar purpose, but it clutters the title and reduces readability. I added a new label to make this clearer. We can also add a label for FRs.

title formatting: I'll revert the title format instructions to how it was before since we can't find consensus on this.

nullsystem (changes requested):
> There is a problem with this template
> 
> title must be of type String and cannot be empty.

https://github.com/NeotokyoRebuild/neo/blob/8844696f3dcda86a6ecea68444d759511c8ccdcd/.github/ISSUE_TEMPLATE/bug_report.yml

Removal of `[BUG]` feels like less clear on separate of bugs vs others but whatever.

nullsystem (approved):

Rainyan (approved):

PR #410, opened 2024-06-28
Make scoreboard avatars optional.

Agiel:
## Description

* Add new cvar neo_show_scoreboard_avatars (default: 1)
* Add map name to top right corner of scoreboard.
* Make live player background slightly dimmer
* Make ping text slightly dimmer so it doesn't blend together with the names when avatar isn't showing.
* Minor tweaks to column widths.

With avatars:
image

Without:
image

## Toolchain

- Windows MSVC VS2022


Agiel:
This includes #402 btw, but if that gets merged first I'll just fix the conflict.

Rainyan (approved):

nullsystem (approved):

Issue #414, opened 2024-06-29
[BUG] Hud stuck at resolution change and map change


nullsystem:
### Build Info

20240629_321f139

### Description

Hud can get stucked at the initial resolution when changing resolution, and hud can get stuck at its previous state, not being refreshed on map change

### To Reproduce

Resolution bug:
1. Join local game
2. ESC -> Settings change resolution
3. See the ammo + health position not updated properly

Map refresh:
1. Join local game
2. Change map
3. See hud stuck and not updating

### Expected behavior

Hud updating properly, position and states

### Actual behavior

Hud not updating properly, including position and states

### Operating System

- [X] Windows
- [X] Linux

### Version/Distro

Linux + Windows 10/11

PR #406, opened 2024-06-27
Round state HUD - Refactor, fix res change, fix round 0 strings

nullsystem:
## Description

* Off-center fixed by moving the variable settings over to `ApplySchemeSettings` (do note the diff looks a bit crap atm so view as the whole file)
* Also fix when supposed to not have visible strings it wasn't cleared, but now will explicitly clear them all properly and won't look stuck on map change
* Refactor
    * Common variables like stars and class are array-based
        * Refactored initialization and fetching of data
    * Reduce amount of variables holding position state
        * Removed `m_iYPos - 2` calculations which always comes down to 0 anyway
        * Renamed, changed types, removed variable same sizes of the other one

## Toolchain

- Linux GCC Distro Native Arch/GCC 14

## Linked Issues

- fixes #405

PR #439, opened 2024-06-30
Partial fix HUD - fix timer and fix hud name

nullsystem:
## Description

* This is a **partial** fix PR not full as other places/PRs will be done to eventually piece up to fix HUDs eventually
* The hud names are all wrong due to macro trip-ups, they're now renamed to the actual one
* During client-side spawning, reset the timer to 0 and slightly tweak the conditionals so it works properly on next map again.
* #414 will be marked as related not fixed as this is a partial PR

## Toolchain

- Windows MSVC VS2022

## Linked Issues

- related #414

### DIFF
--- a/mp/src/game/client/hl2mp/hl2mp_hud_target_id.cpp
+++ b/mp/src/game/client/hl2mp/hl2mp_hud_target_id.cpp
@@ -13,7 +13,7 @@
 using namespace vgui;
 
 #ifdef NEO
-DECLARE_NAMED_HUDELEMENT(CTargetID, UI_ELEMENT_TARGET_ID);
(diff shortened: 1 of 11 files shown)


AdamTadeusz:
image
On a listen server the score hud is still frozen before the round starts. It does fix itself once the game starts however

Rainyan (approved):

nullsystem:
> image On a listen server the score hud is still frozen before the round starts. It does fix itself once the game starts however

@AdamTadeusz I'm going to fully fix that as part of #406 . That's why I named this PR partial.

AdamTadeusz (approved):

PR #452, opened 2024-06-30
Fix resolution HUD update - All proportional_xypos need to be defined in HudLayout

nullsystem:
## Description

Without these definitions, updating the screen resolution is stuck with the default value set in the source which is never adjusted for therefore stuck on the initial screen res proportional sizing. This just defines the default values set for proportional_xypos values so that it can be recognized properly.

## Toolchain

- Windows MSVC VS2022

## Linked Issues

- related #414


AdamTadeusz (approved):
Might be a good idea to just include all the variables read from hudlayoutscheme into hudlayoutscheme so players know what can be adjusted and how

Rainyan (approved):

nullsystem:
> Might be a good idea to just include all the variables read from hudlayoutscheme into hudlayoutscheme so players know what can be adjusted and how

I'm gonna do something about the color variables so will do that as a separate PR.

PR #457, opened 2024-06-30
Add GitHub Actions workflow

Masterkatze:
## Description
Add simple GitHub Actions workflow with building for the Windows/Ubuntu operating system.

## Toolchain
- Linux GCC Archlinux Native (g++ (GCC) 14.1.1 20240522)


Masterkatze:
> wouldnt it be possible to use the steamrt3 docker container for building?

AFAIK it is possible, I just want to test this simpler approach first.

mp/src/cmake/build_info.cmake
```diff
@@ -1,31 +1,37 @@
...
 )
+string(SUBSTRING "${GIT_HASH}" 0 7 GIT_HASH)
+
 execute_process(
     COMMAND git log -1 --format=%H
     WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
     OUTPUT_VARIABLE GIT_LONGHASH
     OUTPUT_STRIP_TRAILING_WHITESPACE
     ERROR_QUIET
 )
```
nullsystem, line 18: I've realized something, `%h` short hash fetch wasn't necessary, could just do `string(SUBSTRING "${GIT_LONGHASH}" 0 7 GIT_HASH)` and not be affected by `core.abbrev` git setting

Masterkatze: Updated

PR #463, opened 2024-07-01
WIP: Movement speeds and footsteps at parity

Rainyan:
Replaces #373


Agiel:
Putting this here so it doesn't get lost in discord chat (values from disassembled server.dll):

base speed is 200

then a class modifier
recon: 1
assault: 0.8
support: 0.8
(you have to drop everything including the knife to get the full speed)

there's also different sprint modifiers
recon: 1.35
assault: 1.6
support: 1.35 (but they have the can sprint boolean set to false)

aim modifier is 0.6
crouch and walk modifiers are both 0.75

then there's the funky bit at the end where if the speed is below 55 it gets capped to 75

Rainyan:
Closing this for now so I can properly rebase by fork.

PR #465, opened 2024-07-02
Enable hot reload on windows

AdamTadeusz:


Rainyan (dismissed):
Confirmed working for latest MSVC, although I imagine the more idiomatic way to accomplish this would be with https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_DEBUG_INFORMATION_FORMAT.html (which we do have in the CMake file, but it alone is not working for whatever reason)

AdamTadeusz:
I just had a look at files changed and it seems that I somehow managed to make a mistake even though I'm only adding two lines ♿, even though the end result seems to be the same. @Rainyan do you think I should change the end of the first line to "${CMAKE_CXX_FLAGS_DEBUG}"), even though it works regardless of whether the bit at the end has or doesn't have DEBUG?

Rainyan:
> I just had a look at files changed and it seems that I somehow managed to make a mistake even though I'm only adding two lines ♿, even though the end result seems to be the same. @Rainyan do you think I should change the end of the first line to "${CMAKE_CXX_FLAGS_DEBUG}"), even though it works regardless of whether the bit at the end has or doesn't have DEBUG?

I'd have to spend some quality time reading CMake docs to answer that. Perhaps @Masterkatze would know better?

Masterkatze:
We need to set `cmake_minimum_required(VERSION 3.25)` for this variable to work. If everyone is fine using more modern version instead of 3.18 then this is preferable way of enabling hot reload.

Rainyan:
> We need to set `cmake_minimum_required(VERSION 3.25)` for this variable to work. If everyone is fine using more modern version instead of 3.18 then this is preferable way of enabling hot reload.

I would be ok with this min version bump.

Rainyan:
> Although this is a Windows-specific PR, it does seem Linux SteamRT3 sniper container and Debian 12 are 3.25 also so I guess it's fine.

If CMake offers a non-complicated way to set platform-specific min CMake version, that could also be an option?

Masterkatze:
I think it's better to add `cmake_policy(SET CMP0141 NEW)` before project call (at line 4) and use `CMAKE_MSVC_DEBUG_INFORMATION_FORMAT` variable instead of removing flags in `CMAKE_CXX_FLAGS_DEBUG` and setting `/ZI` manually.

AdamTadeusz:
> I think it's better to add `cmake_policy(SET CMP0141 NEW)` before project call (at line 4) and use `CMAKE_MSVC_DEBUG_INFORMATION_FORMAT` variable instead of removing flags in `CMAKE_CXX_FLAGS_DEBUG` and setting `/ZI` manually.

Do you know how to do this without flooding the output with this message? 
image

Masterkatze:
I don't see these warnings, can you try to delete the CMake cache and rebuild?

On my side the only change in diff with master is `cmake_policy(SET CMP0141 NEW)`

AdamTadeusz:
> I don't see these warnings, can you try to delete the CMake cache and rebuild?
> 
> On my side the only change in diff with master is `cmake_policy(SET CMP0141 NEW)`

It was the cache indeed

mp/src/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
-cmake_minimum_required(VERSION 3.18.4)
+cmake_minimum_required(VERSION 3.25)
```
Masterkatze, line 1: Don't know why it shows up in the diff, it's 3.25 in the master already.

Masterkatze (approved):

Rainyan:
Would review but I'm just chronically incapable of loading PDBs on Windows with our CMake setup and I still have no idea why...

Capture

The PDB is generated, but Visual Studio just refuses to use it. This problem is not really related to this PR's changes but anyway. Perhaps someone more blessed by Microsoft gods can chime in for the 2nd review.

AdamTadeusz:
For a few days now I noticed hot reload works on any new branch, so some of the recent changes to the build system must have changed something to enable it, ill close this pr for now

Issue #316, opened 2024-05-31
All classes have the correct height


blaberry:
At the moment, eyepos shows incorrect values for at least support.

Rainyan:
Looks like  the client/server heights are off by about 10 - `cl_showerror 2` output whilst crouching:

#### Support ducked:
```
001 CCollisionProperty::m_vecMaxsPreScaled
	- vec[] differs (1st diff)
		(net 16.000000 16.000000 60.000000 - pred 16.000000 16.000000 50.000000)
		delta(0.000000 0.000000 -10.000000)
		
002 CCollisionProperty::m_vecMaxs
	- vec[] differs (1st diff)
		(net 16.783783 16.783783 62.939186 - pred 16.783783 16.783783 52.449322)
		delta(0.000000 0.000000 -10.489864)
```

#### Assault ducked:
```
001 CCollisionProperty::m_vecMaxsPreScaled
	- vec[] differs (1st diff)
		(net 16.000000 16.000000 60.000000 - pred 16.000000 16.000000 50.000000)
		delta(0.000000 0.000000 -10.000000)
		
002 CCollisionProperty::m_vecMaxs
	- vec[] differs (1st diff)
		(net 15.618243 15.618243 58.568413 - pred 15.618243 15.618243 48.807011)
		delta(0.000000 0.000000 -9.761402)
```

#### Recon ducked:
```
001 CCollisionProperty::m_vecMaxsPreScaled
	- vec[] differs (1st diff)
		(net 16.000000 16.000000 60.000000 - pred 16.000000 16.000000 50.000000)
		delta(0.000000 0.000000 -10.000000)
		
002 CCollisionProperty::m_vecMaxs
	- vec[] differs (1st diff)
		(net 15.385135 15.385135 57.694256 - pred 15.385135 15.385135 48.078545)
		delta(0.000000 0.000000 -9.615711)
```

kassibuss:
Should be fixed by https://github.com/NeotokyoRebuild/neo/pull/470, close?

PR #470, opened 2024-07-02
Hull and view offsets match og

AdamTadeusz:
## Toolchain

- Windows MSVC VS2022

## Linked Issues

- related  #316


AdamTadeusz:
The friction when walking in a tight hole in some cardinal directions I believe is caused by this lagging of the player's qphysics or its vphysics shadow when walking in those same cardinal directions which is already present on master and I haven't found the cause of just yet
https://github.com/NeotokyoRebuild/neo/assets/44210793/c9e309e1-39b5-4d9a-a2bd-894210c7c801

mp/src/game/shared/shareddefs.h
```diff
@@ -85,15 +85,19 @@ class CViewVectors
...
-#define VEC_DUCK_VIEW_NEOSCALE(NeoPlayer) (g_pGameRules->GetViewVectors()->m_vDuckView + ((NeoPlayer->GetClass() == NEO_CLASS_RECON) ? NEO_RECON_VIEW_OFFSET : (NeoPlayer->GetClass() == NEO_CLASS_SUPPORT) ? NEO_SUPPORT_VIEW_OFFSET : NEO_ASSAULT_VIEW_OFFSET))
+#define VEC_DUCK_VIEW_NEOSCALE(NeoPlayer) (g_pGameRules->GetViewVectors()->m_vDuckView + ((NeoPlayer->GetClass() == NEO_CLASS_RECON) ? NEO_RECON_DUCKED_VIEW_OFFSET : (NeoPlayer->GetClass() == NEO_CLASS_SUPPORT) ? NEO_SUPPORT_DUCKED_VIEW_OFFSET : NEO_ASSAULT_DUCKED_VIEW_OFFSET))
 
-#define NEO_RECON_DUCK_MAXHULL_OFFSET Vector(0.0f, 0.0f, -2.0f)
-#define NEO_ASSAULT_DUCK_MAXHULL_OFFSET Vector(0.0f, 0.0f, -1.0f)
-#define NEO_SUPPORT_DUCK_MAXHULL_OFFSET Vector(0.0f, 0.0f, 6.0f)
+#define NEO_RECON_DUCK_MAXHULL_OFFSET Vector(0.0f, 0.0f, -2.161614f) // Hull Max of 59 when crouched
+#define NEO_ASSAULT_DUCK_MAXHULL_OFFSET Vector(0.0f, 0.0f, -0.826738f) // Hull Max of 48 when crouched
+#define NEO_SUPPORT_DUCK_MAXHULL_OFFSET Vector(0.0f, 0.0f, 6.244769f) // Hull Max of 46 when crouched
```
Rainyan, line 100: Where do these Z values come from? A fraction of standing hull vs ducked hulls maxs.z?

Rainyan (commented):
GitHub requires me to type something in this field. Great feature.

mp/src/game/shared/shareddefs.h
AdamTadeusz: I don't know what the exact hull maxes in og are, but there is a phenomenon where the player vphysics shadow moves up from the ground slightly after moving (present in base sdk2013 and ce too). These values give the most headroom to the collision box so that the player can still fit into the same hole when right against it and when moving for some time as in the original to the nearest hammer unit, while not being able to fit in the next largest hole even in the worst case scenario (to the nearest hammer unit). the actual offset applied to the hull max will be further modified by the player scale which I have included in the readout on the right side of the screen

AdamTadeusz: I don't know how this increase in the collision box altitude works exactly but it looks like it raises the collision box by about half a hammer unit, so the values can probably be moved by no more than that to find some nice round number. There was the crouch jitter which i was also trying to negate by giving as much headroom as possible while keeping the same hull max to within one hammer unit, but that has since been fixed in a different pr so maybe we can change these values to something more round

AdamTadeusz: @Rainyan Are we confident the current hull widths are parity? If that is the case we can probably use a sloped roof with cl_showpos to figure out the actual height of the playermodel to a far greater degree of accuracy

AdamTadeusz: image
So I guess that's a no. Probably the current x and y hull dimensions do not take the PlayerScale into account or something

AdamTadeusz:
[nt_dimensions_test2.zip](https://github.com/user-attachments/files/16143978/nt_dimensions_test2.zip)

I included a map to test the hull maxes, its a slightly modified version of nt_dimensions_test. Turn 180 degrees and go to the triangle in front of you. Using cl_showpos 1 you can run into the triangle as far as you can and verify that your y position is the same in og tokyo and this branch

mp/src/game/shared/shareddefs.h
AdamTadeusz: Fractions make more sense now, gonna close this

mp/src/game/shared/neo/neo_player_shared.h
```diff
@@ -147,9 +147,9 @@ COMPILE_TIME_ASSERT(NEO_ASSAULT_CROUCH_SPEED == NEO_SUPPORT_CROUCH_SPEED);
 // searching until the hulls matched within 1 unit in-game.
 #define NEO_HULL_TOLERANCE_SCALE 1.078125
 
-#define NEO_RECON_MODEL_SCALE ((NEO_RECON_PLAYERMODEL_HEIGHT / HL2DM_DEFAULT_PLAYERMODEL_HEIGHT) * NEO_HULL_TOLERANCE_SCALE)
-#define NEO_ASSAULT_MODEL_SCALE ((NEO_ASSAULT_PLAYERMODEL_HEIGHT / HL2DM_DEFAULT_PLAYERMODEL_HEIGHT) * NEO_HULL_TOLERANCE_SCALE)
-#define NEO_SUPPORT_MODEL_SCALE ((NEO_SUPPORT_PLAYERMODEL_HEIGHT / HL2DM_DEFAULT_PLAYERMODEL_HEIGHT) * NEO_HULL_TOLERANCE_SCALE)
+#define NEO_RECON_MODEL_SCALE 1.f//((NEO_RECON_PLAYERMODEL_HEIGHT / HL2DM_DEFAULT_PLAYERMODEL_HEIGHT) * NEO_HULL_TOLERANCE_SCALE)
+#define NEO_ASSAULT_MODEL_SCALE 1.f//((NEO_ASSAULT_PLAYERMODEL_HEIGHT / HL2DM_DEFAULT_PLAYERMODEL_HEIGHT) * NEO_HULL_TOLERANCE_SCALE)
+#define NEO_SUPPORT_MODEL_SCALE 1.f//((NEO_SUPPORT_PLAYERMODEL_HEIGHT / HL2DM_DEFAULT_PLAYERMODEL_HEIGHT) * NEO_HULL_TOLERANCE_SCALE)
```
nullsystem, line 152: They seem to all be `1.f` and only used for multiplication, so it seems un-needed now. Any reason to keep them?

mp/src/game/shared/shareddefs.h
```diff
@@ -85,17 +85,25 @@ class CViewVectors
...
+#define NEO_RECON_VIEW_OFFSET Vector(0.0f, 0.0f, -1.f)
+#define NEO_ASSAULT_VIEW_OFFSET Vector(0.0f, 0.0f, -0.5f)
+#define NEO_SUPPORT_VIEW_OFFSET Vector(0.0f, 0.0f, 2.0f)
+#define NEO_RECON_DUCKED_VIEW_OFFSET Vector(0.0f, 0.0f, -4.f)
+#define NEO_ASSAULT_DUCKED_VIEW_OFFSET Vector(0.0f, 0.0f, -4.f)
+#define NEO_SUPPORT_DUCKED_VIEW_OFFSET Vector(0.0f, 0.0f, 5.f)
+
 #define VEC_VIEW_NEOSCALE(NeoPlayer) (g_pGameRules->GetViewVectors()->m_vView + ((NeoPlayer->GetClass() == NEO_CLASS_RECON) ? NEO_RECON_VIEW_OFFSET : (NeoPlayer->GetClass() == NEO_CLASS_SUPPORT) ? NEO_SUPPORT_VIEW_OFFSET : NEO_ASSAULT_VIEW_OFFSET))
-#define VEC_DUCK_VIEW_NEOSCALE(NeoPlayer) (g_pGameRules->GetViewVectors()->m_vDuckView + ((NeoPlayer->GetClass() == NEO_CLASS_RECON) ? NEO_RECON_VIEW_OFFSET : (NeoPlayer->GetClass() == NEO_CLASS_SUPPORT) ? NEO_SUPPORT_VIEW_OFFSET : NEO_ASSAULT_VIEW_OFFSET))
+#define VEC_DUCK_VIEW_NEOSCALE(NeoPlayer) (g_pGameRules->GetViewVectors()->m_vDuckView + ((NeoPlayer->GetClass() == NEO_CLASS_RECON) ? NEO_RECON_DUCKED_VIEW_OFFSET : (NeoPlayer->GetClass() == NEO_CLASS_SUPPORT) ? NEO_SUPPORT_DUCKED_VIEW_OFFSET : NEO_ASSAULT_DUCKED_VIEW_OFFSET))
```
nullsystem, line 96: Nitpick: These which maps class to a value could be refactored to utilize a const array to map class to offset instead.
Although, not ground for blocking PR/can be ignored.

nullsystem (changes requested):

mp/src/game/shared/neo/neo_player_shared.h
AdamTadeusz: They are all indeed 1.f, kept them because it looked like a lot of effort went into calculating the previous values is all xd

AdamTadeusz: removed unused defs

mp/src/game/shared/shareddefs.h
AdamTadeusz: image
Could probably be refactored further to use this NEOViewVectors object which is closer to how the base game works. If it's alright we can do that in a separate pr?

nullsystem: It's fine to do it as a separate PR

nullsystem (approved):
```
nt_dimensions_test2 cl_showpos 1 y pos test:

         OGNT              NTRE

         Stand   Crouch    Stand   Crouch
Recon:   -1454 | -1472  || -1454 | -1472
Assult:  -1453 | -1470  || -1453 | -1470
Support: -1448 | -1459  || -1448 | -1459
```

mp/src/game/shared/shareddefs.h
Rainyan: > @Rainyan Are we confident the current hull widths are parity?

I believe we are off by one, at least for some classes. You can test this using the "funnel" area in the dimensions_test map against the OG NT.

brysondev (approved):
lgtm

PR #487, opened 2024-07-06
Player stuck when leaning

### LINKED ISSUE #152: Leaning while walking up stairs doesn't cause camera jitter
Seems to be partially fixed by a previous hull PR, also see PR #487

AdamTadeusz:
## Description

Video of original problem

Uploading Recording 2024-07-06 165938.mp4…

Since player collision hull does not rotate with the entity, changing the hull min and max y offset when leaning does not have the intended effect here. Instead causes players to get stuck when beginning a lean on a staircase in certain directions

## Toolchain

- Windows MSVC VS2022

## Linked Issues

I think #152 was opened as a result of this problem, don't see any camera jitter when walking up stairs now
- fixes #152

### DIFF
--- a/mp/src/game/shared/gamemovement.cpp
+++ b/mp/src/game/shared/gamemovement.cpp
@@ -756,13 +756,7 @@ Vector CGameMovement::GetPlayerMins( void ) const
 	{
 #ifdef NEO
 		Assert(dynamic_cast<CNEO_Player*>(player));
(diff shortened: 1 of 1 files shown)


AdamTadeusz:
https://github.com/NeotokyoRebuild/neo/assets/44210793/f80ab172-9270-4b8e-9508-35e4f15af637

nullsystem (approved):

brysondev (approved):
https://github.com/user-attachments/assets/4b3de08c-3f78-4ace-a857-b0aa6ff5ef16

lgtm.

PR #492, opened 2024-07-07
Add GitHub Actions workflow

Masterkatze:
## Description

- Add simple GitHub Actions workflow with building for the Windows and Ubuntu operating systems.
- Add CMake presets.
- Add separate debug information (.debug-files) support for Linux.
- Remove `BUILD_DATETIME` from version info to stop rebuilding if not necessary.
- Move version info from header to source file.
- Add CMake `install` calls (primarily used for GitHub Actions)
- Add CPack support, although support of it was not needed in the end.
- Update documentation.


nullsystem (approved):

Masterkatze:
> libraries for both windows and linux seem to still come as `bin/*` instead of `neo/bin/*`, in case this isn't intended

That was unintended, thanks for noticing.

Rainyan:
> couldn't "RelWithDebInfo" be renamed to just "Release"?

RelWithDebInfo seems more accurate, since that's what it is.

AdamTadeusz:
image
What am I doing wrong

Masterkatze:
> ```
> $ file client.so
> client.so: ... with debug_info, not stripped
> ```

It's correct because `(windows/linux)-release` profiles are RelWithDebInfo.

I think there is a confusion possible because now these profiles are more suitable for CI rather than developer use. If we want to 1:1 match between profile name and presence/absence of debug info then I think I'll rework this a little, something like this:

1. With `-DNEO_CI_BUILD=OFF` (default):
- `linux-debug` is Debug, `linux-release` is Release
- `-DNEO_USE_SEPARATE_BUILD_INFO=OFF` (so you can debug it as usual)

2. With -DNEO_CI_BUILD=ON`:
- `linux-debug` is Debug, `linux-release` is RelWithDebInfo (renamed to Release for artifact names)
`-DNEO_USE_SEPARATE_BUILD_INFO=ON` (so we can debug crashdumps)

I don't know if we need to add other CMake build types to profiles, if someone need RelWithDebInfo or MinSizeRel then it's still possible to build without profiles (the usual -S and -B).

Masterkatze:
> doesn't "RelWithDebInfo" CI still have the release binaries separated from the debug information? which is why they're ~9mb in size in your actions. That's why I thought it would do the same if I used the "linux-release" preset.

CI build uses `-DNEO_USE_SEPARATE_BUILD_INFO=ON`

nullsystem (dismissed):

nullsystem (approved):

AdamTadeusz (approved):

brysondev (approved):
image
image

PR #494, opened 2024-07-07
Allow ADS while changing weapons + static_cast GetActiveWeapon

### LINKED ISSUE #356: Add ability to aim whilst changing weapons
In OG NT, players can start aiming as soon as they switch weapons, even before the weapon is ready to be fired.

In rebuild, you cannot start zooming in until the weapon is ready to be fired.

nullsystem:
## Description

* Removed the restriction on preventing ADS while weapon changing
* There doesn't seem to be an auto switch or it causing issues requiring the restriction
* static_cast the usage of GetActiveWeapon since it's all NT weapons now

## Toolchain

## Linked Issues

* fixes #356

### DIFF
--- a/mp/src/game/client/neo/c_neo_player.cpp
+++ b/mp/src/game/client/neo/c_neo_player.cpp
@@ -698,8 +698,7 @@ void C_NEO_Player::ItemPreFrame( void )
 
 	if (m_afButtonPressed & IN_DROP)
 	{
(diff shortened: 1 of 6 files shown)


AdamTadeusz:
I put this in place initially as a bandaid fix for the problem where switching to a weapon with no zoom (like a knife or grenade) when already zoomed in caused some fov problems

AdamTadeusz (approved):

VantoNortim (approved):

PR #495, opened 2024-07-07
Assets as a submodule

AdamTadeusz:
## Description

Remove neo folder, replace it with a git submodule from the assets repository

## Toolchain

- Windows MSVC VS2022


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "mp/game/neo"]
+	path = mp/game/neo
+	url = https://github.com/NeotokyoRebuild/assets
```
nullsystem, line 3: This isn't going to support people using ssh. Use relative path to support both HTTPS and SSH, see: https://www.damirscorner.com/blog/posts/20210423-ChangingUrlsOfGitSubmodules.html

nullsystem (changes requested):
This should be put on hold for now till the PRs merged in as it could miss out some changes if not being careful. However, another thing, I think stuff that are just text files like cfgs/scripts are better off as part of this repo? One thing I think of is if we're doing separation of files is that `mp/game/neo` should just be a clean slate (have nothing in checkout) and using cmake to pull things together during configuration+compile stage.

AdamTadeusz:
> This should be put on hold for now till the PRs merged in as it could miss out some changes if not being careful. However, another thing, I think stuff that are just text files like cfgs/scripts are better off as part of this repo? One thing I think of is if we're doing separation of files is that `mp/game/neo` should just be a clean slate (have nothing in checkout) and using cmake to pull things together during configuration+compile stage.

The way git submodules work by default is it places a folder of the submodule inside the parent repository, so easiest if all the extra assets are in a single folder

nullsystem:
@AdamTadeusz I know how sub-modules work, that's why I've suggested the use of cmake to pull things together and `mp/game/neo` as a clean slate. My main issue is basically the need to co-ordinate between two repos if needing to edit some scripts/cfgs when they're better served in this main repo.

blaberry:
I'd say the divide should be that things that you edit in plain text goes in the neo repo, including scripts, whereas binary formats go in assets.

PR #496, opened 2024-07-07
Client vs server build integrity check + server startup print neo_version info

### LINKED ISSUE #437: [FEATURE] Output of build date + hash in console for server ops
### Description

Add an output of build date + hash in console for server ops on launching the server/map change

### Proposed Solution

As a server op, I would want to see the build date + hash in the console when launching the server or when the map changes, so that I can easily find out what version I'm running, plus it will appear in the logs so that will be helpful too.

### Alternatives Considered

The `neo_version` command, but I will need to enter that every time or include it in the autoexec, which are both options which are not as neat as just having it outputted by itself, as well it will be server-dependent, so if I checkout a log from a different server and they don't have that command in autoexec, I wouldn't know what version they're running.

### Additional Context

_No response_

### LINKED ISSUE #485: [FEATURE] Hash/version check for clients before joining server
### Description

In previous playtests, people have joined with outdated builds, which has caused people to not have 1:1 experiences with others. It has also led to reporting outdated bugs, which have already been fixed

### Proposed Solution

Add a check for the client before it joins a server so that the players cant connect with an outdated version of the game

### Alternatives Considered

_No response_

### Additional Context

_No response_

nullsystem:
## Description

* Server now just prints neo_version on startup
* Server and client long git hashes are now matched up on startup to check build integrity. If they don't, the client cannot connect unless they have matching git hashes.
* Server convar `neo_sv_build_integrity_check` to enable/disable this integrity check feature. Enabled by default.
* Server convar `neo_sv_build_integrity_check_allow_debug` to enable/disable integrity checking for debug clients. Disabled by default for release builds, enabled by default for debug builds.
    * A warning message will indicate if a client bypassed the checking
    * Debug bypass is determined by the first byte's final bit of (8-bits) char. 1 = debug build, 0 = release build.
    * String matching will first mask out that char so even if `neo_sv_build_integrity_check_allow_debug` is disabled, a matching git hash between debug client and server can still be matched.

## Toolchain

- Windows MSVC VS2022

## Linked Issues

* fixes #437
* fixes #485

## Examples
Client-side example of integrity check fail setting bogus "foobar" to trigger it:
clientintegrityfailexample

Server side will log this line:
```
Dropped PLAYER from server (Build integrity failed! Client vs server mis-match: Check your neo_version. Client: foobar | Server: e266b6d)
```

### DIFF
--- a/mp/src/game/shared/neo/neo_version.cpp
+++ b/mp/src/game/shared/neo/neo_version.cpp
@@ -3,12 +3,11 @@
 #include "convar.h"
 #include "dbg.h"
 
-namespace {
-void neoVersionCallback()
+void NeoVersionPrint()
 {
 #if defined(GAME_DLL)
 	static constexpr char HEADER[] = "neo_version (Server's build info):";
-#else defined(CLIENT_DLL)
+#elif defined(CLIENT_DLL)
 	static constexpr char HEADER[] = "neo_version (Client's build info):";
 #endif
 	Msg("%s\n"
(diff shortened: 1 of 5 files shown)


Rainyan:
As a rogue feature request: could we get some kind of "I know what I'm doing, just let me in" magic bit that's toggled if you're trying to join using a debug build? For dev convenience.

nullsystem:
@Rainyan I've tried to do the leading char-bit, but looks like the engine really didn't like that. So probably more like some magic number prefix instead.

nullsystem:
@Rainyan Implemented debug integrity bypass in: https://github.com/NeotokyoRebuild/neo/pull/496/commits/44c04c45a996910be17e8e6f001476d9f9b59035

mp/src/game/shared/neo/neo_version.cpp
```diff
@@ -3,12 +3,11 @@
...
-void neoVersionCallback()
+void NeoVersionPrint()
 {
 #if defined(GAME_DLL)
 	static constexpr char HEADER[] = "neo_version (Server's build info):";
-#else defined(CLIENT_DLL)
+#elif defined(CLIENT_DLL)
 	static constexpr char HEADER[] = "neo_version (Client's build info):";
 #endif
 	Msg("%s\n"
```
brysondev, line 13:
```suggestion
	Msg("%s\n"
```
`Msg(static_cast<tchar>("%s\n"`?

nullsystem: No

brysondev (approved):
lgtm

AdamTadeusz (approved):

PR #499, opened 2024-07-10
Round-end graphic drop shadow

### LINKED ISSUE #472: Round-end graphics have a drop shadow
This would make then universally legible against brighter areas that occur on some maps.

Round end graphics being the "JINRAI WINS" text and "X captured the ghost" text for example.

nullsystem:
## Description

* Set graphics and fonts as non-additive so it doesn't fade on black
* Add 2px offset render behind main render
* Some bit of refactor in the class also

## Toolchain

## Linked Issues

* fixes #472

### DIFF
--- a/mp/game/neo/materials/vgui/hud/roundend_jinrai.vmt
+++ b/mp/game/neo/materials/vgui/hud/roundend_jinrai.vmt
@@ -1,0 +1,10 @@
+"UnlitGeneric"
+{
+	"$basetexture" "vgui\hud\roundend_jinrai"
+	"$translucent" 1
+	"$vertexcolor" 1
+	"$vertexalpha" 1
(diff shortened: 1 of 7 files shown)


AdamTadeusz (approved):
image 
image


Would be nice if the messages written by this util could also have a drop shadow, and the version number drops shadow is a little far from the text itself. The first might require more refactoring that is outside of the scope of this pr, and the second is not important since the version number will be removed / changed in the future to be less obtrusive anyway

nullsystem:
@AdamTadeusz 
1. Probably more of a question if people want drop shadow for the central text or not, so yeah not something I'll want to apply now currently. 
2. I think the short hash looked fine especially considering 1080p res, but again not something to bother about

VantoNortim (approved):

PR #509, opened 2024-07-14
Remove VPC build system

### LINKED ISSUE #313: Drop the VPC build system
As we've moved to CMake with #241, I think we should remove the VPC build files and documentation from the project, because maintaining two build systems is just extra overhead. And I assume we're not going to be building for the Xbox360 any time soon 😄

Masterkatze:
Remove VPC build system, protobuf and some build scripts.

Closes #313


CONTRIBUTING.md
```diff
@@ -113,14 +113,8 @@ Ochii's impressive [reverse engineering project](https://github.com/Ochii/neotok
```
nullsystem, line 113: `Current: CMake` -> `CMake`

Rainyan:
This completely breaks our ability to compile shaders - we need the `neo\mp\src\materialsystem\stdshaders` Perl toolchain for compiling the DirectX shaders.

Rainyan:
Also, is there a need for us to drop Protobuf? It seems useful for encoding networked events efficiently, etc. Or was it only a part of the VPC project but not CMake, or outdated?

Rainyan:
Personally, I'd be happy just to see the .vpc scripts gone for not having to maintain them. Also, I'm not sure I have the confidence to review a -300k loc diff across so many files. Seems like a lot of changes for a single PR.

Masterkatze:
> This completely breaks our ability to compile shaders - we need the `neo\mp\src\materialsystem\stdshaders` Perl toolchain for compiling the DirectX shaders.

I was unsure if we use it, I'll add it back.

> Also, is there a need for us to drop Protobuf? It seems useful for encoding networked events efficiently, etc. Or was it only a part of the VPC project but not CMake, or outdated?

It was not being used even with VPC, if we will need protobuf one day - there are better ways to add it rather than having an old version in our repo directly.

> -300k loc diff across so many files

It's 99% protobuf.

AdamTadeusz:
image
Nice

CONTRIBUTING.md
```diff
@@ -109,18 +109,12 @@ Ochii's impressive [reverse engineering project](https://github.com/Ochii/neotok
 
 ## Good to know
 
-### Current: CMake
+### CMake
 
 This project currently use the [CMake](https://cmake.org/) build system to generate ninja/makefiles and is integrated with IDEs such as VS2022 and Qt Creator. When modifying the project file structure, look into `CMakeLists.txt` and `cmake/*.cmake`.
```
brysondev, line 114: This project ~~currently~~ uses.

Masterkatze: Updated

brysondev (approved):
looks great!

nullsystem (approved):

PR #520, opened 2024-07-25
Original HUD colours

### LINKED ISSUE #455: Hud background opacity should be the same as OG NT
### Build Info

20240629_321f139

### Description

OG NT:
image

NT;RE:
image

### To Reproduce

Launch into a map, spawn on a team and observer the HUD.

### Expected behavior

The HUD background opacity should match OG NT.

### Actual behavior

It's much less transparent.

### Operating System

- [X] Windows
- [ ] Linux

### Version/Distro

Windors 10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

AdamTadeusz:
## Description

Used a custom map with a totally black wall to test hud colours, think these are the original. Didn't touch the compass and round status text colour. Compass text in the original has a fade effect that needs to be implemented still, and round status element needs to be readable so left as is. Technically health sprint aux bar outlines should be drawn outside the bar itself instead of on top of the bars, so the outlines are brighter than they should be when next to actual bar. Can fix if this really bugs someone.

## Toolchain

- Windows MSVC VS2022
## Linked Issues

- fixes #455

### DIFF
--- a/mp/game/neo/scripts/HudLayout.res
+++ b/mp/game/neo/scripts/HudLayout.res
@@ -538,7 +538,7 @@
 		"needle_visible"	"0"
 		"needle_colored"	"0"
 		"objective_visible"	"1"
-		"box_color"		"100 100 100 178"
+		"box_color"		"200 200 200 40"
 	}
 	NHudWeapon
 	{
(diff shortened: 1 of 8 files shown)


AdamTadeusz:
Screenshot 2024-07-25 140440
Screenshot 2024-07-25 141224
Some screenshots of the original and this branch. Image was taken before the fire indicator was also changed to use the same colour as the ammo_text_colour, also using a custom hud layout in the original but that shouldn't have an effect on the colours

VantoNortim (approved):

brysondev (approved):
lgtm

PR #523, opened 2024-07-27
Old Squad HUD

### LINKED ISSUE #145: Squads exist and you can swap between them

AdamTadeusz:
## Description

Old Squad HUD plus some adjustments

## Toolchain

- Windows MSVC VS2022

## Linked Issues

- fixes #145

### DIFF
--- a/mp/src/game/client/neo/ui/neo_hud_round_state.cpp
+++ b/mp/src/game/client/neo/ui/neo_hud_round_state.cpp
@@ -28,7 +28,9 @@ DECLARE_NAMED_HUDELEMENT(CNEOHud_RoundState, NRoundState);
 
 NEO_HUD_ELEMENT_DECLARE_FREQ_CVAR(RoundState, 0.1)
 
-ConVar neo_old_squad_hud("neo_old_squad_hud", "0", FCVAR_ARCHIVE, "Use the old squad HUD", true, 0.0, true, 1.0);
+ConVar neo_cl_squad_hud_original("neo_cl_squad_hud_original", "0", FCVAR_ARCHIVE, "Use the old squad HUD", true, 0.0, true, 1.0);
+
+ConVar neo_cl_squad_hud_star_scale("neo_cl_squad_hud_star_scale", "0", FCVAR_ARCHIVE, "Scaling to apply from 1080p, 0 disables scaling");
 
 namespace {
 constexpr int Y_POS = 2;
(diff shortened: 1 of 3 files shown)


AdamTadeusz:
image
In the original the squad marker doesn't change size regardless of window size and aspect artio

Rainyan:
> In the original the squad marker doesn't change size regardless of window size and aspect artio

We might at least want a cvar for controlling this even if not parity - otherwise players on very large resolutions will have a hard time seeing the HUD, and same with very low resolutions rendering unreasonably large icons.

I'd suggest something like:
* value `0`: hardcoded size (parity)
* other values: scale from baseline, for example `value*res/1080p`, so a value of `1.5` at 1920x1080 would render the icons at 1.5x scaling.

AdamTadeusz:
> > In the original the squad marker doesn't change size regardless of window size and aspect artio
> 
> We might at least want a cvar for controlling this even if not parity - otherwise players on very large resolutions (4k+) will have a hard time seeing the HUD, and same with very low resolutions rendering unreasonably large icons.
> 
> I'd suggest something like:
> 
>     * value `0`: hardcoded size (parity)
> 
>     * other values: scale from baseline, for example `value*(res/1080p)`, so a value of `1.5` would decouple the scaling from resolution, and render the icons at 1.5x scaling compared to what OG NT size would render at 1920x1080. And `1.0` would render at the same scale as OG NT would for 1920x1080, regardless of what your resolution is.

I think at the very least the aspect ratio of the image should remain, the squad marker looks very squished on a 4:3 monitor

brysondev (changes requested):
image
image

Timer seems to resize upon changing to empty squad.

brysondev (approved):
Looks good to me

mp/src/game/client/neo/ui/neo_hud_round_state.cpp
```diff
@@ -270,113 +288,179 @@ void CNEOHud_RoundState::DrawNeoHudElement()
...
+					}
+
+					// Draw player
+					static constexpr int SQUAD_MATE_TEXT_LENGTH = 62; // 31 characters in name without end character max plus 3 in short rank name plus 7 max in class name plus 3 max in health plus other characters
+					char squadMateText[SQUAD_MATE_TEXT_LENGTH];
+					wchar_t wSquadMateText[SQUAD_MATE_TEXT_LENGTH];
+					const char* squadMateRankName = GetRankName(g_PR->GetXP(i), true);
+					const char* squadMateClass = GetNeoClassName(g_PR->GetClass(i));
+					const int squadMateHealth = g_PR->IsAlive( i ) ? g_PR->GetHealth( i ) : 0;
+					V_snprintf(squadMateText, SQUAD_MATE_TEXT_LENGTH, "%s %s  [%s]  Integrity %i\0", g_PR->GetPlayerName( i ), squadMateRankName, squadMateClass, squadMateHealth);
```
nullsystem, line 442: There's no point of the `\0` at the end when it's already applied for C strings. But not going to prevent the PR just for this.

nullsystem (approved):

PR #541, opened 2024-08-04
SourceTV maxplayers fix

### LINKED ISSUE #538: "-maxplayers 33" with SourceTV enabled leads to real max players value to be higher than the MAX_PLAYERS macro
### Build Info

20240801_9d04458

### Description

Current situation:
- MAX_PLAYERS is set as 33 in `mp/src/game/shared/shareddefs.h`
- There exists some kind of "real max players value"
- `-maxplayers` launch argument can be set to 33
- Enabling SourceTV leads to the real max players value being increased by 1
- Having both `-maxplayers` set to 33 and SourceTV enabled leads to the real max players value being 34
- This leads to unintended behavior, crashes, segfaults, etc.

Proposed solutions:
- Clamping `-maxplayers` launch argument to the value of "MAX_PLAYERS - 1", currently it's clamped to just "MAX_PLAYERS" (Seems to be not possible: we don't have access to the code of the `-maxplayers` launch argument)
- Don't allow SourceTV to increase the max players value by 1 if `-maxplayers` = MAX_PLAYERS (Seems to be not possible: we don't have access to the code of the behavior of STV increasing max players by 1)
- Disable the behavior of SourceTV increasing the max players value by 1 entirely (Seems to be not possible: same reason as above)
- Crash/quit the server if the real max players value is higher than MAX_PLAYERS, additionally alert the server host to keep `-maxplayers` under or equal to <dynamic value of MAX_PLAYERS - 1>

### To Reproduce

1. Launch a server with `-maxplayers 33` and SourceTV enabled
2. See the real max players being set to 34
3. Join as a client
4. Client crashes with "Engine Error: UpdatePlayerName with bogus slot 34"
5. Use `changlevel` once so that the SourceTV bot joins
6. Use `bot_add` several times to get the server's player count to 33/34
7. Enter `status`, see "0 humans, 33 bots (34 max)" in the console
8. Use `bot_add` again, witness the segfault

### Expected behavior

Not being able to run the server with 34 max players, because MAX_PLAYERS is set to 33

### Actual behavior

War crimes

### Operating System

- [ ] Windows
- [X] Linux

### Version/Distro

Debian 12 server, Arch and Windows clients

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

AdamTadeusz:
## Toolchain

- Windows MSVC VS2022

## Linked Issues

- fixes #538

### DIFF
--- a/mp/src/game/server/hl2mp/hl2mp_gameinterface.cpp
+++ b/mp/src/game/server/hl2mp/hl2mp_gameinterface.cpp
@@ -19,9 +19,10 @@
 void CServerGameClients::GetPlayerLimits( int& minplayers, int& maxplayers, int &defaultMaxPlayers ) const
 {
 #ifdef NEO
+	static constexpr int MAX_CLIENTS = MAX_PLAYERS - 1;
 	minplayers = 2;
 	defaultMaxPlayers = 24;
-	maxplayers = MAX_PLAYERS;
+	maxplayers = MAX_CLIENTS;
 #else
 	minplayers = defaultMaxPlayers = 2;
 	maxplayers = 16;


Masterkatze (approved):

nullsystem (approved):
@xedmain Dunno why you'd want 33 players without stv though.

AdamTadeusz:
> stv off, -maxplayers 15 = max players is 15 stv on, -maxplayers 15 = max players is 16 stv off, -maxplayers 32 = max players is 32 stv on, -maxplayers 32 = max players is 33 stv off, -maxplayers 33 = max players is 32 stv on, -maxplayers 33 = max players is 33
> 
> works well, although makes it impossible to have 33 max players without stv, but imo not that important or would even be widely used

image
This original comment in the source code above the definition of MAX_PLAYERS states that the value of maxplayers in gamerules should always be less than MAX_PLAYERS, so this is how its supposed to work

Rainyan:
> works well, although makes it impossible to have 33 max players without stv, but imo not that important or would even be widely used

The engine asserts there can never be >MAX_PLAYERS players on the server, so afaik this is not a supported use of the config option. What's supposed to happen for a statically allocated array compiled in the binary that can hold a maximum of MAX_PLAYERS number of items, when a MAX_PLAYERS+1th player joins the server, and their info is being inserted into that array, other than a buffer overrun?

YoYo178:
> a developer for FF13 suggested that wouldnt a better way be to only subtract if the cvar `tv_enable` is set to 1

yeah apparently i misunderstood the part, the fix is indeed correct.

PR #545, opened 2024-08-05
Grenades falling through floor, dropped weapon mechanics

### LINKED ISSUE #233: Picking up Grenades and Detpack
If they fall with a certain orinetation their origin is below the floor so the raycast to check for clear LOS hits the floor before the player. Maybe the origin needs to be adjusted in the model?

AdamTadeusz:
## Description

Grenades don't fall through the floor, picked up grenades can be used, grenades are only picked up if missing a grenade of that type, weapons dropped on death fall from their attachment point and inherit parent velocity and force of damage that led to death

## Toolchain

- Windows MSVC VS2022

## Linked Issues

- related #233


mp/src/game/shared/weapon_parse.cpp
```diff
@@ -431,6 +432,7 @@ void FileWeaponInfo_t::Parse( KeyValues *pKeyValuesData, const char *szWeaponNam
 	Q_strncpy( szBulletCharacter, pKeyValuesData->GetString("BulletCharacter", "a"), MAX_BULLET_CHARACTER);
 	iAimFOV = pKeyValuesData->GetInt("AimFov", 45);
 	m_flPenetration = pKeyValuesData->GetFloat("Penetration", 0);
+	m_bDropOnDeath = pKeyValuesData->GetFloat("DropOnDeath", 1);
```
nullsystem, line 435: * `GetFloat` -> `GetBool`
* Initialize to `true` instead of `1`

mp/src/game/server/neo/neo_player.cpp
```diff
@@ -1768,8 +1766,54 @@ void CNEO_Player::Event_Killed( const CTakeDamageInfo &info )
...
+		}
+	}
+	if (pNeoWeapon->GetNeoWepBits() & NEO_WEP_SMOKE_GRENADE) {
+		if (((m_nButtons & IN_ATTACK) && (!(m_afButtonPressed & IN_ATTACK))) ||
+			((m_nButtons & IN_ATTACK2) && (!(m_afButtonPressed & IN_ATTACK2)))) {
+			auto pWeaponSmoke = static_cast<CWeaponSmokeGrenade*>(pNeoWeapon);
+			pWeaponSmoke->DropLiveGrenade(this, velocity);
+			return;
+		}
+	}
```
nullsystem, line 1795: I think the code around `DropLiveGrenade` can be improved.

- `CheckThrowPosition` used within both grenade's `DropLiveGrenade` doesn't actually utilize any of the method var/func within and is repeated the same twice.  It can be separated out as it own plain function.
- `DropLiveGrenade` body could then be pulled out into this scope where it can be fitted in with the common codes and only differ a bit when creating grenade, and the two `DropLiveGrenade` can be dropped from being a function method.
- The if statement can then be combind with: `... & NEO_WEP_FRAG_GRENADE | NEO_WEP_SMOKE_GRENADE`

nullsystem (changes requested):

mp/src/game/server/neo/neo_player.cpp
AdamTadeusz: DropLiveGrenade is mostly a copy of the existing ThrowGrenade function except the initial velocity of the grenade is the player's velocity at the time of death, CheckThrowPosition potentially modifies the value of vecSrc which is then used in NEOFraggrenade_Create/NEOSmokegrenade_Create

image
I like the idea of including the content of DropLiveGrenade once in here, using for example isFrag to change the behaviour slightly between the two but I would have to drop CheckThrowPosition, and since I no longer see the sv_neo_infinite_frag/smoke_grenades I would first have to call a function to find that server convar to then check its value, unless we just skip that check since its probably not necessary anyway, you loose the weapon_grenade/smokegrenade the moment you throw a grenade

AdamTadeusz: I could just modify ThrowGrenade to take an optional parameter deathDrop or something, or even just checking if owner is alive when throwing to not include the players throw force in vecThrow

AdamTadeusz: actually I just remembered there is a check for if player is alive, but when weapon drop is called the player is still technically alive, hence the new function

nullsystem:
@AdamTadeusz Odd, it seems like the grenades had to be shot at first before they can be picked up? Can't just simply walk over it.

mp/src/game/server/neo/neo_player.cpp
```diff
@@ -1768,8 +1766,58 @@ void CNEO_Player::Event_Killed( const CTakeDamageInfo &info )
...
+{
+	CNEOBaseCombatWeapon* pNeoWeapon = static_cast<CNEOBaseCombatWeapon*>(pWeapon);
+	if (!pNeoWeapon)
+		return;
+	if (!pNeoWeapon->GetWpnData().m_bDropOnDeath)
+		return;
+#if(0) // Remove this #if to enable dropping of live grenades if killed while primed
+
+	// If attack button was held down when player died, drop a live grenade. NEOTODO (Adam) Add delay between pressing an attack button and the pin being fully pulled out. If pin not out when dead do not drop a live grenade
+	if (GetActiveWeapon() == pNeoWeapon && pNeoWeapon->GetNeoWepBits() & NEO_WEP_FRAG_GRENADE) {
```
Rainyan, line 1781: This is not parity, so it should probably go behind a cvar at a minimum.

Rainyan: Ah my bad, it was just a drive-by comment as I was glancing over quickly without time to properly review.

Rainyan (approved):
LGTM

nullsystem (approved):

nullsystem:
I've approved this assuming a fix for needing to shoot before pickup will follow up

PR #550, opened 2024-08-06
Grenade throw source position closer to center of screen

### LINKED ISSUE #466: Thrown grenade behaviour should match OG NT
### Build Info

20240629_321f139

### Description

Doing the same lineups you're used to from OG NT, especially when close to a wall, often results in disaster. It can also be noted that the grenades seem to warp/lag weirdly. Unsure if it's because of the angle or the netcode.

### To Reproduce

Stand close to a wall or corner, aim around the corner (use a gun to get a crosshair), switch to your smoke/frag grenade and throw it.

### Expected behavior

The grenade should fly past the corner following the same arc as in OG NT (would need some specific examples/videos to match this correctly).

### Actual behavior

The grenade hits the wall/corner and ends up killing whatever's close to you.

### Operating System

- [X] Windows
- [ ] Linux

### Version/Distro

Windors 10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

AdamTadeusz:
## Toolchain

- Windows MSVC VS2022

### DIFF
--- a/mp/src/game/shared/neo/weapons/weapon_grenade.cpp
+++ b/mp/src/game/shared/neo/weapons/weapon_grenade.cpp
@@ -287,7 +287,7 @@ void CWeaponGrenade::ThrowGrenade(CBasePlayer *pPlayer)
 	Vector	vForward, vRight;
 
 	pPlayer->EyeVectors(&vForward, &vRight, NULL);
-	Vector vecSrc = vecEye + vForward * 18.0f + vRight * 8.0f;
+	Vector vecSrc = vecEye + vForward * 2.0f;
 	CheckThrowPosition(pPlayer, vecEye, vecSrc);
 	vForward.z += 0.1f;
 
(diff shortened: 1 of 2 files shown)


AdamTadeusz:
https://github.com/user-attachments/assets/49315514-5ac1-4658-9cde-994ef9221bb6
355481342-c411bef7-f2f0-4fcd-b4ea-d8f867fa8162
Initial position of thrown grenade in this branch and the original game. Note pictures were taken with no angular velocity applied, as that seems to add some rotation to the initial grenade position when its spawned client side.

AdamTadeusz:
I can remove the -fixes #466 if there are other problems that we want to fix with the grenades. For example throwing from the same position at the same angle I notice some variation in how far the grenade flies, with there being almost no variation in the original

Agiel:
Idk why you're making it hard on yourself. The NT grenade code is right here https://github.com/NeotokyoRebuild/neo/blob/master/mp/src/game/shared/sdk/weapon_basesdkgrenade.cpp

AdamTadeusz:
> Idk why you're making it hard on yourself. The NT grenade code is right here https://github.com/NeotokyoRebuild/neo/blob/master/mp/src/game/shared/sdk/weapon_basesdkgrenade.cpp

Far as I can tell thats the weapon_basesdkgrenade code, and using the values defined there results in different behaviour to the original

Agiel:
> Far as I can tell thats the weapon_basesdkgrenade code

Which is what OGNT inherits from. Maybe grey tweaked some values, but behavior wise it'll give you parity.

AdamTadeusz:
The grenade's velocity indeed seems to correlate to the angle at which it was thrown , but I don't have the know how to measure the grenades thrown angle compared to the eyes angle or the velocity of the grenade in the original. I'll remove the fixes 466 and leave this pr as is

brysondev (approved):
LGTM

Masterkatze (approved):

PR #557, opened 2024-08-19
Don't drain AUX in underwater, don't tie drowning to AUX in underwater

### LINKED ISSUE #529: Aux depleted when under water
### Build Info

75f1e63

### Description

Aux depleted when under water

### To Reproduce

Download and open the map breakwater
Jump into the water

### Expected behavior

Aux remains at 100%

### Actual behavior

Aux rapidly falls to 0%

### Operating System

- [X] Windows
- [ ] Linux

### Version/Distro

11

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

### LINKED ISSUE #530: Water doing 10 damage per hit instead of 7
### Build Info

75f1e63

### Description

Water doing 10 damage per hit instead of 7

### To Reproduce

Download and launch the map breakwater
Jump in the water

### Expected behavior

After some time the character starts taking 10 damage per drown tick

### Actual behavior

Character takes 7 damage per drown tick

### Operating System

- [ ] Windows
- [ ] Linux

### Version/Distro

11

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

nullsystem:
## Description

* Just a quick ifndef out the device that starts/ends the drain
* Also override `CanBreatheUnderwater` so it doesn't depends on AUX = 0 to drown, just being in water long enough just drowns now
* NOTE: Might want to really separate "NT aux" from "suit power" in the future, but this is good enough for now

## Toolchain

## Linked Issues

* fixes #529
* fixes #530

### DIFF
--- a/mp/src/game/server/hl2/hl2_player.cpp
+++ b/mp/src/game/server/hl2/hl2_player.cpp
@@ -2257,6 +2257,7 @@ void CHL2_Player::CheckFlashlight( void )
 //-----------------------------------------------------------------------------
 void CHL2_Player::SetPlayerUnderwater( bool state )
 {
+#ifndef NEO
 	if ( state )
 	{
 		SuitPower_AddDevice( SuitDeviceBreather );
@@ -2265,6 +2266,7 @@ void CHL2_Player::SetPlayerUnderwater( bool state )
 	{
 		SuitPower_RemoveDevice( SuitDeviceBreather );
 	}
+#endif
 
 	BaseClass::SetPlayerUnderwater( state );
 }
--- a/mp/src/game/server/neo/neo_player.h
+++ b/mp/src/game/server/neo/neo_player.h
@@ -149,6 +149,8 @@ public:
 	void CloakPower_Charge(float flPower);
 	float CloakPower_Cap() const;
 
+	bool CanBreatheUnderwater() const override { return false; }
+
 	float GetNormSpeed_WithActiveWepEncumberment(void) const;
 	float GetCrouchSpeed_WithActiveWepEncumberment(void) const;
 	float GetWalkSpeed_WithActiveWepEncumberment(void) const;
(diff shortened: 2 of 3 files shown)


AdamTadeusz:
genius xD

AdamTadeusz (approved):

AdamTadeusz (changes requested):
I would remove the  -fixes #530

nullsystem:
@AdamTadeusz Oh, so aux is tied to the water damage in OG:NT...

AdamTadeusz:
It's not tied to aux, im just trying to not get swept away by the current. But you definitely start taking damage later and take less damage per tick. Damage taken also seems to either be a bit random or some kind of innacuracy between ticks because sometimes you take 6 and sometimes 7 damage

nullsystem:
@AdamTadeusz Ok OG:NT's damage certainly looks like a mix of 6 and 7. I mean I'd rather just stick with a fixed number rather than replicating this randomness...

nullsystem:
@AdamTadeusz Actually it does seems like it is increasing from a min to maximum value. Adjusted the value that seems to behave close.

AdamTadeusz:
In the original a recon takes between 8 and 7 damage and a support takes between 4 and 3 damage, So I guess the class damage modifier applies to drowning damage too

AdamTadeusz:
Also unrelated but you can initiate a sprint in the original when airborne, swimming and on the floor in water, none of which works in ntre

nullsystem:
@AdamTadeusz I've differ the modifier by class now, although isn't exact due to OG:NT's odd varying damages where NT;RE is consistent, it's close neough.

AdamTadeusz (approved):

brysondev (approved):

brysondev:
image

LGTM

PR #560, opened 2024-08-31
Spread parity

### LINKED ISSUE #107: Recoil is at parity

Agiel:
## Description
Change gun spread mechanics and values to match parity. The exact values were found via a mix of [empirical testing](https://docs.google.com/spreadsheets/d/1wsJ1u05_RdEBO7WAOB_jMboISHDlPjP4z6OMqgsDy5g/edit?pli=1&gid=0#gid=0) and reverse engineering. However, I can't guarantee that there aren't any typos.

In addition, a ton of duplicate code has been ripped out and, where applicable, moved to the base class. Note that during this process I've semi-intentionally broken the dry firing sound/animation. I intend to fix this, but we may decide to do it in a separate PR. The behavior wasn't matching parity anyway.

I've added a new cvar `sv_neo_accuracy_penalty_scale` which can be used to tweak the default 0.8 penalty per shot during future playtesting.

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #107


brysondev (commented):
Nothing outstanding in terms of code. Testing with real players might be necessary for the networking.

Agiel:
Just realized I forgot about slugs 😩

nullsystem (changes requested):
Most of the weapons looks fine... apart from the AA13:

https://github.com/user-attachments/assets/4f84a73d-d917-4058-ae46-4f2dcff60e29

AdamTadeusz:
> Most of the weapons looks fine... apart from the AA13:
> 2024-09-07.13-47-53.mp4
> 
> Really fast firerate now.

nah this is parity

AdamTadeusz:
line 476
image

Could re-introduce dry fire this way again. Originally I believe  it was handled a bit further down when PrimaryAttack() was triggered even when a weapon was empty, but now that is not the case

AdamTadeusz (approved):

nullsystem (approved):

PR #561, opened 2024-09-03
remove FixIncompatibleNeoAssets

AdamTadeusz:
## Description

With changes to how assets are mounted this is no longer necessary I think? Correct me if I'm wrong. Removes some annoying errors in the console.

## Toolchain

- Windows MSVC VS2022


brysondev (dismissed):

AdamTadeusz:
The reason I would like to remove this is because the restoring of the original file doesnt always work, so running ntre can break the rabbitfrog material for players who then go to play the original game. even more important now that we are doing playtests more often. I would even go as far as to say this is top priority.

AdamTadeusz:
we cant be breaking the original game for people who are playtesting ntre

brysondev (approved):

Rainyan (approved):

PR #562, opened 2024-09-03
Startup Errors in console

AdamTadeusz:
## Description

Files in https://github.com/NeotokyoRebuild/neoAssets/pull/3

neo_english.txt is just neotokyo_english.txt renamed from the original game files

The weapon sounds file is the same file from the original, modified as follows

For the weapons sounds channels, I used the same sound channel for the empty weapon sound as the empty weapon sound for the default pistol that was in the same file

For the grenade pin pull and bounce sound volume, I used the defined normal sound level. Needs testing whether this is too loud/too quiet, as we don't have the definition for the item sound volume

Values in hud_layout.res are placeholder to silence the errors, the elements get resized in code anyway

For the convars, it makes sense to me to have commands that modify gameplay be enforced by the server and replicated clientside, and for clientside only commands referring to the commands using a string literal is inline with whats done elsewhere in the code and silences errors related to clientside only commands being defined in server.dll

For the two missing resource files, we could probably silence them by including some dummy files maybe, or perform a check for those two specific files and return before the warning is printed, I don't see these files in the original sdk or the community edition sdk anyway. I will leave the errors in for now, maybe we can discuss what to do about it together. If the lack of those files means the associated elements don't work and we can't find the files anywhere, also can just ifdef the creation of those elements out.

For missing materials when opening the options menu:
the mic test material and definition was grabbed from https://forum.elite-hunterz.com/threads/how-to-fix-missing-materials-in-voice-options.11991/ <-- This one might be a problem
the spray_canned material definition was grabbed from hl2dm, all the other spray material definitions then generate as you select them all in the settings menu
the crosshair material and definition was grabbed from counter strike source vpks, but then replaced with our own

Map related errors and errors that appear during gameplay will be addressed in separate PRs, as they tend to be more complicated to solve

## Toolchain

- Windows MSVC VS2022


AdamTadeusz:
Theres some errors that appear when changing resolution for example that I can fix here too

brysondev:
Erm, so is there any way to test this? Aside from not seeing spam in-game of course.

AdamTadeusz:
You should be able to see a few new things in the options menu like this mic test visual aid 
image

Also the sprays that were originally in the spray drop down now appear in the preview on the left when you select them
image

For the PANEL_SPECMENU, I just did a search through all the code and there are other instances where the specmenu is referenced so I will have to either if neo it out the way that was originally done here 
image
or instead remove the offending parts of the spec menu so we still have the panel, even if it doesn't work, for demo replays etc

brysondev (dismissed):
lgtm

Rainyan:
We can find the mic textures via the Source SDK Base 2013 mount, at `Source SDK Base 2013 Multiplayer/hl2/hl2_textures_dir.vpk`, in the path `materials/vgui/resource`. Maybe it would be better to mount from there, instead of pulling Counter-Strike files?

Rainyan:
Also, I don't think this is related to issue 449, as it's about character animation issues specifically.

AdamTadeusz:
> We can find the mic textures via the Source SDK Base 2013 mount, at `Source SDK Base 2013 Multiplayer/hl2/hl2_textures_dir.vpk`, in the path `materials/vgui/resource`. Maybe it would be better to mount from there, instead of pulling Counter-Strike files?
> 
> Likewise, the `materials/vgui/cursors/crosshair.vtf` can be found inside the same VPK. Although unsure if it's the same crosshair, but probably doesn't really matter in this case.

the crosshair vtf is a background image for the crosshair preview, in this pr we are using our own. The one from the counter strike vtf is an image from mirage or dust 2 I think

brysondev (approved):

AdamTadeusz:
@Rainyan do you know how to grab those files from the vpk or is it ok if I extracted them and put them in neoAssets? They're probably gonna get packaged up into a vpk anyway at some point right?

brysondev (approved):

Masterkatze (approved):

PR #568, opened 2024-09-08
Copy binary files from neoAssets

AdamTadeusz:
## Description

Move all asssets to neoAssets. Now to build the game you need to have the repository https://github.com/NeotokyoRebuild/neoAssets cloned/downloaded to the same folder as the neo repository

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0


Rainyan:
Do we want to pin this to a static `GIT_TAG`? That'd require PRs for both repos whenever there's an asset update on the binary side.

AdamTadeusz:
> Do we want to pin this to a static `GIT_TAG`? That'd require PRs for both repos whenever there's an asset update on the binary side.

So you're just suggesting we use git tags instead of git commit hashes to reference versions of the assets repository? These and even branch names can be used if I understand correctly, unless we use use the git_shallow parameter in which case only branch names can be used I think

AdamTadeusz:
Never mind im retarded, you're suggesting we don't select the version of assets used. So yeah we could point to the master branch for example and add the git shallow tag and that would mean we don't need an extra pr here and it would be even quicker to fetch the git asset repository. The only question is would older versions of this repo break with newer assets. The only things that might break in the future are probably like materials, but also the stuff we are planning with navmeshes might for some reason not be compatible with older version of the repo I don't know

Rainyan:
Yeah, I meant we most likely want to use a tag that points to the most recent change (like `master`), to avoid having to create somewhat superfluous PRs the code repo to manually keep the two repos in sync, whenever the binary repo receives an update.

Masterkatze:
LGTM.

The use of `copy_directory` is fine since we copy it inside the same repo.

If someone wants to use manually cloned neoAssets:

1. Set `FETCHCONTENT_FULLY_DISCONNECTED` to `ON`
2. Set `FETCHCONTENT_SOURCE_DIR_NEO_ASSETS` to the path of cloned neoAssets repo

brysondev (approved):

Rainyan (approved):

PR #572, opened 2024-09-09
SRS Spawns with a round in the chamber

AdamTadeusz:
## Description

I thought it looked cool when the player chambers the srs when one is initially spawned, but maybe its kind of annoying and not parity so we can make the srs spawn with a round chambered maybe

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0


Rainyan (approved):

brysondev (approved):
Finally, no more cock on spawn

PR #577, opened 2024-09-10
Recoil parity

### LINKED ISSUE #107: Recoil is at parity

### LINKED ISSUE #566: Supa7 slugs shouldn't use the same spread as regular pellets
### Build Info

20240907_dfa52da

### Description

Since #560 the Supa7 uses a flat 2.5 degree spread. For the slugs it should be
ADS Min: 1 degree
ADS Max: 3 degrees
Hip Min: 4 degrees
Hip Max: 10 degrees

### To Reproduce

Grab a Supa7, shoot once, hit +attack2 to load a slug. Fire and observe spread.

### Expected behavior

Spread should match the OG values noted in the description.

### Actual behavior

Spread is a flat 2.5 degrees.

### Operating System

- [X] Windows
- [ ] Linux

### Version/Distro

Windows 10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

Agiel:
## Description

Brings view kick and recoil to parity. The values have been dug out of the OG client.dll (see tables in #107)~~, but with the current implementation they feel a bit strong. Need to figure out what's going on there before this can be merged. We'll probably also want to smear the recoil out over several frames even though that's not parity.~~

## Toolchain

- Windows MSVC VS2022

## Linked Issues

- fixes #107 
- fixes #566

### DIFF
--- a/mp/src/game/shared/neo/weapons/weapon_aa13.cpp
+++ b/mp/src/game/shared/neo/weapons/weapon_aa13.cpp
@@ -86,8 +86,6 @@ void CWeaponAA13::PrimaryAttack(void)
 		return;
 	}
 
-	pPlayer->ViewPunchReset();
-
 	// MUST call sound before removing a round from the clip of a CMachineGun
 	WeaponSound(SINGLE);
 
@@ -117,5 +115,7 @@ void CWeaponAA13::PrimaryAttack(void)
 		// HEV suit - indicate out of ammo condition
 		pPlayer->SetSuitUpdate("!HEV_AMO0", FALSE, 0);
 	}
+
+	pPlayer->ViewPunchReset();
 	AddViewKick();
 }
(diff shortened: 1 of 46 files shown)


AdamTadeusz:
Honestly can't tell the difference between this and ognt

AdamTadeusz:
image
Playing around with the weapons in the original I noticed that most of the weapons that end up with the aim point at the same angle as they started with still have recoil, but the character automatically drags the gun back to its original position (weapons like the tachi actually pull down as you fire, and your character pulls them back up). I guess this is what you would call view kick, but what if there was a similar force pulling the recoil of the guns that have actual recoil away from the direction of recoil

Agiel:
Ok it's ready for review now. Decided not to do the smearing here since it's not parity anyway.

brysondev (approved):
lgtm

nullsystem (approved):

PR #587, opened 2024-09-14
Workaround for Linux shaders breakage

### LINKED ISSUE #438: [BUG] Vision modes are sometimes broken on linux
### Build Info

20240629_321f139

### Description

Sometimes, the vision modes shaders are broken on linux and they're either completely black, or the colors are all wack.

### To Reproduce

Unreproducible

### Expected behavior

Normal vision modes

### Actual behavior

Broken vision modes

### Operating System

- [ ] Windows
- [X] Linux

### Version/Distro

Arch

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

Needs screenshots, but @nullsystem had the same problem as well

Rainyan:
## Description
This is a hack around the vision mode shaders sporadically breaking on Linux builds.

It seems that for whatever reason, the shader state cached at the `glshaders.cfg` file somehow leads to the corruption upon successive runs of the game, and write-protecting the file offers a simple way to avoid that breakage.

Ideally, we should figure out  what's really going wrong here at the shader system and/or precaching level, but this should suffice for avoiding the bug.

~~I opted to implement this as a CMakeLists step rather than submit the glshaders file itself, because to my understanding git does not version control the "read-only" file permissions required to keep the game from modifying the file.~~

edit: The hack is now done at runtime, because doing it in the CMake scripts won't carry over to users [as pointed out by nullsystem](https://github.com/NeotokyoRebuild/neo/pull/587#issuecomment-2351554803), and we can't version control the file, because the filesystem specific metadata required for the read-only state won't carry over.

For more discussion, see: https://github.com/NeotokyoRebuild/neo/issues/438#issuecomment-2351068151

### Steps to repro the bug
* Use Linux
* Build the game from master (or test with a recent release)
* Launch the game
* Check whether visions modes are broken for any of the 3 classes
* If they're *not* broken, close the game, relaunch it, and check again
* If you have the bug, you should be seeing shader corruption within a few tries

### Steps to test this patch
* The aforementioned corruption no longer occurs

## Toolchain

- Linux GCC Distro Native [Mint 21.3 Cinnamon + (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0]
- cmake version 3.30.3

## Linked Issues
- Fix #438

### DIFF
--- a/mp/src/game/client/neo/neo_fixup_glshaders.cpp
+++ b/mp/src/game/client/neo/neo_fixup_glshaders.cpp
@@ -17,7 +17,7 @@
 // For details, see: https://github.com/NeotokyoRebuild/neo/pull/587
 void FixupGlShaders(IFileSystem* filesystem)
 {
-	constexpr auto filename = "glshaders.cfg";
+	constexpr char filename[] = "glshaders.cfg";
 	constexpr auto pathID = "MOD";
 	if (filesystem->FileExists(filename, pathID))
 	{
--- a/mp/src/game/client/neo/neo_fixup_glshaders.cpp
+++ b/mp/src/game/client/neo/neo_fixup_glshaders.cpp
@@ -5,6 +5,9 @@
 #include "filesystem.h"
 #include "icvar.h"
 
+// Stdlib
+#include <source_location>
+
 // memdbgon must be the last include file in a .cpp file!!!
 #include "tier0/memdbgon.h"
 
@@ -13,7 +16,8 @@
 void FixupGlShaders(IFileSystem* filesystem, ICvar* cvar)
 {
 	constexpr auto filename = "glshaders.cfg",
-		pathID = "MOD";
+		pathID = "MOD",
+		funcName = std::source_location::current().function_name();
 
 	// This prevents the game from generating the problematic file on exit.
 	// There's also a "mat_autoload_glshaders", but it seems by the time we load,
@@ -24,7 +28,7 @@ void FixupGlShaders(IFileSystem* filesystem, ICvar* cvar)
 	ConVar* autosaveGlShaders;
 	if (!(autosaveGlShaders = cvar->FindVar(cvarname)))
 	{
-		Warning("%s: cvar %s not found\n", __func__, cvarname);
+		Warning("%s: cvar %s not found\n", funcName, cvarname);
 		return;
 	}
 	autosaveGlShaders->SetValue(false);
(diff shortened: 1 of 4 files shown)


nullsystem:
My main concern is for those downloading and installing the builds. The resources.zip is via Windows so it didn't apply here and neither the libraries zip includes this, so it seems to me those will not get the read-only glshaders.cfg.

Rainyan:
> My main concern is for those downloading and installing the builds. Checking the GitHub artifact, the resources.zip is generated via Windows so it didn't apply here and neither the libraries zip includes this, so it seems to me those will not get the read-only glshaders.cfg.

Right, and the chmod permissions are stored on the filesystem, so that's a dead end... I guess I'll have to see if this can be monkey patched at runtime after all.

Rainyan:
@nullsystem I've moved the patch to the client dll initialization, could you test if it's still working correctly for you?

nullsystem:
@Rainyan Looks like it works for me and manages to stub out that file with empty glshadercachev002.

Rainyan:
@xedmain 
> I wonder what will happen if NT;RE is installed on wacky filesystems like NTFS, FAT32/exFAT that have different types of permissions, as well what will happen if the current user can't modify permissions

> nevermind, I see in the code that it just errors out

Did you test this? If so, did it error gracefully at:

https://github.com/Rainyan/rebuild/blob/1d8cdac917cb2cfb480a40eb33ac56337a0b6335/mp/src/game/client/neo/neo_fixup_glshaders.cpp#L45-L49

Rainyan:
I found the cvar `mat_autosave_glshaders` which is used by the shaderapidx9 library. It seems that toggling it off prevents the game from generating the problematic `glshaders.cfg` file altogether.

I made changes such that we toggle the autosave off via cvar, and then delete the file if it already existed. Since we're too late intervene at client load, it'll still be bugged for a run where the buggy file existed, but this should fix it for any future runs. And players first-time launching the game should never run into the problem, as the cvar change should prevent the file from being generated in the first place.

This also removes the chmod filesystem permission flags fiddling, since it wasn't compatible across filesystems as pointed out by @xedmain.

mp/src/game/client/neo/neo_fixup_glshaders.cpp
```diff
@@ -0,0 +1,48 @@
...
+	// There's also a "mat_autoload_glshaders", but it seems by the time we load,
+	// it's too late for that to take an effect, at least via this route.
+	// The client can still override this by including a command in their autoexec,
+	// if they really want to.
+	auto autosaveGlShaders = cvar->FindVar("mat_autosave_glshaders");
+	if (autosaveGlShaders != nullptr)
+	{
+		autosaveGlShaders->SetValue(false);
+		Assert(!autosaveGlShaders->GetBool());
+	}
```
Masterkatze, line 28: No `else` branch with error logging (which always helps with troubleshooting the bug reports)? Same question for `if` statement on line 41.

Masterkatze (approved):

nullsystem (approved):

PR #589, opened 2024-09-15
Enforce con_enable 1 and multiple fixes for keybind handling

### LINKED ISSUE #586: Cannot enable developer console using new menu. It can only be enabled using the legacy menu.
### Build Info

20240914_b2291c5

### Description

New menu can disable developer console mode, but if developer console is disabled in the legacy menu, it cannot be enabled by the new menu. The expected behaviour is that it should be overridden by the option in the new menu.

Developer console mode is:
Disabled in legacy and disabled in new menu = disabled
Disabled in legacy and enabled in new menu = disabled
Enabled in legacy and disabled in new menu = disabled
Enabled in legacy and enabled in new menu = enabled

### To Reproduce

Go to options -> keybinds -> legacy -> disable developer console mode
Go to options -> keybinds -> enable developer console mode

### Expected behavior

Developer console mode is:
Disabled in legacy and disabled in new menu = disabled
Disabled in legacy and enabled in new menu = enabled
Enabled in legacy and disabled in new menu = disabled
Enabled in legacy and enabled in new menu = enabled

### Actual behavior

Developer console mode is:
Disabled in legacy and disabled in new menu = disabled
Disabled in legacy and enabled in new menu = disabled
Enabled in legacy and disabled in new menu = disabled
Enabled in legacy and enabled in new menu = enabled

### Operating System

- [X] Windows
- [ ] Linux

### Version/Distro

Windows 10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

nullsystem:
## Description
* Enforce con_enable 1 to fix console toggle issue
* Instead always enforce the "legacy" toggleconsole, because internally it requires con_enable to be set anyway, although the legacy menu toggles this.
* Basically means the support for legacy menu on this is an unsupported thing anyway
* Write the command file on save
* Fix attempts to write empty keys
* Unbind duplicate key

## Toolchain

## Linked Issues

* fixes #586

### DIFF
--- a/mp/src/game/client/cdll_client_int.cpp
+++ b/mp/src/game/client/cdll_client_int.cpp
@@ -1226,6 +1226,10 @@ void MusicVol_ChangeCallback(IConVar *cvar, const char *pOldVal, float flOldVal)
 	UpdateBgm((ConVar*)cvar);
 }
 
+#ifdef NEO
+extern void NeoToggleConsoleEnforce();
+#endif
+
 //-----------------------------------------------------------------------------
 // Purpose: Called after client & server DLL are loaded and all systems initialized
 //-----------------------------------------------------------------------------
@@ -1266,14 +1270,7 @@ void CHLClient::PostInit()
 		Assert(false);
 	}
 
-	// Rebind ` from toggleconsole to neo_toggleconsole
-	const auto toggleConsoleBind = gameuifuncs->GetButtonCodeForBind("toggleconsole");
-	if (toggleConsoleBind == KEY_BACKQUOTE)
-	{
-		char cmdStr[128];
-		V_sprintf_safe(cmdStr, "bind \"`\" \"neo_toggleconsole\"\n");
-		engine->ClientCmd_Unrestricted(cmdStr);
-	}
+	NeoToggleConsoleEnforce();
 #endif
 }
 
--- a/mp/src/game/client/neo/ui/neo_root.cpp
+++ b/mp/src/game/client/neo/ui/neo_root.cpp
@@ -43,6 +43,8 @@ extern ConVar cl_onlysteamnick;
 CNeoRoot *g_pNeoRoot = nullptr;
 void NeoToggleconsole();
 inline NeoUI::Context g_uiCtx;
+inline ConVar neo_cl_toggleconsole("neo_cl_toggleconsole", "0", FCVAR_ARCHIVE,
+								   "If the console can be toggled with the ` keybind or not.", true, 0.0f, true, 1.0f);
 
 namespace {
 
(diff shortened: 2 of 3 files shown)


Rainyan:
I'm having trouble saving the preference, so I can't really test the change. Using the legacy menu, I can save the "developer console" setting correctly however.

Example:

[bind.webm](https://github.com/user-attachments/assets/4f186055-dd20-4e6e-b72b-184706034715)

nullsystem:
@Rainyan It seems like it failed because 'q' was also used for lean left. Guess that's also why the legacy menu discards the other bind since it seems the behavior for GetButtonCodeForBind is 1:1 and can't handle more.
For now I'll put in a "fix" to match the legacy menu's behavior of unbinding the other key.

Rainyan:
> @Rainyan It seems like it failed because 'q' was also used for lean left. Guess that's also why the legacy menu discards the other bind since it seems the behavior for GetButtonCodeForBind is 1:1 and can't handle more. I'll put in a "fix" to match the legacy menu's behavior of unbinding the other key.

That was the other thing, but my main problem was being unable to turn off developer mode in the new menu.

nullsystem:
@Rainyan I was able to turn it off... but then I checked again and realised the main issue wasn't fixed anyway.
I think I'll just force the legacy option to be on at all times. It's going to be odd to try to deal with the two anyway...

nullsystem:
@Rainyan I've changed the description + title to reflect the way the change of the fix is done now. Basically the game will try to enforce `con_enable 1` when possible and won't be dealing with the situation of using both legacy and new settings since they both collides anyway.

nullsystem:
I've also gone ahead and do the duplicate keybind unbinding enforcement also.

Rainyan:
This seems to fix my previous inability to toggle the developer console setting in the new menu.

However I'm a bit unclear as to what the <code>\`</code> bind stuff in `NeoToggleConsoleEnforce` is achieving. It seems that typing: <code>bind \`</code> in the dev console instantly toggles the console off from under your feet while you're still typing, which feels a bit weird.

Also worth noting that the default backtick [does not work for some keyboard layouts](https://superuser.com/questions/1231849/how-to-open-the-in-game-console-in-source-engine-games/1409661#1409661), for example for the nordic kb layout it's impossible to press the <code>`</code> bind afaik.

AdamTadeusz:
Lets open an issue and merge this in then

AdamTadeusz (approved):

Rainyan (approved):

nullsystem:
@Rainyan I've put in a rough issue for now so it's noted: https://github.com/NeotokyoRebuild/neo/issues/620

PR #570, opened 2024-09-09
Fix initial player tposing

AdamTadeusz:
## Description

Replace weapon_switch with ClientCommand

## Toolchain

- Windows MSVC VS2022
- Linux GCC Distro Native [Specify distro + GCC version]
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #569

PR #598, opened 2024-09-17
Revert "replace weapon_switch with ClientCommand (#570)"

### LINKED ISSUE #569: Player TPosing
### Build Info

yes

### Description

Players t-pose at the start of the round until they switch weapons

### To Reproduce

run a server with nt_dimensions loaded and have two players join in and spawn. The second person will see the first person tposing

### Expected behavior

not tposing

### Actual behavior

tposing

### Operating System

- [X] Windows
- [ ] Linux

### Version/Distro

11

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

Rainyan:
## Description

When spawning in a round, the primary weapon you choose does not automatically activate like it should. This breaks both weapon selections for players, but seems to affect bots too. The player is left in the pistol animation without any weapon selected, and pressing `+attack` does nothing in this state.

This is a regression from d209d44b; the parent commit 51a1ef42 does not observe this behaviour.

I propose we revert #570 (commit d209d44b), because it was intended to fix #569 but doesn't, so it's only introducing new bugs currently.

## To reproduce the problem

* Join a server
* Join a team
* Choose a class
* Choose a loadout

## Expected behaviour
* Your chosen loadout is activated as you spawn in.
* Bots will also spawn with their primary weapon active.

## Actual behaviour
* You spawn empty-handed, until you manually select your weapon.
* Bots also spawn empty-handed.

### Weapon selection behaviour before the bug

This is the behaviour before commit d209d44b. Notice how the weapon is readied as soon as the player spawns, and the same holds true for bots:

[wepselect1.webm](https://github.com/user-attachments/assets/bdeb428c-aa31-4e53-ad38-a7a97f9f2a05)

### Weapon selection behaviour after the bug

Weapon does not become active after spawning, until the player manually selects it:

[wepselect-regression-1.webm](https://github.com/user-attachments/assets/af6cf41b-1f56-42db-971e-f25f515201d9)

Here's a third person view of a player spawning in empty-handed:

[wepselect-regression-2.webm](https://github.com/user-attachments/assets/8af8ca6a-1e83-446a-b43e-6746c6441752)

Here's a bot spawning in empty-handed. There's also the T pose bug #569 occurring here, but that is unrelated to the bugs described here:

[wepselect-regression-3.webm](https://github.com/user-attachments/assets/fbe553d8-3c99-49a9-b428-fbe97ab1a797)

## Linked Issues
- related #569, #570

### DIFF
--- a/mp/src/game/server/neo/neo_player.cpp
+++ b/mp/src/game/server/neo/neo_player.cpp
@@ -2491,22 +2491,21 @@ void CNEO_Player::GiveDefaultItems(void)
 		GiveNamedItem("weapon_knife");
 		GiveNamedItem("weapon_milso");
 		if (this->m_iXP >= 4) { GiveDet(this); }
-		engine->ClientCommand(edict(), "slot2");
+		Weapon_Switch(Weapon_OwnsThisType("weapon_milso"));
 		break;
 	case NEO_CLASS_ASSAULT:
 		GiveNamedItem("weapon_knife");
 		GiveNamedItem("weapon_tachi");
 		GiveNamedItem("weapon_grenade");
-		engine->ClientCommand(edict(), "slot2");
+		Weapon_Switch(Weapon_OwnsThisType("weapon_tachi"));
 		break;
 	case NEO_CLASS_SUPPORT:
 		GiveNamedItem("weapon_kyla");
 		GiveNamedItem("weapon_smokegrenade");
-		engine->ClientCommand(edict(), "slot2");
+		Weapon_Switch(Weapon_OwnsThisType("weapon_kyla"));
 		break;
 	default:
 		GiveNamedItem("weapon_knife");
-		engine->ClientCommand(edict(), "slot3");
 		break;
 	}
 }
@@ -2559,7 +2558,7 @@ void CNEO_Player::GiveLoadoutWeapon(void)
 				RemoveAllItems(false);
 				GiveDefaultItems();
 				pEnt->Touch(this);
-				engine->ClientCommand(edict(), "slot1");
+				Weapon_Switch(Weapon_OwnsThisType(szWep));
 			}
 		}
 		else
@@ -2590,7 +2589,7 @@ void CNEO_Player::GiveAllItems(void)
 
 	GiveNamedItem("weapon_tachi");
 	GiveNamedItem("weapon_zr68s");
-	engine->ClientCommand(edict(), "slot1");
+	Weapon_Switch(Weapon_OwnsThisType("weapon_zr68s"));
 }
 
 // Purpose: For Neotokyo, we could use this engine method


AdamTadeusz:
Would combining weapon_switch and ClientCommand work? Otherwise the first person view model muzzle flashes will be slightly scuffed when initially spawning

AdamTadeusz:
image
bots shoot again and my weapon is readied instantly so maybe just do both?

AdamTadeusz:
> > Would combining weapon_switch and ClientCommand work? Otherwise the first person view model muzzle flashes will be slightly scuffed when initially spawning [Edit] if that pr gets merged in as well
> 
> Could you clarify a bit? Any steps to reproduce the problem? I don't mind looking into it, but I've no idea what & where needs fixing.

image

on the viewmodel muzzleflash pr we update the properties of the muzzle flash when this function is called. Using weapon_switch this function is only called on the server side, when a client switches their own weapon with the slotx command the function is called both on the client and server.

Replace the ClientCommand on the first person view model muzzleflash branch with the original and then spawn in, select support and pick the supa7. The muzzle flash will be cross shaped instead of round and the size of the muzzle flash will be smaller than it should be

Rainyan:
I see, thanks. There's a few problems with the approach.

Firstly, I'm worried it might create a race condition where the behaviour depends on whether the networked command reaches the client before or after they have materialized in the level. You could for example register a think context for this, queuing it up for a future tick, to get around it.

But more critically, because we need to set the player's initial weapon as their primary slot server side, issuing a `slot1` or similar command won't actually have an effect, because they already have that as their active item. You can verify this by calling `slot<n>` with that slot already equipped, and see how it doesn't fire `CBasePlayer::Weapon_Switch` client side.

I think this logic should be moved client side as much as possible. In this case, it's not really the switching of the weapon we're concerned about, but the fact the player has materialized in the word - so we could have the the client side player listening for the `player_spawn` event, and if it's their userid that fired it, do the whatever muzzle flash tweaks from there. This could be as simple as `ListenForGameEvent("player_spawn")` in `C_NEO_Player` ctor, and the corresponding `FireGameEvent` override:

```cpp
void C_NEO_Player::FireGameEvent(IGameEvent *event)
{
	if (FStrEq(event->GetName(), "player_spawn"))
	{
		if (event->GetInt() == GetUserID())
		{
			UpdateMuzzleFlashProperties(GetActiveWeapon());
		}
	}
}
```

But I think it's maybe outside the scope of this PR, and might be more suited as part of #559 , since it's the feature relying on this stuff.

Rainyan:
Actually, looks like even the event listener fires too early (at least on LAN). Well, I'll mess around with a think based approach tomorrow.

AdamTadeusz:
I merged in the view model muzzle flash PR, should make it easier to see what solutions we can come up with for this problem work. Having the initial muzzle flash be a little broken is not a big problem so I suggest merging this in too

AdamTadeusz (approved):

nullsystem (approved):

PR #599, opened 2024-09-17
Settings changed popup - Replace discard with cancel

### LINKED ISSUE #590: [FEATURE] Cancel button in the Root Menu's Options
### Description

Currently, when exiting the options after having modified something but not applied, this screen is shown:
image
As it stands, it's impossible to go back to options. You can only either save the modified settings (which you might not even remember) or discard them.

### Proposed Solution

image
I propose adding a "Cancel" button, with which the user will be able to go back to the options in case they have accidentally clicked the back button, or if they had forgotten/want to verify what they modified.
Additionally, since discarding is a destructive action, I think it's better to have the escape key focus the new "Cancel" button, so that in case the user that accidentally goes back and panics and presses the escape button doesn't erase all of their modifications.

### Alternatives Considered

_No response_

### Additional Context

_No response_

nullsystem:
## Description
* Basically go back to settings instead, in which settings have a "Restore" button anyway

## Toolchain

## Linked Issues

* fixes #590

### DIFF
--- a/mp/src/game/client/neo/ui/neo_root.cpp
+++ b/mp/src/game/client/neo/ui/neo_root.cpp
@@ -641,7 +641,7 @@ void CNeoRoot::MainLoopSettings(const MainLoopParam param)
 				NeoUI::Pad();
 				if (m_ns.bModified)
 				{
-					if (NeoUI::Button(L"Restore").bPressed)
+					if (NeoUI::Button(L"Revert").bPressed)
 					{
 						NeoSettingsRestore(&m_ns);
 					}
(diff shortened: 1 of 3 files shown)


Rainyan (approved):
The "Restore" button in the settings menu should perhaps be renamed "Discard current changes" or similar - I assumed it would restore the defaults, but that does not appear to be the case.

It would also be nice to have a "Restore defaults" button, but that's not strictly related.

nullsystem:
> The "Restore" button in the settings menu should perhaps be renamed "Discard current changes" or similar - I assumed it would restore the defaults, but that does not appear to be the case. This could leave a user confused, where they cannot exit the menu because of pending changes, but they don't want to press the "Restore" button, fearing it'll override everything back to default instead of just discarding the pending changes.
> 
> It'd also be nice to have a "Restore defaults" button, but that's not strictly related.

@Rainyan I would like to try to keep the button short though. Other alternatives like "Reset", "Discard", or "Cancel", although both "Discard"/"Cancel" does sound more like it'll also exit the settings page.

nullsystem:
@xedmain Alright discard back in now

Rainyan (approved):
LGTM.

I'm neutral on the "Discard" button for the popup, but the "Revert" wording is much clearer as to what the button does.

AdamTadeusz (approved):

PR #607, opened 2024-09-18
End reload sound when switching with fast switch enabled

### LINKED ISSUE #606: Player reload sound persists when switching weapons
If the player begins reloading a weapon and then switches to another weapon while using fast switch, the reload sound persists even though the weapon was not reloaded. This does not occur when using the slower selection process as the sounds that the slower weapon selection hud emits play on the same channel "CHAN_ITEM" as the reload sound, effectively cancelling it. An easy solution to halt the reload sound would be to play Player.WeaponSelected when selecting a weapon even with fastswitch, although weapon switching in the original is completely quiet so not really parity, however hearing the sound of a weapon being equipped when using slow switch but not fast switch is also not very consistent.

Alternatively can also use the StopSound command, passing in the channel and name of reload sound.

Note this is also a problem in the original, but I consider the original behaviour a bug that should be fixed. Audibly informing players that their weapon has been reloaded by playing the entire reload sound but not actually reloading their weapon is bad design.

AdamTadeusz:
## Description

Plays WeaponSelected if weapon was succesfully selected using FastWeaponSwitch. Semantically makes sense to play WeaponSelected where DenyWeaponSelection is already played if FastWeaponSwitch returns false, however although the sound WeaponSelected is already played for the other weapon selection methods, it is never audible as in all other cases CloseWeaponSelection is played afterwards in the same channel, overwriting this sound.

If the same sound should be heard regardless of weapon selection method then either the sound added here should be replaced with CloseWeaponSelection which is audibly much louder to the player, is not heard ever when using the default weapon switch method in the original, and is a sound closely associated with the source engine (and perhaps should therefore be avoided). The alternative therefore would be to remove the CloseWeaponSelection sound from the other weapon selection methods, which would make the WeaponSelected sound that is also played in those instances audible.

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes  #606

### DIFF
--- a/mp/src/game/client/hl2/hud_weaponselection.cpp
+++ b/mp/src/game/client/hl2/hud_weaponselection.cpp
@@ -1340,6 +1340,9 @@ void CHudWeaponSelection::FastWeaponSwitch( int iWeaponSlot )
 	{
 		// select the new weapon
 		::input->MakeWeaponSelection( pNextWeapon );
+#ifdef NEO
+		pPlayer->EmitSound("Player.WeaponSelected");
+#endif
 	}
 	else if ( pNextWeapon != pActiveWeapon )
 	{


AdamTadeusz:
Fixed by #610, although maybe should still consider the inconsistencies in sounds heard when switching weapons using the different methods

PR #610, opened 2024-09-18
Spatial reload sounds, stop reload sounds when dropping or switching weapon, fix for weapons reloading on the floor

### LINKED ISSUE #613: Reload sound from other people playing inside head
Regardless of where the person doing a reload is, the person listening hears the sound inside their head, with the only difference being the volume of the reload varies with distance to actual source of sound

### LINKED ISSUE #614: When initiating a reload and then dropping that weapon on the floor, picking it up will complete the reload

### LINKED ISSUE #615: Reload sound persists through weapons switching and dropping
When dropping or switching a weapon, the reload sounds persists, except for the person initiating the reload when using the mousewheel or slowswitch to switch the weapon

AdamTadeusz:
## Description

reload_npc is the same reload sound as reload except with the character that allows for spatial sounds. 

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #613
- fixes #614
- fixes #615


Rainyan (approved):
:+1:

mp/src/game/server/neo/neo_player.cpp
```diff
@@ -1899,6 +1904,7 @@ bool CNEO_Player::Weapon_CanSwitchTo(CBaseCombatWeapon *pWeapon)
     {
         if (!GetActiveWeapon()->CanHolster())
             return false;
+		GetActiveWeapon()->StopWeaponSound(RELOAD_NPC);
```
Masterkatze, line 1907: Bad indentation lol

AdamTadeusz: thats a github problem I think. It's supposed to be an early return if GetActiveWeapon is null

Rainyan: It's a tabs vs spaces problem; line 1907 uses tabs whereas the above two lines are space-indented. Most IDEs format it automagically so it doesn't look hideous. Although it might be worth converting that whole function to tabs, for consistency.

AdamTadeusz: ill add some brackets around it too

Rainyan (approved):

Masterkatze (approved):

brysondev (approved):

PR #616, opened 2024-09-19
Weapon Selection Available 15 seconds after round is live, 25 seconds after spawning

### LINKED ISSUE #287: Players are allowed to select their initial weapon outside of freezetime
### Description

In OGNT, the player does not spawn until they select a weapon. After roughly 30 seconds (including freeze time) the player gets forcefully spawned with the weapon they played with last round.
In NT;RE, you can only select and receive a weapon during freeze time.

### Issue

With the way the weapon selection works currently, people who aren't paying attention or are AFK during the freeze time will only get their secondary, which has proven to be a confusing and annoying mechanic.

### Proposed solution

Return the weapon selection to back how it is in OGNT

AdamTadeusz:
## Description

Useful for gamemodes like tdm also where respawns are enabled

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #287

### DIFF
--- a/mp/src/game/server/neo/neo_player.cpp
+++ b/mp/src/game/server/neo/neo_player.cpp
@@ -58,7 +58,7 @@ SendPropBool(SENDINFO(m_bInVision)),
 SendPropBool(SENDINFO(m_bHasBeenAirborneForTooLongToSuperJump)),
 SendPropBool(SENDINFO(m_bShowTestMessage)),
 SendPropBool(SENDINFO(m_bInAim)),
-SendPropBool(SENDINFO(m_bDroppedAnything)),
+SendPropBool(SENDINFO(m_bIneligibleForLoadoutPick)),
 
 SendPropTime(SENDINFO(m_flCamoAuxLastTime)),
 SendPropInt(SENDINFO(m_nVisionLastTick)),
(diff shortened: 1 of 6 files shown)


mp/src/game/server/neo/neo_player.cpp
```diff
@@ -2533,7 +2533,7 @@ void CNEO_Player::GiveLoadoutWeapon(void)
 {
 	const NeoRoundStatus status = NEORules()->GetRoundStatus();
 	if (!(status == NeoRoundStatus::Idle || status == NeoRoundStatus::Warmup) && 
-		(IsObserver() || IsDead() || m_bDroppedAnything || ((NEORules()->GetRemainingPreRoundFreezeTime(false)) < 0)))
+		(IsObserver() || IsDead() || m_bIneligibleForLoadoutPick || m_aliveTimer.IsGreaterThen(25.f)))
```
Masterkatze, line 2536: Is it fine to hardcode this value?

AdamTadeusz: Are you complaining about magic numbers or saying I should add a convar

Masterkatze: Magic number - yes, convar - dunno, maybe? We need an another opinion.

brysondev: Probably would be nice as a convar, but personally I don't see a use for it in any scenario.

AdamTadeusz: in some other game modes might be worth changing this settings, Ill make it a convar

Masterkatze (approved):

brysondev (approved):

PR #636, opened 2024-09-22
NeoUI - Console toggle key rebind, intended for non-` key layouts

### LINKED ISSUE #620: NeoUI Console - Improve keybind fix
### Build Info

v8.1-prealpha

### Description

Follow up from: https://github.com/NeotokyoRebuild/neo/pull/589#issuecomment-2360454261

Basically the console wasn't check if it's opened or not, so ` ends up only being used for toggling but not able to use for text input.

Another issue is there are keyboard layouts with `, so the current keybind workaround won't work for those layouts.

### To Reproduce

* Can't type ` into console
    1. Open up the console
    2. Try to use `
* Keyboard layout without `, EX: nordic layout
    1. Try to open console where the physical key usually is

### Expected behavior

* Can't type ` into console
    * Be able to type ` into the console
* Keyboard layout without `, EX: nordic layout
    * Be able to open the console

### Actual behavior

* Can't type ` into console
    * It closes the console
* Keyboard layout without `, EX: nordic layout
    * Cannot open the console

### Operating System

- [x] Windows
- [x] Linux

### Version/Distro

Windows + Linux

nullsystem:
## Description
* To workaround console not working for non-BACKQUOTE key layouts, allow the rebinding of the console to any key the user wants. This is also beneficial for people who also just want to put console to a separate key.
* Non-BACKQUOTE keys however requires the root menu to handle the key and toggleconsole itself, whereas BACKQUOTE key will always envoke. The root menu will handle it if it's active without a console opened. Otherwise the engine will envoke it even for non-BACKQUOTE itself if you're in-game without the root menu opened or the console is already opened.
* Nevra, who uses FR AZERTY confirmed this works well for him

## Toolchain

- Windows MSVC VS2022

## Linked Issues

* fixes #620

### DIFF
--- a/mp/src/game/client/neo/ui/neo_root.cpp
+++ b/mp/src/game/client/neo/ui/neo_root.cpp
@@ -372,6 +372,21 @@ void CNeoRoot::FireGameEvent(IGameEvent *event)
 
 void CNeoRoot::OnRelayedKeyCodeTyped(vgui::KeyCode code)
 {
+	if (m_ns.keys.bcConsole <= KEY_NONE)
+	{
+		m_ns.keys.bcConsole = gameuifuncs->GetButtonCodeForBind("neo_toggleconsole");
+	}
+
+	if (code == m_ns.keys.bcConsole && code != KEY_BACKQUOTE)
+	{
+		// NEO JANK (nullsystem): Prevent toggle being handled twice causing it to not really open.
+		// This can happen if using the default ` due to the engine enacting this all the time, however calling
+		// NeoToggleconsole here is required for non-` keys that is toggled while the root panel is opened. The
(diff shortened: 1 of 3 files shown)


AdamTadeusz:
Can't toggle the console off with a rebound key under any situation, in menu, in game or in game with the menu open.

Neither is this possible when using "bind x toggleconsole" originally however so I'm guessing this is something that can't be fixed?

AdamTadeusz (approved):

nullsystem:
@AdamTadeusz I think it depends on what key you rebind. For example F5 worked well but "\" just relays to the console. In this case not much that can be done since the console out of our control, but it's good enough to workaround "`" issue.

brysondev (approved):

PR #638, opened 2024-09-22
Alter lean speed/angle

### LINKED ISSUE #450: Leaning animation shouldn't take longer than OG NT
### Build Info

20240629_321f139

### Description

It's taking too long from pushing the button to full lean.

### To Reproduce

Press any of the lean buttons and observe the screen tilting.

### Expected behavior

The duration until full tilt should be the same as in OG NT (someone will have to time this).

### Actual behavior

It's slower than OG NT.

### Operating System

- [X] Windows
- [ ] Linux

### Version/Distro

Windors 10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

nullsystem:
## Description
Alter the lean speed/angle to make it seem closer to OG:NT? Although it ain't exact as lean is done differently.
Also to note this was from like 2 months ago just rebased now. Only been notified to try for a PR.

## Toolchain

## Linked Issues

* fixes #450

### DIFF
--- a/mp/src/game/shared/neo/neo_predicted_viewmodel.cpp
+++ b/mp/src/game/shared/neo/neo_predicted_viewmodel.cpp
@@ -140,9 +140,8 @@ void CNEOPredictedViewModel::DrawRenderToTextureDebugInfo(IClientRenderable* pRe
 #endif
 
 ConVar neo_lean_debug_draw_hull("neo_lean_debug_draw_hull", "0", FCVAR_CHEAT | FCVAR_REPLICATED);
-ConVar neo_lean_speed("neo_lean_speed", "0.333", FCVAR_REPLICATED | FCVAR_CHEAT, "Lean speed scale", true, 0.0, false, 1000.0);
-// Original Neotokyo with the latest leftlean fix uses 7 for leftlean and 15 for rightlean yaw slide.
-ConVar neo_lean_yaw_peek_left_amount("neo_lean_yaw_peek_left_amount", "7.0", FCVAR_REPLICATED | FCVAR_CHEAT, "How far sideways will a full left lean view reach.", true, 0.001f, false, 0);
+ConVar neo_lean_speed("neo_lean_speed", "0.2", FCVAR_REPLICATED | FCVAR_CHEAT, "Lean speed scale", true, 0.0, false, 1000.0);
+ConVar neo_lean_yaw_peek_left_amount("neo_lean_yaw_peek_left_amount", "15.0", FCVAR_REPLICATED | FCVAR_CHEAT, "How far sideways will a full left lean view reach.", true, 0.001f, false, 0);
 ConVar neo_lean_yaw_peek_right_amount("neo_lean_yaw_peek_right_amount", "15.0", FCVAR_REPLICATED | FCVAR_CHEAT, "How far sideways will a full right lean view reach.", true, 0.001f, false, 0);
 ConVar neo_lean_angle_percentage("neo_lean_angle_percentage", "0.75", FCVAR_REPLICATED | FCVAR_CHEAT, "for adjusting the actual angle of lean to a percentage of lean.", true, 0.0, true, 1.0);
 ConVar neo_lean_tp_exaggerate_scale("neo_lean_tp_exaggerate_scale", "2", FCVAR_REPLICATED | FCVAR_CHEAT, "How much more to scale leaning motion in 3rd person animation vs 1st person viewmodel.", true, 0.0, false, 0);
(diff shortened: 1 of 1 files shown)


AdamTadeusz:
When I suggested this be opened I didn't check whether the values here and resultant behavior were actually correct. Playing around, starting a lean from the same position and angle I notice that our camera in the original has a final roll of 20 degrees in either direction depending on direction on lean. However while adjusting the final body yaw does indeed increase the final camera roll, it primarily increases how much we can see, as shown by the x part of the pos. a value of 20 which almost gets us the right camera roll, and a value of 15 in this PR, both make the peak way too strong.
image
I think, after doing some testing, that the final yaw_peek left and right amounts should both be 7.5 default, and the function needs to be either changed to increase the camera roll at a rate of (20/7.5) 2.666 higher to the body yaw so the final camera roll is 20 when the upper body yaw is 7.5 (looking at the body in thirdperson I guess the body_yaw variable controls how far the camera moves perpendicular to the direction we are looking, although the way that is achieved is the final body roll is changed depending on the body_yaw variable)

AdamTadeusz:
Also I should add that if rolling the upper body 7.5 degrees either way results in exactly the same amount of stuff being visible to the player, and therefore the same amount of the player being exposed, rolling the camera a further 2.66 times to get the same camera angle is kinda pointless. It just makes the screen more sideways without giving the player any more information.

Unless of course the reason why the final camera roll in the original is 20 degrees in either direction is because the character in the original doesn't roll his entire upper body, but only like the top part of his upper body, in which case he can see the same amount of stuff while exposing less of his body, although the camera rotated that much more probably makes it a bit more awkward to aim

AdamTadeusz:
image
Here's what I mean. Both players can see pretty much exactly the same scene, except the image is rolled over more for the original player, so I can only assume the reason is that the original player isnt rotating his body around the waist but instead higher up, maybe around the breastbone area

AdamTadeusz:
image
Ok the bots seem to roll from their waist, so it's probably just a case of the camera roll being exaggerated. Using cl_showpos the body yaw actually moves 7.5 to the left when left peeking and 15 to the right when right peeking. Rotating the camera further around the camera's origin can be done to mimick the original behavior, but its probably only a hindrance to players as it doesn't expose them any more or give them any more info, just makes the screen closer to portrait mode

Agiel:
Less roll is probably better because it's less disorientating as a player. What's important is that the visibility is the same and that the first person and third person matches up so you can't see around a corner without exposing yourself.

AdamTadeusz:
> Less roll is probably better because it's less disorientating as a player. What's important is that the visibility is the same and that the first person and third person matches up so you can't see around a corner without exposing yourself.

I agree, that's why I personally wouldn't bother implementing the camera roll that's present in the base game, and instead keep the camera roll fixed to the body roll. We should probably open a discussion about this and make an entry in the design doc

mp/src/game/shared/neo/neo_predicted_viewmodel.cpp
```diff
@@ -140,9 +140,8 @@ void CNEOPredictedViewModel::DrawRenderToTextureDebugInfo(IClientRenderable* pRe
 #endif
 
 ConVar neo_lean_debug_draw_hull("neo_lean_debug_draw_hull", "0", FCVAR_CHEAT | FCVAR_REPLICATED);
-ConVar neo_lean_speed("neo_lean_speed", "0.333", FCVAR_REPLICATED | FCVAR_CHEAT, "Lean speed scale", true, 0.0, false, 1000.0);
-// Original Neotokyo with the latest leftlean fix uses 7 for leftlean and 15 for rightlean yaw slide.
-ConVar neo_lean_yaw_peek_left_amount("neo_lean_yaw_peek_left_amount", "7.0", FCVAR_REPLICATED | FCVAR_CHEAT, "How far sideways will a full left lean view reach.", true, 0.001f, false, 0);
+ConVar neo_lean_speed("neo_lean_speed", "0.2", FCVAR_REPLICATED | FCVAR_CHEAT, "Lean speed scale", true, 0.0, false, 1000.0);
```
AdamTadeusz, line 143: I think you just made the lean speed slower here. A speed of 0.4 is one frame away from the lean coming to a full stop at the same time as in the original when recorded at host_timescale 0.1

AdamTadeusz: image

AdamTadeusz: since its using lerp we're actually very close to the end of the lean a few frames before 01:10, but 01:10 is when cl_showpos shows no more changes to our position

AdamTadeusz (changes requested):

mp/src/game/shared/neo/neo_predicted_viewmodel.cpp
AdamTadeusz: keep in mind this was recorded with neo_lean_yaw_peek_left_amount 7.5, but im fairly confident that's the correct value, I explained my reasoning earlier

AdamTadeusz: damn im playing this back at host_timescale 1 and its so quick now ;(

PR #648, opened 2024-09-24
NeoUI - Settings add reset to default button + Fix settings applying

### LINKED ISSUE #612: Root NeoUI - Settings missing "Default" reset button
There isn't a button to reset all the settings back to the default ones. Although when this button is clicked, make sure to have a popup dialog so it guards from the user mistakenly clicking on it and double-check if they really want to reset to default.

### LINKED ISSUE #634: Client settings desynchronised
It's possible to de-synchronise a users settings such that the server thinks a client has one value for a convar while that client thinks it has another value for that convar

For example, when enabling auto-reload when empty in the new settings menu but having it disabled through the legacy console via cl_autoreload_when_empty 0, this happens

https://github.com/user-attachments/assets/2ed89d98-ea22-4257-bba2-c6d546e91e3e

You can't quite see it in the recording but in the options menu auto reload is enabled

nullsystem:
## Description
* Have a popup to confirm if the player wants to reset it to default
* Fixes issues with non-ConVarRef's, so turning them to ConVarRef to properly apply the settings

## Toolchain

## Linked Issues

* fixes #612
* fixes #634


brysondev (approved):
lgtm

brysondev (approved):

Masterkatze (approved):

PR #649, opened 2024-09-25
Throwables origin clipping into ground workaround

### LINKED ISSUE #233: Picking up Grenades and Detpack

AdamTadeusz:
## Description

give viewoffset so origin always appears above ground when dropped when performing a cansee calculation

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #233

### DIFF
--- a/mp/src/game/shared/neo/weapons/weapon_detpack.cpp
+++ b/mp/src/game/shared/neo/weapons/weapon_detpack.cpp
@@ -84,6 +84,7 @@ CWeaponDetpack::CWeaponDetpack()
 #ifdef GAME_DLL
 	m_pDetpack = NULL;
 #endif
+	SetViewOffset(Vector(0, 0, 1.0));
 }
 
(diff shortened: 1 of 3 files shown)


AdamTadeusz:
It's easiest to see the change with a detpack. In this video a change was made with hotreload so that the first smoke grenade and first detpack were spawned without the fix, and the next smoke grenade and detpack were spawned with the fix. As you can see I can pick up the detpack when its origin is at the same height above(below) ground as the first detpack. Harder to get the grenade in the right orientation but it works with the two grenades too.

https://github.com/user-attachments/assets/7be0c142-ec71-4b44-93dd-2bfcaae8808d

nullsystem (approved):

brysondev (approved):

PR #662, opened 2024-09-30
Detpack correct sounds

### LINKED ISSUE #584: Sounds for specific actions are not audible at times
### Build Info

20240914_b2291c5

### Description

Some actions do not play a sound when they should be playing one, e.g an explosion from a nade or when picking up the ghost

### To Reproduce

1. Have teammate/yourself/enemy pick up the ghost or have someone throw a nade

### Expected behavior

Sounds for the respective actions should play

### Actual behavior

They do not

### Operating System

- [X] Windows
- [ ] Linux

### Version/Distro

win 10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

MSVC 19.41.3

_No response_

AdamTadeusz:
## Description

Increase main sound cut-off point, use correct explosion sound for detpack

Sounds in the original game can be heard from up to 10k hu away. This is 10 times larger than the current sound cutoff point.

Currently the detpack also uses an explosion sound with a special parameter that switches to a weaker distant explosion in the sounds right channel when the character is a set distance away from the explosion. The original detpack's explosion sound doesn't change much even out to 10k hu.

https://github.com/user-attachments/assets/64dacc92-6c44-4c82-915e-d6a361849581

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #584

### DIFF
--- a/mp/src/game/server/neo/neo_detpack.cpp
+++ b/mp/src/game/server/neo/neo_detpack.cpp
@@ -283,10 +283,10 @@ int CNEODeployedDetpack::OnTakeDamage(const CTakeDamageInfo& inputInfo)
 void CNEODeployedDetpack::InputRemoteDetonate(inputdata_t& inputdata)
 {
 	ExplosionCreate(GetAbsOrigin() + Vector(0, 0, 16), GetAbsAngles(), GetThrower(), GetDamage(), GetDamageRadius(),
-					SF_ENVEXPLOSION_NOSPARKS | SF_ENVEXPLOSION_NODLIGHTS | SF_ENVEXPLOSION_NOSMOKE, 0.0f, this);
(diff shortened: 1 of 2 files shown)


AdamTadeusz:
The original game uses a cpas filter also to work out whether the explosion is in a visleaf that can be heard by the player. Just because the sound cut off point is 10k doesnt mean that all explosions on all levels will be audible now. For example on oilstain, a detpack in the blue spawn cannot be heard from the middle of the parking garage either in the original or on this branch, as there are too many invisible visleafs between the ones visible to the player and the one with the explosion.

brysondev (approved):

Masterkatze (approved):

PR #665, opened 2024-10-02
#635 Cloak weapons when player is cloaked

### LINKED ISSUE #635: Weapons not cloaked when character is cloaked

VantoNortim:
## Description
Cloak weapons when owner is cloaked

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #635

### DIFF
--- a/mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
+++ b/mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
@@ -23,6 +23,7 @@ extern ConVar weaponstay;
 
 // memdbgon must be the last include file in a .cpp file!!!
 #include "tier0/memdbgon.h"
+#include <model_types.h>
 
 IMPLEMENT_NETWORKCLASS_ALIASED( NEOBaseCombatWeapon, DT_NEOBaseCombatWeapon )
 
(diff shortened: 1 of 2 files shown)


AdamTadeusz (changes requested):
Not sure whether we need to be drawing cloaked models twice, but that's something that can be changed later in one go, I have a NEO TODO in a different PR about it already

Just remove the unused import if it's not required, or explain what it's doing and I'm happy.

mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
```diff
@@ -1016,8 +1017,47 @@ int CNEOBaseCombatWeapon::DrawModel(int flags)
 	C_BaseCombatCharacter* localPlayer = C_BasePlayer::GetLocalPlayer();
 	if (GetOwner() == localPlayer && ShouldDrawLocalPlayerViewModel())
 		return 0;
+	
+	auto ret = BaseClass::DrawModel(flags);
 
-	return BaseClass::DrawModel(flags);
+	auto owner = dynamic_cast<C_NEO_Player*>(GetOwner());
```
AdamTadeusz, line 1023: we're using a lot of static_cast in many other places, I think this is a safe assumption now.

PR #678, opened 2024-10-05
Change NVG Shader

### LINKED ISSUE #135: NVGs should work as well as in OGNT
This one is a little hard to quantify, but in rebuild there are fewer distinct shades of color on screen at all time.

map nt_tarmac_ctg 
sv_cheats 1
setpos -272.583618 552.612244 -475.968
setang 5.016034 -15.808

OG

Image

Rebuild

Image

Let's call if "done" when:

1. Base color is dark grey instead of green
2. Highlights/lighter areas become whiter/lighter grey instead of darker

Does that make sense?

AdamTadeusz:
## Description

NVG shader that more closely matches the original. Really dark areas are maybe not as light as they should be, but bright areas are reaching the limit where it hurts the eyes, might need to figure out some curve and method to boost mainly low brightness luminescence

Original:
image
Old:
image
New:
image

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #135

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -216,3 +216,4 @@ mp/src/game/shared/neo/neo_version_info.cpp
 
 # News cache
 mp/game/neo/news.txt
+/mp/src/materialsystem/stdshaders/buildhl2mpshaders.bat
--- a/mp/src/game/shared/neo/weapons/weapon_knife.cpp
+++ b/mp/src/game/shared/neo/weapons/weapon_knife.cpp
@@ -297,14 +297,9 @@ void CWeaponKnife::Hit(trace_t& traceHit, [[maybe_unused]] Activity nHitActivity
 
 		static constexpr int damageToOneShotSupport = (100 * (1 / NEO_SUPPORT_DAMAGE_MODIFIER)) + 1;
 
-		CTakeDamageInfo info(GetOwner(), GetOwner(), KNIFE_DAMAGE, DMG_SLASH);
+		CTakeDamageInfo info(GetOwner(), GetOwner(), (currentAngle > maxBackStabAngle ? KNIFE_DAMAGE : damageToOneShotSupport), DMG_SLASH);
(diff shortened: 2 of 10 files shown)


brysondev (dismissed):

.gitignore
```diff
@@ -216,3 +216,4 @@ mp/src/game/shared/neo/neo_version_info.cpp
 
 # News cache
 mp/game/neo/news.txt
+/mp/src/materialsystem/stdshaders/buildhl2mpshaders.bat
```
Rainyan, line 219: Possibly a typo, since changes to `mp/src/materialsystem/stdshaders/buildhl2mpshaders.bat` are also included in this PR. Did you mean to ignore the path without the leading `/`?

AdamTadeusz: I changed the file before noticing it should be added to gitignore so I assumed the change would be added regardless, but maybe instead this is the reason why, lemme check

AdamTadeusz: I remember adding this file to the gitignore by right clicking it in visual studios Git Changes, which is why its placed so haphazardly and has the leading slash. I removed it and moved the entry to the other shader entries. I don't think the slash has any effect

AdamTadeusz:
I've taken to adding all the shaders to the same files the existing neo shaders used, so actually I can revert the changes to three files hold on

AdamTadeusz:
Apologies, should have done these from the start

brysondev (approved):

Rainyan (approved):
LGTM

AdamTadeusz:
> LGTM
> 
> You should probably create an accompanying PR into the assets repo for the compiled shaders.
> 
> I could look into creating some kind of PR workflow for automating this.

Link's at the top of the description, should probably make it more obvious

PR #679, opened 2024-10-06
Adjust Uplink element for larger resolutions

brysondev:
## Description

Currently, any resolution exceeding 1440 pixels horizontally will result in the uplink icon clipping below the compass rose on the HUD. This aims to mitigate this issue for up to 4K (my max resolution).

## Toolchain

- Windows MSVC VS2022

4K (3840x2160) before changes
image

4K (3840x2160) after changes
Preview of 4K (3840x2160) after changes

## Notes
I have also tested swapping resolutions and watching the `vgui_drawtree` highlight to see if it stays consistent. Not sure if there are any other scenarios I might need to try out but it seems fine.

### DIFF
--- a/mp/src/game/client/neo/ui/neo_hud_ghost_uplink_state.cpp
+++ b/mp/src/game/client/neo/ui/neo_hud_ghost_uplink_state.cpp
@@ -51,7 +51,10 @@ void CNEOHud_GhostUplinkState::ApplySchemeSettings(vgui::IScheme* pScheme)
 	surface()->GetScreenSize(screenWidth, screenHeight);
 	int centerX = screenWidth / 2;
 	int textureXPos = centerX - (m_iUplinkTextureWidth / 2);
-	static constexpr int COMPASS_HEIGHT_PLUS_MARGINS = 30;
+	//NEOJANK <1440p resolutions have the uplink element clip into compass (Bryson)
+	static int COMPASS_HEIGHT_PLUS_MARGINS = 30;
+	if (screenHeight > 1440)
+		COMPASS_HEIGHT_PLUS_MARGINS = 45;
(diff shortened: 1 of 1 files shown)


AdamTadeusz:
The compass and other ui elements could do with a rework, then we can set the compass's bounds to fit tightly around the compass itself, then inside of the uplink ui element's ApplySchemeSettings grab the compass from gHUD, cast to a Panel, grab the bounds and use those to position the uplink accordingly.

If we don't want to do all that, then the following is a viable alternative:

Dimensions of the compass box
```
DrawNeoHudRoundedBox(
	resXHalf - xBoxWidthHalf, m_resY - yBoxHeight - margin,
	resXHalf + xBoxWidthHalf, m_resY - margin,
	m_boxColor);
```

We're interested in the value of y0, which is the second value passed to this function
We can get the value of m_resY in uplinkstate no problem.
yBoxHeight is the height of the font used in the compass. Can grab the size of the same font in uplinkstate no problem.
margin is a value  grabbed from HudLayout.res, with a default coded in value of 3, adjusted for current screen size. Assuming this value never changes (whithout which this workaround doesn't work, unless you can find a way of grabbing this value from HudLayout.res which is not obvious for a vgui element with a different name), the value we need is simply (currentScreenSize/480) * 3

These three values plus an additional custom margin value between the compass and uplink state (which we could also make proportional to the current screen size), are all you need to position the uplink state perfectly regardless of resolution.

AdamTadeusz:
Some other nitpicks:

Since COMPASS_HEIGHT_PLUS_MARGINS is no longer a constant value, it shouldn't be written in all caps

COMPASS_HEIGHT_PLUS_MARGINS also did not need to be a static to begin with, that's my bad

AdamTadeusz (approved):

nullsystem (approved):

PR #695, opened 2024-10-12
Communicate player label in real time

### LINKED ISSUE #137: Spectated players health points is communicated to spectators in real time
Description:
When spectating a player, the amount of health showing up next to their name in the spectator overlay is slow to update, it's delayed by a few seconds. Or maybe I am misremembering and it never updates as long as you spectate them. But if I remember right, it does update, just slowly.
To reproduce:

spectate a player as they are taking damage.

Expected outcome: 

The HP of the player being spectated decreases as soon as they get hit

Actual outcome:

The HP stays the same for a few seconds or until you swap back and forth with another player.

AdamTadeusz:
## Description

Update player label on think

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #137

### DIFF
--- a/mp/src/game/client/game_controls/SpectatorGUI.cpp
+++ b/mp/src/game/client/game_controls/SpectatorGUI.cpp
@@ -590,6 +590,12 @@ void CSpectatorGUI::OnThink()
 			V_sprintf_safe(scoreBuff, "%d", Max(0, Min(99, pNsfTeam->GetRoundsWon())));
 			m_scoreValueLabelNSF->SetText(scoreBuff);
 		}
+
+		// Update player health
+		if (m_pPlayerLabel->IsVisible())
+		{
+			UpdatePlayerLabel();
+		}
 #endif
 	}
 }
@@ -709,7 +715,9 @@ void CSpectatorGUI::Update()
 	m_pPlayerLabel->SetVisible( ShouldShowPlayerLabel(specmode) );
 
 	// update player name filed, text & color
-
+#ifdef NEO
+	UpdatePlayerLabel();
+#else
 	if ( playernum > 0 && playernum <= gpGlobals->maxClients && gr )
 	{
 		Color c = gr->GetTeamColor( gr->GetTeam(playernum) ); // Player's team color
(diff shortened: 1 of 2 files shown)


AdamTadeusz:
Code for UpdatePlayerLabel() taken from ~line 721

brysondev (approved):

mp/src/game/client/game_controls/SpectatorGUI.cpp
```diff
@@ -783,6 +792,47 @@ void CSpectatorGUI::UpdateTimer()
...
+	int playernum = GetSpectatorTarget();
+
+	// update player name filed, text & color
+	if (playernum > 0 && playernum <= gpGlobals->maxClients && gr)
+	{
+		Color c = gr->GetTeamColor(gr->GetTeam(playernum)); // Player's team color
+
+		m_pPlayerLabel->SetFgColor(c);
+
+		wchar_t playerText[80], playerName[64], health[10];
```
Rainyan, line 812: `playerName[MAX_PLAYER_NAME_LENGTH+1]` should suffice - since `MAX_PLAYER_NAME_LENGTH` equals 32, allocating 64*sizeof(wchar_t) bytes is probably more than we need here?

Could also zero-initialize statically as `playerName[MAX_PLAYER_NAME_LENGTH+1]{0}` instead of using the memset, although perhaps the compiler will optimise it regardless.

Rainyan (changes requested):

mp/src/game/client/game_controls/SpectatorGUI.cpp
brysondev: image

But I guess it would make more sense to use that?

Rainyan: Well, looking at the release disassembly it gets optimized into a single stack allocation, so at most there's gonna be a 32 byte padding there which I suppose isn't a big deal.

Rainyan (approved):

PR #700, opened 2024-10-12
Fix possible crash on exit - NeoLoading

nullsystem:
Sometimes it can crash on exit due to a direct delete. Even then it kind of doesnt even matter much if delete or not since the lifetime of this is practically once whole session and OS cleans up on exit anyway.


brysondev (approved):
🙏

Masterkatze (approved):

PR #718, opened 2024-10-13
Damage/killer info duration 60s->10s + Dismiss forever

### LINKED ISSUE #717: Damage/killer info menu duration is too long
### Build Info

v9.0-prealpha

### Description

The current 60s duration till it automatically close is too long. Maybe more like 10s is enough. Could also add an option to dismiss for the rest of the match.

### To Reproduce

1. Get killed by someone (or wait till round end by timer)
2. See the damage/killer info menu
3. Don't interact with it

### Expected behavior

The damage/killer info menu goes away in a reasonable amount of time.

### Actual behavior

It takes 60s for it to automatically go away.

### Operating System

- [X] Windows
- [X] Linux

### Version/Distro

Windows 10 + Arch Linux

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

nullsystem:
## Description
* 60s was too long, turned down to 10s
* Add the ability to dismiss and not have the info for the rest of the match

## Toolchain

## Linked Issues

* fixes #717

### DIFF
--- a/mp/src/game/server/neo/neo_player.cpp
+++ b/mp/src/game/server/neo/neo_player.cpp
@@ -111,7 +111,7 @@ DEFINE_FIELD(m_szNameDupePos, FIELD_INTEGER),
 DEFINE_FIELD(m_bClientWantNeoName, FIELD_BOOLEAN),
 END_DATADESC()
 
-#define SHOWMENU_STRLIMIT (512)
+static constexpr int SHOWMENU_STRLIMIT = 512;
 
 CBaseEntity *g_pLastJinraiSpawn, *g_pLastNSFSpawn;
 CNEOGameRulesProxy* neoGameRules;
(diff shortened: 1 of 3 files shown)


brysondev (dismissed):

mp/src/game/server/neo/neo_player.cpp
```diff
@@ -1298,8 +1308,8 @@ void CNEO_Player::ShowDmgInfo(char* infoStr, int infoStrSize)
 		}
 		UserMessageBegin(filter, "ShowMenu");
 		{
-			WRITE_SHORT(static_cast<short>(m_iDmgMenuNextPage ? 3 : 1)); // Amount of options
-			WRITE_CHAR(static_cast<char>(60)); // 60s timeout
+			WRITE_SHORT(static_cast<short>(((m_iDmgMenuNextPage ? 0b11 : 1) | 1 << 8))); // Amount of options
```
Rainyan, line 1311: What does the magic bitwise math do here? Any chance we could have a constant expression for this to make the intent of the passed data more obvious?

mp/src/game/server/neo/neo_player.cpp
```diff
@@ -1204,9 +1204,14 @@ bool CNEO_Player::ClientCommand( const CCommand &args )
 			switch (slot)
 			{
 			case 1: // Dismiss
+			case 9: // Do not show again
```
Rainyan, line 1207: Could we have an enumeration instead of magic values + comments, to make this clearer

mp/src/game/server/neo/neo_player.cpp
```diff
@@ -1321,7 +1331,7 @@ int CNEO_Player::SetDmgListStr(char* infoStr, const int infoStrMax, const int pl
 	Assert(showMenu != NULL);
 	*showMenu = false;
 	static const int TITLE_LEN = 30; // Rough-approximate
-	static const int POSTFIX_LEN = 15 + 12;
+	static const int POSTFIX_LEN = 16 + 12 + 24;
 	static const int TOTALLINE_LEN = 64; // Rough-approximate
```
Rainyan, line 1335: Could be constexpr

Rainyan (changes requested):

brysondev (approved):
Enums my beloved

Rainyan (approved):
We might wanna eventually move the HUD "never show" option as part of the client's FCVAR_USERINFO, so the preference is not tied to the gamerules entity lifetime, and can be set with a config file by the user.

PR #719, opened 2024-10-13
Hide own view model on death

### LINKED ISSUE #720: Can see own viewmodel after dying and spectating someone else, in the spot where you originally died

AdamTadeusz:
## Description

Will fix the issue where after death your own view model remains visible in the spot where you died. No clue what caused it initially, this is a workaround I guess.

(Edit) Use Rain's solution instead

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #720

### DIFF
--- a/mp/src/game/client/neo/c_neo_player.cpp
+++ b/mp/src/game/client/neo/c_neo_player.cpp
@@ -976,6 +976,11 @@ void C_NEO_Player::PostThink(void)
 			m_bInVision = false;
 			m_bInLean = NEO_LEAN_NONE;
 
+			if (auto vm = GetViewModel(0, false))
+			{
+				vm->AddEffects(EF_NODRAW);
+			}
+
 			if (IsLocalPlayer() && (GetTeamNumber() == TEAM_JINRAI || GetTeamNumber() == TEAM_NSF))
 			{
 				SetObserverMode(OBS_MODE_DEATHCAM);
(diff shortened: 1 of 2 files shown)


brysondev (dismissed):
mmmmm yummy random jank appearing 🤤

Rainyan (changes requested):
This seems to be a regression from 31044a92 as per git bisect (bad `v9.0-prealpha`, good `v8.1-prealpha`).

I would instead suggest:

```diff
diff --git a/mp/src/game/server/neo/neo_player.cpp b/mp/src/game/server/neo/neo_player.cpp
index 5ab5382a..a554c0d8 100644
--- a/mp/src/game/server/neo/neo_player.cpp
+++ b/mp/src/game/server/neo/neo_player.cpp
@@ -1887,11 +1887,12 @@ void CNEO_Player::Weapon_Drop( CBaseCombatWeapon *pWeapon,
                }
        }
 
-       if (!pWeapon)
-               return;
+       if (pWeapon)
+       {
+               pWeapon->m_bInReload = false;
+               pWeapon->StopWeaponSound(RELOAD_NPC);
+       }
 
-       pWeapon->m_bInReload = false;
-       pWeapon->StopWeaponSound(RELOAD_NPC);
        BaseClass::Weapon_Drop(pWeapon, pvecTarget, pVelocity);
 }
```

as the early return from the `BaseClass::Weapon_Drop` chain responsible for the VM nodraw application here seems to be the root cause of the bug.

brysondev (approved):

Rainyan (approved):

PR #730, opened 2024-10-14
Do not delete g_pNeoLoading ourselves, probably already done by engine

nullsystem:
## Description
Still getting crash here (user after free on dangling pointer?). Just don't handle it. It's also called on exit so it's unnecessary to even cleanup here anyway.

## Toolchain

### DIFF
--- a/mp/src/game/client/clientmode_shared.cpp
+++ b/mp/src/game/client/clientmode_shared.cpp
@@ -324,13 +324,6 @@ ClientModeShared::ClientModeShared()
 ClientModeShared::~ClientModeShared()
 {
 	delete m_pViewport; 
-#ifdef NEO
-	if (g_pNeoLoading)
-	{
-		g_pNeoLoading->MarkForDeletion();
-		g_pNeoLoading = nullptr;
-	}
-#endif
 }
 
 void ClientModeShared::ReloadScheme( bool flushLowLevel )


brysondev (approved):

Rainyan:
Do you have a crash dump for this, or repro steps?

AdamTadeusz:
It's probably not noticeable unless you specifically run the game from visual studio in debug mode maybe?
https://github.com/user-attachments/assets/670be906-ef59-4d55-b9ef-f7251a0113e6

nullsystem:
@Rainyan Repro step is just opening then closing, and it happens sometimes, sometimes it doesn't. If it happens it's on shutdown.

@AdamTadeusz Still a crash and rather anonying when on a debugger?

AdamTadeusz:
> @Rainyan Repro step is just opening then closing, and it happens sometimes, sometimes it doesn't. If it happens it's on shutdown.
> 
> @AdamTadeusz Still a crash and rather anonying when on a debugger?

This is on master, not this branch. Haven't verified this fixes the problem yet but yes its a common crash on master on exit when debugging

Rainyan:
> It's probably not noticeable unless you specifically run the game from visual studio in debug mode maybe? https://github.com/user-attachments/assets/670be906-ef59-4d55-b9ef-f7251a0113e6

I don't currently have Windows installed, but do you still get this behaviour if you initialize the

https://github.com/NeotokyoRebuild/neo/blob/9dc5d669b95b11c88924ec963de094963f5df71b/mp/src/game/client/clientmode_shared.cpp#L115-L117

as `nullptr` upon declaration: `CNeoLoading *g_pNeoLoading = nullptr;`?

I see the pointer in your video evaluates as `0xffffffff` (-1), which I don't think any of our code paths are assigning for it, so I'm thinking maybe this is something MSVC decided to default-initialize it as for debug mode?

Rainyan (approved):
Ok, I can reproduce it too. Looking at the backtrace, it seems to indeed be a segfault at line

https://github.com/NeotokyoRebuild/neo/blob/9dc5d669b95b11c88924ec963de094963f5df71b/mp/src/game/client/clientmode_shared.cpp#L330

with the ECX register ("this" ptr of the thiscall) pointing to unaddressable memory at the time of the function call, which then signals SIGSEGV.

Tracing the `~CNeoLoading` base dtor, it's invoked via

https://github.com/NeotokyoRebuild/neo/blob/9dc5d669b95b11c88924ec963de094963f5df71b/mp/src/vgui2/vgui_controls/Panel.cpp#L1682

from the engine library before the `~ClientModeShared` dtor is called, so it seems to indeed be a use-after-free on the pointer. And since the `CNeoLoading` destructor is already being invoked by the engine, we should be safe to remove these lines without a memory leak.

AdamTadeusz:
firefox moment?

Issue #567, opened 2024-09-08
Dry fire behaviour should match OG NT.


Agiel:
### Build Info

0

### Description

After dry firing a gun in OG NT, there's a small delay before the gun is reloaded. Attempting to fire during this time plays the dry fire effect again and the timer is reset. This also means you have a chance to see the dry fire effect on the semi auto guns.

In the current NT;RE version the reload happens as soon as the fire button is released.

### To Reproduce

Fire a gun until it's empty. Observe when the reload happens.

### Expected behavior

There should be a delay.

### Actual behavior

The gun is reloaded immediately.

### Operating System

- [X] Windows
- [ ] Linux

### Version/Distro

Windows 10

Rainyan:
Just so this doesn't get lost in sands of time, here's some empirical data that may be pertinent: https://github.com/NeotokyoRebuild/neo/pull/731#issuecomment-2450508551

AdamTadeusz:
> Just so this doesn't get lost in sands of time, here's some empirical data that may be pertinent: [#731 (comment)](https://github.com/NeotokyoRebuild/neo/pull/731#issuecomment-245050855

3 seconds seems like a lot, even if it is the right value. I think we should make it shorter. Also I think pushing the reload into the future by like 1.25 seconds whenever you dryfire is unnecessary, I think it would feel much nicer if you started reloading immediately after releasing the attack key after this initial period of not reloading

Agiel:
Pressing R to reload faster is a skill.

kassibuss:
something something parity

PR #731, opened 2024-10-14
tachi, milso and zr68l improvements(zr68l still not ideal) plus faster srs firing

### LINKED ISSUE #703: Weapon view models behaving weirdly when sprinting
Specifically the tachi, zr68l and milso. Need to test all weapons for this.

AdamTadeusz:
## Description

use idleTime, allow srs weapon pullback after weapon has finished primary attack activity

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #703

### DIFF
--- a/mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
+++ b/mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
@@ -477,7 +477,8 @@ void CNEOBaseCombatWeapon::ProcessAnimationEvents(void)
 	{
 		m_bLowered = true;
 		SendWeaponAnim(ACT_VM_IDLE_LOWERED);
-		m_flNextPrimaryAttack = max(gpGlobals->curtime + 0.2, m_flNextPrimaryAttack);
+		SetWeaponIdleTime(gpGlobals->curtime + SequenceDuration());
+		m_flNextPrimaryAttack = max(gpGlobals->curtime + SequenceDuration(), m_flNextPrimaryAttack);
 		m_flNextSecondaryAttack = m_flNextPrimaryAttack;
 	}
 	else if (m_bLowered && !(pOwner->IsSprinting() || pOwner->GetMoveType() == MOVETYPE_LADDER))
@@ -491,15 +492,15 @@ void CNEOBaseCombatWeapon::ProcessAnimationEvents(void)
 	{ // For bolt action weapons
 		m_bLowered = false;
 		SendWeaponAnim(ACT_VM_PULLBACK);
-		m_flNextPrimaryAttack = max(gpGlobals->curtime + 0.2, m_flNextPrimaryAttack);
+		m_flNextPrimaryAttack = max(gpGlobals->curtime + 1.2f, m_flNextPrimaryAttack);
 		m_flNextSecondaryAttack = m_flNextPrimaryAttack;
 	}
 
 	if (m_bLowered)
 	{
-		if (gpGlobals->curtime > m_flNextPrimaryAttack)
+		if (gpGlobals->curtime > GetWeaponIdleTime())
 		{
-			SendWeaponAnim(ACT_VM_IDLE_LOWERED);
+			SetWeaponIdleTime(gpGlobals->curtime + 0.2);
 			m_flNextPrimaryAttack = max(gpGlobals->curtime + 0.2, m_flNextPrimaryAttack);
 			m_flNextSecondaryAttack = m_flNextPrimaryAttack;
 		}
--- a/mp/src/game/shared/neo/weapons/weapon_srs.cpp
+++ b/mp/src/game/shared/neo/weapons/weapon_srs.cpp
@@ -56,7 +56,7 @@ void CWeaponSRS::ItemPreFrame()
 		m_bRoundChambered = true;
 	}
 
-	if (m_flNextPrimaryAttack <= gpGlobals->curtime && !m_bRoundChambered && !m_bRoundBeingChambered && m_iClip1 > 0)
+	if (m_flLastAttackTime + SequenceDuration(ACT_VM_PRIMARYATTACK) <= gpGlobals->curtime && !m_bRoundChambered && !m_bRoundBeingChambered && m_iClip1 > 0)
 	{ // Primary attack animation finished, begin chambering a round
 		if (CNEO_Player* pOwner = static_cast<CNEO_Player*>(ToBasePlayer(GetOwner()))) {
 			if (pOwner->m_nButtons & IN_ATTACK)
@@ -66,10 +66,10 @@ void CWeaponSRS::ItemPreFrame()
 			}
 			pOwner->Weapon_SetZoom(false);
 		}
-		m_bRoundBeingChambered = true;
-		m_flChamberFinishTime = gpGlobals->curtime + GetFireRate();
-		WeaponSound(SPECIAL1);
 		SendWeaponAnim(ACT_VM_PULLBACK);
+		WeaponSound(SPECIAL1);
+		m_bRoundBeingChambered = true;
+		m_flChamberFinishTime = gpGlobals->curtime + 1.2f;
 	}
 
 	BaseClass::ItemPreFrame();


AdamTadeusz:
Will take some recordings of the srs and update relevant values so they match, for now leaving as draft

brysondev (approved):
finally, no more srs jank

mp/src/game/shared/neo/weapons/weapon_srs.cpp
```diff
@@ -66,10 +66,10 @@ void CWeaponSRS::ItemPreFrame()
 			}
 			pOwner->Weapon_SetZoom(false);
 		}
-		m_bRoundBeingChambered = true;
-		m_flChamberFinishTime = gpGlobals->curtime + GetFireRate();
-		WeaponSound(SPECIAL1);
 		SendWeaponAnim(ACT_VM_PULLBACK);
+		WeaponSound(SPECIAL1);
+		m_bRoundBeingChambered = true;
+		m_flChamberFinishTime = gpGlobals->curtime + 1.2f;
```
Rainyan, line 72: What's the reason for hardcoding this as 1.2 instead of calling the `GetFireRate`? Is this different from what the SRS WpnData would return for the cycle time?

mp/src/game/shared/neo/weapons/weapon_srs.cpp
```diff
@@ -56,7 +56,7 @@ void CWeaponSRS::ItemPreFrame()
 		m_bRoundChambered = true;
 	}
 
-	if (m_flNextPrimaryAttack <= gpGlobals->curtime && !m_bRoundChambered && !m_bRoundBeingChambered && m_iClip1 > 0)
+	if (m_flLastAttackTime + 0.08f <= gpGlobals->curtime && !m_bRoundChambered && !m_bRoundBeingChambered && m_iClip1 > 0)
```
Rainyan, line 59: Where is the value 0.08 coming from here?

mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
```diff
@@ -500,12 +500,16 @@ void CNEOBaseCombatWeapon::ProcessAnimationEvents()
...
+		m_flNextSecondaryAttack = m_flNextPrimaryAttack;
 	}
 
 	if (m_bLowered && gpGlobals->curtime > m_flNextPrimaryAttack)
 	{
-		next(ACT_VM_IDLE_LOWERED);
+		SetWeaponIdleTime(gpGlobals->curtime + 0.2);
+		m_flNextPrimaryAttack = max(gpGlobals->curtime + 0.2, m_flNextPrimaryAttack);
+		m_flNextSecondaryAttack = m_flNextPrimaryAttack;
 	}
```
Rainyan, line 513: Since every path in the if-else chain above updates the `m_flNextPrimaryAttack` value as larger than `gpGlobals->curtime`, we could probably simplify this as:

```diff
diff --git a/mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp b/mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
index b90b0cec..e9c21c2f 100644
--- a/mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
+++ b/mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
@@ -504,12 +504,10 @@ void CNEOBaseCombatWeapon::ProcessAnimationEvents()
                m_flNextPrimaryAttack = max(gpGlobals->curtime + 1.2f, m_flNextPrimaryAttack);
                m_flNextSecondaryAttack = m_flNextPrimaryAttack;
        }
-
-       if (m_bLowered && gpGlobals->curtime > m_flNextPrimaryAttack)
+       else if (m_bLowered)
        {
                SetWeaponIdleTime(gpGlobals->curtime + 0.2);
-               m_flNextPrimaryAttack = max(gpGlobals->curtime + 0.2, m_flNextPrimaryAttack);
-               m_flNextSecondaryAttack = m_flNextPrimaryAttack;
+               next(ACT_VM_IDLE_LOWERED);
        }
 }
```

mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
```diff
@@ -500,12 +500,16 @@ void CNEOBaseCombatWeapon::ProcessAnimationEvents()
 	else if (m_bLowered && m_bRoundBeingChambered)
 	{ // For bolt action weapons
 		m_bLowered = false;
-		next(ACT_VM_PULLBACK);
+		SendWeaponAnim(ACT_VM_PULLBACK);
+		m_flNextPrimaryAttack = max(gpGlobals->curtime + 1.2f, m_flNextPrimaryAttack);
+		m_flNextSecondaryAttack = m_flNextPrimaryAttack;
```
Rainyan, line 505: We could maybe refactor this such that all the paths use the same `next` lambda for setting the values:

```diff
diff --git a/mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp b/mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
index b90b0cec..dd0e669a 100644
--- a/mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
+++ b/mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
@@ -475,13 +475,12 @@ void CNEOBaseCombatWeapon::ProcessAnimationEvents()
                return;
        }
 
-       const auto next = [this](const int activity) {
+       const auto next = [this](const int activity, const float nextAttackDelay = 0.2) {
                SendWeaponAnim(activity);
                if (GetNeoWepBits() & NEO_WEP_THROWABLE)
                {
                        return;
                }
-               constexpr auto nextAttackDelay = 0.2;
                m_flNextPrimaryAttack = max(gpGlobals->curtime + nextAttackDelay, m_flNextPrimaryAttack);
                m_flNextSecondaryAttack = m_flNextPrimaryAttack;
        };
@@ -500,9 +499,7 @@ void CNEOBaseCombatWeapon::ProcessAnimationEvents()
        else if (m_bLowered && m_bRoundBeingChambered)
        { // For bolt action weapons
                m_bLowered = false;
-               SendWeaponAnim(ACT_VM_PULLBACK);
-               m_flNextPrimaryAttack = max(gpGlobals->curtime + 1.2f, m_flNextPrimaryAttack);
-               m_flNextSecondaryAttack = m_flNextPrimaryAttack;
+               next(ACT_VM_PULLBACK, 1.2f);
        }
```

mp/src/game/shared/neo/weapons/weapon_srs.cpp
AdamTadeusz: 0.08 is the time after firing when the character begins playing the round chamber animation, as I recorded it at 1/10 speed

mp/src/game/shared/neo/weapons/weapon_srs.cpp
AdamTadeusz: Yes

AdamTadeusz: It's different to the sequence length also

mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
AdamTadeusz: I'm purposefully avoiding sending the ACT_VM_IDLE_LOWERED animation here to prevent the jittering of the tachi and milso this pr is primarily fixing

AdamTadeusz:
@Rainyan I found the srs recordings. I'm gonna put together a comparison video and hopefully its small enough to upload here

AdamTadeusz:
https://github.com/user-attachments/assets/88cd6355-ee5e-4e0f-b10f-3306ff43392a

Should have made these two separate prs honestly

brysondev (approved):

Rainyan (changes requested):
Since we're touching up the wep idle time: it seems that currently our post-attack value for it is incorrect. In OG NT, after a primary attack, the idle time gets bumped by (seems like) 2 seconds. But currently, our guns switch to idle too fast after firing, which looks kind of weird, especially when firing semi-auto or small bursts. You can see this visually with the viewmodel's thumb position on the Tachi (at least the NSF one), or using something like `cl_pdump` on the weapon entity index to monitor the value - we currently go to idle at I believe 0.2 seconds or so, which is much faster than OG.

We could probably get this closer to parity with something like:

```diff
diff --git a/mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp b/mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
index ddb736e4..527f6ec0 100644
--- a/mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
+++ b/mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
@@ -814,6 +814,7 @@ void CNEOBaseCombatWeapon::PrimaryAttack(void)
                pPlayer->DoMuzzleFlash();
 
        SendWeaponAnim(GetPrimaryAttackActivity());
+       SetWeaponIdleTime(gpGlobals->curtime + 2.0);
 
        // player "shoot" animation
        pPlayer->DoAnimationEvent(PLAYERANIMEVENT_ATTACK_PRIMARY);
```

The 2.0 value is just eyeballed, and I've only verified it for the Tachi, ZR68s, and Jittescoped, but would imagine all of the guns behave in this way.

Slightly related, it seems that our next primary attack value might be slightly off, at least when exiting sprint whilst holding down mouse1. I'm seeing a few frames discrepancy when recording in OBS, although I haven't tracked down the actual tick counts to verify. To reproduce:

* Equip the ZR68s
* Start sprinting
* Hold down mouse1
* Let go of the sprint key
* See how long it takes until the first bullet is fired

And even more tangentially, it seems that holding down mouse1 in this way will not fire the semi-auto guns like Tachi at all, whereas in OG NT it does fire off the first bullet in such a case. But maybe that belongs in a different PR?

mp/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
```diff
@@ -500,12 +499,15 @@ void CNEOBaseCombatWeapon::ProcessAnimationEvents()
 	else if (m_bLowered && m_bRoundBeingChambered)
 	{ // For bolt action weapons
 		m_bLowered = false;
-		next(ACT_VM_PULLBACK);
+		SendWeaponAnim(ACT_VM_PULLBACK);
+		next(ACT_VM_PULLBACK, 1.2f);
```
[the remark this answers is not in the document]
AdamTadeusz: Screen_Shot_2022-09-23_at_10 40 58_AM

brysondev (approved):

Agiel:
> Since we're touching up the wep idle time: it seems that currently, our post-attack increment for it is incorrect. In OG NT, after a primary attack, the idle time gets bumped by (seems like) 2 seconds. But our guns switch to idle too fast after firing, which looks kind of weird, especially when firing semi-auto or small bursts, and the viewmodel keeps fiddling its fingers back and forth for each shot. You can see this visually with the viewmodel's thumb position on the Tachi (at least the NSF one), or using something like cl_pdump on the weapon entity index to monitor the value - we currently go to idle at I believe 0.2 seconds or so, which is much faster than OG.
I wonder if this might also fix #567 ?

Rainyan:
> I wonder if this might also fix #567 ?

It's probably related, although this doesn't directly fix it. Empirically investigating the values in OG, it seems to be something like:

```
IF (2 SECONDS IDLE) THEN
  IF hasAmmo THEN
    GO IDLE
  ELSE
    idleTime += 1
      
IF (3 SECONDS IDLE) AND NOT hasAmmo THEN
  DO RELOAD
```

Loosely annotated at top-left of the pred dump, in colored text:

[nt_reload_idle.webm](https://github.com/user-attachments/assets/22fac965-f6dc-4640-a526-f40b72587ff5)

Rainyan (approved):

PR #742, opened 2024-10-16
Enemy only collisions

AdamTadeusz:
## Description

Adds collisions with enemies with sv_neo_collision_enemy_collisions 1
Also allows standing on top of friendlies with sv_neo_collision_friendly_head_collisions 1
value sv_neo_collision_friendly_head_collisions_overlap controls the difference between the bottom of the top player and the top of the bottom player before friendly collisions switch off

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

### DIFF
--- a/mp/src/game/shared/util_shared.cpp
+++ b/mp/src/game/shared/util_shared.cpp
@@ -297,6 +297,17 @@ bool CTraceFilterSimple::ShouldHitEntity( IHandleEntity *pHandleEntity, int cont
 
 	// Don't test if the game code tells us we should ignore this collision...
 	CBaseEntity *pEntity = EntityFromEntityHandle( pHandleEntity );
+#ifdef NEO
+	if (m_collisionGroup == COLLISION_GROUP_PLAYER_MOVEMENT && m_pPassEnt && pEntity->IsPlayer())
+	{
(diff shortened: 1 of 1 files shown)


brysondev:
https://github.com/user-attachments/assets/596733f8-de26-4aa7-8dae-b791e166357b

brysondev (approved):

nullsystem (approved):

mp/src/game/shared/util_shared.cpp
```diff
@@ -296,7 +301,43 @@ bool CTraceFilterSimple::ShouldHitEntity( IHandleEntity *pHandleEntity, int cont
...
+					{
+						return true;
+					}
+				}
+			}
+		}
+	}
+#endif
 	if ( !pEntity )
 		return false;
```
nullsystem, line 342: Actually, are there scenarios pEntity and pEntity2 returns nullptr? The SDK code handles that scenario below.

AdamTadeusz: I don't know if there is any scenario where m_pPassEnt is true but pEntity2 isn't, but checking just in case

nullsystem (approved):

brysondev (approved):

PR #745, opened 2024-10-16
Streamer mode

### LINKED ISSUE #150: Streamer Mode
Requested features, preferably with individual toggles: 
- Hide Other Player names (Gives other players generic nicknames)
- Mute Players (ability to mute individual players in VC and Text)
- Hide any information related to Region/IP
- Profanity Filter (Could just be a general setting to toggle)
- Disable Gibs

nullsystem:
## Description
* No avatar
* No names (only generic ones)

- [x] OBS detection (Linux)
- [x] OBS detection (Windows)
- [x] Automatic (OBS detection) mode
- [x] Names are "A...." like Discord streamer mode

* **Won't do:** Custom names

## Toolchain

## Linked Issues

* fixes #150

### DIFF
--- a/mp/src/game/client/c_playerresource.cpp
+++ b/mp/src/game/client/c_playerresource.cpp
@@ -60,6 +60,10 @@ BEGIN_PREDICTION_DATA( C_PlayerResource )
 
 END_PREDICTION_DATA()	
 
+#ifdef NEO
+extern ConVar neo_cl_streamermode;
+#endif
+
 C_PlayerResource *g_PR;
 
 IGameResources * GameResources( void ) { return g_PR; }
@@ -187,6 +191,15 @@ const char *C_PlayerResource::GetPlayerName( int iIndex )
 	}
 
 #ifdef NEO
+	if (neo_cl_streamermode.GetBool() && !IsLocalPlayer(iIndex))
+	{
+		[[maybe_unused]] uchar32 u32Out;
+		bool bError = false;
+		const int iSize = Q_UTF8ToUChar32(m_szName[iIndex], u32Out, bError);
+		if (!bError) V_memcpy(gStreamerModeNames[iIndex], m_szName[iIndex], iSize);
+		return gStreamerModeNames[iIndex];
+	}
+
 	const bool clientWantNeoName = C_NEO_Player::GetLocalNEOPlayer()->ClientWantNeoName();
 	const int dupeIdx = m_iNeoNameDupeIdx[iIndex];
 
--- a/mp/src/game/client/cdll_client_int.cpp
+++ b/mp/src/game/client/cdll_client_int.cpp
@@ -174,8 +174,19 @@ extern vgui::IInputInternal *g_InputInternal;
 #include "neo_version.h"
 #include "neo_mount_original.h"
 #include "ui/neo_loading.h"
+#include "neo_player_shared.h"
 extern bool NeoRootCaptureESC();
 extern CNeoLoading *g_pNeoLoading;
+extern ConVar neo_cl_streamermode_autodetect_obs;
+bool g_bOBSDetected = false;
+
+#ifdef WIN32
+#include <windows.h>
+#include <tchar.h>
+#include <psapi.h>
+#include <tlhelp32.h>
+#undef CreateEvent
+#endif
 
 #ifdef LINUX
 #include "neo_fixup_glshaders.h"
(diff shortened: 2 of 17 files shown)


mp/src/game/shared/neo/neo_player_shared.h
```diff
@@ -311,4 +311,24 @@ struct PlayerXPInfo
 void DMClSortedPlayers(PlayerXPInfo (*pPlayersOrder)[MAX_PLAYERS + 1], int *piTotalPlayers);
 #endif
 
+// NEO TODO: Ubuntu release names
```
brysondev, line 314: Editable txt would be nice :^)

Rainyan:
A few suggestions - obfuscate console printouts:
* IP address prints (your own IP & the server IP - avoid attracting DDOS attacks)
  * This could for example obfuscate the 2 middle octets as: `123.*.*.456`
* Home dir prints by comparing env vars `$HOME` (Linux) and `%HOMEPATH%` (Windows) for the few prints that use these, to avoid self-doxxing ( `/home/johnsmith/...`, `C:\Users\JohnSmith\...`)
  * For example (Linux): `Some printout result: /home/realname/foo` -> `Some printout result: $HOME/foo`

Detect OBS at game launch for cvar initial value:
* Linux: for `<pid>` in `/proc/<pid>`: cat `cmdline` -> if equals "obs"
* Windows: https://learn.microsoft.com/en-us/windows/win32/psapi/enumerating-all-processes -> "obs.exe" (or whatever it's called on Windows)

mp/src/game/shared/neo/neo_player_shared.h
nullsystem: There's an issue of names being dealt with and done server-side. That could mean each player sends over a whole list of names but that seems much, or rather name displays need to defer more to client-side. So I won't bother with it now.

nullsystem:
@Rainyan 
Looks like that "enumerating all processes" WINAPI only shows the names of 32-bit processes/programs due to hl2.exe/NT;RE also only being a 32-bit program. For example in my printout, for the names I can see 32-bit programs like hammer++ and steam showing up, but 64-bit programs like OBS and Chrome not showing up.

nullsystem:
Branch `GH-150_WINAPIStreamerModeAuto` from my Github fork repo added the `Msg` printout in the beginning to try to show the names. Most will be "<unknown>" but you'll only see 32-bit processes if they have names.

nullsystem:
Found a `Process32Next` which actually works so gonna use that instead.

mp/src/game/client/cdll_client_int.cpp
```diff
@@ -1273,6 +1284,64 @@ void CHLClient::PostInit()
...
+				if (fgets(szCmdline, sizeof(szCmdline), fp))
+				{
+					// NEO TODO (nullsystem): Check on OBS flatpak, snaps, nix, and appimage.
+					// Could they give out different cmdline?
+					g_bOBSDetected = (V_strcmp(szCmdline, "obs") == 0);
+				}
+				pclose(fp);
+			}
+		}
+#else
```
brysondev, line 1327: Nit:
```suggestion
#else // LINUX
```

mp/src/game/client/cdll_client_int.cpp
```diff
@@ -1273,6 +1284,64 @@ void CHLClient::PostInit()
...
+#endif
+		ConVarRef("neo_cl_streamermode").SetValue(g_bOBSDetected);
+	}
+
+	// Initialize streamer mode names
+	for (int i = 0; i < MAX_PLAYERS + 1; ++i)
+	{
+		V_memset(gStreamerModeNames[i], '.', 5);
+	}
 #endif
```
brysondev, line 1345: Nit:
```suggestion
#endif // WINDOWS
```

brysondev (approved):

mp/src/game/client/cdll_client_int.cpp
```diff
@@ -1273,6 +1284,64 @@ void CHLClient::PostInit()
...
+					// Could they give out different cmdline?
+					g_bOBSDetected = (V_strcmp(szCmdline, "obs") == 0);
+				}
+				pclose(fp);
+			}
+		}
+#else
+		HANDLE hSnapShot = CreateToolhelp32Snapshot(TH32CS_SNAPPROCESS, 0);
+		PROCESSENTRY32 processInfo = {};
+		processInfo.dwSize = sizeof(PROCESSENTRY32);
```
Rainyan, line 1330: Would be prudent to check the `CreateToolhelp32Snapshot` retval against the `INVALID_HANDLE_VALUE` error value, as the MS docs for `CloseHandle` seem to imply that attempting to close an invalid handle would throw, at least in debug mode.

mp/src/game/client/cdll_client_int.cpp
```diff
@@ -1273,6 +1284,64 @@ void CHLClient::PostInit()
...
+			char szCmdlinePath[256];
+			V_sprintf_safe(szCmdlinePath, "/proc/%s/cmdline", pszFilename);
+			if (!filesystem->FileExists(szCmdlinePath))
+			{
+				continue;
+			}
+
+			// NEO NOTE (nullsystem): At this point, cannot just do filesystem->ReadFile
+			// as it won't output anything. Instead just cat the file and fetch the line.
+			char szCmdCat[256 + 16];
```
Rainyan, line 1312: Might want to couple the alloc size as `[sizeof(szCmdlinePath) + 16]` etc

mp/src/game/client/c_playerresource.cpp
```diff
@@ -187,6 +191,12 @@ const char *C_PlayerResource::GetPlayerName( int iIndex )
 	}
 
 #ifdef NEO
+	if (neo_cl_streamermode.GetBool())
+	{
+		gStreamerModeNames[iIndex][0] = m_szName[iIndex][0];
+		return gStreamerModeNames[iIndex];
```
Rainyan, line 197: Ditto on names with initial MB character

Rainyan (changes requested):

mp/src/game/client/cdll_client_int.cpp
nullsystem: *`// NEO`

mp/src/game/client/cdll_client_int.cpp
```diff
@@ -174,8 +174,19 @@ extern vgui::IInputInternal *g_InputInternal;
 #include "neo_version.h"
 #include "neo_mount_original.h"
 #include "ui/neo_loading.h"
+extern char gStreamerModeNames[MAX_PLAYERS + 1][MAX_PLAYER_NAME_LENGTH + 1];
```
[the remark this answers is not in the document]
nullsystem: I think there was some odd compiler error on Windows causing me to switch it out for extern but I'll relook.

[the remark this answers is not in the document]
Rainyan: If it's needed for compiler quirks then no harm done leaving it be I suppose, was just wondering whether it was intentional.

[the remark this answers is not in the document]
nullsystem: Well I've updated it and that seemed to fixed it for me (and the github action ran fine anyway). Second time this year I've seen MSVC bugged out on nonsense.

Rainyan (approved):

brysondev (approved):

PR #756, opened 2024-10-20
NEO_SLOW_SIDE_MODIFIER

### LINKED ISSUE #757: Player moves at the same speed backwards as when moving backwards while holding a directional key
OG behaviour for comparison:

https://github.com/user-attachments/assets/9c912924-9f42-47c1-a2a1-bec287c715b1

AdamTadeusz:
## Description

Additional modifier for when holding a left/right direction key when moving backwards

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #757

### DIFF
--- a/mp/src/game/client/neo/c_neo_player.h
+++ b/mp/src/game/client/neo/c_neo_player.h
@@ -111,7 +111,7 @@ public:
 
 private:
 	float GetActiveWeaponSpeedScale() const;
-	float GetBackwardsMovementPenaltyScale() const { return ((m_nButtons & IN_BACK) ? NEO_SLOW_MODIFIER : 1.0); }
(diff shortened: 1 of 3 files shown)


AdamTadeusz:
Yeah I don't think max speed affects players, hence why you can bhop so quickly. Ill let you open a pr then since you have the whole solution.

> Like I said in discord. OGNT only has the backwards multiplier of 0.75. The rest is simple trigonometry. If you remove GetBackwardsMovementPenaltyScale and instead just do
> 
> ```
> if (mv->m_flForwardMove < 0.0) {
> 	mv->m_flForwardMove *= 0.75;
> }
> ```
> 
> at the start of CGameMovement::WalkMove, you get the exact behavior as in OGNT.
> 
> Fun fact: If you're in the air you get full speed backwards.

PR #788, opened 2024-11-03
IFF fix when previously using xray

### LINKED ISSUE #783: Friendly indicators don't show up after using x-ray mode
### Build Info

20241102_14c4282

### Description

The little friendly player icons stay hidden on next round if you used xray mode in spectate on the previous round

### To Reproduce

1. Die
2. Spectate
3. Turn on xray
4. Conclude round
5. Look for friendly indicators

### Expected behavior

You see friendly indicators on friendlies

### Actual behavior

Indicators are gone

### Operating System

- [X] Windows
- [ ] Linux

### Version/Distro

Win 10

### Machine's CPU

Ryzen 9 7950X

### Machine's GPU

RTX 3080 FE

### GPU's driver

Nvidia 556.12

### Build's compiler

MSVC 19.41.34123.0

_No response_

AdamTadeusz:
## Description

Returns the IFF indicators above friendly players when player was previously using xray in spectator mode. Also fixes a problem where xray button had to be pressed twice to enable xray in spectator mode if person was using xray previously before respawning.

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #783

### DIFF
--- a/mp/src/game/client/neo/c_neo_player.cpp
+++ b/mp/src/game/client/neo/c_neo_player.cpp
@@ -809,7 +809,7 @@ void C_NEO_Player::PlayStepSound( Vector &vecOrigin,
 }
 
 extern ConVar sv_infinite_aux_power;
-
+extern ConVar glow_outline_effect_enable;
 void C_NEO_Player::PreThink( void )
 {
 	BaseClass::PreThink();
@@ -912,6 +912,7 @@ void C_NEO_Player::PreThink( void )
 					player->SetClientSideGlowEnabled(false);
 				}
 			}
+			glow_outline_effect_enable.SetValue(false);
 #endif // GLOWS_ENABLE
 		}
 		Lean();


Rainyan (approved):
This fixes the issue, but we can't toggle the mode by directly modifying the cvar, which limits what players can do with their configs etc.

I think ideally we would instead have a `glow_outline_effect_enable.InstallChangeCallback` set at init, and set the `SetClientSideGlowEnabled` for players from within that callback. Then, we could simply set the cvar's value for the change to take effect, instead of requiring multiple implementations of this logic:
```cpp
for (int i = 1; i <= gpGlobals->maxClients; i++)
{
	if (auto player = UTIL_PlayerByIndex(i))
	{
		player->SetClientSideGlowEnabled(/* ... */);
	}
}
```

brysondev (approved):

PR #789, opened 2024-11-03
player collisions single command

AdamTadeusz:
## Description

One command to control collision method, Bitwise AND, better description to reflect actual behavior (Boosting already worked both on enemies and friendlies regardless of enemy only collision).

Collision options are ordered in terms of how disruptive they are, with enemy only collisions considered more disruptive than boosting, and the combination of the two being the most disruptive.

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

### DIFF
--- a/src/game/shared/util_shared.cpp
+++ b/src/game/shared/util_shared.cpp
@@ -286,6 +286,12 @@ CTraceFilterSimple::CTraceFilterSimple( const IHandleEntity *passedict, int coll
 }
 
 #ifdef NEO
+enum NeoPlayerCollisionsType
+{
+	NEO_PLAYER_COLLISIONS_TYPE_NONE = 0,
+	NEO_PLAYER_COLLISIONS_TYPE_BOOSTING,
+	NEO_PLAYER_COLLISIONS_TYPE_ENEMY
+};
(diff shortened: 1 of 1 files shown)


brysondev (approved):

mp/src/game/shared/util_shared.cpp
```diff
@@ -312,14 +311,14 @@ bool CTraceFilterSimple::ShouldHitEntity( IHandleEntity *pHandleEntity, int cont
...
-			if (sv_neo_collision_enemy_collisions.GetBool())
+			if (sv_neo_player_collisions.GetInt() & 2)
 			{
 				if (pEntity->GetTeamNumber() != pEntity2->GetTeamNumber())
 				{
 					return true;
 				}
 			}
-			if (sv_neo_collision_friendly_head_collisions.GetBool())
+			if (sv_neo_player_collisions.GetInt() & 1)
```
nullsystem, line 321: Nit: Can you give those 1/2 numbers as an enum so it makes it more clear which bitwise flag check is for

nullsystem (changes requested):

brysondev (approved):

brysondev (dismissed):

DESTROYGIRL (approved):

PR #796, opened 2024-11-03
Fix Ghost Pings

### LINKED ISSUE #779: Ghost beeps don't play for players within ghost radius at time of pickup
### Build Info

20241102_14c4282

### Description

The ghost beep sound doesn't play if a player is already within the ghost detection range

### To Reproduce

1. pick up ghost with enemy player next to you
2. both of you listen for beep

### Expected behavior

Both enemy player and the ghost holder hears a beep

### Actual behavior

No beeps

### Operating System

- [X] Windows
- [ ] Linux

### Version/Distro

Win 10

### Machine's CPU

Ryzen 9 7950X

### Machine's GPU

RTX 3080 FE

### GPU's driver

Nvidia 556.12

### Build's compiler

MSVC 19.41.34123.0

To be confirmed: players entering the ghost radius after the ghost has been picked up make sound? they don't make beep?

AdamTadeusz:
## Description

Been broken for a very long time, now they're back!

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #779


brysondev (approved):

mp/src/game/client/neo/ui/neo_hud_ghost_beacons.cpp
```diff
@@ -112,28 +112,26 @@ void CNEOHud_GhostBeacons::DrawNeoHudElement()
 
 	m_pGhost = ghost;
 	const bool showGhost = (gpGlobals->curtime > m_flNextAllowGhostShowTime);
-	if (!showGhost)
-	{
-		return;
-	}
```
blaberry, line 118: Removing this will be a bit of a slowdown, no? We're calling GetPlayer for N players and doing some raycasts all for nothing if the innermost showGhost bool is false?

AdamTadeusz: showGhost controls whether player beacons are displayed (think first few seconds after picking up the ghost when player beacons still don't display), but ghost pings can happen as soon as you pick up the ghost

AdamTadeusz: I guess the && showGhost argument could be moved to before the IsPoswithinViewDistance

AdamTadeusz: actually we're relying on isPosWithinViewDistance to set the value of distance used later on regardless of whether the beacons are shown or not. So I guess IsPosWithinViewDistance should be the first argument, otherwise the Min of distance and closestEnemy may have undefined behavior?

brysondev (approved):

brysondev:
image

Agiel (approved):

PR #799, opened 2024-11-03
Cloak fixes

### LINKED ISSUE #764: [BUG] After your cloak has ran out, the game does not play the decloaking sound
### Build Info

20241012_bc3a767

### Description

When your cloak runs out in OGNT, there is an audible sound. Currently, in NTRE, the sound does not play when your cloak runs out, only when you decloak manually does the sound play.

### To Reproduce

1. Join the game
2. Cloak
3. Wait for the cloak to run out

### Expected behavior

The decloaking sound should play from your character.

### Actual behavior

No sound is made

### Operating System

- [X] Windows
- [ ] Linux

### Version/Distro

10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

https://github.com/user-attachments/assets/da065c85-6b4a-433b-aee2-371ef84247c9

### LINKED ISSUE #800: Cloak flickers when near out of power

AdamTadeusz:
## Description

Server handles playing cloaking sound for everyone
Server controls when cloak is turned off due to running out of power
Lowered power threshold required to cloak
Can turn cloak off regardless of power left

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #764 
- fixes #800


AdamTadeusz:
I think ill use this pr to add some more cloak related fixes since its quite small

AdamTadeusz:
I think adjusting the cloak flash color will have to be a separate PR, lot of room for disagreement there

brysondev (dismissed):

Rainyan (changes requested):
I don't believe the server should be the one sending the cloak sounds, rather the client should play the cloak sound when the predicted `m_bInThermOpticCamo` property changes. Otherwise, the player's own cloak sound will play with a delay based on client latency. You can test this with `net_fakelag 100` and see how the cloak sounds plays 0.2 seconds (round trip) late, which makes the game feel unresponsive to the user. And for other players too, it would make sense to couple this to the entprop, because else we don't have quarantees of when exactly the sound packet arrives compared to the netprop delta, which could still lead to desync when compared to the visual effect.

AdamTadeusz:
> I don't believe the server should be the one sending the cloak sounds, rather the client should play the cloak sound when the predicted `m_bInThermOpticCamo` property changes. Otherwise, the player's own cloak sound will play with a delay based on client latency. You can test this for this PR on localhost with `net_fakelag 100` and see how the cloak sounds plays 0.2 seconds (round trip) late, which makes the game feel unresponsive (or "laggy") to the user. And for other players too, it would make sense to couple this to the entprop, because else we don't have guarantee of when exactly the sound packet arrives compared to the netprop change, which could still lead to sound desync in comparison to the visual effect - I believe we would want the sound to play exactly when the cloak effect happens, so it feels responsive to the user.
> 
> Thumbs up otherwise, but if you could change the logic to take the entprop into account (somewhat similar to the handling before this PR I believe?), that'd be great.

Yeah the reason I did this was when turning cloak on at very low cloak power, sometimes the cloak sound wouldn't play because the client thought it didnt have enough cloak power, but the cloak actually toggles on server side. Ill just return the original behaviour since its very unlikely to reproduce this during normal gameplay, but Ill leave the option to play for client that generated the noise from server for the case of cloak running out of power without being toggled off

AdamTadeusz:
https://github.com/user-attachments/assets/0f46d60c-5f1f-4af1-b60b-8c22bd4e8a73

Here's a video from master showing the problem, it's hard to reproduce since if youre spamming cloak you will probably switch it off before realising you switched it on without the sound playing, but I got it towards the end of the video, and you can see some flashes with no sound earlier too

brysondev (dismissed):

brysondev (approved):

Agiel (approved):

PR #805, opened 2024-11-04
Automatic leaning

AdamTadeusz:
## Description

Automatic leaning. Currently using automatic leaning prevents the use of manual leaning completely.

### DIFF
--- a/mp/src/game/client/in_main.cpp
+++ b/mp/src/game/client/in_main.cpp
@@ -512,6 +512,10 @@ void IN_AimDown( const CCommand &args ) { KeyDown( &in_aim, args[1] ); }
 void IN_LeanLeftUp( const CCommand &args ) { KeyUp( &in_lean_left, args[1] ); }
 void IN_LeanLeftDown( const CCommand &args ) { KeyDown( &in_lean_left, args[1] ); }
 
+void IN_LeanLeft() { KeyUp(&in_lean_right, nullptr); KeyDown(&in_lean_left, nullptr); }
+void IN_LeanRight() { KeyUp(&in_lean_left, nullptr); KeyDown(&in_lean_right, nullptr); }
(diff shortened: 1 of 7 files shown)


brysondev (approved):
LGTM but how do we incorporate this for the sourcebots branch? I'd love to have this functionality added for them.

AdamTadeusz:
We can add a server only section which checks if a player is a bot, and does the same thing as whats done client side for the local player except with far fewer tracelines. We could add a time variable so that bots stop leaning only after a set number of seconds pass without the conditions for maintaining their current leaning state, to make their behavior look more natural while saving resources.

AdamTadeusz:
> We can add a server only section which checks if a player is a bot, and does the same thing as whats done client side for the local player except with far fewer tracelines. We could add a time variable so that bots stop leaning only after a set number of seconds pass without the conditions for maintaining their current leaning state, to make their behavior look more natural while saving resources.

Actually a smarter way to go about this with a very similar end result would be to only perform the calculation for leaning for bots every x seconds

Rainyan:
> > We can add a server only section which checks if a player is a bot, and does the same thing as whats done client side for the local player except with far fewer tracelines. We could add a time variable so that bots stop leaning only after a set number of seconds pass without the conditions for maintaining their current leaning state, to make their behavior look more natural while saving resources.
> 
> Actually a smarter way to go about this with a very similar end result would be to only perform the calculation for leaning for bots every x seconds

If there's an issue of this causing a perf spike, you could also consider staggering this for example with modulo of the bot index, so that you're guaranteeing at most X expensive operations per think cycle regardless of the bot count. Something like (pseudocode): `if (botindex % test != 0) return; test = (test+1)%MAX_PLAYERS`

Masterkatze (approved):

PR #809, opened 2024-11-06
Fair spawn mechanics

### LINKED ISSUE #793: [BUG] Players are (sometimes) able to respawn
### Build Info

20241102_14c4282

### Description

Sometimes, you are able to respawn into the round, this usually happens at the start of the round.

### To Reproduce

Not exactly sure how to replicate it in a server with players, but replicating it with bots I was able to do it as follows:

1. When you are in the pre-round start freeze time, type kill in console OR throw a nade at yourself and die from it
2. The fade to black will occur and you will be respawned

### Expected behavior

You should not respawn in CTG

### Actual behavior

You get respawned into your spawn

### Operating System

- [X] Windows
- [ ] Linux

### Version/Distro

10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

https://github.com/user-attachments/assets/ee9d0721-ccd3-4398-bb15-29e9d2d34e80

I have to note that the method did not seem the most consistent, sometimes I could not respawn even if I did it fast enough. Also, this happened to me in a playtest we had back in September. This seems to happen usually around the start of the round in the first ~15 seconds (In the second example, I had died at around 2:20 and the freeze time ended at 2:35)

AdamTadeusz:
## Description

Prevents players from spawning if previously spawned during that ongoing round, and enforces the DEATH_ANIMATION_TIME in game modes with respawns enabled.

Ensures people who spawn during freeze time are frozen.

Also stops spectators from being respawned and frozen during round start.

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #793

### DIFF
--- a/mp/src/game/shared/neo/neo_gamerules.h
+++ b/mp/src/game/shared/neo/neo_gamerules.h
@@ -246,6 +246,7 @@ public:
 #endif
 	bool IsRoundPreRoundFreeze() const;
 	bool IsRoundLive() const;
+	bool IsRoundOn() const;
 	bool IsRoundOver() const;
 #ifdef GAME_DLL
 	void GatherGameTypeVotes();
@@ -349,6 +350,8 @@ public:
 	{
 		int xp;
 		int deaths;
+		bool spawnedThisRound;
+		float deathTime;
 	};
 	// AccountID_t <- CSteamID::GetAccountID
 	CUtlHashtable<AccountID_t, RestoreInfo> m_pRestoredInfos;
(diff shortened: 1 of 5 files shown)


brysondev (approved):

mp/src/game/shared/neo/neo_gamerules.h
```diff
@@ -349,6 +350,8 @@ class CNEORules : public CHL2MPRules, public CGameEventListener
 	{
 		int xp;
 		int deaths;
+		bool spawnedThisRound;
+		float deathTime;
```
Masterkatze, line 354: Seems weird that `deathTime` is `float` when we don't use that value other than comparing it to zero (comparing float without epsilon by the way, which is questionable), it can be `bool` then.

AdamTadeusz: deathTime is passed to SetDeathTime in neo_client

looking through the code quickly when searching for instances of m_flDeathTime, I guess it's value may as well be 0 and thus no longer needs to be stored in m_pRestoredInfos once m_flDeathTime + DEATH_ANIMATION_TIME < gpGlobals->curtime, so I could use this comparison instead if you'd like

Masterkatze: Hmm, somehow I have missed this, LGTM

Masterkatze (approved):

PR #811, opened 2024-11-06
Glows refactor, automatically updated for new players and players switching teams

AdamTadeusz:
## Description

Glow effect updated for all players in callback when convar value is changed

Glow effect status and colour now updated when player changes team (except bots)

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0


mp/src/game/client/glow_outline_effect.cpp
```diff
@@ -12,13 +12,35 @@
 #include "materialsystem/itexture.h"
 #include "view_shared.h"
 #include "viewpostprocess.h"
+#ifdef GLOWS_ENABLE
+#ifdef NEO
+#include "neo_gamerules.h"
```
brysondev, line 17: NIT: `#if defined NEO && defined GLOWS_ENABLED`? Would look nicer imo but ignore if we have a standard I am unaware of :^)

mp/src/game/client/glow_outline_effect.cpp
```diff
@@ -12,13 +12,35 @@
...
+#endif // NEO
+#endif // GLOWS_ENABLE
 
 #define FULL_FRAME_TEXTURE "_rt_FullFrameFB"
 
 #ifdef GLOWS_ENABLE
```
brysondev, line 23: NIT: no `#endif` for this?

brysondev (approved):
LGTM, just some nitpicking

mp/src/game/client/glow_outline_effect.cpp
AdamTadeusz: Missing an #endif would be a major problem, in this case its at the end of the file.

brysondev (approved):

mp/src/game/client/clientmode_shared.cpp
```diff
@@ -69,6 +69,7 @@ extern ConVar replay_rendersetting_renderglow;
 #include "c_neo_player.h"
 #include <GameUI/IGameUI.h>
 #include "ui/neo_loading.h"
+#include "neo_gamerules.h"
```
Masterkatze, line 72: `#ifdef NEO` ?

mp/src/game/client/neo/c_neo_player.cpp
```diff
@@ -895,13 +895,6 @@ void C_NEO_Player::PreThink( void )
 			// NEO TODO (Adam) since the stuff in C_NEO_PLAYER::Spawn() only runs the first time a person spawns in the map, would it be worth moving some of the stuff from there here instead?
 #ifdef GLOWS_ENABLE
 			// Disable client side glow effects of all players
```
Masterkatze, line 897: Leftover comment?

mp/src/game/client/clientmode_shared.cpp
AdamTadeusz: yes, on line 68

mp/src/game/client/neo/c_neo_player.cpp
AdamTadeusz: changing the value of this convar calls a callback function which does the same for loop, so the comment is factually correct at least. I can remove it if you think its unnecessary

mp/src/game/client/clientmode_shared.cpp
Masterkatze: real

mp/src/game/client/neo/c_neo_player.cpp
Masterkatze: No, it's fine then

Masterkatze (approved):

PR #815, opened 2024-11-18
Corpses A posing

AdamTadeusz:
## Description

Set the model of the corpse ragdoll after it is created, instead of setting the player model before the ragdoll is created.

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0


AdamTadeusz:
This is a fix to the pr that introduced gibs, can't find it but #252 is also related.

brysondev (approved):

Agiel (dismissed):

brysondev (approved):
https://github.com/user-attachments/assets/5e8d8709-bf25-4b4f-bc78-4c2d6824d795

AdamTadeusz:
ok looks like it works still

Agiel (approved):

PR #816, opened 2024-11-18
Toggle Walk

### LINKED ISSUE #786: [FEATURE] Allow option for walking to be toggled
### Description

Toggling walk is the default setting in OGNT and bringing it back would be parity.

### Proposed Solution

As someone who is used to using walk as a toggle for years in NT, bringing it back would be easier for me to transition to nt;re. Furthermore, giving players more options in terms of controls/settings is a good thing most of the time (especially so in this case in my opinion). Having the option to switch between hold vs toggle to walk in settings would be ideal.

### Alternatives Considered

_No response_

### Additional Context

_No response_

AdamTadeusz:
## Description

can toggle walk, can switch back to hold to walk and still have it work if walking is already toggled on, fixes to sprinting when walking.
## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #786

### DIFF
--- a/mp/src/game/client/in_main.cpp
+++ b/mp/src/game/client/in_main.cpp
@@ -471,6 +471,7 @@ void IN_MoveleftUp( const CCommand &args ) {KeyUp(&in_moveleft, args[1] );}
 void IN_MoverightDown( const CCommand &args ) {KeyDown(&in_moveright, args[1] );}
 void IN_MoverightUp( const CCommand &args ) {KeyUp(&in_moveright, args[1] );}
 void IN_WalkDown( const CCommand &args ) {KeyDown(&in_walk, args[1] );}
+#ifdef NEO
 void IN_WalkUp( const CCommand &args )
 {
 	int k = atoi(args[1]);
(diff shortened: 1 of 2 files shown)


AdamTadeusz:
There's still room for improvement, like for example when toggling on walking with the +walk command, using the (walk_toggle) keybind will do nothing the first time it is pressed, and then its behavior returns to normal, but I think that we don't really need to worry about that, unless we want to make the experience even more smooth for people who use funny aliases to do weird things with movement

brysondev (approved):

mp/src/game/client/in_main.cpp
```diff
@@ -471,7 +471,16 @@ void IN_MoveleftUp( const CCommand &args ) {KeyUp(&in_moveleft, args[1] );}
...
+void IN_WalkUp( const CCommand &args )
+{
+	int k = atoi(args[1]);
+	if (in_walk.down[0] == -1 && in_walk.down[1] == k)
+	{ // Releasing walk button using hold to walk when walking is already toggled using toggle walk, have to release toggle walk // NEO TODO (Adam) Any way to figure out it was exactly toggle walk that is holding this button down? Does it matter?
+		in_walk.down[0] = k;
+		in_walk.down[1] = 0;
+	};
+	KeyUp(&in_walk, args[1] );
+}
```
DESTROYGIRL, line 483: Do we need to contain these changes within a NEO ifdef orrrrr why not

AdamTadeusz: yes we should 😔

DESTROYGIRL: 😔

brysondev (approved):
its not that deep bruh

DESTROYGIRL (approved):

PR #817, opened 2024-11-20
[New Feature Idea] See through smokes when using xray in spectator mode

AdamTadeusz:
## Description

remove smoke grenade overlay and lower smoke particle opacity when spectating and using xray

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

### DIFF
--- a/mp/src/game/client/c_particle_smokegrenade.cpp
+++ b/mp/src/game/client/c_particle_smokegrenade.cpp
@@ -687,7 +687,11 @@ inline void C_ParticleSmokeGrenade::ApplyDynamicLight( const Vector &vParticlePo
 	}
 }
 
-
+#ifdef NEO
+#ifdef GLOWS_ENABLE
+extern ConVar glow_outline_effect_enable;
+#endif // GLOWS_ENABLE
+#endif // NEO
 void C_ParticleSmokeGrenade::RenderParticles( CParticleRenderIterator *pIterator )
 {
 	if (!CanSeeThroughSmokeGrenades())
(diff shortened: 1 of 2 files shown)


Rainyan (changes requested):
I like the idea, if we have consensus for adding this.

But with the current implementation, clients playing in the match can also toggle `glow_outline_effect_enable` to see past smokes, which shouldn't be allowed. We should do a `IsAlive` check in there somewhere so players can't abuse it.

Also, we could improve the visuals so it's more obvious to the spectator (and their audience, in case of a stream observer) that the smoke is being reduced and doesn't reflect what the players see. Increasing the alpha fraction a bit more could help(?) Drawing an outline effect to the smoke puff itself, similar to the xray feature could be another option. But I guess we could iterate on it more later.

AdamTadeusz:
> I like the idea, if we have consensus for adding this.
> 
> But with the current implementation, clients playing in the match can also toggle `glow_outline_effect_enable` to see past smokes, which shouldn't be allowed. We should do a `IsAlive` check in there somewhere so players can't abuse it.
> 
> Also, we could improve the visuals so it's more obvious to the spectator (and their audience, in case of a stream observer) that the smoke is being reduced and doesn't reflect what the players see. Increasing the alpha fraction a bit more could help(?) Drawing an outline effect to the smoke puff itself (similar to the xray feature) could be another option. But I guess we could iterate on it more later.

I'm sure we will restrict the spectator options for people who are playing in a competitive setting, but considering what dead players can do in the original in a casual game, I didn't see the need to draw the line here at preventing dead players from seeing through smoke.

I don't know what the ideal percentage is for opacity compared to the original, I thought this was fine, I would appreciate if you could maybe play around with the branch and find a value that you deem suitable so we don't have to go back and forth as much. It's the 0.25 in c_particle_smokegrenade.cpp

I'm sure we could add outlines to the smoke particles but that would require a significant refactor. Currently i think the glows effect can only be applied to cbasecombatcharacters and their attachments, and I think with a little refactoring which valve developer community has a guide on this can be extended to all cbaseanimating and their derivatives. Particles don't inherit from cbaseanimating so there would be a bit of work there to get that to work.

PR #818, opened 2024-11-20
Spectated player vision mode

### LINKED ISSUE #791: [BUG] When spectating a player and they're using a vision mode, the vision mode is not rendered
### Build Info

20241102_14c4282

### Description

When spectating someone, you do not see their vision mode when they are using one, you can only hear the sound of toggling the mode or see their viewmodel change.

### To Reproduce

1. Spectate someone

### Expected behavior

You should be able to see when theyre using a vision mode.

### Actual behavior

You do not see their vision mode, you can only hear the sound of toggling the vision mode and viewmodels changing colors.

### Operating System

- [X] Windows
- [ ] Linux

### Version/Distro

Windows 10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

The viewmodel lights up from the player using the vision mode, but you cannot see the vision mode itself

AdamTadeusz:
## Description

get local player's spectating target when spectating to use instead of local player when drawing models and vision mode shader. Vision mode shaders disabled when using xray.

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #791


brysondev (approved):
LGTM but maybe we should future proof so its only people on team spectator can use this functionality. (Depending on how we handle comp scene, being able to "ghost" for teammates while using thermal vision will be a problem)

AdamTadeusz:
> LGTM but maybe we should future proof so its only people on team spectator can use this functionality. (Depending on how we handle comp scene, being able to "ghost" for teammates while using thermal vision will be a problem)

To clarify, you are suggesting we disable the option to see the spectated players vision mode for dead players entirely? I can see the argument for enforcing the current spectated targets vision mode and its status for competitive play as opposed to allowing dead players to switch between all vision modes like what casters can currently do in the original with the right plugin. This is also what this branch is doing, on account of us not having the switching between vision modes when spectating functionality. However with vision modes disabled for dead players spectating supports would be very similar to how spectating for dead players currently works in competitive.

Also vision modes are supposed to be limiting in ways that allow them to only be used in certain situations. Disabling vision modes for dead players non-intuitively may allow them to have a clearer picture of the situation and allow for better ghosting (think a spectated support in thermals being unable to see a cloaked recon, but the same cloaked recon could be easier to spot without thermals by a spectating team mate), personally I think the information a dead player receives in a competitive scenario when spectating a team mate should be equal or less, this has the potential of giving them unique information the player being spectated may not have.

brysondev:
Dead tammates have current spec target's vision mode, and people casting/spectating are free to use whatever mode. 

But yeah maybe I was misinterpreting what was happening. If it's not allowing dead teammates to change vision modes then we're fine.

brysondev (approved):

Agiel (approved):

AdamTadeusz:
models in motion vision weren't working when spectating, this branch might be older than the pr that moved motion vision stuff to a base class, anyway that works now too

https://github.com/user-attachments/assets/7361de19-ada5-4acb-9e98-56d7627207b9

Agiel (approved):

PR #822, opened 2024-11-21
roundstatehud uses neoscoreboard image list

### LINKED ISSUE #639: *NEEDS REPRO STEPS!* Steam Avatars sometimes don't load in in the Top Hud
### Build Info

20240919_9e83cc3

### Description

When new players join a server (← SUPPOSEDLY) while you were already playing on it, their Steam Avatars don't load in in the Top Hud and appear as either completely white squares or team icons.

The Steam Avatars do show up in the leaderboard.

Examples:
image
image

### To Reproduce

`*NEEDS INFO*`

### Expected behavior

All players have their steam avatars in the Top Hud.

### Actual behavior

Some players' steam avatars are blank or are default team icons.

### Operating System

- [ ] Windows
- [X] Linux

### Version/Distro

Arch

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

### LINKED ISSUE #792: [BUG] Doing hud_reloadscheme causes the profile picture icons on the hud turn into Jinrai/NSF icons
### Build Info

20241102_14c4282

### Description

When you do hud_reloadscheme, the profile pictures on the new hud seem to get replaced with the Jinrai/NSF icons. 

### To Reproduce

1. Be in game with players
2. Write hud_reloadscheme in console or press a bind if you have one for it

### Expected behavior

The profile pictures should stay

### Actual behavior

The profile pictures turn into Jinrai/NSF icons 

### Operating System

- [X] Windows
- [ ] Linux

### Version/Distro

10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

AdamTadeusz:
## Description

roundstate hud uses avatars defined in the neoscoreboard, added a global pointer to the scoreboard to access them

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #792
- fixes #639


brysondev (dismissed):

mp/src/game/client/neo/ui/neo_scoreboard.cpp
```diff
@@ -738,10 +763,10 @@ void CNEOScoreBoard::GetPlayerScoreInfo(int playerIndex, KeyValues *kv)
 //-----------------------------------------------------------------------------
 // Purpose:
 //-----------------------------------------------------------------------------
-void CNEOScoreBoard::UpdatePlayerAvatar( int playerIndex, KeyValues *kv )
+void CNEOScoreBoard::UpdatePlayerAvatar( int playerIndex, KeyValues *kv, bool skipKey )
```
Rainyan, line 766: Since we're already guarding for the nullptr case for the `kv` parameter, and it seems like the only place where we set the non-default `true` value for the newly introduced `skipKey` parameter also has the `kv` argument passed as nullptr:

```cpp
// At line 326 of mp/src/game/client/neo/ui/neo_scoreboard.cpp
UpdatePlayerAvatar(pPlayer->index, nullptr, true);
```

couldn't we simply rely on that in-band `kv` nullptr check, so we could get away without introducing the third `skipKey` parameter to the prototype, for simplicity? As both `!kv` and `!skipKey` are essentially doing the same work of avoiding the `kv` ptr indirection, unless I misunderstand the parameter's purpose.

Rainyan (changes requested):

mp/src/game/client/neo/ui/neo_scoreboard.cpp
AdamTadeusz: I was hesitant to modify the behavior of this function where it is called without the skipKey value set where kv may be missing, though in hindsight that is probably silly.

brysondev (approved):

Rainyan (approved):

PR #824, opened 2024-12-02
Player class filter for maps

### LINKED ISSUE #655: [FEATURE] Player class filter entity
### Description

Entity to filter by player class (recon,assault,support,VIP,etc.)

### Proposed Solution

Should function very similar to other filter entities (https://developer.valvesoftware.com/wiki/Category:Filters). Proposed name could be "filter_activator_playerclass". This would make some complex logic for maps very trivial.

An entity like this already exists in TF2: https://developer.valvesoftware.com/wiki/Filter_tf_class

### Alternatives Considered

_No response_

### Additional Context

_No response_

DESTROYGIRL:
## Description

A player class filter that grabs it from neo_player... please let me know if this sucks.
## Toolchain

- Windows MSVC VS2022

## Linked Issues

- resolves #655


AdamTadeusz:
I don't know how this would be used, do you have a map where this is implemented?

Semantically I see no issues, I was thinking whether checking the entity index instead of calling IsPlayer would be cheaper since player entities are guaranteed an entity index between 1 and either gpGlobals->maxplayers or  MAX_PLAYERS, but the other filters use IsPlayer so I wouldn't bother trying to optimize this particular bit.

One thing I did notice is 2 out of the 3 preexisting filter class names begin with a C and FilterNeoClass does not. Not sure whether thats important but it seems like this filter is more similar to the filters down below that have a C than the damage filter, but again probably not important

DESTROYGIRL:
> I don't know how this would be used, do you have a map where this is implemented?
> 
> Semantically I see no issues, I was thinking whether checking the entity index instead of calling IsPlayer would be cheaper since player entities are guaranteed an entity index between 1 and either gpGlobals->maxplayers or MAX_PLAYERS, but the other filters use IsPlayer so I wouldn't bother trying to optimize this particular bit.
> 
> One thing I did notice is 2 out of the 3 preexisting filter class names begin with a C and FilterNeoClass does not. Not sure whether thats important but it seems like this filter is more similar to the filters down below that have a C than the damage filter, but again probably not important

I believe the C at the beginning is just a naming scheme for stuff to represent that it's server side, I forgot about that. Not sure how to change it now I've made the pull request

This could be implemented in a variety of ways where you would want players of a specific class to activate something but not others. Like if you wanted only supports to take an elevator, or only recons be light enough not to break a bridge or something. This is nessecary functionality for the tutorial map I'm working on

AdamTadeusz:
> > I don't know how this would be used, do you have a map where this is implemented?
> > Semantically I see no issues, I was thinking whether checking the entity index instead of calling IsPlayer would be cheaper since player entities are guaranteed an entity index between 1 and either gpGlobals->maxplayers or MAX_PLAYERS, but the other filters use IsPlayer so I wouldn't bother trying to optimize this particular bit.
> > One thing I did notice is 2 out of the 3 preexisting filter class names begin with a C and FilterNeoClass does not. Not sure whether thats important but it seems like this filter is more similar to the filters down below that have a C than the damage filter, but again probably not important
> 
> I believe the C at the beginning is just a naming scheme for stuff to represent that it's server side, I forgot about that. Not sure how to change it now I've made the pull request
> 
> This could be implemented in a variety of ways where you would want players of a specific class to activate something but not others. Like if you wanted only supports to take an elevator, or only recons be light enough not to break a bridge or something. This is nessecary functionality for the tutorial map I'm working on

image

Looks like you're working on your master branch, so any new commits you make locally to master that you then push will show up here. In the future consider making a separate branch for each new feature you are working on, that way you can work on multiple features at once

mp/src/game/server/filters.cpp
```diff
@@ -10,6 +10,8 @@
 #include "ai_squad.h"
 #include "ai_basenpc.h"
 
+#include "neo_player.h"
```
Rainyan, line 13: We'd had the general convention of guarding NEO specific blocks of code inside Valve implementation with the:
```cpp
#if NEO
// Here's some NT code
#endif // NEO
```
similar to what Valve does with their games that do this.

So since we are not inside a purely NT unit here, this should be:
```cpp
#if NEO
#include "neo_player.h"
#endif
```
or similar.

Obviously we are only building for NT and not for other Source games, but abiding by this convention lets us locate NT-specific stuff in the code base more easily.

mp/src/game/server/filters.cpp
```diff
@@ -324,6 +326,40 @@ BEGIN_DATADESC( FilterTeam )
 END_DATADESC()
 
 
+// ###################################################################
+//	> FilterNeoClass
```
Rainyan, line 330: Likewise, this class implementation should be guarded by a NEO ifdef.

mp/src/game/server/filters.cpp
```diff
@@ -324,6 +326,40 @@ BEGIN_DATADESC( FilterTeam )
...
+
+public:
+	int m_iFilterNeoClass;
+
+	bool PassesFilterImpl(CBaseEntity* pCaller, CBaseEntity* pEntity)
+	{
+		if (!pEntity || !pEntity->IsPlayer())
+			return false;
+
+		CNEO_Player* pPlayer = dynamic_cast<CNEO_Player*>(pEntity);
```
Rainyan, line 345: Since the only kind of player we have is `CNEO_Player`, and we've already passed `pEntity->IsPlayer` to get here, we can safely assume this must be a `CNEO_Player`. Therefore, this can be a `static_cast` for performance. Or you may use the `assert_cast`, which will assert the `dynamic_cast` for debug builds, but optimize it into `static_cast` for release builds, if you want the guarantee nonetheless.

mp/src/game/server/filters.cpp
```diff
@@ -324,6 +326,40 @@ BEGIN_DATADESC( FilterTeam )
...
+		if (!pPlayer)
+			return false;
```
Rainyan, line 347: Using `static_cast` (or `assert_cast`), this pointer *must* be valid. So we can skip this check entirely.

Rainyan (changes requested):
Thanks for the PR, left some feedback.

DESTROYGIRL:
> Thanks for the PR, left some feedback.

Thank you for your help, hopefully these changes should suffice
I was very wary of this filter being tested against something that was not a player and it causing problems

Rainyan (approved):
So I don't currently have the Source SDK on Windows set up to validate this with an actual Hammer test map compile (or any kind of Windows installation for that matter), but this builds and runs correctly, and it seems like the entity logic does exactly what it's supposed to, so I'm giving it a thumbs up for merge.

AdamTadeusz:
I don't have hammer setup for rebuild, but I guess hammer doesn't stop you from making a custom entity and giving it custom fields. There's no errors in the console but these don't seem to do anything
image

DESTROYGIRL:
> I don't have hammer setup for rebuild, but I guess hammer doesn't stop you from making a custom entity and giving it custom fields. There's no errors in the console but these don't seem to do anything image

You need a trigger entity or another entity to test the activator (which should be a player) before it sends an output
Make a trigger_multiple, enter the targetname of the filter in "filter name", then give the trigger an output like playing a sound or something

AdamTadeusz (approved):

PR #827, opened 2024-12-11
neo_bloom_controller

### LINKED ISSUE #825: neo_bloom_controller is not implemented
### Build Info

20241102_14c4282

### Description

Mappers can use the neo_bloom_controller entity in Hammer to control the amount of bloom in a map. Currently, it is not implemented into ntre

### To Reproduce

.

### Expected behavior

.

### Actual behavior

.

### Operating System

- [X] Windows
- [ ] Linux

### Version/Distro

10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

AdamTadeusz:
## Description

Adds neo_bloom_controller for backwards compatibility with older maps. Values controlled by this entity can be controlled with env_tonemap_controller already, unless the original purpose of this was to control the custom bloom that grey supposedly implemented, in which case this is a bandaid fix untill we can implement the custom bloom (not a clue in the world how we would go about doing that).

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #825


mp/src/game/client/viewpostprocess.cpp
```diff
@@ -786,6 +787,12 @@ static float GetCurrentBloomScale( void )
 	{
 		flCurrentBloomScale = mat_bloomscale.GetFloat();
 	}
+#ifdef NEO
+	if (!g_pMaterialSystemHardwareConfig->GetHDREnabled())
+	{
+		flCurrentBloomScale *= 4; // NEOTOD (Adam) the base values defined in neo_bloom_controller in old maps seem way too small, need to figure out what the right multiplier is for this. 4 seems to be in the right ballpark. The translation from non-hdr to hdr is 0.3 -> 1.0 according to valve developer community, a multiplier of 3.333 could be the right value
```
brysondev, line 793: We should not be going off of "old maps". We're remaking the game, maps "included" (recompile on sdk2013). So we shouldn't be setting this arbitrarily. Lets have a discussion before settling on this.

mp/src/game/server/neo/neo_bloom_controller.cpp
```diff
@@ -0,0 +1,69 @@
...
+//-----------------------------------------------------------------------------
+class CNeoBloomController : public CPointEntity
+{
+	DECLARE_CLASS( CNeoBloomController, CPointEntity );
+public:
+	DECLARE_DATADESC();
+	DECLARE_SERVERCLASS();
+
+	void	Spawn( void );
+	int		UpdateTransmitState( void );
```
brysondev, line 25: NIT: indent

brysondev (changes requested):
LGTM, but as per the comment I made we really should decide on an "update" control value and not base it off "old maps". All of them should be recompiled for the engine update, and therefore should reflect it.

mp/src/game/client/viewpostprocess.cpp
DESTROYGIRL: If this multiplier is based off of how the bloom in the original looked, then I don't see an issue with this. Maybe some comparison shots would be good to have.
If things don't look right after we recompile maps with the new tools, then we can change the bloom value in the entity for those maps specifically, which then won't affect the intended look of custom maps that haven't been compiled for the new engine

brysondev: We planned on recompiling all maps for RE. If that isn't what we're doing then yeah this is fine.

DESTROYGIRL: I'm talking about custom maps for OG specifically - it's down to the original creators of them to recompile those, otherwise we should make an attempt to have the bloom values as close to original as possible for them, since we can simply modify the values for the official maps when we go to recompile them, and can't for those
But it's not really a big deal. The lighting looks pretty different as-is so the changes in bloom probably aren't even noticeable at first glance

brysondev: Yeah good point. Well, whenever Zwia comes back I'll just leave this here for him to say his piece and resolve this.

AdamTadeusz: Finding and using the right value here has the upsides that all old maps should look closer to the original, we don't have to recompile all the maps (which isn't a lossless procedure for maps we only have the bsp) to edit this specific value only (even if this has to be done for other reasons), and mappers can continue making maps using the bloom values that they are used to.

Downsides are that its ugly programmatically, anyone who loads a map that knows about this will sub-consciously know that a value had to be multiplied when the map was loaded to facilitate, I forget what its called but something like the "Technical baggage" i.e the cost of this game being made using preexisting resources, which means they will have a worse time in the game (and I'm being totally serious here). Also we have to consider the run-time cost of multiplying this value once per map load (not so serious here).

I don't know what the correct solution is, but unless we are sure that we can recompile every single map in the existence of neotokyo with an updated value among other things for parity's sake, I'm leaning towards this multiplier. At the very least I should probably make it a constexpr with a name so we can find it easier in the future

mp/src/game/server/neo/neo_bloom_controller.cpp
AdamTadeusz: 😔

brysondev (approved):

mp/src/game/server/neo/neo_bloom_controller.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+private:
+	CNetworkVar( float, m_flBloomAmountSetting);
+};
+
+LINK_ENTITY_TO_CLASS( neo_bloom_controller, CNeoBloomController );
+
+BEGIN_DATADESC( CNeoBloomController )
+	DEFINE_FIELD( m_flBloomAmountSetting, FIELD_FLOAT ),
+	DEFINE_INPUT( m_flBloomAmountSetting, FIELD_FLOAT, "bloomamount"),
```
DESTROYGIRL, line 38:
```suggestion
	DEFINE_KEYFIELD( m_flBloomAmountSetting, FIELD_FLOAT, "bloomamount"),
	DEFINE_INPUTFUNC(FIELD_FLOAT, "SetBloomScale", InputSetBloomScale),
```

On my side, this entity doesn't work at all as you seem to have defined your keyvalue and input incorrectly. This should fix it?

DESTROYGIRL (approved):

brysondev (approved):

PR #832, opened 2024-12-14
[NEW FEATURE] tachi spawns in the automatic mode

AdamTadeusz:
## Description

Not parity, thoughts?


nullsystem:
Convar + UI setting to toggle if the player wants it auto or not by default

kassibuss:
Nothing worse than a tachi left on semi-auto, I would add it in. Also, what marter said.

AdamTadeusz:
> Convar + UI setting to toggle if the player wants it auto or not by default

https://tenor.com/view/thanos-snap-emoji-crying-gif-22298553

AdamTadeusz:
It's a networked entity so that's going to be a little tricky

Rainyan:
Unpopular NT purist contrarian opinion: the Tachi firemode should default to semi-auto, and you should not be able to switch it at all until the freeze-time ends; this is a skill with timing the wepmode toggle keyboard input tick-perfect without losing your spawn movement inputs.

But realistically I realize I am probably the only person in the entire universe who holds this opinion, so meh, whatever. :smile:

Issue #819, opened 2024-11-20
[FEATURE] Input to set player model


DESTROYGIRL:
### Description

Previously it was possible to change a players model by changing the `modelindex` value on the `player` entity. This is no longer functional in the new engine version and was very tedious to work with, so I think it would be better to create some new functionality for this rather than fixing the old method

### Proposed Solution

Give the `player` entity a new input, say `SetPlayerModel`, with the parameter being a the path to the MDL file that the player's model should switch to when recieving this input

PR #736, opened 2024-10-15
make sure m41L has scope behavior

brysondev:
## Description

Add scope functionality to the M41L.
Assets PR: https://github.com/NeotokyoRebuild/neoAssets/pull/10
## Toolchain

- Windows MSVC VS2022

PR #834, opened 2024-12-18
Input for setting custom player model

DESTROYGIRL:
## Description
Input for the player entity that sets the player model as determined by the path included in the input parameters
It will also disable gibbing on that player until they respawn so their corpse is consistent with the model change
Assets PR: https://github.com/NeotokyoRebuild/neoAssets/pull/17

## Toolchain

- Windows MSVC VS2022

## Linked Issues

- related #819

### DIFF
--- a/mp/src/game/server/neo/neo_player.cpp
+++ b/mp/src/game/server/neo/neo_player.cpp
@@ -111,6 +111,10 @@ DEFINE_FIELD(m_szNeoName, FIELD_STRING),
 DEFINE_FIELD(m_szNeoClantag, FIELD_STRING),
 DEFINE_FIELD(m_szNameDupePos, FIELD_INTEGER),
 DEFINE_FIELD(m_bClientWantNeoName, FIELD_BOOLEAN),
+
+// Inputs
+DEFINE_INPUTFUNC(FIELD_STRING, "SetPlayerModel", InputSetPlayerModel),
+
 END_DATADESC()
 
 static constexpr int SHOWMENU_STRLIMIT = 512;
@@ -415,6 +419,7 @@ CNEO_Player::CNEO_Player()
 	m_bPreviouslyReloading = false;
 	m_bLastTickInThermOpticCamo = false;
 	m_bIsPendingSpawnForThisRound = false;
+	m_bAllowGibbing = true;
 
 	m_flNextTeamChangeTime = gpGlobals->curtime + 0.5f;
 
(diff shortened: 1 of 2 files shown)


mp/src/game/server/neo/neo_player.cpp
```diff
@@ -1960,6 +1972,21 @@ void CNEO_Player::SetPlayerTeamModel( void )
 #endif
 }
 
+// Input to set the player's model (not skin)
+void CNEO_Player::InputSetPlayerModel( inputdata_t & inputdata )
+{
+	const char* modelpath = inputdata.value.String();
+
+	if (modelpath && *modelpath)
```
brysondev, line 1980:
```suggestion
	if (modelpath)
```

brysondev (changes requested):
Just some minor things. Also feel free to just move the bool down a few lines instead of adding "private:".

brysondev:
Oh also, if you could update the description to include the referenced NeoAssets PR as well that would be appreciated. 

Example: #736

mp/src/game/server/neo/neo_player.cpp
DESTROYGIRL: This stops the game from crashing when the input is sent and there is no input parameter specified

AdamTadeusz:
Would it be worth adding a check whether the model the player is set to is one of the models for which we do have gibs, so that we can re-enable gibs for a player without having to respawn them?

DESTROYGIRL:
> Would it be worth adding a check whether the model the player is set to is one of the models for which we do have gibs, so that we can re-enable gibs for a player without having to respawn them?

Potentially, but I don't really think it's worth it since players can choose their own skin before spawning so there's not much reason to use this input to pick another one. Unless it's switching back from a custom model to a regular one, there's no way of getting the player's skin from map logic so it wouldn't be consistent with what they chose anyway.

mp/src/game/server/neo/neo_player.cpp
nullsystem: *modelpath (which is equivalent to modelpath[0] == '\0') is for checking empty string/null terminator, so there's no need to change this if statement

brysondev (approved):

mp/src/game/server/neo/neo_player.cpp
Rainyan: The `value.String()` ptr result resolves via:

https://github.com/NeotokyoRebuild/neo/blob/839d747779c929a50a4a16fdc9b3abef8c099282/mp/src/public/string_t.h#L60

which explicitly checks for nullptr, hence the `if (modelpath)` check is redundant. Checking for null terminator of the indirected pointer is probably prudent however.

PR #840, opened 2025-01-10
GH-839 Move grenade detonation without screenshake

### LINKED ISSUE #839: Player's camera/pov shakes when standing near an explosion
Does not happen in OGNT, should be removed(?)

VantoNortim:
## Description
Standing near a grenade explosion no longer triggers screnshake. This is not an issue with detpack.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #839

### DIFF
--- a/mp/src/game/server/neo/neo_grenade.h
+++ b/mp/src/game/server/neo/neo_grenade.h
@@ -26,6 +26,7 @@ class CNEOGrenadeFrag : public CBaseGrenade
 
 public:
 	virtual void	Explode(trace_t* pTrace, int bitsDamageType);
+	virtual	void	Detonate(void);
 
 	void	Spawn(void);
 	void	Precache(void);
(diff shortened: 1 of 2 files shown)


AdamTadeusz:
can probably override GetShakeAmplitude instead to just return a value of 0

mp/src/game/server/neo/neo_grenade.h
```diff
@@ -26,6 +26,7 @@ class CNEOGrenadeFrag : public CBaseGrenade
 
 public:
 	virtual void	Explode(trace_t* pTrace, int bitsDamageType);
+	virtual	void	Detonate(void);
```
Rainyan, line 29: Nit: should mark as `override`. Also, the `(void)` is redundant for modern C++. So preferrably: `virtual void Detonate() override`

Rainyan (changes requested):
Since our `::Detonate` override is identical to BaseClass' except for omitting the

```cpp
if ( GetShakeAmplitude() )
{
    // ...
```

check, I would prefer if instead of overriding this method, we overrode the `::GetShakeAmplitude` virtual instead & returned a zero value there, such that the base class if check did not pass for our explosives.

AdamTadeusz:
https://github.com/NeotokyoRebuild/neo/issues/839#issuecomment-2582640806

Should still override this function but it looks like 0 isn't the right value to return

brysondev (approved):

PR #851, opened 2025-01-18
Raise map preview camera position

### LINKED ISSUE #850: Map preview camera does not start at correct offset from info_player_start
When first loading a map the camera defaults to the origin and angles of info_player_start. In OGNT, the camera starts higher by 64 units, at about crotch height with the player start preview model in hammer

AdamTadeusz:
## Description

Raise map preview camera by 64 units

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #850


DESTROYGIRL:
This also causes living players that spawn from info_player_start to spawn 64 units above where they should, which may affect singleplayer maps, but it's niche and can be accounted for by level designers

brysondev (approved):

DESTROYGIRL (approved):

PR #852, opened 2025-01-18
toolsblockbullets uses surfaceprop to block bullets instead of thickness of brush

### LINKED ISSUE #658: [FEATURE] Blockbullets material with anti-penetration compile flag
### Description

A new blockbullets material that relies on a compile flag to correctly block bullets. The current blockbullets material doesn't actually rely on a compile flag to block bullets- it's just an invisible wall. So in the game bullets will simply penetrate it

### Proposed Solution

Create a new material compile flag (https://developer.valvesoftware.com/wiki/Material_map_compile_flags) that prevents bullets from penetrating the brush

### Alternatives Considered

Make a new surfaceprop (https://developer.valvesoftware.com/wiki/Material_surface_properties) instead that stops bullets penetrating. It looks like CSGO already has this

### Additional Context

toolsblockpenetrate
Proposed tool texture

AdamTadeusz:
## Description

Relies on PR in neoassets ->>>>> https://github.com/NeotokyoRebuild/neoAssets/pull/20 <<<<<-

Give toolsBlockBullets a material we can query to return from HandleShotPenetration early meaning toolsBlockBullets no longer needs to rely on its thickness to stop bullets, plus a fix to HandleShotPenetration so bullets don't skip very thin brushes. There was already a traceline to detect the correct place to do a backwards trace from, the name of the traceline that found this next place is called nextObjectTrace, however this result mistakenly wasn't actually used later in the code.

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #658


DESTROYGIRL (changes requested):
Nittyest pick in existance but instead of replacing the old material you should just make a new one, toolsblockbullets is part of the hl2 base so it may cause conflicts for the compiler/hammer having two of the same material. Did you replace the old one in testing?
I included a tool texture in the issue so the two could be distinguished

AdamTadeusz:
> Nittyest pick in existance but instead of replacing the old material you should just make a new one, toolsblockbullets is part of the hl2 base so it may cause conflicts for the compiler/hammer having two of the same material. Did you replace the old one in testing? I included a tool texture in the issue so the two could be distinguished

I just did it this way because its easier, but I figured because of how we are mounting assets that it wouldn't be a problem, theres only ever one toolsblockbullets vmt at runtime. If you really think this will be a problem then I can look into making a new material instead

DESTROYGIRL:
> > Nittyest pick in existance but instead of replacing the old material you should just make a new one, toolsblockbullets is part of the hl2 base so it may cause conflicts for the compiler/hammer having two of the same material. Did you replace the old one in testing? I included a tool texture in the issue so the two could be distinguished
> 
> I just did it this way because its easier, but I figured because of how we are mounting assets that it wouldn't be a problem, theres only ever one toolsblockbullets vmt at runtime. If you really think this will be a problem then I can look into making a new material instead
> 
> (Edit) also this way any existing maps wouldn't have to be fixed to use the new brush, I imagine any use of toolsblockbullets in existing neotokyo maps was intended to block bullets completely and this slightly different behavior in the rebuild shouldn't cause any novel behavior. I was testing this on a map with the toolsblockbullets material and I found it nice that I could run the map both in neotokyo and in neotokyo rebuild fine
> 
> (Edit 2) we also have some other materials that we are overwriting in neoAssets, like changing the rabbitfrog engine material from an UnlitGeneric to a VertexLitGeneric, is that going to cause problems in hammer too?

It depends on your hammer setup really, half life 2 base assets will always be part of their own game assets folder, and always included, so it's not a good idea to have duplicates of those, whereas with rebuild we haven't solidified any "proper" mapping setup due to the current way assets are spread about. I personally have a copy of all OGNT assets and then have new rebuild assets added into that same game assets folder, so hammer & the compilers will only be looking into that folder rather than one for OGNT and one for rebuild- since ideally we'd like to package all OGNT assets with rebuild.
For the rabbitfrog specifically it doesn't matter since it's a model material and isn't used during compile unless specified

As for maps that already use blockbullets, none of the official maps seem to use it, and I doubt any community ones do either since it's inneffective at actually doing it's job. I used it in tetsujin to fake a prop's collision hull. I would say fair enough if we wanted rebuild maps to be playable in OGNT, but since that's not possible I'd say personally it's a bad practice

DESTROYGIRL (approved):

.github/workflows/cibuild.yml
```diff
@@ -22,7 +22,7 @@ jobs:
 
       matrix:
         platform:
-        - { display_name: 'Linux Native Libraries', os: ubuntu-latest, preset_os: linux }
+        - { display_name: 'Linux Native Libraries', os: ubuntu-24.04, preset_os: linux }
```
brysondev, line 25:
```suggestion
        - { display_name: 'Linux Native Libraries', os: ubuntu-22.04, preset_os: linux }
```
Just so some reasonably older distros can work with the artifacts (Debian 12)

.github/workflows/cibuild.yml
```diff
@@ -36,19 +36,19 @@ jobs:
...
+        build_dir=${workspace_dir//\\//}/src/build/$preset_name
         install_dir=$build_dir/install
         echo "source_dir=$source_dir" >> "$GITHUB_ENV"
         echo "build_dir=$build_dir" >> "$GITHUB_ENV"
         echo "install_dir=$install_dir" >> "$GITHUB_ENV"
         echo "preset_name=$preset_name" >> "$GITHUB_ENV"
 
     - name: Install packages
-      if: ${{ matrix.platform.os == 'ubuntu-latest' }}
+      if: ${{ matrix.platform.os == 'ubuntu-24.04' }}
```
brysondev, line 48:
```suggestion
      if: ${{ matrix.platform.os == 'ubuntu-22.04' }}
```
ditto

.github/workflows/cibuild.yml
```diff
@@ -197,18 +206,18 @@ jobs:
     name: Upload Latest Build
 
     needs: [build, pack-resources]
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
brysondev, line 209:
```suggestion
    runs-on: ubuntu-22.04
```
ditto

.github/workflows/cibuild.yml
```diff
@@ -223,11 +232,11 @@ jobs:
     name: Upload Release
 
     needs: [build, pack-resources]
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
brysondev, line 235:
```suggestion
    runs-on: ubuntu-22.04
```
ditto

.gitignore
```diff
@@ -1,38 +1,43 @@
...
-mp/game/neo/particles
-mp/game/neo/shaders
-mp/game/neo/models
+/game/neo/maps
+/game/neo/mapsrc
+/game/neo/materials
+/game/neo/particles
+/game/neo/shaders
+/game/neo/models
+game/neo/materials/temp
```
brysondev, line 24:
```suggestion
/game/neo/materials/temp
```

.gitignore
```diff
@@ -73,54 +78,54 @@ ipch
 .vs/
 Degub/
```
brysondev, line 79:
```suggestion
Debug/
```

brysondev (approved):

AdamTadeusz:
@brysondev were these comments supposed to be on a different PR?

brysondev:
> @brysondev were these comments supposed to be on a different PR?

I have no idea. They showed in this PR when I clicked review so I wrote comments.

AdamTadeusz:
> > @brysondev were these comments supposed to be on a different PR?
> 
> I have no idea. They showed in this PR when I clicked review so I wrote comments.


image
bro I didn't touch any of these

brysondev:
> > > @brysondev were these comments supposed to be on a different PR?
> 
> > 
> 
> > I have no idea. They showed in this PR when I clicked review so I wrote comments.
> 
> 
> 
> 
> 
> image
> 
> bro I didn't touch any of these

Github moment. Ignore them then lol

PR #855, opened 2025-01-20
single command for comp mode

### LINKED ISSUE #706: [FEATURE] Server variable for comp NT
### Description

Right now, all of the comp related variables have to be enabled separately. If there was a single command which would automatically turn all of them all, itd be easier for the server operators

### Proposed Solution

Adding something like sv_comp, where if enabled, automatically turns on all comp related server variables (reading up, pause mechanics, map veto etc). Basically the comp plugin, but as a native variable

### Alternatives Considered

_No response_

### Additional Context

_No response_

AdamTadeusz:
## Description

not sure what to do about neo_round_limit

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #706

### DIFF
--- a/mp/src/game/shared/neo/neo_gamerules.cpp
+++ b/mp/src/game/shared/neo/neo_gamerules.cpp
@@ -127,9 +127,9 @@ ConVar neo_sv_pausematch_unpauseimmediate("neo_sv_pausematch_unpauseimmediate", 
 
 static void neoSvCompCallback(IConVar* var, const char* pOldValue, float flOldValue)
 {
-	const float flCurrentValue = !flOldValue;
-	neo_sv_readyup_lobby.SetValue(flCurrentValue);
-	mp_forcecamera.SetValue(flCurrentValue); // 0 = OBS_ALLOWS_ALL, 1 = OBS_ALLOW_TEAM. For strictly original neotokyo spectator experience, 2 = OBS_ALLOW_NONE
+	const bool bCurrentValue = !(bool)flOldValue;
+	neo_sv_readyup_lobby.SetValue(bCurrentValue);
+	mp_forcecamera.SetValue(bCurrentValue); // 0 = OBS_ALLOWS_ALL, 1 = OBS_ALLOW_TEAM. For strictly original neotokyo spectator experience, 2 = OBS_ALLOW_NONE
 }
 
 ConVar neo_sv_comp("neo_sv_comp", "0", FCVAR_REPLICATED, "Enables competitive gamerules", true, 0.f, true, 1.f, neoSvCompCallback);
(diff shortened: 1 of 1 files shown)


brysondev (approved):

mp/src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -2719,7 +2728,7 @@ bool CNEORules::RoundIsMatchPoint() const
 	if (teamJinrai && teamNSF && neo_round_limit.GetInt() != 0)
 	{
 		if (RoundIsInSuddenDeath()) return false;
-		const int roundDiff = neo_round_limit.GetInt() - (m_iRoundNumber - 1);
+		const int roundDiff = neo_round_limit.GetInt() - (m_iRoundNumber);
```
nullsystem, line 2731: Why remove the `- 1`? Was that logic wrong?

mp/src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -125,6 +125,15 @@ ConVar neo_sv_readyup_lobby("neo_sv_readyup_lobby", "0", FCVAR_REPLICATED, "If e
 ConVar neo_sv_pausematch_enabled("neo_sv_pausematch_enabled", "0", FCVAR_REPLICATED, "If enabled, players will be able to pause the match mid-game.", true, 0.0f, true, 1.0f);
 ConVar neo_sv_pausematch_unpauseimmediate("neo_sv_pausematch_unpauseimmediate", "0", FCVAR_REPLICATED | FCVAR_CHEAT, "Testing only - If enabled, unpause will be immediate.", true, 0.0f, true, 1.0f);
 
+static void neoSvCompCallback(IConVar* var, const char* pOldValue, float flOldValue)
+{
+	const float flCurrentValue = !flOldValue;
```
nullsystem, line 130: Nit: Cast `flOldValue` to bool and set value on bool instead.

nullsystem (changes requested):

mp/src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -125,6 +125,15 @@ ConVar neo_sv_readyup_lobby("neo_sv_readyup_lobby", "0", FCVAR_REPLICATED, "If e
...
+	neo_sv_readyup_lobby.SetValue(flCurrentValue);
+	mp_forcecamera.SetValue(flCurrentValue); // 0 = OBS_ALLOWS_ALL, 1 = OBS_ALLOW_TEAM. For strictly original neotokyo spectator experience, 2 = OBS_ALLOW_NONE
+}
+
+ConVar neo_sv_comp("neo_sv_comp", "0", FCVAR_REPLICATED, "Enables competitive gamerules", true, 0.f, true, 1.f, neoSvCompCallback);
```
nullsystem, line 135: `neo_round_limit` is set a 15 I think in current OG:NT comp?

nullsystem (changes requested):

mp/src/game/shared/neo/neo_gamerules.cpp
AdamTadeusz: yes, here and theres one other place I think where it shows up but its been a minute

AdamTadeusz: it's a double negative, the -1 in - (m_iRoundNumber - 1) increases the size of roundDiff by 1.
This 1 represents the current round.
Below we are checking what would happen if this 1 current round was won by either team, there represented by the +1 on the left side of the equation.
It doesn't make sense that this current round is both on the left and the right side of the equation, since both teams cannot win the current round, only one team can

mp/src/game/shared/neo/neo_gamerules.cpp
AdamTadeusz: we had a conversation about this and Agiel says this command shouldn't set the neo_round_limit because he might want to introduce half times and other stuff

brysondev (approved):

nullsystem (approved):

PR #856, opened 2025-01-20
Add IO to neo_game_config for custom round events

### LINKED ISSUE #552: [FEATURE] Win trigger logic entity
### Description

Add a way for mappers to trigger a round end and specify which teams win/lose/tie

### Proposed Solution

Being able to activate round-end would enable mappers to create a greater variety of custom maps with unique objectives

### Alternatives Considered

This could be done by adding a logic input to neo_gamerules.cpp, or by creating a new entity for this specific purpose

### Additional Context

_No response_

DESTROYGIRL:
## Description
This adds three inputs and two outputs to neo_game_config to give mappers greater control over the game, potentially to add custom win conditions or special effects.

FireTeamWin - Ends the round with the team specified in the parameter as the winner (2 = Jinrai, 3 = NSF)
FireDMPlayerWin - Ends the round in deathmatch with the activating player as the winner
FireRoundTie - Ends the round as a tie

OnRoundEnd - Fired when any team wins or ties
OnRoundStart - Fired after "- GO! GO! GO! -"

In a scenario where the winning team is ambiguious, the mapper can have an output go through two team filters to then trigger FireTeamWin. I did not check the team of the activator to deterimine the winner instead as this can be more restrictive in some scenarios
I will make an FGD entry for this filter as it is mostly unknown to mappers it appears

## Toolchain

- Windows MSVC VS2022

## Linked Issues

- fixes #552

### DIFF
--- a/mp/src/game/server/neo/neo_game_config.cpp
+++ b/mp/src/game/server/neo/neo_game_config.cpp
@@ -10,6 +10,7 @@ BEGIN_DATADESC(CNEOGameConfig)
 
 	DEFINE_INPUTFUNC(FIELD_INTEGER, "FireTeamWin", InputFireTeamWin),
 	DEFINE_INPUTFUNC(FIELD_VOID, "FireDMPlayerWin", InputFireDMPlayerWin),
+	DEFINE_INPUTFUNC(FIELD_VOID, "FireRoundTie", InputFireRoundTie),
 
 	DEFINE_OUTPUT(m_OnRoundEnd, "OnRoundEnd"),
 	DEFINE_OUTPUT(m_OnRoundStart, "OnRoundStart")
@@ -43,6 +44,11 @@ void CNEOGameConfig::InputFireDMPlayerWin(inputdata_t& inputData)
     {
         static_cast<CNEORules*>(g_pGameRules)->SetWinningDMPlayer(static_cast<CNEO_Player*>(pPlayer));
     }
+}
+
+void CNEOGameConfig::InputFireRoundTie(inputdata_t& inputData)
+{
+	static_cast<CNEORules*>(g_pGameRules)->SetWinningTeam(1, 7, false, true, true, false);
 }
 
 // Outputs
--- a/mp/src/game/server/neo/neo_game_config.cpp
+++ b/mp/src/game/server/neo/neo_game_config.cpp
@@ -27,19 +27,6 @@ void CNEOGameConfig::Spawn()
 
 // Inputs
 
-enum
-{
-	NEO_VICTORY_GHOST_CAPTURE = 0,
-	NEO_VICTORY_VIP_ESCORT,
-	NEO_VICTORY_VIP_ELIMINATION,
-	NEO_VICTORY_TEAM_ELIMINATION,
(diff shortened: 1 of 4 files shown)


DESTROYGIRL:
Also can someone explain to me why for `CNEORules::SetWinningTeam` everything inside the CNEORules class can use the name of the win reason for `iWinReason` but outside of that I strictly have to use an int

AdamTadeusz:
I guess this enum is a property of the CNEORules class, to access it you need an object of this class. CNEORules defines a function NEORules() which returns the global pointer g_pGameRules cast to a CNEORules pointer, you can include neo_gamerules.h wherever you want to query the victory condiftion and do something like
`int victoryCondiftion = NEORules()->NEO_VICTORY_VIP_ESCORT;`

However this is probably not a great solution as we would be doing an unnecessary static_cast every time that code is ran. The other enums like NeoGameType and NeoRoundStatus, are defined in the neo_gamerules.h file too, but they are defined outside the scope of the CNEORules class definition. I guess the victory condition enum was not expected to be used outside the scope of a CNEORules object? I don't know, a refactor seems simple enough and probably should be done since magic numbers. Why some enums have a name and some don't I'm not sure, it might have something to do with the fact that you may accidently define values in multiple separate enums with the same name. By prefixing the value with the name of the enum (so something like NeoRoundStatus::NEO_VICTORY_VIP_ESCORT) it should ensure that you are getting the expected value

mp/src/game/server/neo/neo_game_config.cpp
```diff
@@ -7,4 +7,60 @@ LINK_ENTITY_TO_CLASS(neo_game_config, CNEOGameConfig);
...
+
+	s_pGameRulesToConfig = this;
+}
+
+// Inputs
+
+void CNEOGameConfig::InputFireTeamWin(inputdata_t& inputData)
+{
+	int team = inputData.value.Int();
+	static_cast<CNEORules*>(g_pGameRules)->SetWinningTeam(team, 8, false, true, false, false);
```
AdamTadeusz, line 33: as per the previous comment, i would move the victory condition enum definition outside the CNEORules class and us it here

mp/src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -2868,7 +2880,10 @@ void CNEORules::SetWinningTeam(int team, int iWinReason, bool bForceMapReset, bo
 	UserMessageBegin(filter, "RoundResult");
 	WRITE_STRING(team == TEAM_JINRAI ? "jinrai" : team == TEAM_NSF ? "nsf" : "tie");	// which team won
 	WRITE_FLOAT(gpGlobals->curtime);													// when did they win
-	WRITE_STRING(victoryMsg);															// extra message (who capped or last kill or who got the most points or whatever)
+	if(iWinReason != 8)
```
AdamTadeusz, line 2883: magic number, use enum instead

AdamTadeusz (changes requested):

AdamTadeusz (approved):

brysondev (approved):
Pls remember to squash

Issue #59, opened 2024-05-14
Air control is similar to OGNT


blaberry:

blaberry:
I tried to make this measurable by using some movement commands. Something along the lines of 
1. Stand in open area
2. Walk forward and right
3. Jump
4. Stop walking
And the difference in position could tell whether we're close to parity or not. This is probably overcomplicating the issue and honestly I don't feel too much difference. Setting this as discarded for now.

blaberry:
Oh and here's the aliases and commands I was using.

`alias w1 "wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait;wait"
alias w2 "w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1;w1"
alias w3 "w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2;w2"

+forward;+right;w1;+jump;w1;-jump;w1;-right;-forward`

AdamTadeusz:
if this is discarded should we close?

PR #857, opened 2025-01-23
No ghost bunnyhopping + landing slowdown rework

### LINKED ISSUE #213: Players can't bhop with the ghost
Kino's not gonna love this one.

AdamTadeusz:
## Description

This PR does two things.

1. Disables ghost bunnyhopping by default, following the same methodology as rains anti bunnyhopping plugin except using the current player max speed before slowdown from landing as a non-recon instead of using each classes max wallrun speed as the max.
2. Reworks the deceleration due to landing as a non-recon. After taking some videos in slow motion of an assault character running forward with a zr68s in their hands, jumping and then landing, I have observed that the character's max speed is set to 75 for 0.25 seconds, and then the acceleration is nerfed for a further 1/3 of a second, while the player slowly builds their speed back up to max speed. Given that after the initial 0.25 seconds after landing the player speed dips further below 75 before building back up to max speed, I believe the original game sets the max speed of the player to some value which increases linearly after landing, where a max speed low enough means that after 0.25 seconds this linearly increasing value does not yet pass 75 and so the speed dips for a moment.

This PR also includes the fixes to player speed when forced to duck from PR #846

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #213


brysondev (dismissed):

mp/src/game/server/neo/neo_player.cpp
```diff
@@ -683,30 +678,60 @@ void CNEO_Player::PreThink(void)
 	}
 	if (IsSprinting() && !IsAirborne())
 	{
-		static constexpr float RECON_SPRINT_SPEED_MODIFIER = 0.75;
-		static constexpr float OTHER_CLASSES_SPRINT_SPEED_MODIFIER = 0.6;
-		speed /= m_iNeoClass == NEO_CLASS_RECON ? RECON_SPRINT_SPEED_MODIFIER : OTHER_CLASSES_SPRINT_SPEED_MODIFIER;
+		static constexpr float RECON_SPRINT_SPEED_MODIFIER = 1.35;
```
brysondev, line 681: Isn't this defined in a header somewhere?

mp/src/game/server/neo/neo_player.cpp
```diff
@@ -683,30 +678,60 @@ void CNEO_Player::PreThink(void)
...
+		static constexpr float OTHER_CLASSES_SPRINT_SPEED_MODIFIER = 1.6;
```
brysondev, line 682: Ditto

mp/src/game/client/neo/c_neo_player.cpp
```diff
@@ -816,9 +814,9 @@ void C_NEO_Player::PreThink( void )
 	}
 	if (IsSprinting() && !IsAirborne())
 	{
-		static constexpr float RECON_SPRINT_SPEED_MODIFIER = 0.75;
-		static constexpr float OTHER_CLASSES_SPRINT_SPEED_MODIFIER = 0.6;
-		speed /= m_iNeoClass == NEO_CLASS_RECON ? RECON_SPRINT_SPEED_MODIFIER : OTHER_CLASSES_SPRINT_SPEED_MODIFIER;
+		static constexpr float RECON_SPRINT_SPEED_MODIFIER = 1.35;
```
brysondev, line 817: Ditto

mp/src/game/client/neo/c_neo_player.cpp
```diff
@@ -816,9 +814,9 @@ void C_NEO_Player::PreThink( void )
...
+		static constexpr float OTHER_CLASSES_SPRINT_SPEED_MODIFIER = 1.6;
```
brysondev, line 818: Ditto

brysondev (dismissed):

mp/src/game/server/neo/neo_player.cpp
AdamTadeusz: image
I don't think so? Also I just did this pre-emptively in preperation for your pr to fix these values, if you're defining this in the header somewhere then ill update the pr once i see what that looks like

brysondev: https://github.com/NeotokyoRebuild/neo/blob/0c4eb23085380e72fa66b6fb08f59fdb0ac2cf65/mp/src/game/shared/neo/neo_player_shared.h#L42

AdamTadeusz: technically this is the NEO_SPRINT_MODIFIER which has a value of 1.6, but recons have a different sprint modifier to the other classes

AdamTadeusz: https://github.com/user-attachments/assets/64f7f404-3994-4a75-87c5-bb7c51800992

Here is a recording of the sprint speed of a recon in the original after removing all weapons (hence why the viewmodel is slightly off), the value of 272 is 2 units off and we're currently using the speed with a knife or other very light weapon as the base speed anyway (which despite Agiels protests I don't think is really an issue).

Agiel:
I had a look at the slowdown in OGNT again and it's slightly more convoluted than I remember.
Basically when you jump as a non-recon, your "slowdown factor" is set to `40`.
When grounded this value decays by `timeDelta * 35`.
It is then applied to the max speed with the formula `speed *= MAX(0.01, 1 - slowFactor / 20)`.
Finally, after applying all the other speed multipliers, it does
```
if speed < 55
  speed = 75
```

AdamTadeusz:
> I had a look at the slowdown in OGNT again and it's slightly more convoluted than I remember. Basically when you jump as a non-recon, your "slowdown factor" is set to `40`. This value decays by `timeDelta * 35`. It is then applied to the max speed with the formula `speed *= MAX(0.01, 1 - slowFactor / 20)`. Finally, after applying all the other speed multipliers, it does
> 
> ```
> if speed < 55
>   speed = 75
> ```

So that supports what I've been observing with the player speed in slow motion. I drew a little graph to illustrate it with your code
image

I think this is one of those things where we could improve on the original, I don't think having this hitch where the player speed falls for a bit after player speed exceeds 55 only to start rising again is necessary or feels good

Agiel:
I agree. One of 55 or 75 is probably a typo anyway. I guess making it 75 would be the least noticeable? Unless that would interfere with crouchwalk or something. What's the slowest speed you can reach currently by crouch/walking/aiming? I wouldn't mind making it 55 either, but that'd probably make bhopping harder.

Agiel:
Alternatively, apply the jump slowdown first, and cap it at 75. Then do the walk/crouch/aim modifiers after and cap to whatever feels good.

Agiel:
Right, so we probably shouldn't cap it at 75 then. Best to do the jump slowdown first at 75, and then apply walk/crouch/aim and cap at 55.

AdamTadeusz:
I assume we want the weapon slowdown done before the jump slowdown?

mp/src/game/server/neo/neo_player.cpp
```diff
@@ -667,6 +667,22 @@ extern ConVar neo_ghost_bhopping;
...
+		speed *= pNeoWep->GetSpeedScale();
+	}
+	// Slowdown after landing 
+	if (!IsAirborne() && m_iNeoClass != NEO_CLASS_RECON)
+	{
+		const float timeSinceLanding = gpGlobals->curtime - m_flLastAirborneJumpOkTime;
+		constexpr float SLOWDOWN_TIME = 1.15f;
+		if (timeSinceLanding < SLOWDOWN_TIME)
+		{
+			speed = MAX(75,speed * MAX(0.01, 1 - (((SLOWDOWN_TIME - timeSinceLanding) * 35) * 0.05)));
```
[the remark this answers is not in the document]
AdamTadeusz: maybe we don't need the inner max as well, it shouldn't make a noticeable difference right?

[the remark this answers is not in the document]
AdamTadeusz: you're right, walking off of a ledge in the original does not cause a slowdown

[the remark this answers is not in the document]
Agiel: Oh yeah, that's going to be annoying...

Agiel:
So I figured out #59. The difference is that OGNT applies the jump slowdown even while in the air, which has an effect on the air control. Basically you should be able to just remove the !IsAirborne() check.
We also need to rethink the 75/55 cap. I thought it was fine like this, because it's cleaner and makes more logical sense. But with the airborne check removed it means that crouched bhops have worse aircontrol than in OGNT because your max speed will be 75*0.75 = 56.6 instead of 75. I wish I had a better idea than going back to `if < 55 then 75` but I guess it's what we have to do for parity.

AdamTadeusz:
> So I figured out #59. The difference is that OGNT applies the jump slowdown even while in the air, which has an effect on the air control. Basically you should be able to just remove the !IsAirborne() check. We also need to rethink the 75/55 cap. I thought it was fine like this, because it's cleaner and makes more logical sense. But with the airborne check removed it means that crouched bhops have worse aircontrol than in OGNT because your max speed will be 75*0.75 = 56.25 instead of 75. I wish I had a better idea than going back to `if < 55 then 75` but I guess it's what we have to do for parity.

what if we SetMaxSpeed(MAX(speed, IsAirborne() ? 75 : 55));

Agiel:
An alternative would be to not apply the crouch/walk modifiers while in the air. Idk if there's some tech there that pro recons make use of? I see we're already not applying sprint while in the air, which isn't parity.

Agiel:
> what if we SetMaxSpeed(MAX(speed, IsAirborne() ? 75 : 55));

This would be a nerf to assaults and supports because it makes the slowdown stronger. Remember the cap only drops to 55 for a split second.

Agiel:
Ok after sleeping on it, just moving the jump slowdown to the end seems to work https://github.com/Agiel/neo/commit/fc636994b66eea082c75e257c1c39823ddbc01c2.
Side-note: Why isn't this stuff in neo_player_shared?

AdamTadeusz:
> Ok after sleeping on it, just moving the jump slowdown to the end seems to work [Agiel@fc63699](https://github.com/Agiel/neo/commit/fc636994b66eea082c75e257c1c39823ddbc01c2). Side-note: Why isn't this stuff in neo_player_shared?

I didn't want to refactor this when Rain was already working on what I imagine would have been a more event based speed system

Agiel (approved):

brysondev (approved):

PR #866, opened 2025-01-28
Pre-round hints on HUD

### LINKED ISSUE #863: Missing hints during round freezetime
In OGNT, hints are sometimes displayed to the player before a round begins. These were sprites drawn on the HUD, and can be found in `materials/vgui/hud/hints/`. These are not yet present in rebuild

DESTROYGIRL:
## Description

Hints, hopefully identical to how they appeared in OGNT
Includes a convar to turn them on / off

## Toolchain
- Windows MSVC VS2022

## Linked Issues

- fixes #863

### DIFF
--- a/mp/game/neo/scripts/HudLayout.res
+++ b/mp/game/neo/scripts/HudLayout.res
@@ -676,9 +676,9 @@
 	neo_hint
 	{
 		"fieldName"		"neo_hint"
-		"xpos"	"0"
-		"ypos"	"0"
-		"wide"	"640"
-		"tall"	"480"
+		"xpos"	"466"
+		"ypos"	"59"
+		"wide"	"142"
+		"tall"	"284"
 	}
 }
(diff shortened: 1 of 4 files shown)


brysondev (commented):
I'd like to see this on my local env before approving. Can you send some screenshots to show me what it looks like on your end?

DESTROYGIRL:
> I'd like to see this on my local env before approving. Can you send some screenshots to show me what it looks like on your end?





Vanishes at GOGOGO. Only visible once spawned

AdamTadeusz:
image
image
Neotokyo deals better with different aspect ratios and smaller window sizes

AdamTadeusz (commented):
The ypos isn't quite right on most resolutions but it looks ok. I'm worried what the hints will look like on very wide screens since the height of the hint depends on the width of the hint which depends on the width of the screen, but I have no way of testing that. Perhaps we should add some clamps to ensure the hint never goes off-screen?

AdamTadeusz (approved):

brysondev (approved):
image
lgtm

PR #867, opened 2025-01-31
Correct scoped weapon scope colour

AdamTadeusz:
## Description

Incorrect:
405616081-0a276948-bd51-4657-9e14-0e895fb88da5

Correct:
image

### DIFF
--- a/mp/src/game/client/hud_crosshair.cpp
+++ b/mp/src/game/client/hud_crosshair.cpp
@@ -383,7 +383,7 @@ void CHudCrosshair::Paint( void )
 #ifdef TF_CLIENT_DLL
 			clr
 #else
-			m_clrCrosshair
+			COLOR_WHITE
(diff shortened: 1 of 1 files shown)


AdamTadeusz:
@nullsystem was your intention to be able to customize the colour of the scope same as the crosshair? in that case it should probably be a separate colour selector and the same colour should be passed to the squares drawn around the scope. Just don't let them pass an alpha value or the scope will be see through

kassibuss:
I'd say keep the scope color as is.

nullsystem:
@AdamTadeusz Nah that was accidental, didn't intend for scopes to change color

nullsystem (approved):

brysondev (commented):
Finally someone spelt colour correctly

brysondev (approved):

PR #869, opened 2025-02-04
Correct muzzle flash when server switches weapons

### LINKED ISSUE #571: Weapon view model muzzle flash not updated when calling Weapon_Switch on the server
Making this issue more specific since the animation PR solves most of the problems associated with the earlier issue. 

Essentially when the player first spawns or selects their loadout, the server forces the player to switch to a particular weapon, which doesn't trigger the update to the weapon view model's muzzle flash

AdamTadeusz:
## Description

Ideally the updateViewModelMuzzleFlash function should belong to the CNEOViewModelMuzzleFlash, but when doing it that way changing the muzzle flash skin doesn't seem to have an effect

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #571

### DIFF
--- a/src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.h
+++ b/src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.h
@@ -37,6 +37,7 @@ public:
 		m_flTimeSwitchOffMuzzleFlash = gpGlobals->curtime + 0.01f;
 		BaseClass::ProcessMuzzleFlashEvent();
 	}
+	virtual void			ClientThink() override;
 #endif
 
 	virtual void Spawn(void) override;
--- a/src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.h
+++ b/src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.h
@@ -25,7 +25,9 @@ public:
 		m_iAngleZIncrement = -5;
 		m_flTimeSwitchOffMuzzleFlash = gpGlobals->curtime;
 	}
-
+	void UpdateMuzzleFlashProperties(CBaseCombatWeapon* pWeapon, bool repeat = true);
+	virtual void Spawn(void) override;
+	virtual void Precache(void) override;
 #ifdef CLIENT_DLL
 	virtual ShadowType_t	ShadowCastType() final override { return SHADOWS_NONE; };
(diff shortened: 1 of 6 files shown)


AdamTadeusz:
I should add that I avoided making these variables networked as I was worried they would generate traffic unnecessarily when a weapon is being fired. Having made them networked and observing generated traffic using dtwatchent *entity id of muzzleflash* I observed that these variables weren't being sent over the network, probably something clever decided its not necessary when the same function is ran both client and server side. However the scale of the muzzle flash would not update even though m_flModelScale is also a networked variable that exists both in a send and receive table.

This solution is the first one that I could find that fixed the whole problem in the fewest number of lines, if the reason why m_flModelScale was not being sent could be solved then making all these properties networked instead would be simpler.

AdamTadeusz:
I think ill work on this some more and get one of the two nicer looking methods working

AdamTadeusz:
ok I think we can make this slightly better

brysondev (dismissed):

Agiel (dismissed):

src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.h
```diff
@@ -37,14 +34,21 @@ class CNEOPredictedViewModelMuzzleFlash : public CPredictedViewModel
 		m_flTimeSwitchOffMuzzleFlash = gpGlobals->curtime + 0.01f;
 		BaseClass::ProcessMuzzleFlashEvent();
 	}
+	virtual void			ClientThink() override;
 #endif
 
-	virtual void Spawn(void) override;
-	virtual void Precache(void) override;
+#ifdef CLIENT_DLL
 	bool	m_bActive;
```
brysondev, line 41: Your privates are showing

brysondev (changes requested):

src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.h
AdamTadeusz: its fine

brysondev: lewd

brysondev (approved):

PR #870, opened 2025-02-04
Background map support

AdamTadeusz:
## Description

Hide GUI elements we don't want displayedon background maps

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

### DIFF
--- a/mp/src/game/client/hud_basechat.cpp
+++ b/mp/src/game/client/hud_basechat.cpp
@@ -1802,6 +1802,10 @@ void CBaseHudChat::LevelShutdown( void )
 //-----------------------------------------------------------------------------
 void CBaseHudChat::ChatPrintf( int iPlayerIndex, int iFilter, const char *fmt, ... )
 {
+	if (engine->IsLevelMainMenuBackground())
+	{
+		return;
+	}
 	va_list marker;
 	char msg[4096];
 
--- a/mp/src/game/client/neo/c_neo_player.cpp
+++ b/mp/src/game/client/neo/c_neo_player.cpp
@@ -1289,7 +1289,7 @@ void C_NEO_Player::Spawn( void )
 			}
 		}
 
-		if (GetTeamNumber() == TEAM_UNASSIGNED)
+		if (GetTeamNumber() == TEAM_UNASSIGNED && !engine->IsLevelMainMenuBackground())
 		{
 			engine->ClientCmd(teammenu.GetName());
 		}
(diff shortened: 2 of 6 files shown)


AdamTadeusz:
Still gotta work out whats the best way to hide the chat, and also whether we want to hide the Loading prompt in the bottom right (which we probably do)

DESTROYGIRL:
What disables/enables loading a background map at startup, is it just a case of putting `map_background <mapname>` in your autoexec

AdamTadeusz:
> What disables/enables loading a background map at startup, is it just a case of putting `map_background <mapname>` in your autoexec

see section Disabling Background Maps in https://developer.valvesoftware.com/wiki/Menu_Background_Map
turns out the original game does indeed have this valve.rc file with the exact same contents which disables background maps (or rather any map with the appropriate name defined in chapterbackgrounds.txt) from ever loading. This pr includes our own version of valve.rc which adds a command at the end also grabbed from that website which re-enables that functionality

DESTROYGIRL:
It's a me problem but wondering if there's a way to disable stuff like cl_showpos when there's a background map running. But not hardcoded so it can still be used for debugging.

DESTROYGIRL:
> yes, we could code it in, question is do you want to loose the ability to see your position in the background map
> 
> of course we could code in a separate cl_showpos called cl_showpos_backgroundmap I guess

Also applies to a lot of other commands that show info in that top right corner. Could have one command that enables visibility of all that dev info stuff in background maps specifically. But if they can't all be hidden in one fell swoop it's probably not worth it

AdamTadeusz:
> > yes, we could code it in, question is do you want to loose the ability to see your position in the background map
> > of course we could code in a separate cl_showpos called cl_showpos_backgroundmap I guess
> 
> Also applies to a lot of other commands that show info in that top right corner. Could have one command that enables visibility of all that dev info stuff in background maps specifically. But if they can't all be hidden in one fell swoop it's probably not worth it

I'd wager most if not all of those commands draw to that area via this function g_pMatSystemSurface->DrawColoredText( , unfortunately we can't edit this function so would have to instead change the logic everywhere where this function and others like it (DrawColoredCircle, DrawColoredTextRect) to not actually draw if in a background map unless we enable drawing of these on a background map

DESTROYGIRL (approved):
We can probably solve the issue outside the scope of this PR. Either way it's minor

brysondev (approved):

PR #871, opened 2025-02-05
IFF Crosshair

### LINKED ISSUE #868: Aiming at teammates does not bring up the warning crosshair
### Build Info

20241102_14c4282

### Description

In OGNT, when you aim at teammates, your crosshair will change to a different one which signals that you are aiming at a friendly and you might hit them. In NT;RE that does not happen.

### To Reproduce

1. Aim at your teammate
2. Observe your crosshair

### Expected behavior

Your crosshair should change to a different indicating that you are aiming at a friendly. 

Image

### Actual behavior

Your crosshair doesn't change.

Image

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

I have a slightly modified warning crosshair, but the principle is the same.

AdamTadeusz:
## Description

https://github.com/user-attachments/assets/76020803-4f4e-4261-bbb6-ecdcb21e2868

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #868

### DIFF
--- a/mp/src/game/client/hud_crosshair.cpp
+++ b/mp/src/game/client/hud_crosshair.cpp
@@ -29,6 +29,7 @@
 
 #ifdef NEO
 #include "weapon_neobasecombatweapon.h"
+#include "neo_gamerules.h"
 #endif
 
 // memdbgon must be the last include file in a .cpp file!!!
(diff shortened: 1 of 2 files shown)


DESTROYGIRL (changes requested):
Works, but this doesn't account for the deathmatch gamemode where there are no teams and it's a free for all

DESTROYGIRL (dismissed):

DESTROYGIRL:
Not sure if it's intentional but this will show up on any entity that's the same team as you, not just players. Which is pretty neat
But this does include any weapon you or a teammate has picked up. FYI

AdamTadeusz:
If we controlled what entities are on what team more closely then this would be a cool feature as it would allow us to display this iff crosshair on friendly entities like the IFV on that one map where whichever team controls the ghost controls that IFV, however for now Ill just check that the entity index is a player index

DESTROYGIRL (approved):

Agiel (approved):

PR #879, opened 2025-02-08
Disable screen overlays when in vision modes

DESTROYGIRL:
## Description
Will hide screen overlays (from env_screenoverlay, r_screenoverlay, or any other means) when the player is using their vision mode
Did this because it can mess with brightness in vision modes, especially with NV where the environment is very dark.

## Toolchain

- Windows MSVC VS2022

### DIFF
--- a/mp/src/game/client/viewrender.cpp
+++ b/mp/src/game/client/viewrender.cpp
@@ -77,6 +77,11 @@
 // Projective textures
 #include "C_Env_Projected_Texture.h"
 
+#ifdef NEO
+// For removing screen overlays when in vision modes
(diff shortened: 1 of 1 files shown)


AdamTadeusz:
do you have an example map with a screen overlay where we can see this working?

DESTROYGIRL:
> do you have an example map with a screen overlay where we can see this working?

`r_screenoverlay effects/combine_binocoverlay` Do this command in console and try using a vision mode

AdamTadeusz (approved):

brysondev (approved):

PR #881, opened 2025-02-10
Empty and Story gamemodes, plus game type settings overhaul

### LINKED ISSUE #656: [FEATURE] Empty Gamemode
### Description

A gamemode that removes the round timer and does not let the player(s) win or lose.

### Proposed Solution

This could be utilised for background maps, server hub maps or anything singleplayer. Make this a gamemode option as part of the neo_game_config entity so it can be set by mappers. On player death, let them respawn seconds later much like round warmup in CTG.

### Alternatives Considered

_No response_

### Additional Context

_No response_

AdamTadeusz:
## Description

Also allows map makers to turn off hud elements, allows map makers to enforce team class and weapons. Enforced items cannot be changed by the user and the menus to do so are disabled. In story mode the character spawns only once the item in the last active menu is selected. Also changes the default spawn team to team unassigned, which allows disabling the initial spectator gui easily.

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #656


AdamTadeusz:
@DESTROYGIRL kind of works now, by no means is the settings overhaul exhaustive enough and I would appreciate some feedback on the null and story gamemodes whether they behave like you expected them to

DESTROYGIRL:
Nice I'll look through it. Thanks

mp/src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -187,6 +197,25 @@ static NEOViewVectors g_NEOViewVectors(
...
+};
+
+const NeoGameTypeSettings NEO_GAME_TYPE_SETTINGS[NEO_GAME_TYPE__TOTAL] = {
+//						gametypeName	respawns	neoRulesThink	changeTeamClassLoadoutWhenAlive	comp	capPrevent
+/*NEO_GAME_TYPE_TDM*/	{"TDM",			true,		true,			false,							false,	false},
+/*NEO_GAME_TYPE_CTG*/	{"CTG",			false,		true,			false,							true,	true},
+/*NEO_GAME_TYPE_VIP*/	{"VIP",			false,		true,			false,							true,	true},
+/*NEO_GAME_TYPE_DM*/	{"DM",			true,		true,			false,							false,	false},
+/*NEO_GAME_TYPE_EMT*/	{"NUL",			true,		false,			true,							false,	false},
+/*NEO_GAME_TYPE_TUT*/	{"STR",			false,		false,			false,							false,	false},
```
DESTROYGIRL, line 216: Intentionally left as NUL & STR?

AdamTadeusz: no 😔

AdamTadeusz: wait are those not correct now?

DESTROYGIRL (dismissed):

mp/src/game/server/neo/neo_game_config.cpp
```diff
@@ -7,6 +7,11 @@ LINK_ENTITY_TO_CLASS(neo_game_config, CNEOGameConfig);
 
 BEGIN_DATADESC(CNEOGameConfig)
 	DEFINE_KEYFIELD(m_GameType, FIELD_INTEGER, "GameType"),
+	DEFINE_KEYFIELD(m_HiddenHudElements, FIELD_INTEGER, "HiddenHudElements"),
```
DESTROYGIRL, line 10:
```suggestion
	DEFINE_KEYFIELD(m_HiddenHudElements, FIELD_INTEGER, "spawnflags"),
```
This isn't how it's usually implemented but if we do this we can use the flags menu in Hammer instead of having mappers input a random number into the KV. Tested and it doesn't appear to cause any issues
image

DESTROYGIRL (approved):

AdamTadeusz:
image
ahhh shit

PR #885, opened 2025-02-12
Aim in if holding aim and can aim but not not aiming when using hold to aim

AdamTadeusz:


Agiel:
With this change I'm unable to zoom out with toggle enabled.

Agiel (approved):

brysondev (approved):
Im sober but I struggled to read the title. Approved.

PR #888, opened 2025-02-14
background maps and menu tweaks

AdamTadeusz:
## Description

Relies on neoAssets PR ---->https://github.com/NeotokyoRebuild/neoAssets/pull/23 <-----

By no means a finished menu look, but something I hope will give people a bit more faith in the project. I have hidden the profile and news section from the main menu for now, I plan to re-introduce these later or after the neoui refactor is complete. Also want to add a music player to the main menu (main menu music currently stops playing when loading into any map, including background maps) and buttons for the tutorial and firing range, once those are complete.

I know there were some problems with the title font on linux, this PR removes the linux specific font sizings but introduces y-res values for all the font sizes, which at least on Windows will force the font to the correct size instead of letting the system pick the right font size. If someone on Linux could verify that the main menu looks good on different screen resolutions I would be thankful.

The title font is also of course not final.


Agiel (approved):

brysondev (approved):
lgtm

PR #889, opened 2025-02-14
Fix scaling on higher resolutions for chat and dev console

brysondev:
## Description

Fix scaling on higher resolutions for chat and dev console

## Toolchain

- Windows MSVC VS2022

## Notes
Linux users should perform regression testing.


brysondev:
Unless someone can figure out how the hell to adjust base source engine fonts (dev console, server browser, etc). Those fonts are stuck at their tiny values at higher resolutions. I'm at a loss.

AdamTadeusz (approved):
tbh I thought the chat and console text were fine at 2k resolution this seems kinda weird, but maybe it should be bigger and ill get used to it

Agiel (approved):

PR #890, opened 2025-02-15
Jump height parity

Agiel:
## Description
Makes jump velocities parity, with a lot less convoluted calculations.
(Ignore whitespace changes. I'm getting tired of hitting "save without formatting" all the time.)

The correct view vectors for the classes were grabbed from OGNT. However, OGNT only uses those when accessed via GetPlayerMaxs/GetPlayerMins. It's probably better to use the correct hulls everywhere so I'm keeping it that way, but to get the correct crouch-jump heights we need to subtract the offsets (and that's why I made the support numbers the defaults).

This mismatch might have effects elsewhere in OGNT that we haven't realized, but for now I'm keeping this hack the FinishDuck method.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #
- related #


Agiel:
Assaults jump too high now. Will have another look at the crouch offsets.

AdamTadeusz:
https://github.com/user-attachments/assets/9184de31-b8d0-4425-8ecf-afabd9aea136

Looks like assault and support duck jump boost is a little to strong, recon is fine

Agiel:
I hate doing this but I ended up adding a magic number to reduce crouch-jumps for assaults and supports slightly. Basically binary search to find the smallest value.
Not extremely scientific, but In OGNT I was able to get 2 units higher once out of ~30 jumps. With a fudge factor of 0.025 it seems about the same.

AdamTadeusz:
I wonder if throwing your mouse to the side is a way to increase your jump height ever so slightly, I seem to be able to still make the jump when doing so.

https://github.com/user-attachments/assets/f809ed6a-2218-4c06-867f-2a4ce60e1c2c

Anyway I don't really care about it, if you want to remove the little offset I'm fine with it.

AdamTadeusz (dismissed):

brysondev (approved):

Agiel:
I determined I'd rather not merge the hack. If someone notices it later I'll dig into it again, but inserting a random offset just for the sake of it doesn't feel good.

AdamTadeusz (approved):

brysondev (approved):

PR #896, opened 2025-02-16
Fix incorrect version info string display in main menu

Masterkatze:
## Description
Before:
before
After:
after

## Toolchain
- Windows MSVC VS2022


brysondev (approved):
lgtm

PR #901, opened 2025-02-17
Tutorial - Dummy NPC

DESTROYGIRL:
## Description
This is an NPC for use in the tutorial. It has custom models designed for it, but you can use anything specified in the keyvalue.
It can also be specified for it to hold a weapon.
The NPC itself does totally nothing, as a dummy should. It's primary use is for demonstrative purposes for the player

For testing it is included in the FGD [here](https://github.com/NeotokyoRebuild/neoAssets/blob/f657075f56591654244bf4b351b8278cbc227454/neo/rebuild.fgd)

## Toolchain

- Windows MSVC VS2022

### DIFF
--- a/src/game/client/c_baseplayer.cpp
+++ b/src/game/client/c_baseplayer.cpp
@@ -2081,11 +2081,22 @@ void C_BasePlayer::ThirdPersonSwitch( bool bThirdperson )
 		}
 
 		//Notify weapon.
+#ifndef NEO
 		CBaseCombatWeapon *pWeapon = GetActiveWeapon();
 		if ( pWeapon )
 		{
 			pWeapon->ThirdPersonSwitch( bThirdperson );
 		}
+#else
+		for (int i = 0; i < WeaponCount(); i++)
+		{
+			CBaseCombatWeapon* pWeapon = GetWeapon(i);
+			if (pWeapon)
+			{
+				pWeapon->ThirdPersonSwitch( bThirdperson );
+			}
+		}
+#endif // NEO
 	}
 }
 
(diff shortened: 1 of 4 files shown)


DESTROYGIRL:
This is feature complete but leaving all the base NPC stuff in there when it doesn't actually need it feels wonky. Regardless it won't let me get rid of it, so

AdamTadeusz:
we've recently merged in the spectated player vision modes pr. While its unlikely you will play as a spectator in the tutorial map, other mappers may want to use this dummy outside of the tutorial map (maybe some kind of map set in a half destroyed training facility). I've not tested it yet but I suspect the dummies will not show up in thermals when spectating someone who is currently using thermals. For comparison here is how we are currently getting the player being observer who may or may not be using thermals

image

AdamTadeusz:
of course this doesn't have to work exactly the same. We are not applying the xray glow effect to these dummies, so they perhaps should stay very visible in thermals regardless of whether xray is being used or not

DESTROYGIRL:
Probably just going to leave them on regardless of xray because they'd only be an aesthetic addition in a PVP map. Having them change with the players would be confusing

AdamTadeusz (dismissed):

DESTROYGIRL:
Update. This is not feature complete I forgot I need to make them show up as enemies when holding the ghost

AdamTadeusz (dismissed):

brysondev (approved):

AdamTadeusz:
to be completely honest we might want a third reviewer here because of how much code is from me

DESTROYGIRL:
> to be completely honest we might want a third reviewer here because of how much code is from me

no problem i will review your code in my pr

DESTROYGIRL:
No I didn't

PR #904, opened 2025-02-18
Tutorial - Hints

DESTROYGIRL:
## Description
This is an entity that draws localised text in a box styled after the original hints. Optionally plays a sound when triggered
To use the entity use the ShowMessage input with the localisation key as the parameter with a # before it
In the localised text use \n to start a new line

[neoAssets PR](https://github.com/NeotokyoRebuild/neoAssets/pull/25)

## Toolchain

- Windows MSVC VS2022

### DIFF
--- a/game/neo/resource/ClientScheme.res
+++ b/game/neo/resource/ClientScheme.res
@@ -1862,6 +1862,28 @@ Scheme
 				//"additive"	"1"
 			}
 		}
+		NHudMessage
+		{
+			"1"
+			{
+				"name"		"Arial"
+				"tall"		"10"
+				"weight"	"100"
+				"additive"	"1"
+				"antialias" "1"
+			}
+		}
+		NHudMessageTitle
+		{
+			"1"
+			{
+				"name"		"Arial"
+				"tall"		"7"
+				"weight"	"1000"
+				"additive"	"1"
+				"antialias" "1"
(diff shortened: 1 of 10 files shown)


mp/src/game/client/neo/ui/neo_hud_message.cpp
```diff
@@ -0,0 +1,180 @@
...
+    {
+        return;
+    }
+
+    // Background box
+    surface()->DrawSetColor(Color(0, 0, 0, 150));
+    surface()->DrawFilledRect(0, 0, m_BoxWidth, m_BoxHeight);
+
+    // Background crosses
+    surface()->DrawSetColor(Color(255, 255, 255, 50));
```
AdamTadeusz, line 84: perhaps these colours should be defined somewhere, or we should use some existing colour to stick to a theme

AdamTadeusz (approved):

mp/src/game/client/neo/ui/neo_hud_message.cpp
DESTROYGIRL: Yea maybe, I was just making it look as similar to OG as possible from eye

AdamTadeusz (approved):

brysondev (approved):

PR #933, opened 2025-02-22
Lean speed and angle parity

Agiel:
## Description
Gets leaning as close to parity as possible, while keeping some of the improvements made.
* Lean angle has been decoupled from the view shift amount. It's now a fixed 20 degrees for both first and third person, same as OGNT. These can be tweaked with the _neo_lean_fp_angle_ and _neo_lean_tp_angle_ cvars.
* The lean amount cvars have been renamed slightly so they make more sense. _neo_lean_peek_left_amount_, _neo_lean_peek_right_amount_ which are set to 7.5 and 15 respectively, same as OGNT.
* _neo_lean_fp_lower_eyes_ is also decoupled and determines how many units to lower the screen when leaning. This is 0 in OGNT, but I decided to keep it at 2 because it feels pretty nice, especially if you have view rotation disabled. We should keep an eye on this though, because it might affect gameplay.
* _neo_lean_speed_ now precisely determines how many seconds it takes to reach full lean. In OGNT it's 0.15s, but I increased it to 0.2s to account for the easing.
* An easing function replaces the Lerp hack to make the lean speed consistent.

## Toolchain

- Windows MSVC VS2022

## Linked Issues

- fixes #
- related #


AdamTadeusz:
https://github.com/user-attachments/assets/0c5d9265-2ee0-4f81-b86c-b381266d155b

I've noticed a problem, you gain a bit of forward or backward velocity when you strafe while leaning. I am currently checking out @nullsystem 's tf2 sdk branch, and I've not checked whether this is the case on master either but leaving a note somewhere that I will come back to later to check

Agiel:
Ok that's funny. I guess we shouldn't be using eyeAngles for the lean since it affects your movement angle when looking up or down 😂.

I'll look into it after the TF2 PR is merged.

AdamTadeusz (approved):

nullsystem (approved):

PR #934, opened 2025-02-23
Add labels to release binaries

blaberry:
## Description

This should let me more easily select for dedicated server/client libraries when building for steampipe.

## Toolchain

N/A
## Linked Issues

N/A

### DIFF
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -122,6 +122,7 @@ jobs:
       with:
         name: ${{ env.libraries }}
         path: ${{ env.install_dir }}/${{ env.libraries }}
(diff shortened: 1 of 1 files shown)


Masterkatze:
`Unexpected input(s) 'label', valid inputs are ['name', 'path', 'if-no-files-found', 'retention-days', 'compression-level', 'overwrite', 'include-hidden-files']`
https://github.com/NeotokyoRebuild/neo/actions/runs/13472139207

AdamTadeusz:
if it adds the label then should we ignore the warning?

blaberry:
I looked at the API data and no, it doesn't actually work, I gotta read the docs and try to figure out how we can mark the assets, or I just go off of the asset name, it'll be fine until we change it and then I just have to change the signature it looks for, nbd really

PR #940, opened 2025-02-24
Fixes strafe direction change when leaning

AdamTadeusz:
## Description

I know modifying gamemovement.cpp is kinda eh, but we already do it anyway, and I can't think of any issues that this would cause currently.


Agiel (dismissed):
It isn't pretty, but it works I guess.

PR #947, opened 2025-02-26
MP3 Player

### LINKED ISSUE #987: If sound setup is set to "Headphones" then the main music will not play on startup
### Build Info

20250306_9c0a8f2

### Description

When the sound setup is changed, the main menu music stops playing. If you connect to a server, disconnect, and return to the main menu, the music will start playing again. Furthermore, unless the sound setup is "Auto" on startup, the main menu music will not play on startup when you launch the game.

### To Reproduce

1. **Options -> Audio -> Sound Setup->** _Change the value_ - main menu music should stop playing
2. Connect to a server and then disconnect - main menu music should resume playing again
3. Exit the game and relaunch - music will not play on startup unless the sound setup was "Auto" when you last exited

### Expected behavior

Main menu music should be playing regardless of whether you changed the sound setup or not. The music should automatically start when the game is launched regardless of souind setup.

### Actual behavior

See above.

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

AdamTadeusz:
## Description

Adds an MP3 player, with the music files in the soundtrack folder of the original being available by default.

-fixes #987


AdamTadeusz:
when the keybind cache gets reset both keys should be reset, so probably only need to check the value of one of them and then update all the cached bindings if the first one isn't cached in retrospect

src/game/client/weapon_selection.cpp
```diff
@@ -21,7 +21,7 @@
 #define HISTORY_DRAW_TIME	"5"
 
 ConVar hud_drawhistory_time( "hud_drawhistory_time", HISTORY_DRAW_TIME, 0 );
-ConVar hud_fastswitch( "hud_fastswitch", "0", FCVAR_ARCHIVE | FCVAR_ARCHIVE_XBOX );
+ConVar hud_fastswitch( "hud_fastswitch", "1", FCVAR_ARCHIVE | FCVAR_ARCHIVE_XBOX );
```
brysondev, line 24: Id argue this might wanna be done in its own PR along with enabling console by default. Just so we can find this again

brysondev (dismissed):
Just that one nit. Also might wanna enable console by default as well while we're at it.

brysondev:
Oh also as a question/general feedback: is there any way to double click a song to have it play instead of spamming "next" or "previous" to get the sound you want?

src/game/client/weapon_selection.cpp
AdamTadeusz: This was supposed to be in reply to your other suggestion

brysondev (approved):

brysondev (approved):

brysondev (approved):

AdamTadeusz:
figured out why the frame close button wasn't working, now the player can be once again closed with the default close button

src/game/client/mp3player.cpp
```diff
@@ -6,7 +6,7 @@
 
 #include "cbase.h"
 
-#if 0
+#if 1
```
Rainyan: Nit: I guess this is technically base engine stuff, so since we're overriding that here, would it make more sense to say `#ifdef NEO` instead of hardcoding this condition if `#if 1`(?)

Since traditionally we've used this to mark our changes to the SDK Base stuff. Although I have no issue with this, if you feel directly flipping this preprocessor is the more appropriate choice.

src/game/client/mp3player.cpp
```diff
@@ -296,6 +311,19 @@ void CMP3FileListPage::OnCommand( char const *cmd )
 	}
 }
 
+#ifdef NEO
+class CMP3PlayListListPanel : public ListPanel
+{
+public:
+	CMP3PlayListListPanel(Panel* parent, const char* panelName) : ListPanel(parent, panelName) {}
+
+	void OnMouseDoublePressed(vgui::MouseCode code)
```
Rainyan: Nit: Maybe mark this `virtual void override` for clarity

src/game/client/mp3player.cpp
```diff
@@ -703,11 +768,30 @@ class CMP3SongProgress : public Slider
...
+#ifdef NEO
+		SetFgColor(random->RandomInt(0, 1) ? COLOR_JINRAI : COLOR_NSF);
+#else
 		SetFgColor( TREE_TEXT_COLOR );
+#endif // NEO
 		SetBgColor( pScheme->GetColor( "BorderDark", Color( 0, 0, 0, 255 ) ) );
 	}
+
+#ifdef NEO
+	void OnMousePressed(MouseCode code)
```
Rainyan: Nit: We could have this as `virtual override` for clarity

src/game/client/mp3player.cpp
```diff
@@ -703,11 +768,30 @@ class CMP3SongProgress : public Slider
...
+#endif // NEO
 };
 
+#ifdef NEO
+ConVar cl_neo_radio_shuffle("cl_neo_radio_shuffle", "0", FCVAR_CLIENTDLL| FCVAR_ARCHIVE | FCVAR_DONTRECORD | FCVAR_HIDDEN, "Randomize song order", true, 0, true, 1);
+ConVar cl_neo_radio_mute("cl_neo_radio_mute", "0", FCVAR_CLIENTDLL | FCVAR_ARCHIVE | FCVAR_DONTRECORD | FCVAR_HIDDEN, "Turn down sound volume as far as possible", true, 0, true, 1);
+ConVar cl_neo_radio_game_pause("cl_neo_radio_game_pause", "0", FCVAR_CLIENTDLL | FCVAR_ARCHIVE | FCVAR_DONTRECORD | FCVAR_HIDDEN, "Pause NEO Radio song when loading into a game", true, 0, true, 1);
+ConVar cl_neo_radio_volume("cl_neo_radio_volume", "80", FCVAR_CLIENTDLL | FCVAR_ARCHIVE | FCVAR_DONTRECORD | FCVAR_HIDDEN, "NEO Radio song volume", true, MUTED_VOLUME, true, 100);
+ConVar cl_neo_radio_volume_ingame("cl_neo_radio_volume_ingame", "20", FCVAR_CLIENTDLL | FCVAR_ARCHIVE | FCVAR_DONTRECORD | FCVAR_HIDDEN, "NEO Radio song volume in game", true, MUTED_VOLUME, true, 100);
+ConVar cl_neo_radio_unused_variable("cl_neo_radio_unused_variable", "20", FCVAR_CLIENTDLL | FCVAR_ARCHIVE | FCVAR_DONTRECORD, "cl_neo_radio_unused_variable", true, MUTED_VOLUME, true, 100);
```
Rainyan: Q: What's the `cl_neo_radio_unused_variable` used for?

src/game/client/mp3player.cpp
```diff
@@ -48,7 +55,11 @@ vgui::Panel *GetSDKRootPanel();
 
 #define SOUND_ROOT "sound"
 
+#ifdef NEO
+#define MUTED_VOLUME		3
+#else
 #define MUTED_VOLUME		0.02f
+#endif // NEO
```
Rainyan: Question: What's the reason for muted volume not being 0 here?

src/game/client/mp3player.cpp
```diff
@@ -786,15 +889,37 @@ CMP3Player::CMP3Player( VPANEL parent, char const *panelName ) :
...
+		PlaySong(randomSong);
+	}
+#endif // NEO
 }
 
 CMP3Player::~CMP3Player()
 {
 	if ( m_bDirty )
 	{
+#ifndef NEO // NEO TODO (Adam) This file is just going to grow non stop, work out what we're doing wrong, for now re-make the db on every launch
```
Rainyan, line 919: Q: Is there a bug tracking this? Maybe we'd wanna create one in the tracker if not, so it doesn't get lost/forgotten.

src/game/client/mp3player.cpp
```diff
@@ -1759,11 +2093,13 @@ void CMP3Player::PopulateLists()
 		return;
 	}
 
+#ifndef NEO
```
Rainyan, line 2099: This preprocessor makes `int i` declared at line 2089 unused in our code path, which (at least for MSVC 2022) emits a warning to the compile log. Perhaps we could mark that declaration `[[maybe_unused]]` to silence the warning, and make our compile log less noisy.

edit: or I guess just ifdef guard that var declaration away entirely, alternatively

src/game/client/neo/ui/neo_root.cpp
```diff
@@ -814,6 +820,28 @@ void CNeoRoot::MainLoopRoot(const MainLoopParam param)
 	}
 	NeoUI::EndSection();
 #endif
+	g_uiCtx.dPanel.x = param.wide - 128;
+	g_uiCtx.dPanel.y = param.tall - 48;
+	g_uiCtx.dPanel.wide = 128;
+	g_uiCtx.dPanel.tall = 1;
+	NeoUI::BeginSection();
+	g_uiCtx.eButtonTextStyle = NeoUI::TEXTSTYLE_CENTER;
+	const auto musicPlayerBtn = NeoUI::Button(L"MP3");
```
Rainyan: Nit: Maybe this `NeoUI::Button(L"MP3");` text could instead read something like "Music"? Since the end user will think about accessing music settings, and not about file formats, which could be confusing for some.

And well, it would be nice to have the phrase itself moved in `mp3player_english.txt` so we can get it outside the code, but perhaps that's out of scope for this PR.

Rainyan (changes requested):
Mostly looks good, but I think we should fix the compiler warning being emitted by the unused var.

Also, I noticed this PR changes the default music behaviour such that music always plays, not only in main menu but also in-game. This seems like a pretty big change, since it affects players' ability to hear in-game sound cues like footsteps, etc. Obviously the player could change the default settings, but still I think such a change should go through the design doc process before being implemented. For example, we could fade out the music post-warmup; but that's probably something for another PR, to not get too sidetracked here. Regardless, I would prefer we default to no music playing in-match, unless the user explicitly accesses the music menu to start the playback.

As for the nits, feel free to ignore if you don't think they're worth changing.

src/game/client/mp3player.cpp
AdamTadeusz: that would make more sense yes

src/game/client/mp3player.cpp
AdamTadeusz: 👍

src/game/client/mp3player.cpp
AdamTadeusz: 👍

src/game/client/mp3player.cpp
AdamTadeusz: I don't have a scooby what I was smoking when writing this

src/game/client/mp3player.cpp
AdamTadeusz: setting the volume to 0 pauses music playback, a limitation of the source engine's sound playback system I guess. I could have sworn at a value of 3 (0.299999...7) the music was still playing but now that I'm testing this again that doesn't seem to be the case anymore so might have to bump it up to 4?

src/game/client/mp3player.cpp
AdamTadeusz: 👍

src/game/client/neo/ui/neo_root.cpp
AdamTadeusz: 👍

src/game/client/mp3player.cpp
AdamTadeusz: done

#1199

src/game/client/neo/ui/neo_root.cpp
```diff
@@ -480,11 +480,13 @@ void CNeoRoot::FireGameEvent(IGameEvent *event)
 	}
 }
 
+#include "mp3player.h";
```
Rainyan: As per the Valve memory debug shenanigans at https://github.com/NeotokyoRebuild/neo/blob/17a3f0c89ba53a171fa99248046185f4488fbadf/src/game/client/neo/ui/neo_root.cpp#L31

I think we should do this include above that line, ideally.

Rainyan (changes requested):

src/game/client/neo/ui/neo_root.cpp
```diff
@@ -23,6 +23,7 @@
 #include "ui/neo_utils.h"
 #include "neo_gamerules.h"
 #include "neo_misc.h"
+#include "mp3player.h";
```
Rainyan: The semicolon at the end of the include is making MSVC unhappy here, could you delete that? Lgtm otherwise

`src\game\client\neo\ui\neo_root.cpp(26): warning C4067: unexpected tokens following preprocessor directive - expected a newline`

Rainyan (changes requested):

AdamTadeusz:
> It bugs me that the music button isn't perfectly in the corner but other than that cool

with the main menu being offset I though it looked weird when the mp3 button was exactly in the corner. I don't expect it to continue looking like this, ideally I wanted something that shows whats being played and the ability to pause while its still mostly hidden, something to think about when replacing this menu with something more consistent with mzync's ui

PR #962, opened 2025-03-02
Fix line endings in new server browser

brysondev:
## Description

Fixes text showing null byte on windows for server browser text elements.

### Before:
image
### After: 
image

### After (linux):
image
Thanks [Ahura](https://discord.com/channels/1235346473827434517/1235346474280161302/1345636721265479754)
## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- related #957


AdamTadeusz (approved):

nullsystem (changes requested):
~~Fix Window's ConvertANSIToUnicode return instead~~
~~Fix Linux's ConvertANSIToUnicode return instead~~
Actually it's worse than we thought

nullsystem (approved):

AdamTadeusz (approved):

PR #964, opened 2025-03-02
Update issue template

brysondev:
## Description

Corrected syntax of the issue/feature templates. Simplified the bug reporting template as most of the fields were unnecessary and can be asked when needed.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,6 +1,6 @@
-name: Bug report
+name: Bug Report
 description: General bugs, program crashes, broken files, wrong values, etc...
-labels: "Bug Report"
+labels: ["Bug Report", "Bug"]
 projects: ["NeotokyoRebuild/1"]
 body:
   - type: markdown
@@ -14,20 +14,19 @@ body:
         description and reproducible steps using the form below.
 
         Make sure the title describes the issue with the current state
-        of the game ie. if the bug is that recons can see through
+        of the game.
+        Examples:
+        - if the bug is that recons can see through
         walls when using their NVGs, the title should
-        be "recons can see through walls with NVGs" and if the
-        bug is that NVGs don't light up dark areas the title should be
-        "recons can't see in the dark with NVGs").
+        be "recons can see through walls with NVGs".
+        - if the bug is that NVGs don't light up dark areas 
+        the title should be "recons can't see in the dark with NVGs".
   - type: input
     id: build
     attributes:
       label: Build Info
       description: |
-        In which build was this bug seen? Replace [DATE] with YYYYMMDD
-        and [GIT HASH] with the Git short hash the build was built on.
-        The main menu contains this in the bottom left corner
-        and the `neo_version` command can also be used to retrieve this
+        In which build was this bug seen? Type `neo_version` command into the console and copy the first line of output labeled "Build version" to retrieve this
         information.
       placeholder: "[DATE]_[GIT HASH] (EX: 20240609_759f6e5)"
     validations:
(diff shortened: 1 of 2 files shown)


AdamTadeusz:
the machine specifics are not required fields, and the operating system probably should stay a required field. I'm not sure if this template should be changed really

brysondev:
> the machine specifics are not required fields, and the operating system probably should stay a required field. I'm not sure if this template should be changed really

https://github.com/NeotokyoRebuild/neo/pull/964/files#diff-637f7b97bba458badb691a1557c3d4648686292e948dbe3e8360564378b653efR77

I left it the way it was. And personally we should discuss this as a team, but I'd still would prefer bug reports to not have so much clutter and just get to the point. [Asking the same question twice shouldn't be apart of the process.](https://github.com/NeotokyoRebuild/neo/pull/964/files#diff-637f7b97bba458badb691a1557c3d4648686292e948dbe3e8360564378b653efL65)

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -88,24 +77,6 @@ body:
       placeholder: "For Windows: EX: 10 or 11; For Linux: EX: Arch, Debian 12, Fedora 40, SteamOS, etc..."
     validations:
       required: true
-  - type: input
```
blaberry, line 91: Let's not delete this, but we don't have to make it required.

blaberry (approved):
LGTM, left one nit

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -62,14 +61,6 @@ body:
...
     validations:
       required: true
-  - type: textarea
-    id: behavior_actual
-    attributes:
-      label: Actual behavior
-      description: |
-        A clear and concise description of what happened instead of the expected behavior.
-    validations:
-      required: true
```
nullsystem, line 72: Not always redundant, just make it non-required

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -88,24 +77,6 @@ body:
...
-    attributes:
-      label: Machine's GPU
-      description: "If applicable, put in the model of the GPU."
-      placeholder: "EX: AMD RX 580, NVIDIA RTX 4070, Intel iGPU, AMD APU, etc..."
-  - type: input
-    id: gpu_driver
-    attributes:
-      label: GPU's driver
-      description: "If applicable, put in the driver for the GPU."
-      placeholder: "EX: NVIDIA 555.99, AMD Adrenalin 24.5.1, Mesa, etc..."
```
nullsystem, line 108: It wasnt even required in the first place, dont delete it

nullsystem (changes requested):

.github/ISSUE_TEMPLATE/bug_report.yml
brysondev: :ebussi:

/gen tho like when has this served a purpose.

nullsystem: Whenever it could be a driver/OS/hardware issue

.github/ISSUE_TEMPLATE/bug_report.yml
brysondev: Description =/= Actual behavior? 

But I agree ill readd then

nullsystem (approved):

kassibuss:
Maybe we should allow to just write in the version in which the bug occurred at, unless its more useful to have the hash.

brysondev:
> Maybe we should allow to just write in the version in which the bug occurred at, unless its more useful to have the hash.

That's what the hash provides? It's the tag hash and the date of release...?

kassibuss:
> > Maybe we should allow to just write in the version in which the bug occurred at, unless its more useful to have the hash.
> 
> That's what the hash provides? It's the tag hash and the date of release...?

lack of thinking moment

brysondev:
> > > Maybe we should allow to just write in the version in which the bug occurred at, unless its more useful to have the hash.
> > 
> > 
> > That's what the hash provides? It's the tag hash and the date of release...?
> 
> lack of thinking moment

I do it a lot but yes please click the funny approve button

kassibuss (approved):

PR #968, opened 2025-03-02
Various HUD tweaks.

### LINKED ISSUE #944: Enabling OGNT style hud should toggle avatars but keep round info
Mockup below, currently when you enable the old hud, it does not show the round info, so you are at somewhat of a disadvantage when using the old hud, since you do not see how many players are alive or how many rounds does X team have, you only see info of your teammates and the round timer.

Image

Agiel:
## Description

* _squad_hud_original_ is enabled by default.
* Display round info even when _squad_hud_original_ is enabled.
* Add _squad_hud_original_ to the settings menu as 'Classic squad list'.
* Show non-squad members in smaller text below squad members.
* Change _squad_hud_ cvar prefixes from _neo_cl_ to _cl_neo_.
* Add smaller and large HudOCR font used by the round info and classic squad list.
* Make rounded corners match OGNT, and configurable in HudLayout.res.
* Change 'AUX' to 'AUX POWER'.

image
image
image

## Toolchain

- Windows MSVC VS2022

## Linked Issues

- fixes #944

### DIFF
--- a/game/neo/resource/ClientScheme.res
+++ b/game/neo/resource/ClientScheme.res
@@ -708,7 +708,7 @@ Scheme
 			}
 			// It seems that 127 is the largest a font can be, so we can't make the title any bigger than this
 		}
-		
+
 		ClientTitleFontSmall
 		{
 			"1"
@@ -1111,7 +1111,7 @@ Scheme
 				"antialias" "1"
 				"custom" "1" [$OSX]
 			}
-		}	
+		}
 		NHudText
 		{
 			"1"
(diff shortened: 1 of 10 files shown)


brysondev (approved):
image
image

4K:
image
image

Id say the 4K scaling is not really there, but that's more of an overall issue that can be addressed later.

brysondev (approved):

Agiel:
> Id say the 4K scaling is not really there, but that's more of an overall issue that can be addressed later.

Yeah I don't have a 4k monitor to test so I kinda just followed the pattern for the other fonts. But I guess we'll look into proper HUD scaling in a separate PR.

brysondev (approved):

AdamTadeusz:
If we're changing the fonts I would also make the UI use non-additive versions of the fonts, should make it a bit clearer on bright backgrounds
image

image

AdamTadeusz (approved):

Agiel:
I created #970 for looking into non-additive fonts and tweaking the background.

PR #973, opened 2025-03-06
Set detpack parent when settled

AdamTadeusz:
## Description

Fixes https://discord.com/channels/1235346473827434517/1235349179296120946/1347227822783074406


DESTROYGIRL:
I imagine if one lands on a weapon and someone picks up that weapon they will have a det glued to them. This will work fine for now but I think in the future we should totally be looking at vphysics projectiles rather than ye-olde-qphysics ones. Post parity, of course...

DESTROYGIRL (dismissed):

AdamTadeusz:
hmm I wonder if you can attach a detpack to the ghost this way 🤔

DESTROYGIRL (approved):

brysondev (approved):

PR #1003, opened 2025-03-09
Recon super jump parity

### LINKED ISSUE #903: Recon long jump parity
### Build Info

v11.0-prealpha

### Description

The recon boost jump currently has a few differences from OGNT. Some minor and some that could have major consequences to recon meta.

### To Reproduce

Hold sprint and jump with > 45 aux while pressing different combinations of WASD.

### Expected behavior

Here's a description of the OGNT behavior:

The boost jump is _always_ initiated when sprinting and you have enough aux. However, this doesn't mean you can boost sideways. The added velocity is calculated as _forwardView * forwardMove_, where _forwardMove_ is your current _max speed_ (229.5 with the fastest guns) when holding W/S, _max speed / sqrt(2)_ if you're also holding A/D, and _0_ if you're only holding A/D.

The fact that it drains aux without giving a boost has an effect on bhopping (where you're usually only holding A/D), because in return for extra air control you use up aux, unless you're really skilled and release sprint before each jump.

### Actual behavior

In NT;RE you currently only boost jump when holding W/S and the velocity added is a flat 250. This means you get no loss in boost when moving diagonally, and when bhopping you can hold sprint without having to worry about draining aux at all.

To allow aux draining for parkour a special case is added where if you hold both W and S you use up aux without getting a boost.

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

AdamTadeusz:
I'm a little skeptical about the sideways nerf, maybe in the original its more something like with how we're applying the backwards movement penalty, where the penalty is applied only to the backwards component of the movement vector, so moving back and to the right is not as slow as moving straight backwards? Anyway this follows what Agiel says in #903 and having jumped around for a bit I think it feels the same as the original, will need some recon gurus to double check

- fixes #903


Agiel (changes requested):
OGNT handles this in GameMovement::CheckJumpButton, where it has access to mv->m_flForwardMove and just adds that to the current velocity. I don't think code like this belongs in PreThink and it's time to stop duplicating code between client and server.

AdamTadeusz:
> OGNT handles this in GameMovement::CheckJumpButton, where it has access to mv->m_flForwardMove and just adds that to the current velocity. I don't think code like this belongs in PreThink and it's time to stop duplicating code between client and server.

I put it here because it was already here, refactoring is a separate issue, one I might be inclined to undertake

blaberry (approved):
Let's open a separate code-cleanup issue for deduping and merge this? I'm sure there's more code needing to be cleaned up.

Rainyan (approved):

Masterkatze:
@AdamTadeusz

DESTROYGIRL:
Press the finish parity button

AdamTadeusz:
@Agiel

PR #1013, opened 2025-03-12
Viewmodel only lean move gun out of face

### LINKED ISSUE #961: Viewmodel only lean visibility issues
When leaning to the right, on some weapons the viewmodel is in the way of the crosshair

Image

One might expect the problematic leaning direction to be inverted when using cl_righthand 0, but there is another problem,

When using cl_rigthand 0 and viewmodel only lean, the viewmodel rotation is not inverted and looks weird

Image

Here is what a left lean looks like

AdamTadeusz:
## Description

Inverts lean angle direction when using cl_righthand 0 and adds an offset when leaning in problematic direction

## Linked Issues

- fixes #961

### DIFF
--- a/src/game/shared/neo/neo_predicted_viewmodel.cpp
+++ b/src/game/shared/neo/neo_predicted_viewmodel.cpp
@@ -400,7 +400,7 @@ float CNEOPredictedViewModel::lean(CNEO_Player *player){
 	player->SetViewOffset(viewOffset);
 
 #ifdef CLIENT_DLL
-	viewAng.z = -m_flLeanRatio * neo_lean_fp_angle.GetFloat();;
+	viewAng.z = -m_flLeanRatio * neo_lean_fp_angle.GetFloat();
 	engine->SetViewAngles(viewAng);
 #endif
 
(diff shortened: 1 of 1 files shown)


src/game/shared/neo/neo_predicted_viewmodel.cpp
```diff
@@ -481,21 +482,38 @@ void CNEOPredictedViewModel::CalcViewModelView(CBasePlayer *pOwner,
 				vOffset = (playerAiming) ? data.m_vecVMAimPosOffset : data.m_vecVMPosOffset;
 				angOffset = (playerAiming) ? data.m_angVMAimAngOffset : data.m_angVMAngOffset;
 			}
+
+#ifdef CLIENT_DLL
+			if (cl_neo_lean_viewmodel_only.GetBool())
+			{ // extra viewmodel offset for when gun rotation would obstruct center of the screen. Should probably be done on a per weapon basis instead
+				QAngle angles = pOwner->EyeAngles();
+				float percent = angles.z / neo_lean_fp_angle.GetFloat();
```
Agiel, line 490: Can't you use m_flLeanRatio here instead of calculating a new percent?

AdamTadeusz: It seems to not be a simple drop in, it is stuttering more now than when using the current method, would have to figure out why that's the case

https://github.com/user-attachments/assets/bc2f61b1-5ff2-40d1-abbc-82342a5b26ea

VantoNortim (approved):

brysondev (commented):
Finally, lean

brysondev (approved):

PR #1025, opened 2025-03-16
weapon switching problem

AdamTadeusz:
weapon switching problem


AdamTadeusz:
added to a branch that was already merged, but I don't think this should be a problem right?

brysondev (approved):

nullsystem (approved):

PR #1026, opened 2025-03-17
OnCompetitive

DESTROYGIRL:
## Description
Output for neo_game_config that is fired on map load and every map reset is sv_neo_comp = 1

This is added so that mappers do not need to manage multiple maps to cater for casual & competitive environments. Everything can be toggled via this output

## Toolchain

- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/neo_game_config.cpp
+++ b/src/game/server/neo/neo_game_config.cpp
@@ -18,7 +18,8 @@ BEGIN_DATADESC(CNEOGameConfig)
 	DEFINE_INPUTFUNC(FIELD_VOID, "FireRoundTie", InputFireRoundTie),
 
 	DEFINE_OUTPUT(m_OnRoundEnd, "OnRoundEnd"),
-	DEFINE_OUTPUT(m_OnRoundStart, "OnRoundStart")
+	DEFINE_OUTPUT(m_OnRoundStart, "OnRoundStart"),
+	DEFINE_OUTPUT(m_OnCompetitive, "OnCompetitive")
 END_DATADESC()
 
 CNEOGameConfig* CNEOGameConfig::s_pGameRulesToConfig = nullptr;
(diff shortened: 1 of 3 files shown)


AdamTadeusz:
I wonder if this is perhaps not the best idea, do we want maps to change between competitive and normal play? I'm leaning towards making gameplay more consistent

DESTROYGIRL:
> do we want maps to change between competitive and normal play

This is already the case, see turmuk / tarmac, ballistrade / ballistremade, saitama / redux, etc
All this does is compress it into one file instead of two imo

Agiel:
> > do we want maps to change between competitive and normal play
> 
> This is already the case, see turmuk / tarmac, ballistrade / ballistremade, saitama / redux, etc All this does is compress it into one file instead of two imo

The only reason we're doing this is because they're official maps. In nt;re we would just replace the original. There's no reason to keep both.

DESTROYGIRL:
> > > do we want maps to change between competitive and normal play
> > 
> > 
> > This is already the case, see turmuk / tarmac, ballistrade / ballistremade, saitama / redux, etc All this does is compress it into one file instead of two imo
> 
> The only reason we're doing this is because they're official maps. In nt;re we would just replace the original. There's no reason to keep both.

Yea I don't plan for this to be used in official maps it's just to make an existing system I'm using more trivial

Agiel:
I must be missing something then. Where and why would this be used if we can replace the original map?

DESTROYGIRL:
In my use case, it's on rogue to remove the tank. This couldn't be utilised on map edits where someone is not the owner of the map, this is for the original creators to make tweaks that cater to both casual and competitive play while only having to work on a single map.

AdamTadeusz:
> In my use case, it's on rogue to remove the tank. This couldn't be utilised on map edits where someone is not the owner of the map, this is for the original creators to make tweaks that cater to both casual and competitive play while only having to work on a single map.

No fun allowed in competitive :(

DESTROYGIRL:
> No fun allowed in competitive :(

Hey, if it was up to me yomi would be in the WW map pool

Masterkatze (approved):

brysondev (approved):

PR #1029, opened 2025-03-17
[Regression Fix] Active weapons falling through floor on death

### LINKED ISSUE #1028: Primary weapons dropped on death don't collide with environment
Regression caused by #1006

AdamTadeusz:
## Description

Move ragdoll creation from hl2mp_player::Event_Killed to neo_player::Event_Killed, move call to baseclass Event_Killed to after our custom weapon dropping code is ran, back to where it was originally

## Linked Issues

- fixes #1028


brysondev (approved):
Praying A posing doesn't somehow come back again

AdamTadeusz:
I don't know how much you guys played but it very well could be the case that a-posing still wasn't fixed

brysondev:
> I don't know how much you guys played but it very well could be the case that a-posing still wasn't fixed

All I can say is, I never noticed it once. Before, and after your fix. So I genuinely don't get it. And even after this latest play-test nobody was seeing any A posing.

PR #1032, opened 2025-03-19
original number of smoke layers

### LINKED ISSUE #1030: Maximum of 2 smokes can be active at a time
Only two smokes can be active at the same time. Attempting to deploy a third will not show any particles
The env_smokestack entity contributes to a theorised particle limit that may be being hit, causing the issue

### LINKED ISSUE #1031: Smoke overlay too aggressive
Smoke overlay too strong when near a smoke grenade

AdamTadeusz:
## Description

New Assets in -------> https://github.com/NeotokyoRebuild/neoAssets/pull/32 <-------

Reduces the opacity of the smoke overlay when walking near a smokegrenade smoke, and reduced the number of particles used to render the smoke grenade effect, allowing for more smoke grenades to be active at the same time by 2.5x.

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #1030
- fixes #1031


DESTROYGIRL:
Not raising the max particles? Without doing so I think there will still be issues on pissalley.

AdamTadeusz:
> Not raising the max particles? Without doing so I think there will still be issues on pissalley.

I don't want the discussion on whether we should or shouldn't and by how much to delay fixing the smokes

PR #1064, opened 2025-03-28
Ready up - Add countdown stage before the match starts

### LINKED ISSUE #1022: Comp mode does not have a countdown after all players have readied up

nullsystem:
## Description
* A round status that goes between when all teams are fully ready and before the first round goes live
* Also beep sound (clientside) during countdown

## Toolchain

## Linked Issues

* fixes #1022

### DIFF
--- a/src/game/client/neo/ui/neo_hud_round_state.cpp
+++ b/src/game/client/neo/ui/neo_hud_round_state.cpp
@@ -6,6 +6,7 @@
 
 #include "iclientmode.h"
 #include "ienginevgui.h"
+#include "engine/IEngineSound.h"
 
 #include <vgui/ILocalize.h>
 #include <vgui/ISurface.h>
@@ -33,6 +34,8 @@ ConVar cl_neo_squad_hud_original("cl_neo_squad_hud_original", "1", FCVAR_ARCHIVE
 ConVar cl_neo_squad_hud_star_scale("cl_neo_squad_hud_star_scale", "0", FCVAR_ARCHIVE, "Scaling to apply from 1080p, 0 disables scaling");
 extern ConVar sv_neo_dm_win_xp;
 extern ConVar cl_neo_streamermode;
+extern ConVar snd_victory_volume;
+extern ConVar sv_neo_readyup_countdown;
 
 namespace {
 constexpr int Y_POS = 0;
@@ -253,7 +256,7 @@ void CNEOHud_RoundState::UpdateStateForNeoHudElementDraw()
 	}
 	else if (NEORules()->GetRoundStatus() == NeoRoundStatus::Countdown)
 	{
-		m_iWszRoundUCSize = V_swprintf_safe(m_wszRoundUnicode, L"COUNTDOWN");
+		m_iWszRoundUCSize = V_swprintf_safe(m_wszRoundUnicode, L"STARTING");
 	}
 	else
 	{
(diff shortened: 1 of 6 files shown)


brysondev (dismissed):

src/game/client/neo/ui/neo_hud_round_state.cpp
```diff
@@ -242,6 +251,10 @@ void CNEOHud_RoundState::UpdateStateForNeoHudElementDraw()
 		m_iWszRoundUCSize = V_swprintf_safe(m_wszRoundUnicode, L"PAUSED");
 		roundTimeLeft = NEORules()->m_flPauseEnd.Get() - gpGlobals->curtime;
 	}
+	else if (NEORules()->GetRoundStatus() == NeoRoundStatus::Countdown)
+	{
+		m_iWszRoundUCSize = V_swprintf_safe(m_wszRoundUnicode, L"COUNTDOWN");
```
AdamTadeusz, line 256: Maybe "STARTING"? Or even "GAME.INIT" or something really cyberpunky

AdamTadeusz:
I think having a beep at the end as the first round starts would be good too, maybe even higher pitched and louder than the normal beeps.

https://github.com/user-attachments/assets/fb7370a8-ffdf-499b-874c-0bccf499e288

Also I understand the beeps are currently delayed because they are being sent by the server because the neorules object doesn't actually do any thinking client side, I do think it would be better if the beep occurred the moment the value on the timer changed. Since we're not printing a countdown to the chat like in the og plugin, relying instead on the timer in the round state hud element, perhaps it wouldn't be a bad idea to tie this beeping noise to this timer literally in code and have the beeps be generated client side. We have access to the neorules object client side and should still be able to check the round status and round remaining time there

AdamTadeusz:
Also there's something really weird going on with sv_neo_comp 1. Theres a bunch of errors related to steam networking for bots when leaving a local server and I also get kicked out of my own local server very often due to some steam authentication error. I got kicked like 5 times when I was trying to record this video

image
image

when disconnecting:
image

AdamTadeusz:
tbf we should probably use a separate sound file for the beeps, and if we do have a final beep then it should be another separate sound file, that way people can very easily edit what those noises sound like without breaking the tutorial maps

AdamTadeusz:
> Also there's something really weird going on with sv_neo_comp 1. Theres a bunch of errors related to steam networking for bots when leaving a local server and I also get kicked out of my own local server very often due to some steam authentication error. I got kicked like 5 times when I was trying to record this video
> 
> image image
> 
> when disconnecting: image

this is actually a problem on master too

AdamTadeusz (dismissed):
In the end I guess these are all questions of style, the feature as it currently is works

nullsystem:
@brysondev @AdamTadeusz I've changed the beeps to be client side with higher pitch in the final beep.

brysondev (approved):

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -2071,37 +2079,49 @@ void CNEORules::StartNextRound()
...
-		bool loopbackSkipWarmup = false;
-		if (!sv_neo_loopback_warmup_round.GetBool())
+		if (bLobby)
 		{
-			for (int i = 1; i <= gpGlobals->maxClients; i++)
+			SetRoundStatus(NeoRoundStatus::Countdown);
+			m_iRoundNumber = 0;
+			m_flNeoRoundStartTime = gpGlobals->curtime;
+			m_flNeoNextRoundStartTime = gpGlobals->curtime + sv_neo_readyup_countdown.GetFloat();
+			UTIL_CenterPrintAll("- ALL PLAYERS NOW READY: MATCH STARTING SOON... -\n");
```
AdamTadeusz, line 2088: I would drop the "now" and just say "ALL PLAYERS READY: MATCH STARTING SOON"

AdamTadeusz (approved):

PR #1073, opened 2025-04-09
Rain improvements

DESTROYGIRL:
## Description
Rain now creates splash particles when hitting a surface, and reacts to env_wind rather than two commands that could not be utilised outside of cheats.

Rain splash code provided by [Tony Sergi](https://developer.valvesoftware.com/wiki/User:Omega)

neoAssets >>> https://github.com/NeotokyoRebuild/neoAssets/pull/39

## Toolchain

- Windows MSVC VS2022

### DIFF
--- a/src/game/client/c_effects.cpp
+++ b/src/game/client/c_effects.cpp
@@ -26,14 +26,20 @@
 // memdbgon must be the last include file in a .cpp file!!!
 #include "tier0/memdbgon.h"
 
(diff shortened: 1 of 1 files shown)


brysondev (approved):

AdamTadeusz:
increases the number of allocated particles only by about 70 on ntre_rise_ctg, I think thats very much an acceptable value for this kind of visual upgrade

AdamTadeusz (approved):

PR #1075, opened 2025-04-10
add shader detail very high option

### LINKED ISSUE #1074: Missing Shader detail option
the legacy video options page has a "Very High" option for the shader quality setting, which I believe should, among other things, enable r_lightmap_bicubic

AdamTadeusz:
## Description

See issue for description

## Linked Issues

- fixes #1074

### DIFF
--- a/src/game/client/neo/ui/neo_root_settings.cpp
+++ b/src/game/client/neo/ui/neo_root_settings.cpp
@@ -348,7 +348,15 @@ void NeoSettingsRestore(NeoSettings *ns, const NeoSettings::Keys::Flags flagsKey
 		pVideo->iCoreRendering = (queueMode == -1 || queueMode == 2) ? THREAD_MULTI : THREAD_SINGLE;
 		pVideo->iModelDetail = 2 - cvr->r_rootlod.GetInt(); // Inverse, highest = 0, lowest = 2
 		pVideo->iTextureDetail = 3 - (cvr->mat_picmip.GetInt() + 1); // Inverse+1, highest = -1, lowest = 2
-		pVideo->iShaderDetail = 1 - cvr->mat_reducefillrate.GetInt(); // Inverse, 1 = low, 0 = high
+		// Shader detail
+		//					mat_reducefillrate			r_lightmap_bicubic
+		// Low:						1							0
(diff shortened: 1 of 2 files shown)


AdamTadeusz:
@nullsystem following this convention for water detail and shadow detail, we should set the label for shader detail to very high if r_lightmap_bicubic is set to 1, without then checking what mat_reducefillrate is set to. Wouldn't it be a good idea however to somehow indicate to the player that, if for example both r_lightmap_bicubic and mat_reducefillrate is set to 1 that even though the value is set to "Very High", that the graphical settings aren't at their highest?

AdamTadeusz:
I guess thats not really a big issue, updating a single setting should update all of these convars that could have a conflict to the values they are supposed to have

src/game/client/neo/ui/neo_root_settings.cpp
```diff
@@ -568,7 +576,8 @@ void NeoSettingsSave(const NeoSettings *ns)
 		cvr->mat_queue_mode.SetValue((pVideo->iCoreRendering == THREAD_MULTI) ? 2 : 0);
 		cvr->r_rootlod.SetValue(2 - pVideo->iModelDetail);
 		cvr->mat_picmip.SetValue(2 - pVideo->iTextureDetail);
-		cvr->mat_reducefillrate.SetValue(1 - pVideo->iShaderDetail);
+		cvr->mat_reducefillrate.SetValue(pVideo->iShaderDetail == QUALITY_LOW);
+		cvr->r_lightmap_bicubic.SetValue(pVideo->iShaderDetail == 2); // No Medium option, label will say "Very High"
```
nullsystem, line 580: Can you instead define an enum that is `QUALITY2_LOW = 0, QUALITY2_HIGH = 1, QUALITY2_VERYHIGH = 2` rather than on comments

nullsystem (changes requested):

nullsystem:
@AdamTadeusz I wouldn't bother with that specific case, it's basically an "odd" setting if it has it likely if the player somewhat tweaked both of those settings. Just let it do "very high".

Masterkatze (approved):

PR #1080, opened 2025-04-11
Props Penetrateable again

### LINKED ISSUE #1056: Props are non-penetrable
Bullets do not appear to pass through any props whatsoever.

AdamTadeusz:
## Description

A while back I opened a PR which changed how bullet penetration worked to what I had in mind originally, which is that when hitting an object, a traceline would be made starting inside the object to find a good place to then do a backwards trace from, instead of from a set point 12 hammer units past the bullet entrance. This would theoretically ensure that any really thin walls of a few hammer units in thickness did not get skipped during the bullets trajectory. This worked great for brushes where making a call to UTIL_traceline with the starting position slightly inside of the brush would move the starting position of the traceline along the path it would take until the end of the brush was found. However this does not work for props and hitboxes (and perhaps other objects), where starting a traceline inside of these would result in the traceline immediately hitting the object it started inside of and the fraction representing the distance traveled by the traceline returning 0.0f meaning it never moved.

This pr checks the value of this fraction for this value of 0.0, and reverts the behaviour of how a bullet exit is found to the original functionality, which again was to simply look back from a point 12 hammer units past the bullet entrance. This isn't ideal, since it means that placing a prop right up against a thin wall that is resistant enough to normally prevent bullet penetration (either because of the material bullet resistance or because its a blockbullets material), will cause the bullet to exit out the other side of the wall anyway. I drew a little diagram here to explain the problem better

image

## Linked Issues

- fixes #1056

### DIFF
--- a/src/game/shared/baseentity_shared.cpp
+++ b/src/game/shared/baseentity_shared.cpp
@@ -2213,7 +2213,7 @@ void CBaseEntity::HandleShotPenetration(const FireBulletsInfo_t& info,
 	}
 
 	if (tr.m_pEnt && tr.m_pEnt->IsPlayer())
-	{ // If hit entity is player 1. We don't want them taking damage more than once 2. Tracelines get stuck inside of hitboxes at the start of the traceline unlike other objects where they pass straight through
+	{ // If hit entity is player 1. We don't want them taking damage more than once 2. Tracelines get stuck inside of hitboxes at the start of the traceline, like in props and unlike in brushes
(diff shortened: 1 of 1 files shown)


DESTROYGIRL (approved):
Interesting!
TBH I could only see the issue showing itself when using our blockbulletspenetrate brush around props, since nothing else is typically that thin and non-penetrable

Masterkatze (approved):

PR #1081, opened 2025-04-11
Fix broken unicode text in killfeed

Masterkatze:
## Description
Before:
before
After:
after

## Toolchain
- Linux GCC Native [Arch, g++ (GCC) 14.2.1 20250207]


nullsystem (approved):
Works when paired with the fixed fonts/clientscheme.res

brysondev (approved):

PR #1082, opened 2025-04-14
More contrast in thermals in dark places, hot things in thermals

### LINKED ISSUE #1051: Corpses do not fade out in thermal vision
Corpses should remain visible for a short time after death in thermal vision before fading out

AdamTadeusz:
## Description

[Changes in NeoAssets](https://github.com/NeotokyoRebuild/neoAssets/pull/41)

[Discord discussion](https://discord.com/channels/124171198245502976/1361036657041866772)

[Change proposal document](https://docs.google.com/document/d/1Yjt-7CUdDtHxyDU7QH93i0EaKVL5dBRy7vAMGnK0qSM/edit?usp=sharing)

Improves the contrast of thermals in darker areas by increasing the base brightness multiplier of thermals so darker colours match the original more closely, and then adding a brightness nerf threshold above which the brightness is toned down so that brighter areas still match the original more closely.

Gave trailparticles and simpleparticles a boolean which, when set to true, will draw those particles as white when using thermals so that they show up in thermals more. This is a novel feature and has some implications on gameplay, from adding more visual clutter to the screen when using thermals, to also improving a players situational awareness by making it easier to see where bullets are landing, highlighting environmental hazards like fire.

Player ragdolls, player gibs, grenade projectiles and weapons now also show up in thermals (the last already did when it was the active weapon of a player), ragdolls gibs and grenade projectiles cool down over time, weapon temperature falls when without an owner (on the floor), and grows to a threshold when carried by a player, increasing even further when fired. This is visible both on the weapon model, and on the world model (in the original this is only visible on the view model). The rate at which objects cool down is fixed for all entities that do so, set at 5 seconds per unit which closely resembles  the rate at which corpses cool down in the original (though I only eyeballed this), a weapon that is as hot as possible due to continuous use would cool down fully in 7.5 seconds if then immediately dropped.

(for devs) the temperature field counterintuitively is hotter the lower its value, as this was originally just the lifetime of a player corpse, with a player corpse being colder the longer its lifetime. I may change this if it becomes a problem to understand.

Some notes about the shaders:
1. I tried the srgb conversion in the thermalvision shader, I don't know if I used it incorrectly but it gave me wildly different results instead of nudging the colour slightly, I think the colours look fine now with the tweaks without the srgb conversion
2. There is a bunch of changes to the include files for all of the custom neo shaders that I've added to the new buildneoshaders bat file. I didn't make these changes by hand, thus I'm assuming this was done automagically by misyltoads new shader building process? I didn't look into it but everything still works? I should probably see if either misyltoad or anyone else mentioned this in the 2013 source sdk github but those files were mostly copy and paste anyway I don't think we added any custom stuff there so again probably safe to ignore
3. I also added neo shaders that I wasn't working on in this PR, which I assume as part of some changes to the shader build workflow means those also generated some files that we didn't have / or had but not in the right spot.

## Linked Issues

- fixes #1051


AdamTadeusz:
I think people mostly either are happy with the change or want to see it in action, so I'll mark this as ready for review

src/materialsystem/stdshaders/buildneoshaders.bat
brysondev, line 1: Erm, linux?

brysondev (approved):

src/materialsystem/stdshaders/buildneoshaders.bat
AdamTadeusz: I think shader authoring is only available on windows atm

brysondev: Can't spell linux without L

src/game/shared/sdk/sdk_basegrenade_projectile.cpp
```diff
@@ -84,6 +90,23 @@ END_NETWORK_TABLE()
 				}
 			}
 		}
+#endif // NEO
+#ifdef NEO
+#ifdef GLOWS_ENABLE
+		auto pTargetPlayer = glow_outline_effect_enable.GetBool() ? C_NEO_Player::GetLocalNEOPlayer() : C_NEO_Player::GetTargetNEOPlayer();
```
DESTROYGIRL, line 96: This will fail to compile on master cuz `GetTargetNEOPlayer` was changed to `GetVisionTargetNEOPlayer` in #1104 I think

DESTROYGIRL (changes requested):

DESTROYGIRL:
Dummy corpses and server-side corpses also do not have the fade out effect

DESTROYGIRL:
> how do I into server side corpse

I'm pretty sure these are just `prop_ragdoll`s and are created via `CBaseCombatCharacter::BecomeRagdoll`

AdamTadeusz:
@DESTROYGIRL I think dummy ragdolls are server ragdolls, so this should hopefully have fixed all your issues. Also I had a commit in neoassets that I forgot to push which fixed the name of a material to the one thats used here

AdamTadeusz:
interestingly enough bryson's review wasn't dismissed and the commit is showing up as being committed 2 months ago, probably because the commit was created 2 months ago and only now pushed? Seems like a serious vulnerability if I can create a commit before putting a pr up for review and then wait until I get 2 approvals before quickly pushing the local commit. Or maybe it just looks like that because brysons approvals aren't considered anymore

DESTROYGIRL:
> @DESTROYGIRL I think dummy ragdolls are server ragdolls, so this should hopefully have fixed all your issues. Also I had a commit in neoassets that I forgot to push which fixed the name of a material to the one thats used here

You can test the server sided ragdolls if you get threadplate from https://github.com/NeotokyoRebuild/neoAssets/pull/50 and suicide over the elevator
I don't think they're the same thing, but I'll test it out soon

DESTROYGIRL (changes requested):
Dummy ragdolls worked but not server ragdolls. I got them to show by just copying what you did with `C_ClientRagdoll::DrawModel` and pasting it into `C_ServerRagdoll` in ragdoll.cpp

Edit: this will probably make every newly created ragdoll show in thermals though, so every ragdoll placed in the map or created via a template. They are used very rarely in nt maps as a whole so probably not an issue if they fade out before people spawn

DESTROYGIRL (approved):

src/game/client/hl2mp/c_hl2mp_player.h
```diff
@@ -214,7 +214,11 @@ class C_HL2MPRagdoll : public C_BaseAnimatingOverlay
 	virtual void SetupWeights( const matrix3x4_t *pBoneToWorld, int nFlexWeightCount, float *pFlexWeights, float *pFlexDelayedWeights );
 #ifdef NEO
 	inline const Vector& GetRagdollVelocity(void) const { return m_vecRagdollVelocity.Get(); }
-#endif
+#ifdef CLIENT_DLL
+	virtual int DrawModel(int flags) override;
+#endif // CLIENT_DLL
+	float m_flNeoCreateTime; // don't know what m_flCreateTime is used for, better to be safe I guess
```
blaberry, line 220: Seems like it's used for the lifetime calculation? in proxyplayer.cpp:
const float lifeTime = gpGlobals->curtime - pEntity->m_flNeoCreateTime;

src/materialsystem/stdshaders/buildneoshaders.bat
Masterkatze: Shader compiling on Linux? Someday.

src/game/client/hl2mp/c_hl2mp_player.h
AdamTadeusz: Yes I added that myself, I wanted to use m_flCreateTime but I didn't see it used anywhere outside the replay browser, wasn't sure if there was some in engine stuff that I might break

PR #1089, opened 2025-04-18
Fix NeoUI spray issues

nullsystem:
## Description
* Fix spray files not properly created in Windows
* Omit invalid spray files from shown in the picker and deleter
* Fix image in confirm dialog not showing
* Add flag to NeoUI::Texture
* Replace currently used spray with default lambda if it matches the deleting spray

## Toolchain

- Windows MSVC VS2022

### DIFF
--- a/src/game/client/neo/ui/neo_root.cpp
+++ b/src/game/client/neo/ui/neo_root.cpp
@@ -1365,8 +1365,8 @@ void CNeoRoot::MainLoopSprayPicker(const MainLoopParam param)
 		vecStaticSprays.Purge();
 		FileFindHandle_t findHdl;
 		for (const char *pszFilename = filesystem->FindFirst("materials/vgui/logos/ui/*.vmt", &findHdl);
-			 pszFilename;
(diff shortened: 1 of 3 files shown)


Masterkatze:
Tested on Linux - works without any issues.
Question: If we delete the currently used spray, should we also pick the default one automatically instead of keeping using it and removing it only from spray library? In my opinion we should, as it feels like a bug right now.

Masterkatze (approved):

nullsystem:
> Tested on Linux - works without any issues.
> Question: If we delete the currently used spray, should we also pick the default one automatically instead of keeping using it and removing it only from spray library? In my opinion we should, as it feels like a bug right now.

Guess might as well, I suppose with bytes check on the image if they match right before the deletion then override with default HL Lamda spray.

nullsystem:
@Masterkatze I've applied the current spray replacement by matching bytes and sizes

AdamTadeusz (approved):

Masterkatze (approved):
LGTM

PR #1097, opened 2025-04-22
Silent Walking

### LINKED ISSUE #992: [FEATURE] UI element for walking
### Description

Steps are silenced most of the time when ADSing as an assault but this is not consistent. If you don't know whether walk is toggled or not, you might unintentionally make noise and give yourself away. Less experienced supports may not know if they are moving at the correct speed. They cannot cancel walk by sprinting like the other two classes can so it's not as easy to confirm.

### Proposed Solution

Having a UI element that tells you if walking is toggled would be useful - when moving off a neutral position it will signal if you're potentially about to make noise or not. For supports, it would indicate whether they are moving at the intended speed or not.

### Alternatives Considered

_No response_

### Additional Context

_No response_

AdamTadeusz:
## Description

Holding the walk button ensures quiet walking below a raised threshold

Discord discussion: https://discord.com/channels/124171198245502976/1364177734401523814

## Linked Issues

- fixes #992

### DIFF
--- a/src/game/server/hl2/hl2_player.h
+++ b/src/game/server/hl2/hl2_player.h
@@ -345,7 +345,11 @@ private:
 	float				m_fAutoSprintMinTime;	// Minimum time to maintain autosprint regardless of player speed. 
 
 	CNetworkVar( bool, m_fIsSprinting );
+#ifdef NEO
+	CNetworkVar( bool, m_fIsWalking );
+#else
 	CNetworkVarForDerived( bool, m_fIsWalking );
+#endif // NEO
 
 protected:	// Jeep: Portal_Player needs access to this variable to overload PlayerUse for picking up objects through portals
 	bool				m_bPlayUseDenySound;		// Signaled by PlayerUse, but can be unset by HL2 ladder code...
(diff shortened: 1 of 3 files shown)


AdamTadeusz:
I made walking limit the speed at least to what the speed would have to be for the player to walk silently in that stance, this also means that pressing the walk button when already below the noise threshold doesn't have any effect (not sure if there are any situations where that may be the case).

I also added a little something extra. There is a run threshold of 220, any footsteps above this threshold result in louder footsteps. Since the speed of a sprinting assault with a minimal encumbrance weapon such as the srm results in a speed of ~216 (and even lower with higher encumbrance weapons), assaults would be making normal sounding footsteps when running unless turning the mouse or changing direction. In the spirit of making the player controls control the sound the player makes more consistently, I made the prerequisite for making quieter but still audible footsteps both to have a speed below 220 and also to not be sprinting.

brysondev (approved):

DESTROYGIRL (changes requested):
When using some heavy primaries as support, they move so slowly that they don't produce footsteps when walking normally now. Try supa7, PZ

AdamTadeusz:
> When using some heavy primaries as support, they move so slowly that they don't produce footsteps when walking normally now. Try supa7, PZ

I networked m_fIsWalking for this exact purpose and forgot to use it xD

DESTROYGIRL (approved):
I think everyone is in agreement that this is a good change so

sunzenshen (approved):
Looks fine.  If anything, silent walking could be balanced in other ways if players find it too weird that Support is too quiet.

PR #1101, opened 2025-04-27
Fix crash on player disconnect when just killed

AdamTadeusz:
## Description

Saves the team of the player who died at the moment the event was received (should this be instead sent by the server to ensure the team is 100% accurate?)

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #


brysondev (approved):

blaberry (approved):
LGTM

Issue #975, opened 2025-03-06
[SUGGESTION] Enable Developer console by default


brysondev:
### Description

Enable dev console by default

### Proposed Solution

Enable dev console by default

AdamTadeusz:
and weapon fastswitch 🙏

PR #1116, opened 2025-04-30
Enable console and fastswitch by default

brysondev:
## Description

Enables developer console and fast weapon switching by default
## Toolchain

- Windows MSVC VS2022

## Linked Issues

- related #975


Masterkatze (dismissed):

AdamTadeusz:
image

default value for hud_fastswitch can be changed at the definition of hud_Fastswitch I think

AdamTadeusz:
I'm also a bit confused about the con_enable part. That code runs everytime on launch, but regardless of what value it sets con_enable to, that has no effect on whether I can toggle the console or not, either when using '`', and also tested with 'l'.

blaberry (approved):

AdamTadeusz (approved):

PR #1139, opened 2025-05-11
correct class team on class selection menu, disable cancel button when unassigned

AdamTadeusz:
## Description

the previous fix to class selection showing up in spectate plus disable cancel button when unassigned, set picture in class selection menu on team join event

for reference this is how the team selection menu looks in the original when unassigned
image

### DIFF
--- a/src/game/client/neo/c_neo_player.cpp
+++ b/src/game/client/neo/c_neo_player.cpp
@@ -323,7 +323,7 @@ public:
 			return;
 		}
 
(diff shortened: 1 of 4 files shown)


Rainyan (approved):

sunzenshen (approved):
Dunno why the dead code around 
//if (!Q_strcmp(type, "player_team"))
in FireGameEvent() is sticking around but nothing looks horribly wrong (famous last words)

PR #1146, opened 2025-05-14
Hide third person muzzle flash for spectators in eye, simplify particle emission, fix first person muzzle flash not rotating, rotate muzzle flash with viewmodel roation

AdamTadeusz:
## Description

The function in C_BaseAnimating that calls CNEOBaseCombatWeapon::ProcessMuzzleFlashEvent is supposed to exit early if the weapon isn't visible to the local player, but this doesn't work for weapons which are hidden to spectators in first person observer mode. This seems to be at least known about considering the check in C_BaseCombatWeapon::DrawModel, so adding the same check to CNEOBaseCombatWeapon::ProcessMuzzleFlashEvent to exit early.

When trying to figure out why third person muzzle flashes were being drawn at world origin as well (turned out to be a particle error), I figured out a way to emit a particle client side only thats much simpler (see how "vortigaunt_beam_charge" is being emitted). Since we don't need to pass anything to the one and only particle controller in ntr_muzzle_source since we're using an attachment point, it becomes a one liner.

I also noticed two problems with the first person muzzle flash.
- When using viewmodel only rotation, the muzzle flash does not rotate with the viewmodel. This is contrary to what a player would expect, as a muzzle flash is generated in part by the shape of a muzzle device. If a muzzle device is rotated in relation to a camera, then the muzzle flash would rotate too. Adding the muzzle flash rotation onto the initial rotation instead of overwriting it fixes this.
- The first person muzzle flash rotation wasn't working. I'm certain this worked in the past, probably an issue with the muzzle flash current angle not being set server side. Since this variable doesn't really need to be networked, only reset to 0 when changing weapons, I removed it from the send table and receive tables which fixes the problem. In the future I imagine both the view model and view model muzzle flash should should be entirely predicted client side, since we should be able to work out all the viewmodel properties from the current active weapon of the local player or target player, as soon as we work out how to work around the server changing our active weapon for us.

### DIFF
--- a/src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.cpp
+++ b/src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.cpp
@@ -29,13 +29,11 @@ BEGIN_NETWORK_TABLE(CNEOPredictedViewModelMuzzleFlash, DT_NEOPredictedViewModelM
 #ifndef CLIENT_DLL
 SendPropEHandle(SENDINFO_NAME(m_hMoveParent, moveparent)),
 SendPropBool(SENDINFO(m_bActive)),
-SendPropInt(SENDINFO(m_iAngleZ)),
 SendPropInt(SENDINFO(m_iAngleZIncrement)),
 SendPropBool(SENDINFO(m_bScaleChangeFlag))
 #else
 RecvPropInt(RECVINFO_NAME(m_hNetworkMoveParent, moveparent), 0, RecvProxy_IntToMoveParent),
 RecvPropBool(RECVINFO(m_bActive)),
-RecvPropInt(RECVINFO(m_iAngleZ)),
 RecvPropInt(RECVINFO(m_iAngleZIncrement)),
 RecvPropEHandle(RECVINFO(m_bScaleChangeFlag), RecvProxy_ScaleChangeFlag)
 #endif
@@ -45,7 +43,6 @@ END_NETWORK_TABLE()
 BEGIN_PREDICTION_DATA(CNEOPredictedViewModelMuzzleFlash)
 DEFINE_PRED_FIELD(m_hNetworkMoveParent, FIELD_EHANDLE, FTYPEDESC_INSENDTABLE),
 DEFINE_PRED_FIELD(m_bActive, FIELD_BOOLEAN, FTYPEDESC_INSENDTABLE),
-DEFINE_PRED_FIELD(m_iAngleZ, FIELD_INTEGER, FTYPEDESC_INSENDTABLE),
 DEFINE_PRED_FIELD(m_iAngleZIncrement, FIELD_INTEGER, FTYPEDESC_INSENDTABLE),
 DEFINE_PRED_FIELD(m_bScaleChangeFlag, FIELD_BOOLEAN, FTYPEDESC_INSENDTABLE),
 END_PREDICTION_DATA()
@@ -58,7 +55,9 @@ constexpr const char* MUZZLE_FLASH_ENTITY_MODEL = "models/effect/fpmf/fpmf01.mdl
 CNEOPredictedViewModelMuzzleFlash::CNEOPredictedViewModelMuzzleFlash()
 {
 	m_bActive = true;
+#ifdef CLIENT_DLL
 	m_iAngleZ = 0;
+#endif // CLIENT_DLL
 	m_iAngleZIncrement = -5;
 	m_flTimeSwitchOffMuzzleFlash = gpGlobals->curtime;
 	m_bScaleChangeFlag = false;
@@ -108,7 +107,7 @@ int CNEOPredictedViewModelMuzzleFlash::DrawModel(int flags)
 		vm->GetAttachment(iAttachment, localOrigin, localAngle);
 		UncorrectViewModelAttachment(localOrigin); // Need position of muzzle without fov modifications & viewmodel offset
 		m_iAngleZ = (m_iAngleZ + m_iAngleZIncrement) % 360; // NEOTODO (Adam) ? Speed of rotation depends on how often DrawModel() is called. Should this be tied to global time?
-		localAngle.z = m_iAngleZ;
+		localAngle.z += m_iAngleZ;
 		SetAbsOrigin(localOrigin);
 		SetAbsAngles(localAngle);
 
@@ -145,6 +144,9 @@ void CNEOPredictedViewModelMuzzleFlash::UpdateMuzzleFlashProperties(CBaseCombatW
 		return;
 	}
 
+#ifdef CLIENT_DLL
+	m_iAngleZ = 0;
+#endif // CLIENT_DLL
 	const auto neoWepBits = neoWep->GetNeoWepBits();
 	if (neoWepBits & (NEO_WEP_THROWABLE | NEO_WEP_GHOST | NEO_WEP_KNIFE | NEO_WEP_SUPPRESSED))
 	{
@@ -154,7 +156,6 @@ void CNEOPredictedViewModelMuzzleFlash::UpdateMuzzleFlashProperties(CBaseCombatW
 	{
 		m_bActive = true;
 		m_nSkin = 1;
-		m_iAngleZ = 0;
 		m_iAngleZIncrement = -100;
 		SetModelScale(0.75);
 		m_bScaleChangeFlag = !m_bScaleChangeFlag;
@@ -163,7 +164,6 @@ void CNEOPredictedViewModelMuzzleFlash::UpdateMuzzleFlashProperties(CBaseCombatW
 	{
 		m_bActive = true;
 		m_nSkin = 1;
-		m_iAngleZ = 0;
 		m_iAngleZIncrement = -90;
 		SetModelScale(2);
 		m_bScaleChangeFlag = !m_bScaleChangeFlag;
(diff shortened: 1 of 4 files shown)


brysondev (approved):
LGTM

DESTROYGIRL (approved):

src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.cpp
```diff
@@ -29,13 +29,11 @@ BEGIN_NETWORK_TABLE(CNEOPredictedViewModelMuzzleFlash, DT_NEOPredictedViewModelM
 #ifndef CLIENT_DLL
 SendPropEHandle(SENDINFO_NAME(m_hMoveParent, moveparent)),
 SendPropBool(SENDINFO(m_bActive)),
-SendPropInt(SENDINFO(m_iAngleZ)),
 SendPropInt(SENDINFO(m_iAngleZIncrement)),
```
Rainyan: Since we're not networking `m_iAngleZ`, I doubt we need to network `m_iAngleZIncrement` either. Could maybe change as:

```diff
diff --git a/src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.cpp b/src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.cpp
index 8d2cea5d..85bf3ecd 100644
--- a/src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.cpp
+++ b/src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.cpp
@@ -29,12 +29,10 @@ BEGIN_NETWORK_TABLE(CNEOPredictedViewModelMuzzleFlash, DT_NEOPredictedViewModelM
 #ifndef CLIENT_DLL
 SendPropEHandle(SENDINFO_NAME(m_hMoveParent, moveparent)),
 SendPropBool(SENDINFO(m_bActive)),
-SendPropInt(SENDINFO(m_iAngleZIncrement)),
 SendPropBool(SENDINFO(m_bScaleChangeFlag))
 #else
 RecvPropInt(RECVINFO_NAME(m_hNetworkMoveParent, moveparent), 0, RecvProxy_IntToMoveParent),
 RecvPropBool(RECVINFO(m_bActive)),
-RecvPropInt(RECVINFO(m_iAngleZIncrement)),
 RecvPropEHandle(RECVINFO(m_bScaleChangeFlag), RecvProxy_ScaleChangeFlag)
 #endif
 END_NETWORK_TABLE()
@@ -57,8 +55,8 @@ CNEOPredictedViewModelMuzzleFlash::CNEOPredictedViewModelMuzzleFlash()
 	m_bActive = true;
 #ifdef CLIENT_DLL
 	m_iAngleZ = 0;
-#endif // CLIENT_DLL
 	m_iAngleZIncrement = -5;
+#endif // CLIENT_DLL
 	m_flTimeSwitchOffMuzzleFlash = gpGlobals->curtime;
 	m_bScaleChangeFlag = false;
 }
@@ -169,7 +167,9 @@ void CNEOPredictedViewModelMuzzleFlash::UpdateMuzzleFlashProperties(CBaseCombatW
 	{
 		m_bActive = true;
 		m_nSkin = 1;
+#ifdef CLIENT_DLL
 		m_iAngleZIncrement = -100;
+#endif // CLIENT_DLL
 		SetModelScale(0.75);
 		m_bScaleChangeFlag = !m_bScaleChangeFlag;
 	}
@@ -177,7 +177,9 @@ void CNEOPredictedViewModelMuzzleFlash::UpdateMuzzleFlashProperties(CBaseCombatW
 	{
 		m_bActive = true;
 		m_nSkin = 1;
+#ifdef CLIENT_DLL
 		m_iAngleZIncrement = -90;
+#endif // CLIENT_DLL
 		SetModelScale(2);
 		m_bScaleChangeFlag = !m_bScaleChangeFlag;
 	}
@@ -185,7 +187,9 @@ void CNEOPredictedViewModelMuzzleFlash::UpdateMuzzleFlashProperties(CBaseCombatW
 	{
 		m_bActive = true;
 		m_nSkin = 0;
+#ifdef CLIENT_DLL
 		m_iAngleZIncrement = -90;
+#endif // CLIENT_DLL
 		SetModelScale(0.75);
 		m_bScaleChangeFlag = !m_bScaleChangeFlag;
 	}
@@ -193,7 +197,9 @@ void CNEOPredictedViewModelMuzzleFlash::UpdateMuzzleFlashProperties(CBaseCombatW
 	{
 		m_bActive = true;
 		m_nSkin = 0;
+#ifdef CLIENT_DLL
 		m_iAngleZIncrement = -100;
+#endif // CLIENT_DLL
 		SetModelScale(0.6);
 		m_bScaleChangeFlag = !m_bScaleChangeFlag;
 	}
@@ -201,7 +207,9 @@ void CNEOPredictedViewModelMuzzleFlash::UpdateMuzzleFlashProperties(CBaseCombatW
 	{
 		m_bActive = true;
 		m_nSkin = 0;
+#ifdef CLIENT_DLL
 		m_iAngleZIncrement = -90;
+#endif // CLIENT_DLL
 		SetModelScale(0.75);
 		m_bScaleChangeFlag = !m_bScaleChangeFlag;
 	}
diff --git a/src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.h b/src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.h
index 6189e307..a3f88704 100644
--- a/src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.h
+++ b/src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.h
@@ -45,7 +45,6 @@ public:
 	bool	m_bScaleChangeFlag;
 #else
 	CNetworkVar(bool, m_bActive);
-	CNetworkVar(int, m_iAngleZIncrement);
 	CNetworkVar(bool, m_bScaleChangeFlag);
 #endif // CLIENT_DLL
 	float	m_flTimeSwitchOffMuzzleFlash;
```

src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.cpp
```diff
@@ -130,6 +112,36 @@ void CNEOPredictedViewModelMuzzleFlash::ClientThink()
 
 #endif //CLIENT_DLL
 
+void CNEOPredictedViewModelMuzzleFlash::CalcViewModelView(CBasePlayer* pOwner,
+	const Vector& eyePosition, const QAngle& eyeAngles)
+{
+#ifdef GAME_DLL
+	return;
+#endif // GAME_DLL
```
Rainyan: Since this is a no-op for server, we could simplify the ifdef guards here:
```diff
diff --git a/src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.cpp b/src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.cpp
index 8d2cea5d..2ee4a03b 100644
--- a/src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.cpp
+++ b/src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.cpp
@@ -115,10 +115,7 @@ void CNEOPredictedViewModelMuzzleFlash::ClientThink()
 void CNEOPredictedViewModelMuzzleFlash::CalcViewModelView(CBasePlayer* pOwner,
 	const Vector& eyePosition, const QAngle& eyeAngles)
 {
-#ifdef GAME_DLL
-	return;
-#endif // GAME_DLL
-
+#ifdef CLIENT_DLL
 	auto vm = static_cast<CNEOPredictedViewModel*>(GetMoveParent());
 	if (!vm)
 	{
@@ -133,13 +130,14 @@ void CNEOPredictedViewModelMuzzleFlash::CalcViewModelView(CBasePlayer* pOwner,
 	Vector localOrigin;
 	QAngle localAngle;
 	vm->GetAttachment(iAttachment, localOrigin, localAngle);
-#ifdef CLIENT_DLL
+
 	UncorrectViewModelAttachment(localOrigin); // Need position of muzzle without fov modifications & viewmodel offset
 	m_iAngleZ = (m_iAngleZ + m_iAngleZIncrement) % 360; // NEOTODO (Adam) ? Speed of rotation depends on how often CalcViewModel() is called. Should this be tied to global time?
 	localAngle.z += m_iAngleZ;
-#endif //CLIENT_DLL
+
 	SetAbsOrigin(localOrigin);
 	SetAbsAngles(localAngle);
+#endif // CLIENT_DLL
 }
 
 void CNEOPredictedViewModelMuzzleFlash::UpdateMuzzleFlashProperties(CBaseCombatWeapon* pWeapon, bool repeat)
```

because the whole function body is CLIENT_DLL stuff.

src/game/shared/neo/neo_predicted_viewmodel_muzzleflash.h
```diff
@@ -20,6 +20,7 @@ class CNEOPredictedViewModelMuzzleFlash : public CPredictedViewModel
 	DECLARE_PREDICTABLE();
 	CNEOPredictedViewModelMuzzleFlash();
 	virtual ~CNEOPredictedViewModelMuzzleFlash();
+	virtual void CalcViewModelView(CBasePlayer* pOwner,	const Vector& eyePosition, const QAngle& eyeAngles);
```
Rainyan: Nit: Would prefer the `override` annotation for this declaration.

Rainyan (changes requested):

sunzenshen (approved):
On a local branch, rebased on top of https://github.com/NeotokyoRebuild/neo/commit/e80e9b0e98ca1f9410a05f477f4ae8a903df5e22 and then observed bot matches in both first version view and 3rd person follow mode.  I guess muzzle flashes look fine?  But admittedly I didn't notice the problems mentioned in the PR description, so my observations might be unsophisticated.

PR #1148, opened 2025-05-15
Ease menu in on game start

AdamTadeusz:
## Description

Hides the main menu element when loading a background map. Fades the menu in once loaded in with a delay. Also hides the main menu when joining or creating a server through the main menu

### DIFF
--- a/src/game/client/neo/ui/neo_loading.cpp
+++ b/src/game/client/neo/ui/neo_loading.cpp
@@ -31,6 +31,9 @@ CNeoLoading::CNeoLoading()
 
 	vgui::IScheme *pScheme = vgui::scheme()->GetIScheme(neoscheme);
 	ApplySchemeSettings(pScheme);
+
+	m_pHostMap = g_pCVar->FindVar("host_map");
+	Assert(m_pHostMap != nullptr);
 }
 
 CNeoLoading::~CNeoLoading()
@@ -79,8 +82,6 @@ void CNeoLoading::OnMessage(const KeyValues *params, vgui::VPANEL fromPanel)
 {
 	BaseClass::OnMessage(params, fromPanel);
 	const char *pSzMsgName = params->GetName();
-	m_bIsBackgroundMap = false;
-	m_bFoundMap = false;
 	g_pNeoRoot->m_flTimeLoadingScreenTransition = gpGlobals->realtime;
 	if (V_strcmp(pSzMsgName, "Activate") == 0)
 	{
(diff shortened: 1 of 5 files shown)


brysondev (approved):

AdamTadeusz:
To clarify that last point, the menu does normally go away when creating or joining a server through the main menu ui, but with a very slight delay so you can see the main menu flicker for a split second before the loading screen hides it, that's what I am hiding in that case

src/game/client/neo/ui/neo_root.h
```diff
@@ -223,6 +226,7 @@ class CNeoRoot : public vgui::EditablePanel, public CGameEventListener
 	MESSAGE_FUNC_CHARPTR(OnFileSelected, "FileSelected", fullpath);
 
 	bool m_bOnLoadingScreen = false;
+	float m_flTimeLoadingScreenTransition = 0;
```
nullsystem, line 229: `= 0.0f`

src/game/client/neo/ui/neo_root.h
```diff
@@ -115,6 +115,9 @@ struct SprayInfo
 	char szVtf[MAX_PATH];
 };
 
+#define NEO_MENU_SECONDS_DELAY 0.25
+#define NEO_MENU_SECONDS_TILL_FULLY_OPAQUE 0.75
```
nullsystem, line 119: Turn to static constexpr float, use EX: `0.25f` not `0.25` for float.

src/game/client/neo/ui/neo_loading.cpp
```diff
@@ -171,6 +178,36 @@ void CNeoLoading::Paint()
...
+				if (Q_stristr(hostMapName, "background_"))
+				{ // Would be nice if we could figure this out quicker without checking the name
+					m_bIsBackgroundMap = true;
+				}
+				m_bFoundMap = true;
+				// NEO TODO (Adam) Replace the background image with an appropriate one for this map?
+			}
+		}
+		return;
+	}
```
nullsystem, line 209: I'm really unsure what's the deal with the map check bools and if early returns? I've also seen the issue that is visible on when you create a server, it goes blank for a bit until it shows the loading screen which wasn't an issue before.

nullsystem (changes requested):

src/game/client/neo/ui/neo_loading.cpp
AdamTadeusz: there are no references anywhere in the codebase to the command background_map, so I figured the only way to figure out when/if the user loads into a background map (you can choose to not use a background map in the main menu) is to run the command progress_enable, which will hide the main menu for us with the loading screen if the user is loading into a background map. I'm not actually interested in showing the loading screen if the map the user is loading into is the background map, but when the loading screen is first enabled I don't know whether the user is loading into a background map or an actual map, so I hide the loading screen until the map we're loading into is determined. I thought that was an acceptable solution given how natural in my opinion this transition was (I was reminded of playing tf2 where the loading screen is mostly empty for the first few seconds until the map the player is loading into is known at which point the "You are going to x" text, as well as the hint box and class stats show up on screen)

AdamTadeusz: the booleans are there so we don't do a search through the host map name every time this function is called to check whether we are loading into a background map, but testing again now I see that doesn't quite work as when the user is loaded into a background map, the first time they connect to a different server the convar still contains the name of the background map so the loading screen just never shows up, I guess I could make this simpler and just check this convar every time the loading screen tries to draw which should fix this problem.

The loading screen does also get stuck several times which makes fading the loading screen in in the OnMainLoop look a bit weird, if this happens before we've determined what map we're loading into then the loading screen will not appear for significantly longer, but in my testing that's never been a problem, the stuttering starts right after we've determined what map we're loading into

DESTROYGIRL (dismissed):

src/game/client/neo/ui/neo_loading.cpp
```diff
@@ -171,6 +179,28 @@ void CNeoLoading::Paint()
...
+		auto hostMapName = m_pHostMap->GetString();
+		if (V_strlen(hostMapName) == 0)
+		{
+			return;
+		}
+
+		if (Q_stristr(hostMapName, "background_"))
+		{
+			return;
+		}
```
nullsystem: This is causing the UI to look squished on first loading screen for a second. Instead of early return, set a bool to instead still run through the NeoUI context and sections at least once. More of the way NeoUI initializes the context but this should do anyway.

```
static bool bStaticInitNeoUI = false;
bool bSkipRender = false;
if (...)
{
    auto hostMapName = m_pHostMap->GetString();
    if (V_strlen(hostMapName) == 0)
    {
        bSkipRender = true;
    }

    if (Q_stristr(hostMapName, "background_"))
    {
        bSkipRender = true;
    }

    if (bSkipRender && bStaticInitNeoUI)
    {
        return;
    }
}
// ...
NeoUI::BeginContext()
if (bSkipRender)
{
    NeoUI::BeginSection();
    NeoUI::EndSection();
    bStaticInitNeoUI = true;
}
else
{
   // ...
}
NeoUI::EndContext()
```

nullsystem (changes requested):

nullsystem (approved):

Masterkatze (approved):

PR #1149, opened 2025-05-16
World pos hud elements ignore texture size

### LINKED ISSUE #787: HUD elements not scaling properly at higher resolutions
### Build Info

20241102_14c4282

### Description

Some of the icons and HUD elements are not scaling properly at higher resolutions. The ghost capture icon for instance is extremely small and barely visible. Other elements such as the ghost and text messages are smaller as well though not as extreme.

### To Reproduce

Set game's resolution to 2560 x 1440 and play.

### Expected behavior

Icons and HUD elements should cover a similar, if not the same, amount of space in the player's screen at different resolutions.

### Actual behavior

Icons and HUD elements are considerably smaller at higher resolutions.

### Operating System

- [X] Windows
- [ ] Linux

### Version/Distro

Windows 10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

nt;re small cap 2
nt;re small cap
nt;re small ghost and text

AdamTadeusz:
## Description

Size of elements now depends solely on convar settings for those hud elements (and resolution for the ghost beacons when scaling with distance is enabled I guess)

## Linked Issues

- fixes #787

### DIFF
--- a/src/game/client/neo/ui/neo_hud_ghost_beacons.cpp
+++ b/src/game/client/neo/ui/neo_hud_ghost_beacons.cpp
@@ -18,8 +18,11 @@
 
 using vgui::surface;
 
-ConVar neo_ghost_beacon_scale_toggle("neo_ghost_beacon_scale_toggle", "0", FCVAR_ARCHIVE,
-	"Toggles the scaling of ghost beacons.", true, 0, true, 1);
+ConVar neo_ghost_beacon_scale("neo_ghost_beacon_scale", "1", FCVAR_ARCHIVE,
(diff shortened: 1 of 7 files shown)


AdamTadeusz:
I might add a convar to control the size of ghost beacons as well without using the scaling with distance option, and also lower the half texture const for the ghost beacons so the convar has a value of 1 by default (instead of current 0.5 in code)

brysondev (approved):

Rainyan (approved):

PR #1154, opened 2025-05-17
NeoUI Main menu - Fix engine tools having no mouse reaction

### LINKED ISSUE #1137: Engine tools are broken
Attempting to load the game with the `-tools` launch option will make the main menu unusuable and nothing else.
It was previously functional when I made the muzzle flash particle effect (about 1yr ago)

This should be fixed, otherwise all NT;RE assets will need to be extracted into another game to use the particle/material editors with our assets

nullsystem:
## Description
* Fix main menu having no reaction to the mouse when using the engine tools launch options
* Tested using `-tools -nop4` extra CLI opts
* Just applied the same "relay" workaround for keyboard as with the mouse also as prior to the fix it was at least reacting to the keyboard without the mouse. Quick test seems OK without engine tools launch also post-fix.

## Toolchain

- Linux GCC Distro Native - Gentoo GCC 14.2.1

## Linked Issues

* fixes #1137

### DIFF
--- a/src/game/client/neo/ui/neo_root.cpp
+++ b/src/game/client/neo/ui/neo_root.cpp
@@ -118,7 +118,7 @@ CNeoRootInput::CNeoRootInput(CNeoRoot *rootPanel)
 {
 	MakePopup(true);
 	SetKeyBoardInputEnabled(true);
-	SetMouseInputEnabled(false);
+	SetMouseInputEnabled(true);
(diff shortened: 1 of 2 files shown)


Masterkatze (approved):
LGTM, and mouse works.

DESTROYGIRL (approved):

PR #1158, opened 2025-05-19
func_clip_vphysics crash fix

DESTROYGIRL:
## Description
Very very specific crash made apparent when using trigger_serverragdoll and this entity

## Toolchain

- Windows MSVC VS2022

## Linked Issues
- related #1141

### DIFF
--- a/src/game/server/bmodels.cpp
+++ b/src/game/server/bmodels.cpp
@@ -1446,15 +1446,15 @@ void CFuncVPhysicsClip::Activate( void )
 
 bool CFuncVPhysicsClip::EntityPassesFilter( CBaseEntity *pOther )
 {
-#ifdef NEO
-	if ( !pOther->VPhysicsGetObject() ) // DG: Server ragdolls appear to be MOVETYPE_VPHYSICS but aren't actually a physics object right when they're spawned
-		return false;
-#endif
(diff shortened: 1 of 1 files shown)


src/game/server/bmodels.cpp
```diff
@@ -1446,6 +1446,10 @@ void CFuncVPhysicsClip::Activate( void )
 
 bool CFuncVPhysicsClip::EntityPassesFilter( CBaseEntity *pOther )
 {
+#ifdef NEO
+	if ( !pOther->VPhysicsGetObject() ) // DG: Server ragdolls appear to be MOVETYPE_VPHYSICS but aren't actually a physics object right when they're spawned
+		return false;
+#endif
 	CBaseFilter* pFilter = (CBaseFilter*)(m_hFilter.Get());
 
 	if ( pFilter )
```
AdamTadeusz, line 1455: Is this check unnecessary if this entity passes this filter? Would it make more sense to move the additional check lower where its done right before IsMoveable() is called on a nullptr?

DESTROYGIRL: I assumed that if it's null it wouldn't work/collide anyway so may as well return early. But you're probably right to move it down since it's such a rare occurrence

AdamTadeusz (approved):

Masterkatze (approved):

PR #1159, opened 2025-05-20
Dynamic Crosshair

### LINKED ISSUE #660: [FEATURE] Custom crosshair - Dynamic crosshair
### Description

* Basically have a dynamic crosshair which can be configured to change/not-change based on movement, weapon, and/or spray/recoil.

### Proposed Solution

* The crosshair HUD should be refactored entirely at first, then try to tie CNeoWeapons values to the crosshair

### Alternatives Considered

_No response_

### Additional Context

_No response_

AdamTadeusz:
## Description

[Changes in neoAssets](https://github.com/NeotokyoRebuild/neoAssets/pull/51)

Adds an option for a dynamic crosshair. Depending on the type of dynamic crosshair chosen, different parts of the crosshair will grow and shrink with the player accuracy.

Adds an option to show accuracy when using scoped weapons. The same offset used in the normal crosshair is used to draw two more scopes inside the scope view.

Adds an option for a hip fire crosshair for testing

## Linked Issues

- fixes #660


AdamTadeusz:
I'm determined to figure out this fov business before I open this up to be reviewed

src/game/client/hud_crosshair.cpp
```diff
@@ -60,6 +60,17 @@ void CVGlobal_NeoClCrosshair(IConVar *var, [[maybe_unused]] const char *pOldStri
...
+CHudCrosshair* g_pCHudCrosshair = nullptr;
+void cl_neo_crosshair_hip_fire_callback([[maybe_unused]] IConVar* var, [[maybe_unused]] const char* pOldString, float flOldValue)
+{
+	if (g_pCHudCrosshair)
+	{
+		g_pCHudCrosshair->SetHiddenBits(HIDEHUD_PLAYERDEAD | (flOldValue ? HIDEHUD_CROSSHAIR : 0));
+	}
+}
+ConVar cl_neo_crosshair_hip_fire("cl_neo_crosshair_hip_fire", "1", FCVAR_ARCHIVE, "Show the crosshair when not aiming", true, 0, true, 1, cl_neo_crosshair_hip_fire_callback);
+ConVar cl_neo_crosshair_scope_inaccuracy("cl_neo_crosshair_scope_inaccuracy", "1", FCVAR_ARCHIVE, "Show the player's inaccuracy when scoped", true, 0, true, 1);
```
nullsystem: `cl_neo_crosshair_hip_fire` sounds more like an experimental feature? `cl_neo_crosshair_scope_inaccuracy` sounds like it should be within the `cl_neo_crosshair` serialization?

src/game/client/neo/ui/neo_hud_crosshair.cpp
```diff
@@ -203,10 +251,11 @@ bool ImportCrosshair(CrosshairInfo *crh, const char *pszSequence)
 void ExportCrosshair(const CrosshairInfo *crh, char (&szSequence)[NEO_XHAIR_SEQMAX])
 {
 	V_sprintf_safe(szSequence,
-			"%d;%d;%d;%d;%d;%.3f;%d;%d;%d;%d;%d;%d;%d;",
+			"%d;%d;%d;%d;%d;%d;%.3f;%d;%d;%d;%d;%d;%d;%d;",
```
nullsystem: Add it to the end, not in the middle. Also you must bump the serialization enum.

src/game/client/neo/ui/neo_hud_crosshair.cpp
```diff
@@ -186,6 +233,7 @@ bool ImportCrosshair(CrosshairInfo *crh, const char *pszSequence)
 
 	crh->iStyle = vars[NEOXHAIR_SEGMENT_I_STYLE].iVal;
 	crh->color.SetRawColor(vars[NEOXHAIR_SEGMENT_I_COLOR].iVal);
+	crh->iEDynamicType = vars[NEOXHAIR_SEGMENT_I_DYNAMICTYPE].iVal;
```
nullsystem: For any additional features need to deal with serialization versions checking.

src/game/client/neo/ui/neo_hud_crosshair.cpp
```diff
@@ -14,34 +16,78 @@ const char **CROSSHAIR_FILES = INTERNAL_CROSSHAIR_FILES;
...
 const wchar_t **CROSSHAIR_LABELS = INTERNAL_CROSSHAIR_LABELS;
 
+static const wchar_t* INTERNAL_CROSSHAIR_DYNAMICTYPE_LABELS[CROSSHAIR_DYNAMICTYPE_TOTAL] = { L"Disabled", L"Gap", L"Circle", L"Size"};
+const wchar_t **CROSSHAIR_DYNAMICTYPE_LABELS = INTERNAL_CROSSHAIR_DYNAMICTYPE_LABELS;
+
 static const wchar_t *INTERNAL_CROSSHAIR_SIZETYPE_LABELS[CROSSHAIR_SIZETYPE__TOTAL] = { L"Absolute", L"Screen halves" };
 const wchar_t **CROSSHAIR_SIZETYPE_LABELS = INTERNAL_CROSSHAIR_SIZETYPE_LABELS;
 
-void PaintCrosshair(const CrosshairInfo &crh, const int x, const int y)
+void PaintCrosshair(const CrosshairInfo &crh, CNEO_Player *player, const int x, const int y)
```
nullsystem: I don't think `player` be given in the `PaintCrosshair` function, instead just give inaccuracy int. Only `CHudCrosshair` really deals with `player`.

nullsystem (changes requested):
Also, may want to hold this off till after networked crosshair as that has changed `hud_crosshair` and settings.

src/game/client/hud_crosshair.cpp
AdamTadeusz: yes little experimental test? https://tenor.com/view/pwead-onplanet-plead-babyplead-gif-22115440

I think cl_neo_crosshair_scope_inaccuracy should be removed if everyone's happy with the implementation, don't see why you would want not to use it, but also in the case where we do keep it you probably don't want to have it be disabled when spectating someone with it disabled when you have it enabled but also have spectated player crosshair enabled, or have it disable itself when copying someone's crosshair

src/game/client/neo/ui/neo_hud_crosshair.cpp
AdamTadeusz: why bump it multiple times between the same two releases?

AdamTadeusz: I don't even understand why we're doing this before the game is released at all, since it already fails gracefully if the crosshair cannot be read

src/game/client/neo/ui/neo_hud_crosshair.cpp
```diff
@@ -196,14 +240,16 @@ bool ImportCrosshair(CrosshairInfo *crh, const char *pszSequence)
 	crh->bTopLine = vars[NEOXHAIR_SEGMENT_B_TOPLINE].bVal;
 	crh->iCircleRad = vars[NEOXHAIR_SEGMENT_I_CIRCLERAD].iVal;
 	crh->iCircleSegments = vars[NEOXHAIR_SEGMENT_I_CIRCLESEGMENTS].iVal;
+	
+	crh->iEDynamicType = vars[NEOXHAIR_SEGMENT_I_VERSION].iVal >= NEOXHAIR_SERIAL_ALPHA_V17 ? vars[NEOXHAIR_SEGMENT_I_DYNAMICTYPE].iVal : 0;
```
nullsystem: You're supposed to add a new enum and check on that instead. Anyway ok I get your comment about "pointless" to bump in between version but you need to bare in mind next time. Remove the check for now. I still rather newer serializations gets added to the end rather than in the middle.

src/game/client/hud_crosshair.cpp
nullsystem: After checking around, I think I'd rather `cl_neo_crosshair_hip_fire` disabled by default and just ask on playtest for others to try it out. IMO it feels odd for NT especially with default texture crosshair not differing in aim vs non-aim.
Also regarding `cl_neo_crosshair_scope_inaccuracy` I personally would not use it, it looks distracting to me but that's my preference. At most I'd rather give it as an optional customization like the other crosshair customizations.

src/game/client/neo/ui/neo_root_settings.cpp
```diff
@@ -880,7 +886,7 @@ void NeoSettings_Crosshair(NeoSettings *ns)
 		}
 		else
 		{
-			PaintCrosshair(pCrosshair->info,
+			PaintCrosshair(pCrosshair->info, MAX(0, (int)(sin(gpGlobals->curtime) * 24) + 16),
```
nullsystem: Add a buttom to toggle the animation, with this on it made it harder to configure the crosshair with dynamic on.

nullsystem (changes requested):

src/game/client/neo/ui/neo_hud_crosshair.cpp
AdamTadeusz: what about crosshairs versioned NEOXHAIR_SERIAL_PREALPHA_V8_2?

AdamTadeusz:
@nullsystem I also added FCVAR_HIDDEN to cl_neo_crosshair, I think that change makes sense

src/game/client/neo/ui/neo_hud_crosshair.cpp
```diff
@@ -217,5 +263,18 @@ void ExportCrosshair(const CrosshairInfo *crh, char (&szSequence)[NEO_XHAIR_SEQM
...
+int HalfInaccuracyConeInScreenPixels(C_NEO_Player* pPlayer, C_NEOBaseCombatWeapon* pWeapon, int halfScreenWidth)
+{
+	// Hor+ FOV, e.g at an aspect ratio of 16:9 and horizontal fov of 110, the actual horizontal fov is ~120
+	const float scaledFov = ScaleFOVByWidthRatio(pPlayer->GetFOV(), engine->GetScreenAspectRatio() * 0.75f); //  4 / 3
+	const float halfInaccuracy = pWeapon && pWeapon->GetNeoWepBits() & NEO_WEP_FIREARM ? RAD2DEG(asin(pWeapon->GetBulletSpread().x)) : 0;
+	// No clue, just found a value which works well (fired some shots at 15 fov, then increased fov to 120 and scaled the circle down until it worked)
+	// NEO TODO (Adam) I welcome any suggestions on how to improve this, I assume its something to do with how higher fields of view distort an image.
+	constexpr float MAGIC_FOV_DISTORTION_VALUE = 0.4f / 120.0f;
+	const int size = halfInaccuracy ? (halfScreenWidth / ((scaledFov * 0.5f) / halfInaccuracy)) * (1 - (scaledFov * MAGIC_FOV_DISTORTION_VALUE)) : 0;
+	return size;
```
sunzenshen, line 281: I hope we don't have any players that still use almost square screens these days, but maybe we could wait for someone to complain before we recruit them to playtest a more robust handling of different screen ratios.

src/game/client/hud_crosshair.cpp
```diff
@@ -60,6 +60,17 @@ void CVGlobal_NeoClCrosshair(IConVar *var, [[maybe_unused]] const char *pOldStri
 		crosshair->m_bRefreshCrosshair = true;
 	}
 }
+
+CHudCrosshair* g_pCHudCrosshair = nullptr;
+void cl_neo_crosshair_hip_fire_callback([[maybe_unused]] IConVar* var, [[maybe_unused]] const char* pOldString, float flOldValue)
+{
+	if (g_pCHudCrosshair)
+	{
+		g_pCHudCrosshair->SetHiddenBits(HIDEHUD_PLAYERDEAD | (flOldValue ? HIDEHUD_CROSSHAIR : 0));
```
Rainyan: This logic seems a bit faulty, since it's assuming a change in the cvar will always toggle the boolean, which isn't guaranteed, because the value can be set to anything despite the numeric value clamp.

Consider the following:
* `cl_neo_crosshair_hip_fire 1 // Turns it on`
* `cl_neo_crosshair_hip_fire bagels // Turns it off(?)`
* `cl_neo_crosshair_big_fire 0 // Turns it on(???)`

Another case:
* `cl_neo_crosshair_hip_fire 1 // Turns it on`
* `cl_neo_crosshair_hip_fire 2 // Turns it off(?)`
* `cl_neo_crosshair_hip_fire 1 // It's still off(???)`

So you can get these weird sequences where values `0` and `1` do the same thing.

I think rather than using the old value, we should query the cvar for its current value and just simply evaluate it as boolean, using `ConVarRef(var).GetBool()`. And also drop the `[[maybe_unused]]` annotation from `IConVar* var` since we would definitely be using it then.

src/game/client/hud_crosshair.cpp
```diff
@@ -60,6 +60,17 @@ void CVGlobal_NeoClCrosshair(IConVar *var, [[maybe_unused]] const char *pOldStri
...
+	}
+}
+ConVar cl_neo_crosshair_hip_fire("cl_neo_crosshair_hip_fire", "0", FCVAR_ARCHIVE, "Show the crosshair when not aiming", true, 0, true, 1, cl_neo_crosshair_hip_fire_callback);
```
Rainyan: Since this is not parity to begin with, should we default-enable this cvar? Because it's giving a potential advantage to players who know about its existence.

src/game/client/hud_crosshair.cpp
```diff
@@ -463,6 +487,28 @@ void CHudCrosshair::Paint( void )
 			COLOR_WHITE
 #endif
 		);
+
+		if (cl_neo_crosshair_scope_inaccuracy.GetBool())
```
Rainyan: With SRS, this looks a bit strange, because it needs to un-scope in order to bolt the next bullet in, so as soon as you shoot, the inaccuracy thing pops up but there's nothing the player could possibly do until their next bullet is bolted in, so it's just noise on their scope distracting them. I suggest we don't show the inaccuracy until the SRS player has their next bullet in the chamber:
```diff
@@ -488,25 +488,29 @@ void CHudCrosshair::Paint( void )
 #endif
                );

-               if (cl_neo_crosshair_scope_inaccuracy.GetBool())
+               const auto isSRS = pWeapon->GetNeoWepBits() & NEO_WEP_SRS;
+               if (!isSRS || pWeapon->GetRoundChambered())
                {
-                       const int size = pWeapon ? HalfInaccuracyConeInScreenPixels(pPlayer, pWeapon, m_iHalfScreenWidth) : 0;
-                       if (size)
+                       if (cl_neo_crosshair_scope_inaccuracy.GetBool())
                        {
-                               surface()->DrawSetTexture(m_hCrosshairLight);
-                               surface()->DrawSetColor(COLOR_FADED_WHITE);
-                               surface()->DrawTexturedRect(
-                                       x - size - m_iCrosshairLightWidth,
-                                       y - size - m_iCrosshairLightHeight,
-                                       x - size + m_iCrosshairLightWidth,
-                                       y - size + m_iCrosshairLightHeight
-                               );
-                               surface()->DrawTexturedRect(
-                                       x + size - m_iCrosshairLightWidth,
-                                       y + size - m_iCrosshairLightHeight,
-                                       x + size + m_iCrosshairLightWidth,
-                                       y + size + m_iCrosshairLightHeight
-                               );
+                               const int size = HalfInaccuracyConeInScreenPixels(pPlayer, pWeapon, m_iHalfScreenWidth);
+                               if (size)
+                               {
+                                       surface()->DrawSetTexture(m_hCrosshairLight);
+                                       surface()->DrawSetColor(COLOR_FADED_WHITE);
+                                       surface()->DrawTexturedRect(
+                                               x - size - m_iCrosshairLightWidth,
+                                               y - size - m_iCrosshairLightHeight,
+                                               x - size + m_iCrosshairLightWidth,
+                                               y - size + m_iCrosshairLightHeight
+                                       );
+                                       surface()->DrawTexturedRect(
+                                               x + size - m_iCrosshairLightWidth,
+                                               y + size - m_iCrosshairLightHeight,
+                                               x + size + m_iCrosshairLightWidth,
+                                               y + size + m_iCrosshairLightHeight
+                                       );
+                               }
```

src/game/client/hud_crosshair.cpp
```diff
@@ -390,7 +414,7 @@ void CHudCrosshair::Paint( void )
 
 	// NEO TODO (nullsystem): Probably be better if the entire class refreshed to our own
 	// thing and can do away with that CHudTexture nonsense entirely
-	const bool bIsScoped = pNeoWep && pNeoWep->GetNeoWepBits() & NEO_WEP_SCOPEDWEAPON;
+	const bool bIsScoped = pWeapon && pWeapon->GetNeoWepBits() & NEO_WEP_SCOPEDWEAPON;
```
Rainyan: Nit: Bit of a misnomer; could we rename this `bIsScopedWep` or something, because it's not about whether or not the player "is scoped", but rather the abilities of their equipped gun (whether or not the gun is able to scope, regardless of current scope status).

src/game/client/neo/ui/neo_hud_crosshair.cpp
Rainyan: > I hope we don't have any players that still use almost square screens these days, but maybe we could wait for someone to complain before we recruit them to playtest a more robust handling of different screen ratios.

Tested on both 640x640 and 2000x640 resolutions, seems fine.

src/game/client/neo/ui/neo_root_settings.cpp
```diff
@@ -983,6 +998,7 @@ void NeoSettings_Crosshair(NeoSettings *ns)
 			NeoUI::RingBoxBool(L"Draw top line", &pCrosshair->info.bTopLine);
 			NeoUI::SliderInt(L"Circle radius", &pCrosshair->info.iCircleRad, 0, CROSSHAIR_MAX_CIRCLE_RAD);
 			NeoUI::SliderInt(L"Circle segments", &pCrosshair->info.iCircleSegments, 0, CROSSHAIR_MAX_CIRCLE_SEGMENTS);
+			NeoUI::RingBox(L"Dynamic type", CROSSHAIR_DYNAMICTYPE_LABELS, CROSSHAIR_DYNAMICTYPE_TOTAL, &pCrosshair->info.iEDynamicType);
```
Rainyan, line 1283: Is there a way we could decouple the dynamic type from the crosshair style? I could see people using the "default", or the "alt" crosshairs, wanting to also use the dynamic inaccuracy display.

Rainyan (changes requested):

src/game/client/neo/ui/neo_hud_crosshair.cpp
```diff
@@ -1,12 +1,14 @@
...
+#include "weapon_neobasecombatweapon.h"
 
 #include "filesystem.h"
 #include "vgui/ISurface.h"
 #include "vgui_controls/Controls.h"
 #include "ui/neo_root.h"
+#include "view.h"
 
-ConVar cl_neo_crosshair("cl_neo_crosshair", CL_NEO_CROSSHAIR_DEFAULT, FCVAR_ARCHIVE | FCVAR_USERINFO, "Serialized crosshair setting");
+ConVar cl_neo_crosshair("cl_neo_crosshair", CL_NEO_CROSSHAIR_DEFAULT, FCVAR_ARCHIVE | FCVAR_USERINFO | FCVAR_HIDDEN, "Serialized crosshair setting");
```
nullsystem: I disagree, we have an export feature that exposes the contents of this so I don't see a point of hiding this if someone want to define it through a config/in console.

src/game/client/neo/ui/neo_hud_crosshair.cpp
nullsystem: Don't bother with that, also now we're past V17 create a V18 enum and set dynamic crosshair serialization on that.

src/game/client/hud_crosshair.cpp
AdamTadeusz: I wonder if we should skip the cast since cl_neo_crosshair_hip_fire is in scope here and the value is updated before the callback is called

src/game/client/hud_crosshair.cpp
AdamTadeusz: Not sure if this feature will make it in or in what form but I did originally want to test this with people with the value set to 1

src/game/client/neo/ui/neo_root_settings.cpp
AdamTadeusz: I could display the dynamic type on any crosshair style, and depending on the type chosen display relevant custom crosshair options too, and refactor how the crosshair is drawn to draw the "dynamic" parts of the crosshair above/underneath the texture. Not sure if that's the best way to tackle this though, perhaps something to consider for a separate pr?

AdamTadeusz: We should also consider whether we want the texture crosshairs to modify the custom crosshair values or have their own values so you can switch between the texture and custom crosshair without having to redo your custom crosshair or import it from a file. That would mean another crosshair string that we might want to network for spectators.

sunzenshen (approved):

src/game/client/neo/ui/neo_hud_crosshair.h
```diff
@@ -71,9 +84,12 @@ struct CrosshairInfo
...
+	int iEDynamicType; // int NeoHudCrosshairDynamicType
 };
-void PaintCrosshair(const CrosshairInfo &crh, const int x, const int y);
+void PaintCrosshair(const CrosshairInfo &crh, int inaccuracy, const int x, const int y);
 
 // NEO NOTE (nullsystem): (*&)[NUM] enforces array size
 bool ImportCrosshair(CrosshairInfo *crh, const char *pszSequence);
 void ExportCrosshair(const CrosshairInfo *crh, char (&szSequence)[NEO_XHAIR_SEQMAX]);
+
+int HalfInaccuracyConeInScreenPixels(C_NEO_Player* player, C_NEOBaseCombatWeapon* pWeapon, int halfScreenWidth);
```
nullsystem: Nit: Inconsistent pointer styling with rest of `neo_hud_crosshair.h`

src/game/client/neo/ui/neo_hud_crosshair.h
```diff
@@ -27,14 +28,25 @@ enum NeoHudCrosshairSizeType
...
 extern ConVar cl_neo_crosshair;
 extern ConVar cl_neo_crosshair_network;
 
 extern const char **CROSSHAIR_FILES;
 extern const wchar_t **CROSSHAIR_LABELS;
 extern const wchar_t **CROSSHAIR_SIZETYPE_LABELS;
+extern const wchar_t **CROSSHAIR_DYNAMICTYPE_LABELS;
 
-#define CL_NEO_CROSSHAIR_DEFAULT "2;0;-1;0;6;0.000;2;4;0;0;1;0;0;"
+#define CL_NEO_CROSSHAIR_DEFAULT "2;0;-1;0;6;0.000;2;4;0;0;1;0;0;2;"
```
nullsystem: `NEOXHAIR_SERIAL_ALPHA_V18` makes the first part `3` not `2`

src/game/client/neo/ui/neo_hud_crosshair.cpp
```diff
@@ -15,34 +17,75 @@ const char **CROSSHAIR_FILES = INTERNAL_CROSSHAIR_FILES;
 static const wchar_t *INTERNAL_CROSSHAIR_LABELS[CROSSHAIR_STYLE__TOTAL] = { L"Default", L"Alt", L"Custom" };
 const wchar_t **CROSSHAIR_LABELS = INTERNAL_CROSSHAIR_LABELS;
 
+static const wchar_t* INTERNAL_CROSSHAIR_DYNAMICTYPE_LABELS[CROSSHAIR_DYNAMICTYPE_TOTAL] = { L"Disabled", L"Gap", L"Circle", L"Size"};
```
nullsystem: Nit: Inconsistent pointer style

src/game/client/neo/ui/neo_root_settings.cpp
```diff
@@ -887,19 +894,23 @@ void NeoSettings_Crosshair(NeoSettings *ns)
 		}
 		else
 		{
-			PaintCrosshair(pCrosshair->info,
+			const int iPreviewDynamicAccuracy = (pCrosshair->bPreviewDynamicAccuracy) ? (MAX(0, (int)(sin(gpGlobals->curtime) * 24) + 16)) : 0;
```
nullsystem: Nit: `MAX` -> SDK's `Max`

nullsystem (changes requested):

src/game/client/hud_crosshair.cpp
```diff
@@ -60,6 +60,16 @@ void CVGlobal_NeoClCrosshair(IConVar *var, [[maybe_unused]] const char *pOldStri
...
 }
+
+CHudCrosshair* g_pCHudCrosshair;
+ConVar cl_neo_crosshair_hip_fire("cl_neo_crosshair_hip_fire", "0", FCVAR_ARCHIVE, "Show the crosshair when not aiming", true, 0, true, 1,
+	[]([[maybe_unused]] IConVar* var, [[maybe_unused]] const char* pOldString, [[maybe_unused]] float flOldValue)->void{
+		if (g_pCHudCrosshair)
+		{
+			g_pCHudCrosshair->SetHiddenBits(HIDEHUD_PLAYERDEAD | (cl_neo_crosshair_hip_fire.GetBool() ? 0 : HIDEHUD_CROSSHAIR));
+		}
+	});
```
nullsystem: `g_pCHudCrosshair` is unncessary, `GET_HUDELEMENT(CHudCrosshair)` exists

nullsystem (changes requested):

Rainyan (approved):

src/game/shared/neo/weapons/weapon_bits.h
```diff
@@ -32,20 +32,15 @@
 	X(SCOPEDWEAPON, scopedwep, "") \
 	X(THROWABLE, throwable, "") \
 	X(SUPPRESSED, suppressed, "") \
+	X(FIREARM, firearm, "") \
```
Rainyan, line 36: Just curious, where does this FIREARM define come from? Since it doesn't seem related to this PR(?), and it's not present in the master branch which I presume this conflict was rebased against: https://github.com/NeotokyoRebuild/neo/blob/02319c32473b6b5e70740252198b402d79beaf33/src/game/shared/neo/weapons/weapon_bits.h#L32-L35

AdamTadeusz: image
I use it in this PR to calculate crosshair inaccuracy only if the active weapon is a firearm

Rainyan: Ah, that's my bad on sloppy grepping again the new X macro syntax. That makes sense.

Rainyan (dismissed):
LGTM

sunzenshen (dismissed):
Checked out the different dynamic crosshair types and lgtm.

Rainyan:
> It would also be interesting to change the bot crosshairs to be a dynamic one, to advertise the feature to new players, but that's possibly a subjective opinion. (not parity)

Bikeshedding, but IMO bots should ideally display the crosshair that the player has chosen as their preference, rather than being replicated.

AdamTadeusz:
> > It would also be interesting to change the bot crosshairs to be a dynamic one, to advertise the feature to new players, but that's possibly a subjective opinion. (not parity)
> 
> Bikeshedding, but IMO bots should ideally display the crosshair that the player has chosen as their preference, rather than being replicated.

Since we have the bot profiles PR branch open, we could add some personalized crosshairs to all the bots. I don't know if the original devs would be interested in hopping on and making a crosshair for themselves though

AdamTadeusz:
@nullsystem

PR #1162, opened 2025-05-21
Gamepad NeoUI navigation + settings

### LINKED ISSUE #601: [FEATURE] Root NeoUI - Gamepad support + settings
### Description

Subtask of the root menu/NeoUI - Add support for gamepad and its own settings

### Proposed Solution

* Gamepad navigation, be able to navigate from the main menu, settings, server list, etc...
* Gamepad configuration, have its own dedicated settings just for configuring the controller
* Gamepad keys UI hints, so the user will know what to press

### Alternatives Considered

_No response_

### Additional Context

_No response_

nullsystem:
## Description
* Can now navigate the NeoUI widgets using a controller
* Basic controller settings put into settings
* Removed legacy button from settings as it's all complete

## Toolchain

- Linux GCC Distro Native (Gentoo - GCC 14)

## Linked Issues

* fixes #601


Masterkatze:
Overall it works great, but there are some issues:

1. Unable to exit main menu using Circle button

2. I'm using PS5 controller, but these symbols (`A`, `X`, `B`) are from XBox? I'm not an expert of controller gaming on Steam but I think we should detect the controller model and draw corresponding button symbols:
apply_settings

3. It would be great to have a second column for keybinds so it's possible to use controller or kb+mouse without rebinding the keys every time:
settings

4. `L_TRIGGER`, `X_BUTTON`, etc. should be `L Trigger` and `X Button`?
5. Not NeoUI related, but I can't look up/down using the stick, only left/right.

DESTROYGIRL:
> It would be great to have a second column for keybinds so it's possible to use controller or kb+mouse without rebinding the keys every time

Probably a better idea than what I suggested

nullsystem:
@DESTROYGIRL @Masterkatze 
DG+1-2) Yeah those are XBox keys and the basis of controller support. There does seem to be Steam Input support in the SDK but looks like those will have to be handled different to how it's done here at the moment. I'm not seeing ability to look at names to determine DualShock/DualSense controllers, but maybe through steam inputs, although ABXY fairly standard on most controllers.
3) May have to deal with secondary bindings different as the SDK API is done in a very singular binding way, but could be done and what I'll intend to try to do later.
4) Naming directly from the SDK, at most could change capitalization + _.
5) Yeah I had that issue, maybe something need to be set.

nullsystem:
I'm bringing back this PR up now, now updated and deals with secondary keybindings (for any keybind KB+M and controllers). I think I'll just go ahead with this PR for now for legacy/fallback gamepad controls because at least it does means NeoUI can drop the "legacy" settings button, but bare in mind it doesn't deal with the proper "steam inputs".

AdamTadeusz:
hmm using a ps3 controller I can aim left and right but not up and down, regardless of whether I switch the sticks or not. I can move in all directions however with either stick

AdamTadeusz:
image

It looks like its detecting the pitch input but even after resetting controls to default the aim point doesn't move up or down

nullsystem:
@AdamTadeusz Applied an ifndef over that line to fix up/down joystick and fixed conflict

AdamTadeusz:
Also are the in game team class selection menus supposed to be navigable?

nullsystem:
@AdamTadeusz Well if ifndef it out doesn't seem to cause an issue then I don't see why that doesn't suffice as a fix.

This PR only focuses on the NeoUI so wont be doing the team-class-weapon menu.

AdamTadeusz:
> @AdamTadeusz Well if ifndef it out doesn't seem to cause an issue then I don't see why that doesn't suffice as a fix.
> 
> This PR only focuses on the NeoUI so wont be doing the team-class-weapon menu.

Chesterton's Fence. I have no idea what that code does, it would be good to know before we ifndef it out even if everything seems to work fine

nullsystem:
@AdamTadeusz I've added +jlook in autoexec.cfg, probably an OK way on allowing up/down look by default

AdamTadeusz (approved):

AdamTadeusz:
I think removing the command and all references to it would be fine, but that's beyond the scope of this PR

nullsystem:
@DESTROYGIRL Raised #1356 and #1357
@Masterkatze The issue is that the "secondary" keybind is really just a keybind not picked up as the main keybind through the SDK's API. The most I could change is to disable secondary if primary not for less confusion. NeoUI widget overlap related #1245 and raised #1358

PR #1167, opened 2025-05-31
Configurable Xray

AdamTadeusz:
## Description

[NeoAssets PR](https://github.com/NeotokyoRebuild/neoAssets/pull/53)

Switched to a full frame buffer for the outline, for a more accurate player outline Made a copy of the haloaddoutline_ps20b shader, with the only difference being the width of the shader now being configurable from the material and/or in code via convar. Also made the opacity of the fill shader a convar too, this way both can be configured and disabled with values of 0

### DIFF
--- a/src/game/client/glow_outline_effect.cpp
+++ b/src/game/client/glow_outline_effect.cpp
@@ -212,8 +212,6 @@ void CGlowObjectManager::ApplyEntityGlowEffects( const CViewSetup *pSetup, int n
 
 	int iNumGlowObjects = 0;
 
-	m_GlowObjectDefinitions.Sort();
-
 	for ( int i = 0; i < m_GlowObjectDefinitions.Count(); ++ i )
 	{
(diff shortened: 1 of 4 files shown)


Masterkatze (approved):
LGTM, works in game too.

src/game/client/glow_outline_effect.cpp
```diff
@@ -329,7 +340,7 @@ void CGlowObjectManager::ApplyEntityGlowEffects( const CViewSetup *pSetup, int n
 		ShaderStencilState_t stencilState;
 		stencilState.m_bEnable = true;
 		stencilState.m_nWriteMask = 0x0; // We're not changing stencil
-		stencilState.m_nTestMask = 0xFF;
+		stencilState.m_nTestMask = 0xFF;
```
DESTROYGIRL, line 343: 🤔

AdamTadeusz: just pretend you didn't see that

PR #1174, opened 2025-06-04
fix loading ui crash getting an non valid adress for UText

Nbc66:
## Description

There is a crash that happens in the neo_loading UI which is caused by m_bInfoLabel being an invalid memory address
This fixes it by just removing the info label step 

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #
- related #

### DIFF
--- a/src/game/client/neo/ui/neo_loading.cpp
+++ b/src/game/client/neo/ui/neo_loading.cpp
@@ -209,7 +209,7 @@ void CNeoLoading::OnMainLoop(const NeoUI::Mode eMode)
 		NeoUI::BeginSection(true);
 		{
 			NeoUI::Label(L"Press ESC to cancel");
-			if (m_pLabelInfo) NeoUI::Label(m_pLabelInfo->GetTextImage()->GetUText());
+			if (m_pLabelInfo) //NeoUI::Label(m_pLabelInfo->GetTextImage()->GetUText());
 			if (m_pProgressBarMain) NeoUI::Progress(m_pProgressBarMain->GetProgress(), 0.0f, 1.0f);
 		}
 		NeoUI::EndSection();


src/game/client/neo/ui/neo_loading.cpp
```diff
@@ -209,7 +209,7 @@ void CNeoLoading::OnMainLoop(const NeoUI::Mode eMode)
 		NeoUI::BeginSection(true);
 		{
 			NeoUI::Label(L"Press ESC to cancel");
-			if (m_pLabelInfo) NeoUI::Label(m_pLabelInfo->GetTextImage()->GetUText());
+			if (m_pLabelInfo) //NeoUI::Label(m_pLabelInfo->GetTextImage()->GetUText());
```
nullsystem, line 212: But this will cause it to if over the next line and no one else seem to have this issue so I wouldnt completely cut it out? You altered the loading dialog UI/res file, localization difference, or something? At most can just expand the if statement to check more pointers.

nullsystem (changes requested):

src/game/client/neo/ui/neo_loading.cpp
Nbc66: the thing i would recomend is just get the uText at the beginning when you cast it to a label there is no other way of doing this also I haven't touched a single resfile for this. this only happens cause the pointer becomes invalid after moving out of the scope when casting the pointer and the pointer becoming corupted

nullsystem: I'm not too sure if the text gets properly updated on each onmessage activate. You could also try to nullptr out the panel pointers on deactivate and just before refetching panels to see if it helps somewhat

AdamTadeusz:
so it's a problem with the source sdk 2013 mp version, close?

PR #1177, opened 2025-06-04
NeoUI Loading screen - Do not hold pointer to panels

### LINKED ISSUE #1176: [CURRENT prerelease TODO] NeoUI Loading screen - Can hold invalid pointer to panel and crash
### Build Info

master

### Description

See PR by Nbc66 #1174 somehow for Nbc66 `m_pLabelInfo` can end up holding an invalid pointer and crash during loading screen. I cannot reproduce the crash but I think doing an alternative way on not holding panel pointers can mitigate this.

### To Reproduce

nbc66's setup, just creating a server to load up loading screen just crashes

### Expected behavior

Not crash

### Actual behavior

_No response_

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

Windows

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

nullsystem:
## Description
* Instead, fetch panels on OnThink (250ms intervals) and just copy over the panels text/values data instead
* I cannot reproduce the crash Nbc66 having but I think this could at least mitigate the invalid pointer issue
* Panel fetching at an 250ms interval so shouldn't be too much of a performance issue, plus requiring it to be within activate/deactivate states
* Should be more robust as NeoUI widgets just fetches from plain data that's always in allocation

## Toolchain

- Linux GCC Distro Native - Gentoo Linux/GCC 14.2.1

## Linked Issues

* fixes #1176

### DIFF
--- a/src/game/client/neo/ui/neo_loading.cpp
+++ b/src/game/client/neo/ui/neo_loading.cpp
@@ -4,6 +4,7 @@
 #include "ienginevgui.h"
 #include "ui/neo_root.h"
 #include "vgui/ISurface.h"
+#include "vgui/IVGui.h"
 #include "vgui_controls/ProgressBar.h"
 #include "vgui_controls/Label.h"
 #include "vgui_controls/TextImage.h"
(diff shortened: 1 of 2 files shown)


nullsystem:
Seems to be a setup issue instead.

PR #1181, opened 2025-06-07
Crosshair - Change to single string convar + serialization

### LINKED ISSUE #1180: Change custom crosshair serialization to single convar string instead of multiple convar + files
* Easier for people to share crosshair and similar to other games with custom crosshair features
* Don't need to span multiple convars, just a single convar deals with custom crosshair
* No more crosshair file
* Will make it easier to implement spectating with other player's crosshair online eventually

nullsystem:
## Description
* All crosshair convars now converge into just `cl_neo_crosshair`
* String based serialization instead of file
* Copy/paste crosshair import/export functionality

## Toolchain

- Windows MSVC VS2022
- Linux GCC Distro Native Gentoo/GCC 14.2.1

## Linked Issues

* fixes #1180

### DIFF
--- a/src/game/client/neo/ui/neo_hud_crosshair.h
+++ b/src/game/client/neo/ui/neo_hud_crosshair.h
@@ -25,30 +25,38 @@ enum NeoHudCrosshairSizeType
 	CROSSHAIR_SIZETYPE__TOTAL,
 };
 
-extern ConVar cl_neo_crosshair_style;
-extern ConVar cl_neo_crosshair_color_r;
-extern ConVar cl_neo_crosshair_color_g;
-extern ConVar cl_neo_crosshair_color_b;
-extern ConVar cl_neo_crosshair_color_a;
-extern ConVar cl_neo_crosshair_size;
-extern ConVar cl_neo_crosshair_size_screen;
-extern ConVar cl_neo_crosshair_size_type;
-extern ConVar cl_neo_crosshair_thickness;
-extern ConVar cl_neo_crosshair_gap;
-extern ConVar cl_neo_crosshair_outline;
-extern ConVar cl_neo_crosshair_center_dot;
-extern ConVar cl_neo_crosshair_top_line;
-extern ConVar cl_neo_crosshair_circle_radius;
-extern ConVar cl_neo_crosshair_circle_segments;
+extern ConVar cl_neo_crosshair;
 
 extern const char **CROSSHAIR_FILES;
 extern const wchar_t **CROSSHAIR_LABELS;
 extern const wchar_t **CROSSHAIR_SIZETYPE_LABELS;
 
-#define NEO_XHAIR_EXT "neoxhr"
+#define CL_NEO_CROSSHAIR_DEFAULT "2;0;-1;0;2;0.000;1;2;0;0;1;0;0;"
+static constexpr const int NEO_XHAIR_SEQMAX = 256;
+
+enum NeoXHairSegment
+{
+	NEOXHAIR_SEGMENT_I_VERSION = 0,
+	NEOXHAIR_SEGMENT_I_STYLE,
+	NEOXHAIR_SEGMENT_I_COLOR,
+	NEOXHAIR_SEGMENT_I_SIZETYPE,
+	NEOXHAIR_SEGMENT_I_SIZE,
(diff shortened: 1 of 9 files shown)


src/game/client/neo/ui/neo_root_settings.cpp
```diff
@@ -906,27 +886,37 @@ void NeoSettings_Crosshair(NeoSettings *ns)
...
-						g_pNeoRoot->m_pFileIODialog->MarkForDeletion();
-					}
-					pCrosshair->eFileIOMode = bPresImport ? vgui::FOD_OPEN : vgui::FOD_SAVE;
-					g_pNeoRoot->m_eFileIOMode = CNeoRoot::FILEIODLGMODE_CROSSHAIR;
-					g_pNeoRoot->m_pFileIODialog = new vgui::FileOpenDialog(g_pNeoRoot,
-																		   bPresImport ? "Import crosshair" : "Export crosshair",
-																		   pCrosshair->eFileIOMode);
-					g_pNeoRoot->m_pFileIODialog->AddFilter("*." NEO_XHAIR_EXT, "NT;RE Crosshair", true);
-					g_pNeoRoot->m_pFileIODialog->DoModal();
+					vgui::system()->GetClipboardText(0, szClipboardCrosshair, NEO_XHAIR_SEQMAX);
```
Rainyan: This `GetClipboardText` API call returns what seems to be the number of bytes consumed (or written? I don't think we have access to the implementation of it sadly), so I think we should verify that it is not zero. Otherwise, if the user tries this with a completely empty clipboard, the `szClipboardCrosshair` output ptr would remain uninitialized, so then later when we do `V_strlen(pszSequence)` on that ptr inside the `ImportCrosshair(&pCrosshair->info, szClipboardCrosshair);` call site, it would be eventually calling the C stdlib strlen on a wild uninitialized ptr of raw memory, and I imagine there's a nonzero chance that could crash the program or do something weird with enough effort:

>  The behavior is undefined if str is not a pointer to a null-terminated byte string.
as per: https://en.cppreference.com/w/c/string/byte/strlen

So I guess we might wanna either bail out if `!GetClipboardText`, or alternatively zero-init the memory such that we never pass an uninitialized ptr, etc, to be prudent?

Edit: to be more specific, this was the behaviour on Windows 11. I was seeing `V_strlen` returning seemingly nonsensical values like 704, when I had an empty clipboard (presumably because that just happened to be the distance to the next 0 in memory). I imagine Linux or other platforms may do something slightly different with regards to their clipboard handling.

Looking at the ISystem header, it seems there's also a `GetClipboardTextCount` available, so I guess another alternative would be to query that. But assuming we get sane return values from the `GetClipboardText` function, maybe that's the simplest place to check after all.

Rainyan (changes requested):

src/game/client/neo/ui/neo_root_settings.cpp
nullsystem: I've added in more guardrails now, I also observed Windows seemed to return in `GetClipboardText`'s `char *` version returning in UTF-16 bytes format rather than `char` bytes so changed it to using wide char and converting down to char on Windows. Linux will keep just using the char version.

Rainyan (approved):

src/game/client/neo/ui/neo_hud_crosshair.h
```diff
@@ -25,30 +25,38 @@ enum NeoHudCrosshairSizeType
...
-extern ConVar cl_neo_crosshair_circle_radius;
-extern ConVar cl_neo_crosshair_circle_segments;
+extern ConVar cl_neo_crosshair;
 
 extern const char **CROSSHAIR_FILES;
 extern const wchar_t **CROSSHAIR_LABELS;
 extern const wchar_t **CROSSHAIR_SIZETYPE_LABELS;
 
-#define NEO_XHAIR_EXT "neoxhr"
+#define CL_NEO_CROSSHAIR_DEFAULT "2;0;-1;0;2;0.000;1;2;0;0;1;0;0;"
```
AdamTadeusz, line 34: Default crosshair looks kinda bad, also when I import from clipboard the crosshair types switches back to the default crosshair

AdamTadeusz: and now I can't replicate the crosshair switching to default issue, maybe a problem with the saved user settings?

PR #1184, opened 2025-06-07
Improve VS Studio 2022 instructions

Rainyan:
## Description
Improve the documentation for setting up VS 2022 debugging, since the current instructions using the CMake Targets View can apparently fail for some unknown reason.

## Toolchain
- Windows MSVC VS2022


nullsystem (approved):

DESTROYGIRL (approved):

Rainyan:
Possibly related VS bug: https://developercommunity.visualstudio.com/t/Cant-create-launchvsjson-for-CMake-ex/10913931?sort=active

> We made a fix in the next version of Visual Studio; we will let you know when this is available.
> Jun 11, 2025 8:55 PM

PR #1185, opened 2025-06-08
NEONextBots

AdamTadeusz:
## Description

Simple HL2MP bots that navigate between spawns and shoot at bad guys.

Requires [navigation meshes](https://github.com/NeotokyoRebuild/neoAssets/pull/60)

Stuff that still needs to be added:
- Vision Modes
- Cloak
- Obstructing view through smoke
- Using utility
- Sprinting
- Class specific movement and navigation
- Game mode specific behavior (Ghost call-outs, protecting the ghost carrier, protecting the VIP and avoiding combat as the VIP)
- Better reaction to sound (navigating to sound sources)
- Check generator working
- Probably lots of other things


AdamTadeusz:
I would appreciate if some c++ devs would take a look at the changes to NextBotBehavior.cpp. It was crashing on level change when bots were connected before, I'm not sure if the changes I made are now gonna result in memory leaking

src/game/server/NextBot/NextBotBehavior.h
```diff
@@ -190,12 +190,12 @@ class Behavior : public INextBotEventResponder, public IContextualQuery
...
-		for( bottomAction = m_action; bottomAction && bottomAction->m_buriedUnderMe; bottomAction = bottomAction->m_buriedUnderMe )
-			;
-
-		if ( bottomAction )
+		for (bottomAction = m_action; bottomAction && bottomAction->m_buriedUnderMe; bottomAction = bottomAction->m_buriedUnderMe)
 		{
-			delete bottomAction;
+			if ( bottomAction )
+			{
+				delete bottomAction;
```
nullsystem: This just looks like a use-after-free when `bottomAction` deleted but then goes to `bottomAction = bottomAction->m_buriedUnderMe`. If somewhat the aim is to delete all actions in here (to fix the crash?) then you would have a variable pointer to hold for the next one. Also the if is redundent since you're already checking for it in the for-loop.

nullsystem (changes requested):

src/game/server/NextBot/NextBotBehavior.h
AdamTadeusz: yeah I was a bit confused about the original formatting, but I can see now that's obviously the problem. Perhaps I should raise an issue in the sdk, but I'm not fully confident this wasn't somehow caused by me elsewhere.

src/game/server/neo/bot/behavior/nav_entities/neo_bot_nav_ent_destroy_entity.h
```diff
@@ -0,0 +1,25 @@
...
+	virtual const char *GetName( void ) const	{ return "NavEntDestroyEntity"; };
+
+private:
+	CHandle< CFuncNavPrerequisite > m_prereq;
+	PathFollower m_path;				// how we get to the target
+	CountdownTimer m_repathTimer;
+	bool m_wasIgnoringEnemies;
+
+	//void DetonateStickiesWhenSet( CNEOBot *me, CWeapon_SLAM *stickyLauncher ) const; // NEO TODO (Adam) remotedet
+	bool m_isReadyToLaunchSticky;
```
Rainyan: This `m_isReadyToLaunchSticky` member variable seems unused

src/game/server/neo/bot/behavior/nav_entities/neo_bot_nav_ent_move_to.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+	Extent targetExtent;
+	targetExtent.Init( target );
+
+	// pick random ground position within target entity as move-to goal
+	m_goalPosition = targetExtent.lo + Vector( RandomFloat( 0.0f, targetExtent.SizeX() ), RandomFloat( 0.0f, targetExtent.SizeY() ), targetExtent.SizeZ() );
+
+	TheNavMesh->GetSimpleGroundHeight( m_goalPosition, &m_goalPosition.z );
+
+	m_pGoalArea = (CNavArea*)TheNavMesh->GetNavArea( m_goalPosition );
```
Rainyan: redundant C cast; `GetNavArea` already returns type `CNavArea*`

src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
```diff
@@ -0,0 +1,645 @@
...
+
+	if ( me->HasAttribute( CNEOBot::IGNORE_ENEMIES ) )
+		return;
+
+	if ( !neo_bot_fire_weapon_allowed.GetBool() )
+	{
+		return;
+	}
+
+	CNEOBaseCombatWeapon* myWeapon = dynamic_cast<CNEOBaseCombatWeapon*>( me->GetActiveWeapon() );
```
Rainyan: We should be safe to do `static_cast` here for perf, since we know all our guns are `CNEOBaseCombatWeapon`.

There's also the `assert_cast<Type*>(ptr)` convenience oneliner available: https://github.com/NeotokyoRebuild/neo/blob/19bc9a6ba39f5981393bc7bb121ee7a82a8ecce0/src/public/tier0/dbg.h#L573-L578 if we wanna debug-assert it for good measure.

src/game/server/CMakeLists.txt
```diff
@@ -1384,6 +1384,54 @@ target_sources_grouped(
     neo/neo_dm_spawn.h
 )
 
+target_sources_grouped(
```
Rainyan, line 1392: In the `neo/bot/behavior` directory, it seems at least the `neo_bot_get_ammo` and `neo_bot_get_health` files are not included here; are those intentionally left out of the compile?

src/game/server/neo/bot/behavior/neo_bot_get_health.cpp
```diff
@@ -0,0 +1,229 @@
...
+/** 
+ * Return true if this Action has what it needs to perform right now
+ */
+bool CNEOBotGetHealth::IsPossible( CNEOBot *me )
+{
+	VPROF_BUDGET( "CNEOBotGetHealth::IsPossible", "NextBot" );
+
+	float healthRatio = (float)me->GetHealth() / (float)me->GetMaxHealth();
+
+	float t = ( healthRatio - neo_bot_health_critical_ratio.GetFloat() ) / ( neo_bot_health_ok_ratio.GetFloat() - neo_bot_health_critical_ratio.GetFloat() );
```
Rainyan, line 107: This could be a div by zero is both of these cvars in the divisor subtraction have the same value; maybe we should guard against that since these values are changeable by the user.

src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
```diff
@@ -0,0 +1,400 @@
...
+#endif
+
+	// Fallback and roam random spawn points if we have all weapons.
+	{
+		CNextSpawnFilter spawnFilter( me, 128.0f );
+
+		CUtlVector<CBaseEntity*> pSpawns;
+
+		CBaseEntity* pSearch = NULL;
+		while ( ( pSearch = gEntList.FindEntityByClassname( pSearch, "info_player_*", &spawnFilter ) ) != NULL )
```
Rainyan, line 356: We have players spawning to `info_player_attacker` and `info_player_defender` entities, but there is also the `info_player_start` entity that matches this query, which is the camera spot when joining a server. Since `info_player_start` could be outside playable area (or near environment hazards like a long fall/fire pit/etc), we might not wanna add it to the spawns list here so the bots don't try to access it(?)

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
```diff
@@ -0,0 +1,264 @@
...
+	virtual const char* GetName( void ) const { return "Despawn"; };
+};
+
+
+ActionResult< CNEOBot > CNEODespawn::Update( CNEOBot* me, float interval )
+{
+	// players need to be kicked, not deleted
+	if ( me->GetEntity()->IsPlayer() )
+	{
+		CBasePlayer* player = dynamic_cast< CBasePlayer* >( me->GetEntity() );
```
Rainyan: I believe the faster `static_cast` would be safe to do here (or `assert_cast`)

src/game/server/neo/bot/map_entities/neo_bot_proxy.h
```diff
@@ -0,0 +1,46 @@
...
+
+protected:
+	// Output
+	COutputEvent m_onSpawned;
+	COutputEvent m_onInjured;
+	COutputEvent m_onKilled;
+	COutputEvent m_onAttackingEnemy;
+	COutputEvent m_onKilledEnemy;
+
+	char m_botName[64];
```
Rainyan: We could define this size as `MAX_PLAYER_NAME_LENGTH` I believe, to avoid allocating too many bytes

src/game/server/neo/bot/map_entities/neo_bot_proxy.h
```diff
@@ -0,0 +1,46 @@
...
+	char m_teamName[64];
```
Rainyan: Could use `MAX_TEAM_NAME_LENGTH` here probably

src/game/server/neo/bot/map_entities/neo_bot_generator.cpp
```diff
@@ -0,0 +1,437 @@
...
+		++i;
+	}
+
+	if ( m_spawnedBotVector.Count() >= m_maxActiveCount )
+	{
+		SetNextThink( gpGlobals->curtime + 0.1f );
+		return;
+	}
+
+	char name[256];
```
Rainyan: I believe we could use `MAX_PLAYER_NAME_LENGTH` here

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -0,0 +1,2263 @@
...
+
+ConVar neo_bot_debug_tags("neo_bot_debug_tags", "0", FCVAR_CHEAT, "ent_text will only show tags on bots");
+
+ConVar neo_bot_ignore_real_players("neo_bot_ignore_real_players", "0", FCVAR_CHEAT);
+
+ConVar neo_bot_shotgunner_range("neo_bot_shotgunner_range", "320", FCVAR_NONE);
+ConVar neo_bot_recon_ratio("neo_bot_recon_ratio", "0.2", FCVAR_NONE);
+ConVar neo_bot_support_ratio("neo_bot_support_ratio", "0.2", FCVAR_NONE);
+
+extern ConVar neo_bot_fire_weapon_min_time;
```
Rainyan, line 33: `neo_bot_fire_weapon_min_time` is unused here

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -0,0 +1,2263 @@
...
+extern ConVar neo_bot_difficulty;
+extern ConVar neo_bot_farthest_visible_theater_sample_count;
```
Rainyan, line 35: seems like this is the only reference to (nonexisting?) `neo_bot_farthest_visible_theater_sample_count` - can remove as unused?

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -0,0 +1,2263 @@
...
+	// NEO TODO (Adam) Translate difficulty level
+	CFmtStr name("%sBOT %s", pDifficultyString, GetRandomBotName());
+
+	int i = 0;
+	for (int j = 1; j < gpGlobals->maxClients; j++)
+	{
+		auto player = UTIL_PlayerByIndex(j);
+		if (!player)
+		{
+			break;
```
Rainyan: Since the intent here is to count all named players, this should probably be a `continue` instead of `break`. Because even if the client index `j` isn't connected, there may have been a `j+1`th player who had joined before `j` disconnected, etc. And we can't know how many valid clients there may be in the `j+1...MaxClients` range without checking all the way.

Also, we should check the index from 1 up to maxclients and including maxclients, because it's effectively one-indexed (the ent index 0 is reserved to server/world). So `j <= maxclients` instead of `<`, because otherwise we wouldn't iterate over the highest valid client index.

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -0,0 +1,2263 @@
...
+	{
+		CTeam* pJinrai = GetGlobalTeam(TEAM_JINRAI);
+		CTeam* pNSF = GetGlobalTeam(TEAM_NSF);
+		const int numJinrai = pJinrai->GetNumPlayers();
+		const int numNSF = pNSF->GetNumPlayers();
+
+		iTeam = numJinrai < numNSF ? TEAM_JINRAI : numNSF < numJinrai ? TEAM_NSF : RandomInt(TEAM_JINRAI, TEAM_NSF);
+	}
+
+	char name[256];
```
Rainyan: Nit: Could probably use `MAX_PLAYER_NAME_LENGTH` for `name` size here

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -0,0 +1,2263 @@
...
+				if (bMoveToSpectatorTeam)
+				{
+					player->ChangeTeam(TEAM_SPECTATOR, false, true);
+				}
+				else
+				{
+					engine->ServerCommand(UTIL_VarArgs("kickid %d\n", player->GetUserID()));
+				}
+				CNEOBot* pBot = dynamic_cast<CNEOBot*>(player);
+				if (pBot && pBot->HasAttribute(CNEOBot::QUOTA_MANANGED))
```
Rainyan, line 340: It seems we don't have the `::ServerCommand` engine source code available sadly, but I wonder if it's safe for us to use the `player` (or `pBot`) ptr here after having already called `ServerCommand("kickid ...")` on that player's index at line 336, i.e. is there a risk that the player is no longer valid by the time we dereference it?

Maybe it would be safer to do this before kicking the player? Unless the server command is guaranteed to be processed after this and never before, but I'm not quite sure whether that's always the case.

Maybe something like:
```cpp
if (bMoveToSpectatorTeam)
{
	player->ChangeTeam(TEAM_SPECTATOR, false, true);
}

CNEOBot* pBot = ToNEOBot(player);
if (pBot && pBot->HasAttribute(CNEOBot::QUOTA_MANANGED))
{
	++iNumKicked;
}

if (!bMoveToSpectatorTeam)
{
	engine->ServerCommand(UTIL_VarArgs("kickid %d\n", player->GetUserID()));
}
```
with the `if (bMoveToSpectatorTeam) ... else` condition's `else` part becoming `!bMoveToSpectatorTeam`, so we could move the `iNumKicked` incrementing in-between the if/else cases?

src/game/server/neo/bot/map_entities/neo_bot_generator.cpp
```diff
@@ -0,0 +1,437 @@
...
+
+//------------------------------------------------------------------------------
+void CNEOBotGenerator::InputRemoveBots( inputdata_t &inputdata )
+{
+	for( int i = 0; i < m_spawnedBotVector.Count(); i++ )
+	{
+		CNEOBot *pBot = m_spawnedBotVector[i];
+		if ( pBot )
+		{
+			pBot->Remove();
```
Rainyan: AFAIK we should *not* call `UTIL_Remove` for players (which is what this `Remove` method invokes); just doing the kickid server command alone at line 216 should be enough.

src/game/server/NextBot/Player/NextBotPlayer.h
```diff
@@ -598,6 +602,38 @@ inline void NextBotPlayer< PlayerType >::PhysicsSimulate( void )
...
+#ifdef NEO
+	if (bot_mimic.GetBool())
+	{
+		auto pPlayerMimicked = UTIL_PlayerByIndex(bot_mimic.GetInt());
+		auto pThisBot = static_cast<CBasePlayer*>(GetEntity());
+		{
+			if (pPlayerMimicked && pThisBot)
+			{
+				CUserCmd cmd;
+				Q_memset(&cmd, 0, sizeof(cmd));
```
Rainyan: It looks like the ctor already zeroes the members, so we probably don't need to explicitly zero the memory here(?) Especially since we might be running quite frequently in `::PhysicsSimulate`, so it would be ideal to avoid redundancy to make this as fast as possible.

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -0,0 +1,2263 @@
...
+		"Grey",
+		"Glasseater",
+		"Ivy",
+		"KillahMo",
+		"pushBAK",
+		"Tatsur0",
+	};
+
+	return nameList[RandomInt(0, NUM_NAMES-1)];
+}
```
Rainyan: Nit: Could simplify as:
```diff
 const char* GetRandomBotName(void)
 {
-       constexpr int NUM_NAMES = 10;
-       static const char* nameList[NUM_NAMES] =
+       static const char* nameList[] =
        {
                "Deej",
                "Ed",
@@ -118,7 +117,7 @@ const char* GetRandomBotName(void)
                "Tatsur0",
        };

-       return nameList[RandomInt(0, NUM_NAMES-1)];
+       return nameList[RandomInt(0, ARRAYSIZE(nameList) - 1)];
 }
 ```
 so we don't need to hardcode the amount of entries as `NUM_NAMES`.

File: src/game/server/neo/bot/neo_bot.cpp [hunk 20]
```diff
@@ -0,0 +1,2263 @@
...
+	if (i)
+	{
+		name.AppendFormat("(%i)", i);
+	}
+
+	Q_strncpy(pBuffer, name.Access(), iBufferSize);
+}
+
+//-----------------------------------------------------------------------------------------------------
+CON_COMMAND_F(neo_bot_add, "Add a bot.", FCVAR_GAMEDLL)
```
Rainyan, line 156: For parity with the OGNT bot spawn command, it might be nice to have a `"bot_add"` command alias for `neo_bot_add`. Since that is the command name in OGNT to spawn a bot.

This could be declared as `CON_COMMAND_EXTERN(bot_add, neo_bot_add, "Add a bot.");` below the `neo_bot_add` implementation here for example.

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -0,0 +1,2263 @@
...
+
+
+
+//-----------------------------------------------------------------------------------------------------
+bool IsTeamName(const char* string)
+{
+	if (!stricmp(string, TEAM_STR_JINRAI))
+		return true;
+
+	if (!stricmp(string, TEAM_STR_NSF))
```
Rainyan: `stricmp` is MSVC specific and non-standard; we should probably prefer `V_stricmp` for better compat with Linux compilers etc

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -0,0 +1,2263 @@
...
+
+
+//-----------------------------------------------------------------------------------------------------
+int Bot_GetTeamByName(const char* string)
+{
+	if (!stricmp(string, TEAM_STR_JINRAI))
+	{
+		return TEAM_JINRAI;
+	}
+	else if (!stricmp(string, TEAM_STR_NSF))
```
Rainyan: `stricmp` -> `V_stricmp`

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -0,0 +1,2263 @@
...
+	if (!stricmp(string, "easy"))
+		return CNEOBot::EASY;
+
+	if (!stricmp(string, "normal"))
+		return CNEOBot::NORMAL;
+
+	if (!stricmp(string, "hard"))
+		return CNEOBot::HARD;
+
+	if (!stricmp(string, "expert"))
```
Rainyan: `stricmp` -> `V_stricmp`

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -0,0 +1,2263 @@
...
+	{
+		CNEOBot::DifficultyType trySkill = StringToDifficultyLevel(args.Arg(i));
+		int nArgAsInteger = atoi(args.Arg(i));
+
+		// each argument could be a classname, a team, a difficulty level, a count, or a name
+		if (IsTeamName(args.Arg(i)))
+		{
+			teamname = args.Arg(i);
+		}
+		else if (!stricmp(args.Arg(i), "noquota"))
```
Rainyan: `stricmp` -> `V_stricmp`

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -0,0 +1,2263 @@
...
+}
+static ConCommand neo_bot_warp_team_to_me("neo_bot_warp_team_to_me", CMD_BotWarpTeamToMe, "", FCVAR_GAMEDLL | FCVAR_CHEAT);
+
+
+//-----------------------------------------------------------------------------------------------------
+IMPLEMENT_INTENTION_INTERFACE(CNEOBot, CNEOBotMainAction);
+
+
+//-----------------------------------------------------------------------------------------------------
+static CEntityFactory<CNEOBot> neo_bot("neo_bot");;
```
Rainyan: Nit: To stay consistent with the base engine code style, perhaps we could use https://github.com/NeotokyoRebuild/neo/blob/748891e0681786cf2ef84f2c8f3fd2e1c596de6c/src/game/server/util.h#L149-L150 here, which is an alias for this. Since it seems nowhere else in the code base do we directly declare the factory without using the macro.

So something like `LINK_ENTITY_TO_CLASS(neo_bot, CNEOBot);`

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -0,0 +1,2263 @@
...
+	int m_enemyTeamIndex;
+	CNavArea* m_vantageArea;
+};
+
+
+//-----------------------------------------------------------------------------------------------------
+// Return a nearby area where we can see a member of the enemy team
+CNavArea* CNEOBot::FindVantagePoint(float maxTravelDistance) const
+{
+	CFindVantagePoint find(GetTeamNumber() == TEAM_COMBINE ? TEAM_REBELS : TEAM_COMBINE);
```
Rainyan: We should probably be using the NEO team defs (`TEAM_JINRAI`/`TEAM_NSF`) here, instead. Although the actual values of Rebels/Combine do appear to match, by happenstance.

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -0,0 +1,2263 @@
...
+{
+	CNEOBaseCombatWeapon* myWeapon = static_cast<CNEOBaseCombatWeapon *>(GetActiveWeapon());
+	if (!myWeapon)
+	{
+		return 0.0f;
+	}
+
+	if (myWeapon->GetNeoWepBits() & NEO_WEP_KNIFE)
+	{
+		return 51.0f; // KNIFE_RANGE
```
Rainyan: Maybe use `KNIFE_RANGE` here, so if that value ever changes, this value doesn't go out-of-sync.

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -0,0 +1,2263 @@
...
+	{
+		return false;
+	}
+
+	if (weapon->GetNeoWepBits() & NEO_WEP_GHOST)
+	{
+		return false;
+	}
+
+	return weapon != NULL;
```
Rainyan: This can simplify to `return true` because of the `if (!weapon)` check above (and `GetNeoWepBits` being `const`) - we know `(weapon != NULL)` must always equal true here.

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -0,0 +1,2263 @@
...
+// Return true if given weapon doesn't make much sound when used (ie: spy knife, etc)
+bool CNEOBot::IsQuietWeapon(CNEOBaseCombatWeapon* weapon) const
+{
+	if (!weapon)
+	{
+		return false;
+	}
+
+	if (!IsCombatWeapon(weapon))
+		return false;
```
Rainyan: Nit: `IsCombatWeapon` checks `if (!weapon)` so we could omit the `if (!weapon)` check here safely.

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -0,0 +1,2263 @@
...
+	{
+		CNEO_Player* other = otherVector[i];
+
+		if (other->IsBot())
+			continue;
+
+		Vector otherEye, otherForward;
+		other->EyePositionAndVectors(&otherEye, &otherForward, NULL, NULL);
+
+		Vector toMe = const_cast<CNEOBot*>(this)->EyePosition() - otherEye;
```
Rainyan: Casting away the const here runs the risk of being UB if we ever introduced vehicles (eg. the JGR mode etc...). Because `EyePosition->CacheVehicleView` may mutate members of `this`.

Maybe we should remove the `const` qualifier from this method:

```diff
--- a/src/game/server/neo/bot/neo_bot.cpp
+++ b/src/game/server/neo/bot/neo_bot.cpp
@@ -1693,7 +1693,7 @@ bool CNEOBot::IsEntityBetweenTargetAndSelf(CBaseEntity* other, CBaseEntity* targ

 //-----------------------------------------------------------------------------------------------------
 // Return the nearest human player on the given team who is looking directly at me
-CNEO_Player* CNEOBot::GetClosestHumanLookingAtMe(int team) const
+CNEO_Player* CNEOBot::GetClosestHumanLookingAtMe(int team)
 {
        CUtlVector< CNEO_Player* > otherVector;
        CollectPlayers(&otherVector, team, COLLECT_ONLY_LIVING_PLAYERS);
@@ -1711,7 +1711,7 @@ CNEO_Player* CNEOBot::GetClosestHumanLookingAtMe(int team) const
                Vector otherEye, otherForward;
                other->EyePositionAndVectors(&otherEye, &otherForward, NULL, NULL);

-               Vector toMe = const_cast<CNEOBot*>(this)->EyePosition() - otherEye;
+               Vector toMe = EyePosition() - otherEye;
                float range = toMe.NormalizeInPlace();

                if (range < closeRange)
```

```diff
--- a/src/game/server/neo/bot/neo_bot.h
+++ b/src/game/server/neo/bot/neo_bot.h
@@ -168,7 +168,7 @@ public:

        bool IsEntityBetweenTargetAndSelf(CBaseEntity* other, CBaseEntity* target);     // return true if "other" is positioned inbetween us and "target"

-       CNEO_Player* GetClosestHumanLookingAtMe(int team = TEAM_ANY) const;     // return the nearest human player on the given team who is looking directly at me
+       CNEO_Player* GetClosestHumanLookingAtMe(int team = TEAM_ANY);   // return the nearest human player on the given team who is looking directly at me

        enum AttributeType
        {
```

Just in case we ever did add vehicles of some sort, so we don't introduce undefined behaviour that could be hard to debug in such a case.

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -0,0 +1,2263 @@
...
+		StartIdleSound();
+	}
+}
+
+
+//-----------------------------------------------------------------------------------------------------
+void CNEOBot::PhysicsSimulate(void)
+{
+	if (g_fGameOver)
+	{ // Trouble with Actions on game over, Action pointer read access violation either in Behavior::Reset() or Behaviour::~Behavior() when starting new round after game over, or Action<CNEOBot>::IsOutOfScope() if previous is fixed
```
Rainyan: I can't seem to repro the issue described by this comment, with `neo_score_limit 1` and then capping the ghost to end the match. Do you have any steps to reproduce the crash? And do you know if this is a NT;RE specific problem?

Seems like `CHL2MPBot::PhysicsSimulate` in the SDK 2013 repo doesn't need to do this: https://github.com/ValveSoftware/source-sdk-2013/blob/39f6dde8fbc238727c020d13b05ecadd31bda4c0/src/game/server/hl2mp/bot/hl2mp_bot.cpp#L695-L712

But in any case, I guess it doesn't hurt to have this check here for safety if there is some kind of a problem with this. Just wondering if there's some underlying problem here that we could try fixing.

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -0,0 +1,2263 @@
...
+		return false;
+
+	if (neo_bot_ignore_real_players.GetBool())
+	{
+		CBasePlayer* pPlayer = (CBasePlayer*)them;
+		if (!pPlayer->IsBot())
+			return false;
+	}
+
+	if (HL2MPRules() && HL2MPRules()->IsTeamplay())
```
Rainyan: Nit: It's the same underlying singleton so it doesn't really matter here, but maybe we would prefer to call `NEORules()` instead.

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -0,0 +1,2263 @@
...
+	{
+		if (them->GetTeamNumber() == TEAM_COMBINE && this->GetTeamNumber() == TEAM_REBELS)
+			return true;
+
+		if (them->GetTeamNumber() == TEAM_REBELS && this->GetTeamNumber() == TEAM_COMBINE)
```
Rainyan: Use team defs `TEAM_JINRAI` and `TEAM_NSF` preferrably for NEO code

src/game/server/neo/bot/neo_bot.h
```diff
@@ -0,0 +1,991 @@
...
+
+#define NEO_BOT_TYPE	1338
+
+class CNEOBotActionPoint;
+class CNEOBotGenerator;
+class CNEOBot;
+
+extern ConVar hl2_sprintspeed;
+extern ConVar hl2_normspeed;
+extern ConVar hl2_walkspeed;
```
Rainyan: The sprint and walk speed extern HL2 cvar declarations are unused.

For the normspeed, should we refactor such that the HL2 cvar is not used?

For example, perhaps at `float CNEOBotLocomotion::GetRunSpeed( void ) const`, instead do:
```cpp
return BaseClass::GetRunSpeed();
```
which would call the base class:
```cpp
float PlayerLocomotion::GetRunSpeed( void ) const
{
	return m_player->MaxSpeed();
}
```
from which we could override `virtual float MaxSpeed() const` for `CNEO_Player` (the `m_player`) such that it then calls the relevant NEO speed getter: https://github.com/NeotokyoRebuild/neo/blob/748891e0681786cf2ef84f2c8f3fd2e1c596de6c/src/game/server/neo/neo_player.h#L180-L188 

I suppose we eventually might want to also enable sprint/walk for the bots, but maybe that doesn't need to happen in this PR since it's complicated enough already.

src/game/server/neo/bot/neo_bot.h
```diff
@@ -0,0 +1,991 @@
...
+
+extern int Bot_GetTeamByName(const char* string);
+
+inline int GetEnemyTeam(int team)
+{
+	if (team == TEAM_COMBINE)
+		return TEAM_REBELS;
+
+	if (team == TEAM_REBELS)
+		return TEAM_COMBINE;
```
Rainyan: Use NEO team defs.

src/game/server/neo/bot/neo_bot.h
```diff
@@ -0,0 +1,991 @@
...
+	static bool IsBludgeon(CNEOBaseCombatWeapon* pWeapon);
+	static bool IsCloseRange(CNEOBaseCombatWeapon* pWeapon);
+	static bool IsRanged(CNEOBaseCombatWeapon* pWeapon);
+	static bool PrefersLongRange(CNEOBaseCombatWeapon* pWeapon);
+
+	enum WeaponRestrictionType
+	{
+		ANY_WEAPON = 0,
+		MELEE_ONLY = 0x0001,
+		GRAVGUN_ONLY = 0x0002,
```
Rainyan: The `GRAVGUN_ONLY` enum value is unnecessary since we don't have the gravity gun available at all for our mod.

src/game/server/neo/bot/neo_bot.h
```diff
@@ -0,0 +1,991 @@
...
+		m_closePlayer = NULL;
+		m_team = team;
+	}
+
+	bool operator() (CBasePlayer* player)
+	{
+		if (!player->IsAlive())
+			return true;
+
+		if (player->GetTeamNumber() != TEAM_REBELS && player->GetTeamNumber() != TEAM_COMBINE && player->GetTeamNumber() != TEAM_UNASSIGNED)
```
Rainyan: Change Rebel/Combine teams to NEO team defs preferably

src/game/server/neo/bot/neo_bot.h
```diff
@@ -0,0 +1,991 @@
...
+	"models/humans/group03/male_07.mdl",
+	"models/humans/group03/female_07.mdl",
+	"models/humans/group03/male_08.mdl",
+	"models/humans/group03/male_09.mdl",
+
+	"models/combine_soldier.mdl",
+	"models/combine_soldier_prisonguard.mdl",
+	"models/combine_super_soldier.mdl",
+	"models/police.mdl",
+};
```
Rainyan: Unused string declarations; we should be able to safely delete these.

src/game/server/neo/bot/neo_bot_locomotion.cpp
```diff
@@ -0,0 +1,120 @@
...
+// Move directly towards the given position
+void CNEOBotLocomotion::Approach( const Vector& pos, float goalWeight )
+{
+	BaseClass::Approach( pos, goalWeight );
+}
+
+
+//-----------------------------------------------------------------------------------------
+// Distance at which we will die if we fall
+float CNEOBotLocomotion::GetDeathDropHeight( void ) const
```
Rainyan, line 42: I think the correct value of this would depend on the player's class, since the max HP differs for each? Possibly the approximation at line 52 times HP/100.0 to account for NEO class health(?)

src/game/server/neo/bot/neo_bot_locomotion.h
```diff
@@ -0,0 +1,37 @@
...
+	virtual bool IsEntityTraversable( CBaseEntity *obstacle, TraverseWhenType when = EVENTUALLY ) const;
+
+protected:
+	virtual void AdjustPosture( const Vector &moveGoal ) { }	// never crouch to navigate
+};
+
+inline float CNEOBotLocomotion::GetMaxJumpHeight( void ) const
+{
+	// https://developer.valvesoftware.com/wiki/Dimensions_(Half-Life_2_and_Counter-Strike:_Source)#Jumping
+	return 56.0f;
```
Rainyan: We should probably refer to the appropriate per-class NEO defs here for jump height

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -3802,3 +3808,10 @@ const char *CNEORules::GetTeamClantag(const int iTeamNum) const
 	default: return "";
 	}
 }
+
+#ifdef GAME_DLL
+void CNEORules::OnNavMeshLoad(void)
+{
+	TheNavMesh->SetPlayerSpawnName("info_player_defender");
```
Rainyan, line 3818: Do we need to somehow support also `info_player_attacker` (the default Jinrai spawns) for this, or is this working fine as-is?

Well, is seems like all references (at least non-external) to `m_spawnName` don't actually use this value so ¯\_(ツ)_/¯ Maybe the nav generation system refers to its value, somewhere in engine code...

src/game/server/neo/bot/neo_bot_manager.cpp
```diff
@@ -0,0 +1,721 @@
...
+#include "neo_bot.h"
+#include "neo_gamerules.h"
+
+
+//----------------------------------------------------------------------------------------------------------------
+
+// Creates and sets CNEOBotManager as the NextBotManager singleton
+static CNEOBotManager sNEOBotManager;
+
+ConVar neo_bot_difficulty( "neo_bot_difficulty", "2", FCVAR_NONE, "Defines the skill of bots joining the game.  Values are: 0=easy, 1=normal, 2=hard, 3=expert." );
```
Rainyan: Nit: we could clamp this cvar value to `0 ... (NUM_DIFFICULTY_LEVELS-1)` range using the ctor with `bool bMin, float fMin, bool bMax, float fMax`

src/game/server/neo/bot/neo_bot_manager.cpp
```diff
@@ -0,0 +1,721 @@
...
+
+static bool UTIL_KickBotFromTeam( int kickTeam )
+{
+	int i;
+
+	// try to kick a dead bot first
+	for ( i = 1; i <= gpGlobals->maxClients; ++i )
+	{
+		CNEO_Player *pPlayer = ToNEOPlayer( UTIL_PlayerByIndex( i ) );
+		CNEOBot* pBot = dynamic_cast<CNEOBot*>(pPlayer);
```
Rainyan: Nit: we could avoid the `dynamic_cast` RTTI perf cost by using the `ToNEOBot` helper func here.

Same thing at line 58 below in this file, in the second loop.

src/game/server/neo/bot/neo_bot_manager.cpp
```diff
@@ -0,0 +1,721 @@
...
+		if ( pPlayer == NULL )
+			continue;
+
+		if ( FNullEnt( pPlayer->edict() ) )
+			continue;
+
+		if ( !pPlayer->IsConnected() )
+			continue;
+
+		CNEOBot* pBot = dynamic_cast<CNEOBot*>( pPlayer );
```
Rainyan: Prefer `ToNEOBot` for perf

src/game/server/neo/bot/neo_bot_manager.cpp
```diff
@@ -0,0 +1,721 @@
...
```
Rainyan: Prefer `ToNEOBot` for perf

src/game/server/neo/bot/neo_bot_manager.cpp
```diff
@@ -0,0 +1,721 @@
...
+	int nTotalNonNEOBots = nConnectedClients - nNEOBots;
+
+	if ( FStrEq( neo_bot_quota_mode.GetString(), "fill" ) )
+	{
+		desiredBotCount = MAX( 0, desiredBotCount - nNonNEOBotsOnGameTeams );
+	}
+	else if ( FStrEq( neo_bot_quota_mode.GetString(), "match" ) )
+	{
+		// If bot_quota_mode is 'match', we want the number of bots to be bot_quota * total humans
+		desiredBotCount = (int)MAX( 0, neo_bot_quota.GetFloat() * nNonNEOBotsOnGameTeams );
```
Rainyan: Nit: prefer https://github.com/NeotokyoRebuild/neo/blob/748891e0681786cf2ef84f2c8f3fd2e1c596de6c/src/public/tier0/basetypes.h#L116-L122

src/game/server/neo/bot/neo_bot_manager.cpp
```diff
@@ -0,0 +1,721 @@
...
+	}
+
+	// if bots will auto-vacate, we need to keep one slot open to allow players to join
+	if ( neo_bot_auto_vacate.GetBool() )
+	{
+		desiredBotCount = MIN( desiredBotCount, gpGlobals->maxClients - nTotalNonNEOBots - 1 );
+	}
+	else
+	{
+		desiredBotCount = MIN( desiredBotCount, gpGlobals->maxClients - nTotalNonNEOBots );
```
Rainyan: Nit: prefer https://github.com/NeotokyoRebuild/neo/blob/748891e0681786cf2ef84f2c8f3fd2e1c596de6c/src/public/tier0/basetypes.h#L108-L114

src/game/server/neo/bot/neo_bot_manager.cpp
```diff
@@ -0,0 +1,721 @@
...
+	}
+
+	// add bots if necessary
+	if ( desiredBotCount > nNEOBotsOnGameTeams )
+	{
+		CNEOBot::DifficultyType skill = clamp((CNEOBot::DifficultyType)neo_bot_difficulty.GetInt(), CNEOBot::EASY, CNEOBot::EXPERT);
```
Rainyan: Nit: prefer https://github.com/NeotokyoRebuild/neo/blob/748891e0681786cf2ef84f2c8f3fd2e1c596de6c/src/public/tier0/basetypes.h#L93-L106

src/game/server/neo/bot/neo_bot_manager.cpp
```diff
@@ -0,0 +1,721 @@
...
+		char name[256];
```
Rainyan: Nit: can save some bytes with `MAX_PLAYER_NAME_LENGTH`

src/game/server/neo/bot/neo_bot_manager.cpp
```diff
@@ -0,0 +1,721 @@
...
+			kickTeam = TEAM_JINRAI;
+		}
+		else if (pJinrai->GetScore() < pNSF->GetScore() )
+		{
+			kickTeam = TEAM_NSF;
+		}
+		else
+		{
+			// teams and scores are equal, pick a team at random
+			kickTeam = (RandomInt( 0, 1 ) == 0) ? TEAM_JINRAI : TEAM_NSF;
```
Rainyan: Nit: can simplify as `kickTeam = RandomInt(TEAM_JINRAI, TEAM_NSF)`

src/game/server/neo/bot/neo_bot_manager.cpp
```diff
@@ -0,0 +1,721 @@
...
+		}
+
+		// attempt to kick a bot from the given team
+		if ( UTIL_KickBotFromTeam( kickTeam ) )
+			return;
+
+		// if there were no bots on the team, kick a bot from the other team
+		UTIL_KickBotFromTeam( kickTeam == TEAM_JINRAI ? TEAM_JINRAI : TEAM_NSF);
```
Rainyan: I think the ternary is the wrong way around here; if `kickTeam == TEAM_JINRAI`, we'd want to kick from `TEAM_NSF` and vice versa.

Or alternatively:
```cpp
UTIL_KickBotFromTeam(NEORules()->GetOpposingTeam(kickTeam));
```

src/game/server/neo/bot/neo_bot_manager.cpp
```diff
@@ -0,0 +1,721 @@
...
+		}
+	}
+
+	// if we made it this far, then they must all be bots!
+	if ( pTeam->GetNumPlayers() != 0 )
+	{
+		return true;
+	}
+
+	return true;
```
Rainyan, line 425: Redundant condition since the final two paths both return `true`. Did we mean to return `false` for the empty team case? Or if this logic is indeed correct, then we could simplify this as just `return true`.

src/game/server/neo/bot/neo_bot_manager.cpp
```diff
@@ -0,0 +1,721 @@
...
+void CNEOBotManager::SetIsInOfflinePractice(bool bIsInOfflinePractice)
+{
+	neo_bot_offline_practice.SetValue( bIsInOfflinePractice ? 1 : 0 );
+}
+
+
+//----------------------------------------------------------------------------------------------------------------
+bool CNEOBotManager::IsInOfflinePractice() const
+{
+	return neo_bot_offline_practice.GetInt() != 0;
```
Rainyan: Nit: can simplify as `return neo_bot_offline_practice.GetBool();`

src/game/server/neo/bot/neo_bot_manager.cpp
```diff
@@ -0,0 +1,721 @@
...
+}
+
+
+//----------------------------------------------------------------------------------------------------------------
+CNEOBot* CNEOBotManager::GetAvailableBotFromPool()
+{
+	for ( int i = 1; i <= gpGlobals->maxClients; ++i )
+	{
+		CNEO_Player *pPlayer = ToNEOPlayer( UTIL_PlayerByIndex( i ) );
+		CNEOBot* pBot = dynamic_cast<CNEOBot*>(pPlayer);
```
Rainyan: Prefer `ToNEOBot` for perf

src/game/server/neo/bot/neo_bot_manager.h
```diff
@@ -0,0 +1,109 @@
...
+		m_id = id;
+		Q_strncpy( m_name, name, 256 );
+	}
+
+	bool IsMatch( int id, const char *name )
+	{
+		return ( id == m_id && FStrEq( name, m_name ) );
+	}
+
+	char m_name[256];
```
Rainyan: Probably don't need more than `MAX_PLAYER_NAME_LENGTH` chars allocated for `m_name`

src/game/server/neo/bot/neo_bot_manager.h
```diff
@@ -0,0 +1,109 @@
...
+class CNEOBotManager : public NextBotManager
+{
+public:
+	CNEOBotManager();
+	virtual ~CNEOBotManager();
+
+	virtual void Update();
+	void LevelShutdown();
+
+	virtual void OnMapLoaded( void );						// when the server has changed maps
```
Rainyan: Nit: would prefer if we declared `Update` and `OnMapLoaded` as `override`

src/game/server/neo/bot/neo_bot_squad.cpp
```diff
@@ -0,0 +1,278 @@
...
+			// aux power, and see if they are allowed to sprint here.
+			//
+			// I am not planning on using squads, just pointing it out if someone
+			// else wants to use this code.
+			float memberSpeed = hl2_normspeed.GetFloat();
+			if ( memberSpeed < speed )
+			{
+				speed = memberSpeed;
+			}
+		}
```
Rainyan: Since the `hl2_normspeed` is likely not relevant for us, perhaps we could instead do something like:
```diff
diff --git a/src/game/server/neo/bot/neo_bot_squad.cpp b/src/game/server/neo/bot/neo_bot_squad.cpp
index 5f53fd57..604246ce 100644
--- a/src/game/server/neo/bot/neo_bot_squad.cpp
+++ b/src/game/server/neo/bot/neo_bot_squad.cpp
@@ -168,14 +168,16 @@ float CNEOBotSquad::GetSlowestMemberIdealSpeed( bool includeLeader ) const

        for( ; i<m_roster.Count(); ++i )
        {
-               if ( m_roster[i].Get() != NULL && m_roster[i]->IsAlive() )
+               auto member = m_roster[i].Get();
+
+               if ( member != NULL && member->IsAlive() )
                {
                        // TODO(misyl): One could make this consider all the members of the roster's
                        // aux power, and see if they are allowed to sprint here.
                        //
                        // I am not planning on using squads, just pointing it out if someone
                        // else wants to use this code.
-                       float memberSpeed = hl2_normspeed.GetFloat();
+                       float memberSpeed = member->GetNormSpeed();
                        if ( memberSpeed < speed )
                        {
                                speed = memberSpeed;
```

src/game/server/neo/bot/neo_bot_vision.cpp
```diff
@@ -0,0 +1,150 @@
+#include "cbase.h"
+#include "vprof.h"
+
+#include "neo_bot.h"
+#include "neo_bot_vision.h"
+#include "neo_player.h"
+#include "neo_gamerules.h"
+
+ConVar neo_bot_choose_target_interval( "hl2mp_bot_choose_target_interval", "0.3f", FCVAR_CHEAT, "How often, in seconds, a NEOBot can reselect his target" );
+ConVar neo_bot_sniper_choose_target_interval( "hl2mp_bot_sniper_choose_target_interval", "3.0f", FCVAR_CHEAT, "How often, in seconds, a zoomed-in Sniper can reselect his target" );
```
Rainyan: Possible typo; should these cvar name literals be `neo_...` instead of `hl2mp_...`? As typically the cvar name would match the declared name.

src/game/server/neo/bot/neo_bot_squad.h
```diff
@@ -0,0 +1,111 @@
...
+
+class CNEOBotSquad : public INextBotEventResponder
+{
+public:
+	CNEOBotSquad( void );
+	virtual ~CNEOBotSquad() { }		
+
+	// EventResponder ------
+	virtual INextBotEventResponder *FirstContainedResponder( void ) const;
+	virtual INextBotEventResponder *NextContainedResponder( INextBotEventResponder *current ) const;
```
Rainyan: Preferably declare these as `override`

src/game/server/neo/bot/neo_bot_vision.h
```diff
@@ -0,0 +1,32 @@
...
+	/**
+	 * Populate "potentiallyVisible" with the set of all entities we could potentially see. 
+	 * Entities in this set will be tested for visibility/recognition in IVision::Update()
+	 */
+	virtual void CollectPotentiallyVisibleEntities( CUtlVector< CBaseEntity * > *potentiallyVisible );
+
+	virtual bool IsIgnored( CBaseEntity *subject ) const;		// return true to completely ignore this entity (may not be in sight when this is called)
+
+	virtual float GetMaxVisionRange( void ) const;				// return maximum distance vision can reach
+	virtual float GetMinRecognizeTime( void ) const;			// return VISUAL reaction time
```
Rainyan: Preferably declare as `override`

Rainyan (changes requested):
* Perhaps we should also add a GUI slider for max bots quota under https://github.com/NeotokyoRebuild/neo/blob/748891e0681786cf2ef84f2c8f3fd2e1c596de6c/src/game/client/neo/ui/neo_root.cpp#L944 so that players who are not comfortable with console can intuitively adjust the amount when pressing the server creation button from the main menu.

* Are the bots supposed to move? For me, they appear to stand still at spawn. Or do we need to generate nav meshes for this? And if so, should we commit these meshes to neoAssets in sync with merging this PR?

src/game/server/CMakeLists.txt
AdamTadeusz: Those are non functional right now since we don't have any way to heal or resupply ammunition, I didn't know if it would be better to copy this functionality across in case we want to use it, change the names etc so it would compile and then leave them out or just not copy them across. Since bots can't use remote dets we probably should do the same for neo_bot_nav_ent_destroy_entity as well

AdamTadeusz: now im curious how neo_bot_behavior, neo_bot_scenario_monitor and neo_bot_tactical_monitor are compiling when the external neo_bot_health_ok_ratio is defined in neo_bot_get_health which is not being built

Rainyan: > Those are non functional right now since we don't have any way to heal or resupply ammunition, I didn't know if it would be better to copy this functionality across in case we want to use it, change the names etc so it would compile and then leave them out or just not copy them across. Since bots can't use remote dets we probably should do the same for neo_bot_nav_ent_destroy_entity as well

Ok, that seems reasonable. Wasn't sure if these were left out by accident, but it probably makes sense to not compile them in if we don't have the functionality currently.

src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
AdamTadeusz: NamesMatch only supports a trailing wildcard, so can't add "er" to the end to filter out "start", will just do a second check if found

src/game/server/neo/bot/neo_bot.cpp
AdamTadeusz: my understanding is that ServerCommand appends a command to some string which gets executed once per tick so this should be fine as is

src/game/server/neo/bot/neo_bot.cpp
AdamTadeusz: thats mb, I was expanding these to find the source of a problem and forgot to undo all those expansions

src/game/server/neo/bot/neo_bot.cpp
AdamTadeusz: I'll test again later but just wondering if you had a navmesh generated when you were testing this? I know bots weren't moving for you at some point and it might be the case that Actions don't get added if the bots can't move anywhere. Also from what I remember it was kind of a hit or miss whether this would be a problem, I assumed sometimes neo game rules would try to start a new round and sometimes it wouldn't have enough time to do so, this was some time ago so need to double check

src/game/shared/neo/neo_gamerules.cpp
AdamTadeusz: so far the only use for this that I've found is during automatic nav mesh generation, where generation fails with the default value since info_player_start is often in the air overlooking the map. Usually there is a clear path from all of the info_player_defender entities to the info_player_attacker entities so I've never had issues, save for bullet where navmeshes wouldn't generate in the thin doorways between carriages, though that wasn't a problem since it would then generate a disconnected navmesh in the next carriage anyway so all I had to do was create like 2 or three extra squares and connect them, and then bots would proceed to navigate between the spawn points.

src/game/server/neo/bot/neo_bot.h
AdamTadeusz: I noticed that bots set to the recon class would pull away from assaults and supports, also when using bot_mimic, so I assume we're accidentally doing our set speed per tick for bots as well. I thought I would leave this as is until we take a look at refactoring that code so these GetNorm/Crouch/Sprint speeds functions are used for human players too

src/game/server/neo/bot/neo_bot_locomotion.cpp
AdamTadeusz: I don't think the hl2mp bots could survive a fall from 311 meters at 11 health, perhaps the name of this function isn't ideal for what its actual purpose is which is to limit how far they're prepared to fall from to reach their destination. 
Maybe we want to set this value such that the bots never take damage, which is what the hl2mp bots are programmed to do at 10 health or less, maybe we never want them to take damage, and indeed fall damage taken also takes class damage resistance into account (heavier things tend to take more damage from falls in video games so maybe this isn't ideal). Anyway there's a lot of different things we can do here and I would rather just do what the default is instead of making changes and making people angry

src/game/shared/neo/neo_gamerules.cpp
AdamTadeusz: if info_player_attack is the default JINRAI spawn, then perhaps thats what we should use since TEAM_JINRAI has a lower index to TEAM_NSF

src/game/server/neo/bot/neo_bot_locomotion.h
AdamTadeusz: Can you help me out I can't find them

DESTROYGIRL: https://github.com/NeotokyoRebuild/neo/blob/748891e0681786cf2ef84f2c8f3fd2e1c596de6c/src/game/shared/gamemovement.cpp#L2497-L2510

This is all we have afaik

src/game/server/neo/bot/neo_bot_manager.cpp
AdamTadeusz: this isn't used anywhere so its difficult to say what is the correct case

src/game/server/neo/bot/neo_bot_locomotion.h
AdamTadeusz: with this being inline, if I want to cast GetBot()->GetEntity() to CNEOBot, do I have to include neo_bot.h everywhere where this is used?

src/game/server/CMakeLists.txt
AdamTadeusz: I guess its because they aren't actually used.

src/game/server/neo/bot/neo_bot_locomotion.cpp
Rainyan: I would assume the height units are in feet (i.e. Hammer units) and not meters. But yeah, I couldn't say what the actual values are without some testing. I'm assuming this is used by the algorithm for determining paths where the bots don't kill themselves by dropping when looking for the shortest valid path.

src/game/server/neo/bot/neo_bot_locomotion.h
Rainyan: You can do the `sqrt(2 * GetCurrentGravity() * GAMEMOVEMENT_JUMP_HEIGHT * isRecon?1:1.5)` for now I guess. Or hardcode the movementOptimisations values, since that math should evaluate as either `293.9387691339814f` or `240`. Thought it would be nice to refactor this into some helper function.

Rainyan: > with this being inline, if I want to cast GetBot()->GetEntity() to CNEOBot, do I have to include neo_bot.h everywhere where this is used?

Maybe, but I suspect you don't really need such a cast all too often(?)

src/game/server/neo/bot/neo_bot.cpp
Rainyan: I didn't go out of my way to generate any navmesh, no. Should I do that using the [VDC article](https://developer.valvesoftware.com/wiki/Nav_Mesh_Editing) instructions?

Fwiw, I placed a breakpoint in this func and it was being hit, so the simulate was running. But it could be that the navmesh is required to hit the crash. I'll test more later with the navmesh actually generated.

src/game/server/neo/bot/neo_bot_locomotion.cpp
Rainyan: Anyway, we could just let this be for now, and revisit if the bots do weird things when calculating their drop paths.

src/game/server/neo/bot/neo_bot_manager.cpp
Rainyan: Hmm. Well, I guess we could keep it for now and revisit if it ever poses a problem.

src/game/server/neo/bot/neo_bot_locomotion.h
```diff
@@ -0,0 +1,50 @@
...
+
+inline float CNEOBotLocomotion::GetMaxJumpHeight( void ) const
+{
+	auto me = (CNEO_Player*)GetBot()->GetEntity();
+	switch (me->GetClass())
+	{
+		case NEO_CLASS_RECON:
+			return NEO_RECON_CROUCH_JUMP_HEIGHT;
+		case NEO_CLASS_ASSAULT:
+		case NEO_CLASS_SUPPORT:
```
Rainyan: Nit: maybe mark the implicit fallthrough:
```cpp
case NEO_CLASS_ASSAULT:
	[[fallthrough]];
case NEO_CLASS_SUPPORT:
	[[fallthrough]];
```

Rainyan:
It seems that when the bot runs out of primary wep ammo, they switch to pistol correctly, but they won't actually shoot with pistol. Could be that they incorrectly hold down +attack before the pistol is ready to be fired?

AdamTadeusz:
> It seems that when the bot runs out of primary wep ammo, they switch to pistol correctly, but they won't actually shoot with pistol. Could be that they incorrectly hold down +attack before the pistol is ready to be fired?

I'll add a check whether gpGlobals->curtime <= nextprimaryattacktime, see if that fixes it

src/game/server/neo/bot/neo_bot.cpp
Rainyan: Well, I still can't reproduce this crash even with the navmesh generated. But since the simple if guard is really cheap in terms of perf, I guess we can leave it in just to be safe.

src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
```diff
@@ -0,0 +1,419 @@
...
+
+	if ( m_bGoingToTargetEntity )
+	{
+		bool bEntityVisible = false;
+		if ( m_hTargetEntity )
+		{
+			bEntityVisible = true;
+
+			CBaseCombatWeapon* pWeapon = dynamic_cast<CBaseCombatWeapon*>( m_hTargetEntity.Get() );
+			if ( pWeapon )
```
Rainyan: Nit: We should be able to do:

```cpp
CBaseCombatWeapon* pWeapon = m_hTargetEntity.Get()->MyCombatWeaponPointer();
if ( pWeapon )
{
	// ...
```

such that we don't need the dynamic_cast.

Rainyan:
> > It seems that when the bot runs out of primary wep ammo, they switch to pistol correctly, but they won't actually shoot with pistol. Could be that they incorrectly hold down +attack before the pistol is ready to be fired?
> 
> I'll add a check whether gpGlobals->curtime <= nextprimaryattacktime, see if that fixes it

It seems now the bot will take 1 shot with the pistol whenever they've lost track of you. But only 1. I'm assuming it's because they hold down the fire button as long as they can see the enemy.

Rainyan (changes requested):

AdamTadeusz:
> > > It seems that when the bot runs out of primary wep ammo, they switch to pistol correctly, but they won't actually shoot with pistol. Could be that they incorrectly hold down +attack before the pistol is ready to be fired?
> > 
> > 
> > I'll add a check whether gpGlobals->curtime <= nextprimaryattacktime, see if that fixes it
> 
> It seems now the bot will take 1 shot with the pistol whenever they've lost track of you, and you then re-appeared in their field of view. But only 1.
> 
> I'm assuming it's because they hold down the fire button as long as they can see the enemy, which blocks the semi-auto firing until they let go of the trigger.
> 
> Maybe check if `m_nOldButtons & IN_ATTACK`, and if true, `m_nButtons &= ~IN_ATTACK`? (Or would that be `m_afButtonPressed`?) Or wherever the bot-related site for toggling the fire button is.

yeah the bots were holding the attack button down. I thought I added the pistols to the iscontinuousfire function but turned out i only added the kyla and forgot about the others, tested with bots by getting them to drop their primaries and it seems to work fine now

src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
```diff
@@ -83,7 +83,7 @@ ActionResult< CNEOBot >	CNEOBotSeekAndDestroy::Update( CNEOBot *me, float interv
 		{
 			bEntityVisible = true;
 
-			CBaseCombatWeapon* pWeapon = dynamic_cast<CBaseCombatWeapon*>( m_hTargetEntity.Get() );
+			CBaseCombatWeapon* pWeapon = m_hTargetEntity.Get()->MyCombatWeaponPointer();
```
Rainyan: Actually, it seems this `.Get()->` could be a nullptr dereference, because it calls: https://github.com/NeotokyoRebuild/neo/blob/128973d974755341285b138bc6d9e8b808c8ca7d/src/game/shared/entitylist_base.h#L136-L146 and that may return `NULL`.

For safety, could we instead do an `if` check on the `Get()` ptr returnval before dereferencing the `ptr->MyCombatWeaponPointer()`? Something like:

```diff
diff --git a/src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp b/src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
index f773fead..84f3c9d8 100644
--- a/src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
@@ -83,18 +83,22 @@ ActionResult< CNEOBot >     CNEOBotSeekAndDestroy::Update( CNEOBot *me, float interv
                {
                        bEntityVisible = true;

-                       CBaseCombatWeapon* pWeapon = m_hTargetEntity.Get()->MyCombatWeaponPointer();
-                       if ( pWeapon )
+                       CBaseEntity* ent = m_hTargetEntity.Get();
+                       if (ent)
                        {
-                               if ( pWeapon->IsEffectActive( EF_NODRAW ) )
-                                       bEntityVisible = false;
-
-                               if ( pWeapon->GetOwner() != NULL )
-                                       bEntityVisible = false;
-
-                               // I don't want it anymore.
-                               if ( me->Weapon_OwnsThisType( pWeapon->GetClassname() ) )
-                                       bEntityVisible = false;
+                               CBaseCombatWeapon* pWeapon = ent->MyCombatWeaponPointer();
+                               if (pWeapon)
+                               {
+                                       if (pWeapon->IsEffectActive(EF_NODRAW))
+                                               bEntityVisible = false;
+
+                                       if (pWeapon->GetOwner() != NULL)
+                                               bEntityVisible = false;
+
+                                       // I don't want it anymore.
+                                       if (me->Weapon_OwnsThisType(pWeapon->GetClassname()))
+                                               bEntityVisible = false;
+                               }
                        }
                }
```

Rainyan (changes requested):

Rainyan (dismissed):

Rainyan (approved):

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
```diff
@@ -0,0 +1,261 @@
...
+#include "bot/neo_bot_manager.h"
+
+#include "bot/behavior/neo_bot_tactical_monitor.h"
+#include "bot/behavior/neo_bot_scenario_monitor.h"
+
+#include "bot/behavior/neo_bot_seek_and_destroy.h"
+#include "bot/behavior/neo_bot_retreat_to_cover.h"
+#if 0 // NEO TODO (Adam) Fix picking up weapons, search for dropped weapons to pick up ammo
+#include "bot/behavior/neo_bot_get_ammo.h"
+#endif 0
```
Rainyan: Preferably add comment syntax after endif to silence the warning:
`warning: extra tokens at end of #endif directive`

Also, these are unrelated to the PR, but I grepped the repo & we have these two similar issues in the code base, if you wanna sneak a fix in for those too: https://github.com/NeotokyoRebuild/neo/blob/8c50153ade38979542af9f0e2dfdb90935672809/src/game/server/hl2mp/hl2mp_player.cpp#L123 https://github.com/NeotokyoRebuild/neo/blob/8c50153ade38979542af9f0e2dfdb90935672809/src/game/client/c_baseanimating.h#L691

Rainyan:
Also submitted a [x64 related PR](https://github.com/AdamTadeusz/neo/pull/1) for this branch; if that change seems sensible then merging that should make GitHub automatically add it here, too.

Rainyan:
Reapplied the ptr patch as per https://github.com/NeotokyoRebuild/neo/pull/1185#issuecomment-3048518684

Rainyan (approved):

nullsystem (approved):

PR #1188, opened 2025-06-09
Networked crosshair

### LINKED ISSUE #1187: [FEATURE] Networked crosshair
### Description

Add the ability to view other players crosshair configuration (typically differs with custom crosshair and color configurations), toggle-able to view other players or not.

### Proposed Solution

* A toggle to view or not view other players crosshair
* When spectating in 1st person + other player ADS, should see their crosshair
* Networking is very similar to how `neo_clantag` and `neo_name` is done

### Alternatives Considered

_No response_

### Additional Context

_No response_

nullsystem:
## Description
* Since the change to string-based serialization is applied, it's now more straightforward to implement networked crosshair
* The networking is similar to how it's done for `neo_name` and `neo_clantag` as it's just strings passed around
* `cl_neo_crosshair_network` 0 = disable viewing other players crosshair, 1 = show other players crosshair
* "Other players crosshair" meaning when you're spectating the other person in 1st person while they ADS
* NeoUI settings - Slight alterations, added set to default button but also crosshair no longer part of the full-on reset button, so like the volume setting won't reset on full-on settings reset
* NeoUI settings - Added toggle for viewing networked crosshair
* NeoUI settings - Export/Import(/Default) buttons no longer restricted to only custom
* Improved default crosshair
* Tested with 2 computers that it's generally working

## Toolchain

- Linux GCC Distro Native Gentoo/GCC 14.2.1

## Linked Issues

* fixes #1187

### DIFF
--- a/src/game/client/cdll_client_int.cpp
+++ b/src/game/client/cdll_client_int.cpp
@@ -1264,7 +1264,8 @@ void MusicVol_ChangeCallback(IConVar *cvar, const char *pOldVal, float flOldVal)
 #ifdef NEO
 extern void NeoToggleConsoleEnforce();
 
-static void NeoName_ChangeCallback(IConVar *cvar, [[maybe_unused]] const char *pOldVal, [[maybe_unused]] float flOldVal)
+template <int STR_LIMIT_SIZE>
+static void NeoConVarStrLimitChangeCallback(IConVar *cvar, [[maybe_unused]] const char *pOldVal, [[maybe_unused]] float flOldVal)
 {
 	static bool bStaticCallbackChangedCVar = false;
 	if (bStaticCallbackChangedCVar)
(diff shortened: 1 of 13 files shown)


AdamTadeusz:
maybe instead of doing all those InstallChangeCallback s, pass the function as a parameter in the convar definition

src/game/client/cdll_client_int.cpp
```diff
@@ -1300,6 +1300,26 @@ static void NeoClantag_ChangeCallback(IConVar *cvar, [[maybe_unused]] const char
...
+
+static void ClNeoCrosshair_ChangeCallback(IConVar *cvar, [[maybe_unused]] const char *pOldVal, [[maybe_unused]] float flOldVal)
+{
+	static bool bStaticCallbackChangedCVar = false;
+	if (bStaticCallbackChangedCVar)
+	{
+		return;
+	}
+
+	ConVarRef cvr_cl_neo_crosshair(cvar->GetName());
```
Rainyan: There's a ctor with prototype `ConVarRef::ConVarRef( IConVar *pConVar )` available, so this could be:
```cpp
ConVarRef cvr_cl_neo_crosshair(cvar);
```
instead of doing a cvar string name lookup.

src/game/client/neo/ui/neo_root_settings.cpp
```diff
@@ -969,6 +982,8 @@ void NeoSettings_Crosshair(NeoSettings *ns)
 			NeoUI::SliderInt(L"Circle radius", &pCrosshair->info.iCircleRad, 0, CROSSHAIR_MAX_CIRCLE_RAD);
 			NeoUI::SliderInt(L"Circle segments", &pCrosshair->info.iCircleSegments, 0, CROSSHAIR_MAX_CIRCLE_SEGMENTS);
 		}
+		NeoUI::HeadingLabel(L"NETWORKING");
+		NeoUI::RingBoxBool(L"Show other players crosshair", &pCrosshair->bNetworkCrosshair);
```
Rainyan: Language nit: Maybe use the plural possessive, so `players' crosshairs`

Same with the description text at cvar `cl_neo_crosshair_network` declaration in file `src/game/client/neo/ui/neo_hud_crosshair.cpp`

Rainyan (changes requested):
Tested on LAN, seems to work well. But left some language nits that we might wanna fix before merging.

src/game/client/cdll_client_int.cpp
nullsystem: I've also refactored the whole string limiter functions into a single templated function used by the 3 convars.

src/game/client/neo/ui/neo_hud_crosshair.cpp
```diff
@@ -7,6 +7,7 @@
 #include "ui/neo_root.h"
 
 ConVar cl_neo_crosshair("cl_neo_crosshair", CL_NEO_CROSSHAIR_DEFAULT, FCVAR_ARCHIVE | FCVAR_USERINFO, "Serialized crosshair setting");
+ConVar cl_neo_crosshair_network("cl_neo_crosshair_network", "1", FCVAR_ARCHIVE | FCVAR_USERINFO, "Network crosshair - 0 = disable, 1 = show other players' crosshairs", true, 0.0f, true, 1.0f);
```
AdamTadeusz: I don't think we need the flag FCVAR_USERINFO for this convar since we're not using it server side

AdamTadeusz:
I assumed ChangeStringValue was only used for chars, but that's an easy fix

AdamTadeusz:
could also add FCVAR_NEVER_AS_STRING to every single float and int convar just in case

nullsystem:
Can't really be sure non-open part of SDK will always utilize `FCVAR_NEVER_AS_STRING` for int/float only convars or gets affected negatively by this change. I'd rather not alter tier1 library like this.

Rainyan (approved):

PR #1189, opened 2025-06-10
Entity that converts between variable types

DESTROYGIRL:
## Description
Basic feature for specific things

neoAssets >>> https://github.com/NeotokyoRebuild/neoAssets/pull/55
## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/neo_typeconverter.cpp
+++ b/src/game/server/neo/neo_typeconverter.cpp
@@ -2,7 +2,8 @@
 
 enum ConvertType_t
 {
-    CONV_FLOAT = 0,
+    CONV_INVALID = -1,
+    CONV_FLOAT,
     CONV_INTEGER,
     CONV_STRING,
 };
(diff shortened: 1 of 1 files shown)


Rainyan:
> Basic feature for specific things

Could you describe the problem that this change addresses in a little more detail?

Also, do you have an example map where this new entity is used, for verifying this PR?

Looking at the change, it seems like you're implementing an explicit int/string/float type converter, I'm guessing to get around the failure case at: https://github.com/NeotokyoRebuild/neo/blob/23343df7f2033b9bd3a93fa5acfda7b7eae64702/src/game/server/baseentity.cpp#L4420-L4428

But that failure check calls: https://github.com/NeotokyoRebuild/neo/blob/23343df7f2033b9bd3a93fa5acfda7b7eae64702/src/game/server/cbase.cpp#L1383-L1390 which already handles conversions to boolean, float or integer for `FIELD_INTEGER`, `FIELD_FLOAT`, and `FIELD_STRING`.

Is there some specific case where this engine-native functionality is insufficient?

DESTROYGIRL:
> Could you describe the problem that this change addresses in a little more detail?

There are various entities that store numerical values as strings, or only accept strings when you would want to send an int/float value. The engine seems to convert between floats/ints fine from what I can tell (dont know if it gets rounded when dealing floats), but will always fail when trying to use a string output on a float input or vice versa.

There are a couple entities that have this problem but at the time of making this entity it was so I could display a variable number in `point_worldtext`. Rather than changing that entity specifically I thought it would be a better idea to make a unique entity that handles all cases where nessecary

I also considered adding functionality for vectors but wasn't really sure how to best implement it in a way that would be useful for the entities that support vector inputs/outputs

Rainyan (approved):

AdamTadeusz (approved):

PR #1190, opened 2025-06-21
Killer damage HUD (replacing menu)

### LINKED ISSUE #667: [FEATURE] Replace killer/damage info menu with a proper UI
### Description

Currently, the killer/damage info mimicks SourceMod plugins of using the left-side menu to interact and show information. Replace this with a "proper" GUI.

### Proposed Solution

* Killer-info goes on the center, highlighting name, avatar, weapon, damage
* Other damage info goes on the side which can disappear by keybind or automatically
* Past killer/damage infos can be shown in the in-game ESC-menu

### Alternatives Considered

_No response_

### Additional Context

_No response_

nullsystem:
## Description

* Replace the menu-based killer/damage info with HUD
* Right-side, generally gives slight more killer/damage info than menu
* Keybind toggle
* Pagination
* Convar+settings on how it pops up/interacts (`cl_neo_kdinfo_toggletype`)

## Toolchain

- Linux GCC Distro Native Gentoo/GCC 14.3.0

## Linked Issues

* fixes #667


src/game/client/neo/c_neo_player.cpp
```diff
@@ -142,18 +145,52 @@ static void __MsgFunc_DamageInfo(bf_read& msg)
...
 	if (killerIdx > 0 && killerIdx <= gpGlobals->maxClients)
 	{
 		auto *neoAttacker = assert_cast<C_NEO_Player*>(UTIL_PlayerByIndex(killerIdx));
 		if (neoAttacker && neoAttacker->entindex() != thisIdx)
 		{
-			KillerLineStr(killByLine, sizeof(killByLine), neoAttacker, localPlayer, foundKilledBy ? killedBy : NULL);
+			g_pVGuiLocalize->ConvertANSIToUnicode(neoAttacker->GetNeoPlayerName(),
+												  g_neoKDmgInfos.wszKillerName,
+												  sizeof(g_neoKDmgInfos.wszKillerName));
+			g_neoKDmgInfos.iKillerNameSize = wcslen(g_neoKDmgInfos.wszKillerName);
```
sunzenshen: Maybe we want to standardize on using V_wcslen?

src/game/client/neo/ui/neo_hud_killer_damage_info.cpp
```diff
@@ -0,0 +1,346 @@
...
+
+			// General attacker/inflictor infos
+			static const int ROWLAYOUT_INFO[] = {
+				25, 50, -1
+			};
+			NeoUI::SetPerRowLayout(ARRAYSIZE(ROWLAYOUT_INFO), ROWLAYOUT_INFO);
+			m_uiCtx.eLabelTextStyle = NeoUI::TEXTSTYLE_CENTER;
+
+			const int iCurPageStart = m_iCurPage * PAGE_MAX_SHOW;
+			const int iCurPageEnd = min(m_iAttackersTotalsSize, iCurPageStart + PAGE_MAX_SHOW);
```
sunzenshen: Maybe instead use Min from \neo\src\public\tier0\basetypes.h

src/game/client/neo/ui/neo_hud_killer_damage_info.cpp
```diff
@@ -0,0 +1,346 @@
...
+		"NeoUILarge"
+	};
+	for (int i = 0; i < NeoUI::FONT__TOTAL; ++i)
+	{
+		m_uiCtx.fonts[i].hdl = pScheme->GetFont(FONT_NAMES[i], true);
+	}
+
+	m_uiCtx.layout.iDefRowTall = tall / 27;
+	m_uiCtx.iMarginX = wide / 192;
+	m_uiCtx.iMarginY = tall / 108;
```
sunzenshen, line 123: I have no idea where these numbers come from, so maybe these could be constants?

src/game/client/neo/c_neo_player.cpp
sunzenshen: Just a sanity check question, but is there any chance that ConvernANSIToUnicode could error out and not write a null terminator, causing this len check to read past the buffer?

nullsystem: `GetNeoPlayerName` and `wszKillerName` both have `MAX_PLAYER_NAME_LENGTH + 1` as array length and `g_neoKDmgInfos` zeroed out before this section, so don't think I'll be too concerned about it.

src/game/client/neo/ui/neo_hud_killer_damage_info.cpp
nullsystem: TODO for issue #1253

src/game/client/neo/c_neo_player.cpp
Rainyan: > Just a sanity check question, but is there any chance that ConvernANSIToUnicode could error out and not write a null terminator, causing this len check to read past the buffer?

For Linux, the call resolves to: https://github.com/NeotokyoRebuild/neo/blob/f5cf0e8c24ca28c5c7b8ac0670840ffafdceb538/src/tier1/strtools_unicode.cpp#L233 which looks like it'll null-terminate the output even in error state.

For Windows, it resolves to: https://learn.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-multibytetowidechar which has the following trap:
> MultiByteToWideChar does not null-terminate an output string if the input string length is explicitly specified without a terminating null character.

But for reasons nullsystem stated above, probably not an issue here either way.

sunzenshen (dismissed):

nullsystem:
Rebased for master, but also altered some stuff slightly:

* Made sure to only update HUD infos when necessary
* Changed default from match session to always
* Add bracket to toggle label on round vs match type toggle
* Minor refactor/cleanups

sunzenshen (dismissed):

Rainyan (commented):
Screenshot

Would it be possible for us to completely cut/not render the killfeed for the pixels overlapping the KID panel? Or if that's unreasonably much effort (not too familiar with the UI details), perhaps having the KID panel bg take full (or fuller) alpha opacity, and rendering the killfeed under it, so that it doesn't distract the reading as much at areas where the two overlap.

Another possible option would be to make it such that:

* If the Killer Info panel is open
* And if there's a killfeed item that would overlap with it
* Expire any upper killfeed items for as long as it takes to resolve the overlap

in which case we could dodge this problem. But I'm not sure which is the best approach.

And I guess another option would be to simply move the panel way more to the left, such that it will not overlap no matter what carnage is happening in the killfeed. Maybe this would be cleaner actually?

---

It's also a bit complicated to understand since it's quite busy visually, and many elements/info is packed very close together, but that could maybe be a thing to improve in the future for some visual design/UX pass, instead of getting too fixated on it for this PR.

---

One thing I might like to see changed is the `Killed by: <weapon>` language, since we have two of them appearing at the same time:
* `Killed by <Player name>`
* `Killed by: <Weapon name>`

which makes it a bit hard to read, since they're so similar.

Maybe it'd be better to have:

* `Killed by <Player name>`
* `Weapon: <weapon name>`

AdamTadeusz:
> Screenshot
> 
> Would it be possible for us to completely cut/not render the killfeed for the pixels overlapping the KID panel? Or if that's unreasonably much effort (not too familiar with the UI details), perhaps having the KID panel bg take full (or fuller) alpha opacity, and rendering the killfeed under it, so that it doesn't distract the reading as much at areas where the two overlap.
> 
> Another possible option would be to make it such that:
> 
>     * If the Killer Info panel is open
> 
>     * And if there's a killfeed item that would overlap with it
> 
>     * Expire any upper killfeed items for as long as it takes to resolve the overlap
> 
> 
> in which case we could dodge this problem. But I'm not sure which is the best approach.
> 
> And I guess another option would be to simply move the panel way more to the left, such that it will not overlap no matter what carnage is happening in the killfeed. Maybe this would be cleaner actually?
> 
> It's also a bit complicated to understand since it's quite busy visually, and many elements/info is packed very close together, but that could maybe be a thing to improve in the future for some visual design/UX pass, instead of getting too fixated on it for this PR.
> 
> One thing I might like to see changed is the `Killed by: <weapon>` language, since we have two of them appearing at the same time:
> 
>     * `Killed by <Player name>`
> 
>     * `Killed by: <Weapon name>`
> 
> 
> which makes it a bit hard to read, since they're so similar.
> 
> Maybe it'd be better to have:
> 
>     * `Killed by <Player name>`
> 
>     * `Weapon: <weapon name>`

A few extra pixels of space between each player in the report would be nice

src/game/server/neo/neo_player.cpp
```diff
@@ -1804,182 +1768,31 @@ bool CNEO_Player::BecomeRagdollOnClient( const Vector &force )
...
+		const char *killedWithName = "";
 		if (neoAttacker && neoAttacker->entindex() != entindex())
 		{
 			attackerIdx = static_cast<short>(neoAttacker->entindex());
-			auto* killedWithInflictor = info->GetInflictor();
-			const bool inflictorIsPlayer = killedWithInflictor ? !Q_strcmp(killedWithInflictor->GetDebugName(), "player") : false;
-			killedWithName = killedWithInflictor ? (inflictorIsPlayer ? neoAttacker->m_hActiveWeapon->GetPrintName() : killedWithInflictor->GetDebugName()) : "";
-			if (!Q_strcmp(killedWithName, "neo_grenade_frag")) { killedWithName = "Frag Grenade"; }
-			if (!Q_strcmp(killedWithName, "neo_deployed_detpack")) { killedWithName = "Remote Detpack"; }
+			if (neoAttacker->entindex() != entindex())
```
sunzenshen: This if logic might be redundant if line 1782 already checks `neoAttacker->entindex()`

Did you mean to check the assignment of attackerIdx somehow?

src/game/client/neo/ui/neo_hud_killer_damage_info.cpp
```diff
@@ -0,0 +1,559 @@
...
+			.dealtHits = neoAttacker->GetAttackerHits(thisIdx),
+			.takenDmgs = localPlayer->GetAttackersScores(pIdx),
+			.takenHits = localPlayer->GetAttackerHits(pIdx),
+		};
+		if (attackerInfo.dealtDmgs > 0 || attackerInfo.takenDmgs > 0)
+		{
+			const char *dmgerClass = GetNeoClassName(neoAttacker->GetClass());
+
+			char infoLine[128];
+			if (totals.dealtDmgs > 0 && totals.takenDmgs > 0)
```
sunzenshen: Maybe I'm kind of confused here, but I don't see where totals was ever updated before we hit this check?  Because totals was initialized with an empty {}, by the time we hit this check for the first player index, we will always skip this conditional block because of the initially empty totals.  Perhaps I'm reasoning wrong here, but it seems like the first player indexed would not have their stats printed.

game/neo/scripts/kb_def.lst
```diff
@@ -44,6 +44,9 @@
 "F8" "joinstar 4"
 "F9" "joinstar 5"
 "F10" "joinstar 0"
+"F11" "kdinfo_toggle"
+"p" "kdinfo_page_prev"
+"n" "kdinfo_page_next"
```
AdamTadeusz, line 49: maybe place next previous next to each other by default? like "n" and "m", or "o" and "p".

Actually, since n and p were chosen here presumably for "next" and "previous", it might make more sense to use the arrow keys? I don't think anyone ever uses those to look around, might be nice to support next and previous in other menus as well, for example when switching between tabs in the options menu, instead of f1 and f3

src/game/client/neo/ui/neo_hud_killer_damage_info.cpp
```diff
@@ -0,0 +1,559 @@
...
+
+void CNEOHud_KillerDamageInfo::ApplySchemeSettings(vgui::IScheme *pScheme)
+{
+	vgui::Panel::ApplySchemeSettings(pScheme);
+
+	m_hFont = pScheme->GetFont("NHudOCRSmall", true);
+
+	int wide, tall;
+	vgui::surface()->GetScreenSize(wide, tall);
+	SetBounds(0, 0, wide, tall);
```
AdamTadeusz, line 106: image

You make the hud element take up the entire screen here, but lower down in UpdateStateForNeoHudElementDraw you make the width of the panel one third the width of the screen. The result is that some of the text will sometimes be drawn outside the background. Is this intentional? I don't think it's necessarily wrong, it's difficult to work with text in these kinds of elements so a little overflow in some edge cases might be fine, especially for now.

You could simplify some of the text so it's less likely to be too large for the element. For the keybind hints at the top I would just have "Toggle x | Previous y | Next z".

For the Dealt and Taken columns, its common to just say "75 in 3" for example, though having "in 3" on its own below the total damage for that player might not look great.

The colours in the two columns already make it clear who killed who in that exchange, if anybody did, the extra info text just adds unnecessary clutter in my opinion. Of course if someone is colour-blind that extra information is absolutely necessary, but I think it makes it more difficult to read for someone who isn't. A different way to determine who killed who for colour-blind people (like perhaps a line underneath the damage number that was fatal in addition to the colour), or a way to toggle the extra info off would be appreciated.

Not sure what to do about the Weapon part at the top, but considering that in most exchanges the amount of real-estate that we have going down the column is going to be far greater than the width of the column, and that the information at the very top is likely of more importance to the player than the information at the bottom, since we have pages anyway I would make the top "Killed by" section bigger. Making it stand out from the list below would improve readability. Placing the Key and its value on separate rows would contribute to that and have the added benefit of making it less likely that the weapon name goes off the edge of the element (the "HP" text on its own would probably look a bit lonely, I would use "Health" instead, we're not running out of space on the hp class row anyway). I have a PR somewhere where I defined a nhudocrevensmaller font, might want to look at that if you consider separate font sizes for the top and bottom half of the panel (It hasn't been merged yet as of the time of writing this comment).

Finally I want to stress that these are all suggestions, I will approve this anyway as long as there aren't any issues with the code elsewhere

src/game/client/neo/ui/neo_hud_killer_damage_info.cpp
```diff
@@ -0,0 +1,559 @@
...
+	}
+
+	int wide, tall;
+	vgui::surface()->GetScreenSize(wide, tall);
+	SetBounds(0, 0, wide, tall);
+
+	SetVisible(false);
+
+	SetFgColor(COLOR_TRANSPARENT);
+	SetBgColor(COLOR_TRANSPARENT);
```
AdamTadeusz, line 90: I think we've done this in a few hud elements but its actually unnecessary. In cdll_client_int.cpp once gHud.Init(); initialises all the panels, IGameSystem::InitAllSystems() a few lines down IGameSystem::InitAllSystems reloads the scheme of the base viewport which calls ApplySchemeSettings on all of its children. Then again maybe its safer to do this, I guess not all elements have to be a child of the baseviewport, and that could change for this element somewhere down the line at which point this would break https://developer.valvesoftware.com/wiki/VGUI_Documentation

src/game/client/neo/ui/neo_hud_killer_damage_info.cpp
```diff
@@ -0,0 +1,559 @@
...
+	{
+	case KDMGINFO_TOGGLETYPE_ROUND:
+		m_bShowInfo = true;
+		pwszToggleTypeLabel = L"round";
+		break;
+	case KDMGINFO_TOGGLETYPE_NEVER:
+		m_bShowInfo = false;
+		pwszToggleTypeLabel = L"round";
+		break;
+	case KDMGINFO_TOGGLETYPE_MATCH:
```
AdamTadeusz: rain suggested a [[fallthrough]] once elsewhere, should that be used here?

src/game/client/neo/ui/neo_hud_killer_damage_info.cpp
AdamTadeusz: image

Looks like this nicely skips the info in the killedByLine if we died to this player, but in other instances it looks like it indeed might be skipping the first person, unless the reason for that is that the console skips the last bullet to have killed the local player or the last bullet to have been fired before the round ends?

AdamTadeusz: Also whats with those question marks

[edit] to clarify, the question marks in the console as seen in the image above

AdamTadeusz (changes requested):
Seems fine mostly. Some design suggestions that you can ignore, but I think it would be good if the mismatch between the information in the panel and in the console could be fixed, also confused about the indexing like @sunzenshen mentioned around line 529. It seems like its handled separately before the for loop but like I showed on that image in the third test it seems that a person is missing, but that could be linked to the last bullet mismatch between the panel and console

src/game/client/neo/ui/neo_hud_killer_damage_info.cpp
nullsystem: Good catch, it's supposed to be attackerInfo not total. Not sure how it went like that probably just some too quick edits.

AdamTadeusz: hmm actually yeah in the second case its also wrong, I guess the intention was to have the list after the question marks be exhaustive then

game/neo/scripts/kb_def.lst
nullsystem: If the arrow keys used in the main menu this is going to collide with the already existing slider and left-right widgets movement, so it should not. I'd also rather not mixup keybindings with main menu bindings, they're both should be separated and will screw up TextEdit if someone binds those to alphanumerical keys.

src/game/client/neo/ui/neo_hud_killer_damage_info.cpp
nullsystem: Again, see this comment: https://github.com/NeotokyoRebuild/neo/pull/1256#discussion_r2270555654 why I rejected that suggestion

My comment:

They're empty cases though, so they're never warned by the compilers, even in gcc -Wimplicit-fallthrough=5 at max level fallthrough warning doesn't flag KEY_HOME, KEY_LEFT, and KEY_BACKSPACE up at all. I think it's similar with MSVC also: https://learn.microsoft.com/en-us/cpp/code-quality/c26819?view=msvc-170 they even says it's fine to not mark them as fallthrough. I'd rather not mark fallthrough attribute for empty cases as they're not like the non-empty ones and looks like code noise if putting the fallthrough attribute on them.

Also see: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-break

src/game/client/neo/ui/neo_hud_killer_damage_info.cpp
nullsystem: The question marks seems like corrupt data, but it doesn't seem to make much sense why it happens for you?

AdamTadeusz: Ill have a look at it in the debugger see if I can sniff it out

src/game/client/neo/ui/neo_hud_killer_damage_info.cpp
nullsystem: I think we did this way in the past because the HUD elements were giving a darker tint to the entire screen? Either way just file this as something to take a look if doing all this really needed or not, but not in this PR.

src/game/client/neo/ui/neo_hud_killer_damage_info.cpp
nullsystem: Question marks also related to the usage of totals issue.

src/game/client/neo/ui/neo_hud_killer_damage_info.cpp
nullsystem: Took some of the suggestions to reduce the issue of text overflowing bounds in lower resolution, anyway it's intentional the HUD element take the entire screen so that NeoUI can positioning and sizing can do it relatively to the screen, as long as it's not tinting the entire screen I don't see the issue with that.

sunzenshen (approved):
LGTM.  Damage from Jeff isn't counted (blank results if you only took damage from non-player entities), but I think that kind of detail is optional.  Otherwise results look good.

nullsystem:
@AdamTadeusz @sunzenshen Will look at it separately: #1275

Issue #883, opened 2025-02-11
Weapon entities do not spawn as physics objects or replenish ammo


DESTROYGIRL:
When trying to add a weapon entity to a map, they can be picked up and used, however they have no physics and will stand upright and motionless. This isn't the case in OGNT
Edit: They don't replenish ammo either as they should.

AdamTadeusz:
same with weapons created with the ent_create command

PR #1195, opened 2025-06-25
Weapons vphysics when created with commands

AdamTadeusz:
## Description

Add vphysics to weapons on activate like how its done in fallthink when weapons are dropped by player

## Linked Issues

- related #883

Two issues in one


DESTROYGIRL (changes requested):
It doesn't seem to respect the SF_ITEM_START_CONSTRAINED spawnflag. This will be needed for the tutorial

DESTROYGIRL (approved):

Rainyan:
@DESTROYGIRL Could you please share some example map file and steps to test this constraint flag for the review?

PR #1197, opened 2025-06-26
JGR Gamemode

DESTROYGIRL:
## Description
Gamemode where players must hold control of a mech to win
neoAssets >>> https://github.com/NeotokyoRebuild/neoAssets/pull/68 <<<

## Toolchain
- Windows MSVC VS2022


DESTROYGIRL:
image

DESTROYGIRL:
I am going to move all the JGR functionality outside of ghost related functions and vars to save me from the inevitable wrath of the reviewers

src/game/shared/neo/weapons/weapon_knife.cpp
```diff
@@ -106,7 +106,7 @@ void CWeaponKnife::PrimaryAttack()
 
 bool CWeaponKnife::CanBePickedUpByClass(int classId)
 {
-	return classId != NEO_CLASS_SUPPORT;
+	return classId != (NEO_CLASS_SUPPORT || NEO_CLASS_JUGGERNAUT);
```
sunzenshen: I suspect this will not function as intended. In C++, the logical OR operator `||` on two non-zero integer values (enum constants here) will always evaluate to true. This means the condition will effectively become classId != true, which seems unlikely to be what you want.

To properly check if the classId is neither a Support nor a Juggernaut, you will need to check against both cases individually):

```cpp
(classId != NEO_CLASS_SUPPORT) && (classId != NEO_CLASS_JUGGERNAUT)
```

src/game/shared/gamemovement.cpp
```diff
@@ -2497,16 +2497,33 @@ bool CGameMovement::CheckJumpButton( void )
...
-		flMul = neoPlayer->GetClass() == NEO_CLASS_RECON ?
-			293.9387691339814f: // sqrt(2 * 800 * 54)
-			240.0f; 			// sqrt(2 * 800 * 36)
+		switch (neoPlayer->GetClass())
+		{
+			case NEO_CLASS_RECON:
+				flMul = 293.9387691339814f;	// sqrt(2 * 800 * 54)
+				break;
+			case NEO_CLASS_JUGGERNAUT:
+				flMul = 283.9718295887816f;	// sqrt(2 * 800 * 1.4) DG: did i do this correctly
```
sunzenshen: My result from python doesn't seem to match:

```python
>>> math.sqrt(2 * 800 * 1.4)
47.32863826479693
```

What is the intent of the 1.4 value?

src/game/shared/neo/neo_enums.h
```diff
@@ -17,6 +17,7 @@ enum NeoClass {
 	// NOTENOTE: VIP *must* be last, because we are
```
sunzenshen: This comment might need to be updated to clarify that the VIP or Juggernaut entries must be after the block of recon/assault/support, as the array offsets for the latter are significant for logic elsewhere.

src/game/shared/neo/neo_juggernaut.cpp
```diff
@@ -0,0 +1,241 @@
...
+	}
+	else
+	{
+		SetPlaybackRate(0.0f);
+	}
+
+	SetMoveType(MOVETYPE_STEP);
+	SetSolid(SOLID_BBOX);
+	UTIL_SetSize(this, -Vector(15, 15, 0), Vector(15, 15, 90)); // Needs to be equal or smaller than the player's girth to avoid getting stuck
+	SetCollisionGroup(COLLISION_GROUP_PLAYER);
```
sunzenshen: Maybe add a comment regarding why this isn't COLLISION_GROUP_WEAPON?

I'm guessing based on context that it's because the weapon is so big that it makes sense for players to collide with it when they try to walk through.

DESTROYGIRL: It's the opposite actually, players collide with it the same as they would other players, because it eventually becomes a player I think it's the behaviour people would expect. Also prevents getting stuck inside it

src/game/shared/gamemovement.cpp
DESTROYGIRL: Oh it was `36 * 1.4` aka default jump height by the multiplier
So `2 * 800 * 50.4` really

src/game/client/neo/ui/neo_hud_ammo.cpp
```diff
@@ -296,3 +302,30 @@ void CNEOHud_Ammo::DrawNeoHudElement()
 		DrawAmmo();
 	}
 }
+
+void CNEOHud_Ammo::DrawHeatMeter(C_NEOBaseCombatWeapon* activeWep) const
+{
+	float flHeatAmount = (1.0f - (activeWep->GetPrimaryAmmoCount() / (float)activeWep->GetDefaultClip1()));
```
AdamTadeusz, line 308: possible division by 0?

AdamTadeusz:
I'm assuming you're not supposed to be able to pick the ghost up by walking over it in a juggernaut, but you can still pick it up with the use key.

image

src/game/client/neo/c_neo_player.cpp
```diff
@@ -1350,14 +1352,34 @@ void C_NEO_Player::PostThink(void)
...
+		if (m_hDroppedJuggernautItem)
+		{
+			vTarget = m_hDroppedJuggernautItem->WorldSpaceCenter();
+		}
+		else
+		{
+			vTarget = NEORules()->GetJuggernautMarkerPos();
+
+		}
+		eyeOrigin = vTarget + Vector(80, 80, 80);
```
AdamTadeusz, line 1377: you could do a traceline here to make sure the camera never goes out of bounds, if that's something you care about.

Something like
```

		eyeOrigin = vTarget + Vector(80, 80, 80);
		trace_t tr;
		CTraceFilterWorldOnly traceFilter;
		UTIL_TraceLine(vTarget, eyeOrigin, MASK_OPAQUE, &traceFilter, &tr);
		eyeOrigin = vTarget + ((Vector(75, 75, 75) * tr.fraction));
```
for example

src/game/client/neo/ui/neo_hud_ghost_marker.cpp
```diff
@@ -83,126 +83,204 @@ void CNEOHud_GhostMarker::ApplySchemeSettings(vgui::IScheme *pScheme)
 extern ConVar cl_neo_hud_worldpos_verbose;
 void CNEOHud_GhostMarker::UpdateStateForNeoHudElementDraw()
```
AdamTadeusz, line 84: Can this function be refactored to use more general terms like NEORules()->ObjectiveExists(), NEORuleS()->GetObjectivePos(), m_objectiveInPVS? Theres a lot of repeating code here, we're basically doing the same thing regardless of whether we're tracking the juggernaut or the ghost, only the text differs

src/game/client/neo/ui/neo_hud_ghost_marker.cpp
```diff
@@ -83,126 +83,204 @@ void CNEOHud_GhostMarker::ApplySchemeSettings(vgui::IScheme *pScheme)
...
 	}
 }
 
 void CNEOHud_GhostMarker::resetHUDState()
 {
 	m_ghostInPVS = nullptr;
+	m_jgrInPVS = nullptr;
 }
 
 void CNEOHud_GhostMarker::DrawNeoHudElement()
```
AdamTadeusz, line 134: Same as above. I guess for the ghost we are changing the colour and hiding the distance text as well but that can be a simple check nested deeper inside more general code related to the game objective

src/game/client/neo/ui/neo_hud_ghost_marker.h
```diff
@@ -30,6 +32,7 @@ class CNEOHud_GhostMarker : public CNEOHud_WorldPosMarker
 
 private:
 	C_WeaponGhost *m_ghostInPVS = nullptr;
+	CNEO_Juggernaut *m_jgrInPVS = nullptr;
```
AdamTadeusz, line 35: We don't call any special functions on this pointer, can probably have a single m_pObjectiveInPVS pointer of type CBaseAnimating, maybe @sunzenshen can confirm?

I guess neo_hud_ghost_marker could also me renamed to neo_hud_objective_marker in that case

src/game/client/neo/ui/neo_hud_round_state.cpp
```diff
@@ -373,6 +373,9 @@ void CNEOHud_RoundState::UpdateStateForNeoHudElementDraw()
 			}
 		}
 		break;
+	case NEO_GAME_TYPE_JGR:
+		V_sprintf_safe(szGameTypeDescription, "Keep control of the Juggernaut\n");
```
AdamTadeusz: I was trying to stick to short 3 word objectives where possible, like in Warhammer 40K Space Marine, can we put "Control the Juggernaut" here for consistency?

src/game/server/neo/neo_player.h
```diff
@@ -73,8 +73,10 @@ class CNEO_Player : public CHL2MP_Player
 	virtual void ChangeTeam(int iTeam) OVERRIDE;
 	virtual void PickupObject(CBaseEntity *pObject, bool bLimitMassAndSize) OVERRIDE;
 	virtual void PlayStepSound(Vector &vecOrigin, surfacedata_t *psurface, float fvol, bool force) OVERRIDE;
+	virtual const char* GetOverrideStepSound(const char* pBaseStepSound) OVERRIDE;
```
AdamTadeusz: doesn't need to be virtual

DESTROYGIRL:
> I notice there is a problem where the juggernaut struggles to step up into tight spaces where its bounding box is small enough to let it fit in there, requiring a jump forward move instead

This is more of a map design thing. CTG maps aren't made to accomidate the big boy. In the JGR maps the spaces are smoothed out to avoid situations like this

AdamTadeusz:
well the doorway is more than large enough for the jug to fit so ideally the jug should fit

DESTROYGIRL:
If you have a juggernaut specific solution we could include it here

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -228,6 +231,9 @@ BEGIN_NETWORK_TABLE_NOBASE( CNEORules, DT_NEORules )
 	RecvPropInt(RECVINFO(m_iEscortingTeam)),
 	RecvPropBool(RECVINFO(m_bGhostExists)),
 	RecvPropVector(RECVINFO(m_vecGhostMarkerPos)),
+	RecvPropInt(RECVINFO(m_iJuggernautPlayerIndex)),
+	RecvPropBool(RECVINFO(m_bJuggernautItemExists)),
+	RecvPropVector(RECVINFO(m_vecJuggernautMarkerPos)),
```
AdamTadeusz, line 236: reuse ghost player index and pos? (m_iObjectivePlayerIndex etc)

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -1709,6 +1784,118 @@ void CNEORules::SpawnTheGhost(const Vector *origin)
 			m_vecPreviousGhostSpawn.z);
 }
 
+// Very similar to above.
```
AdamTadeusz, line 1787: SpawnObjective instead and combine the two?

AdamTadeusz:
> I'm assuming you're not supposed to be able to pick the ghost up by walking over it in a juggernaut, but you can still pick it up with the use key.
> image

proabbly a case of editing CNEO_Player::Weapon_CanSwitchTo to return false if player class is JGR, assuming that doesn't break the initial weapon switch to the balc

src/game/client/neo/ui/neo_hud_ammo.cpp
DESTROYGIRL: I guess if people mess up the weapon scripts but is it necessary to guard against that?

src/game/client/neo/ui/neo_hud_ghost_marker.cpp
DESTROYGIRL: Initially it was like this but I was told to change it lol...

DESTROYGIRL:
> proabbly a case of editing CNEO_Player::Weapon_CanSwitchTo to return false if player class is JGR, assuming that doesn't break the initial weapon switch to the balc

Where is the unique +use pick up functionality handled? I'd prefer if we can test against `CanBePickedUpByClass()` on the weapon itself

src/game/client/neo/ui/neo_hud_ghost_marker.cpp
DESTROYGIRL: I'm going to generalise some vars but functions in neo_gamerules will probably remain seperate because they don't all work the same between the gamemodes.

DESTROYGIRL: Changed my mind

src/game/server/neo/neo_player.h
```diff
@@ -73,8 +73,10 @@ class CNEO_Player : public CHL2MP_Player
 	virtual void ChangeTeam(int iTeam) OVERRIDE;
 	virtual void PickupObject(CBaseEntity *pObject, bool bLimitMassAndSize) OVERRIDE;
 	virtual void PlayStepSound(Vector &vecOrigin, surfacedata_t *psurface, float fvol, bool force) OVERRIDE;
+	const char* GetOverrideStepSound(const char* pBaseStepSound) OVERRIDE;
```
AdamTadeusz: we're also using the lowercase override keyword now, same in c_neo_player.h

src/game/client/neo/c_neo_player.h
```diff
@@ -66,6 +66,7 @@ class C_NEO_Player : public C_HL2MP_Player
 	virtual bool ShouldReceiveProjectedTextures( int flags );
 	virtual void PostDataUpdate( DataUpdateType_t updateType );
 	virtual void PlayStepSound( Vector &vecOrigin, surfacedata_t *psurface, float fvol, bool force );
+	const char* GetOverrideStepSound(const char* pBaseStepSound) OVERRIDE;
```
AdamTadeusz: override

src/game/client/neo/ui/neo_hud_ammo.cpp
AdamTadeusz: maybe not

AdamTadeusz:
> > proabbly a case of editing CNEO_Player::Weapon_CanSwitchTo to return false if player class is JGR, assuming that doesn't break the initial weapon switch to the balc
> 
> Where is the unique +use pick up functionality handled? I'd prefer if we can test against `CanBePickedUpByClass()` on the weapon itself

well CanBePickedUpByClass doesn't get called in CHL2_Player::PlayerUse but you can add it in there too probably, especially if its a method unique to neoplayer as that is least likely to break anything

image

DESTROYGIRL:
@AdamTadeusz I decided to move the bit that handles the weapon +use switch in `CHL2_Player::PlayerUse` to `CNEOBaseCombatWeapon::Use` instead because that gets called on our weapons anyway and it makes more sense to me to handle it in one continuity. Good idea?

AdamTadeusz:
This seems like a neater solution than what it was originally. make sure you remove the ifdefs NEO and ifndef NEOs that are now redundant.

Only thing that worries me is that now we're bumping and trying to pickup a weapon that we potentially previously equipped, but that doesn't seem to be the problem with the only weapon that currently can be used, i.e the ghost. Also funny things happen if you comment some of the code out and also enable hl2mp_allow_pickup

https://github.com/user-attachments/assets/7b5731aa-4b1d-4658-b166-85208ae0910f

src/game/server/hl2/hl2_player.cpp
```diff
@@ -3250,21 +3250,21 @@ void CHL2_Player::PlayerUse ( void )
...
-#endif // NEO
+//#endif
 				{
 					Weapon_DropSlot( pWeapon->GetSlot() );
 					Weapon_Equip( pWeapon );
-#ifdef NEO
+//#ifdef NEO
 					pWeapon->RemoveEffects(EF_BONEMERGE);
-#endif // NEO
+//#endif
```
AdamTadeusz:
```suggestion
				//Try to take ammo or swap the weapon
				if ( Weapon_OwnsThisType( pWeapon->GetClassname(), pWeapon->GetSubType() ) )
				{
					Weapon_EquipAmmoOnly( pWeapon );
				}
				else
				{
					Weapon_DropSlot( pWeapon->GetSlot() );
					Weapon_Equip( pWeapon );
```

AdamTadeusz: github multi line suggestions confuse the hell out of me but I think that looks right?

DESTROYGIRL: yolo

DESTROYGIRL: Okay didn't work

AdamTadeusz: damn didnt know you could commit these directly

AdamTadeusz (approved):

sunzenshen (approved):

src/game/server/neo/neo_player.cpp
```diff
@@ -3259,3 +3290,100 @@ void CNEO_Player::ModifyFireBulletsDamage(CTakeDamageInfo* dmgInfo)
 {
 	dmgInfo->SetDamage(dmgInfo->GetDamage() * sv_neo_wep_dmg_modifier.GetFloat());
 }
+
+void CNEO_Player::BecomeJuggernaut()
+{
+	NEORules()->JuggernautActivated(this);
+	if (m_iNextSpawnClassChoice = -1)
```
sunzenshen, line 3297: Doh, I wonder if we accidentally missed this comparison.  Is this supposed to override `m_iNextSpawnClassChoice` or just to compare it against -1?

DESTROYGIRL: Will be fixed in #1378

PR #1198, opened 2025-06-27
Fix #1165

### LINKED ISSUE #1165: Sprinting as recon uses aux power in some situations
Getting blocked by objects, especially physics objects can cause aux power to start going down as long as the player remains sprinting

https://github.com/user-attachments/assets/037b6f48-12b7-4fa7-8684-dc2b0009f077

Rainyan:
## Description
Fix the bug with recon sprint costing AUX.

It seems the AUX draining stems from an anti-exploit logic in the engine base code regarding sprinting. Since recons have infinite sprint, we can just skip that part for them to avoid the problem.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- Fixes #1165

### DIFF
--- a/src/game/client/hl2mp/c_hl2mp_player.cpp
+++ b/src/game/client/hl2mp/c_hl2mp_player.cpp
@@ -551,12 +551,22 @@ bool C_HL2MP_Player::SuitPower_RemoveDevice( const CSuitPowerDevice &device )
 	if( !IsSuitEquipped() )
 		return false;
 
+#ifdef NEO
+	// Recons can spam sprint whenever and it doesn't matter.
+	// Fixes https://github.com/NeotokyoRebuild/neo/issues/1165
+	// where this could drain AUX for recons stuck in geometry whilst attempting to sprint.
+	const bool isRecon = static_cast<CNEO_Player*>(this)->GetClass() == NEO_CLASS_RECON;
+	const float antiExploitDrainRate = isRecon ? 0 : device.GetDeviceDrainRate() * 0.1f;
+#else
+	const float antiExploitDrainRate = device.GetDeviceDrainRate() * 0.1f;
+#endif
+
 	// Take a little bit of suit power when you disable a device. If the device is shutting off
 	// because the battery is drained, no harm done, the battery charge cannot go below 0. 
 	// This code in combination with the delay before the suit can start recharging are a defense
 	// against exploits where the player could rapidly tap sprint and never run out of power.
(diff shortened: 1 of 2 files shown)


src/game/client/hl2mp/c_hl2mp_player.cpp
```diff
@@ -551,12 +551,22 @@ bool C_HL2MP_Player::SuitPower_RemoveDevice( const CSuitPowerDevice &device )
...
 		return false;
 
+#ifdef NEO
+	// Recons can spam sprint whenever and it doesn't matter.
+	// Fixes https://github.com/NeotokyoRebuild/neo/issues/1165
+	// where this could drain AUX for recons stuck in geometry whilst attempting to sprint.
+	const bool isRecon = static_cast<CNEO_Player*>(this)->GetClass() == NEO_CLASS_RECON;
+	const float antiExploitDrainRate = isRecon ? 0 : device.GetDeviceDrainRate() * 0.1f;
+#else
+	const float antiExploitDrainRate = device.GetDeviceDrainRate() * 0.1f;
```
AdamTadeusz: technically this is changing hl2mp code outside an ifdef NEO block, perhaps do the check below, that way the original comment which now isn't entirely factual can also be amended?

AdamTadeusz: then we also can just not call device.GetDeviceDrainRate, I don't see why we should for recons unless that clamp is not present elsewhere where power is drained

Rainyan: Well, I was feeling adventurous so I changed it to a goto that jumps over the offending block entirely for recons. We do probably wanna keep hitting that code for assaults though, so that they can't tamper with their AUX sprint.

AdamTadeusz: whoa

this is like so radical

AdamTadeusz (approved):

src/game/client/hl2mp/c_hl2mp_player.cpp
AdamTadeusz: I'm not sure if Agiel would approve of all this code duplication

Rainyan: It's unideal that we're doing the same dance on both cli/srv, but I figured it's outside the scope for this to start refactoring entirely here.

Masterkatze (approved):

PR #1204, opened 2025-07-05
remove weapons and gibs after some time on maps with respawns enabled

AdamTadeusz:


Rainyan (dismissed):

src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
```diff
@@ -1262,9 +1263,10 @@ void CNEOBaseCombatWeapon::SetPickupTouch(void)
 		return;
 	}
 
-	if (!weaponstay.GetBool())
-	{
-		BaseClass::SetPickupTouch();
+	if ((!weaponstay.GetBool() || NEORules()->RespawnsEnabled()) && GetSpawnFlags() & SF_NORESPAWN)
+	{ // regardless of the value of mp_weaponstay, disappear weapons in game modes with respawns enabled. Otherwise things can get too chaotic
+		SetThink(&CBaseEntity::SUB_Remove); // NEO TODO (Adam) roll our own sub_remove to check for players in pvs?
```
Rainyan: I didn't test this, but there's also https://github.com/NeotokyoRebuild/neo/blob/da15e5bcfe2105c0b3e7257ce5e5551fa22ea503/src/game/server/baseentity.cpp#L8183-L8186 which might be appropriate for doing this

src/game/shared/neo/neo_gamerules.h
```diff
@@ -198,8 +198,8 @@ class CNEORules : public CHL2MPRules, public CGameEventListener
 	int GetForcedSkin();
 	int GetForcedWeapon();
 	virtual const char* GetGameTypeName(void) OVERRIDE;
-	const bool RespawnsEnabled();
-	bool CanChangeTeamClassWeaponWhenAlive();
+	virtual const bool CanChangeTeamClassLoadoutWhenAlive();
+	virtual const bool CanRespawnAnyTime();
```
Rainyan: Since this is the base implementation, and we probably won't have any class overriding the NEO gamerules, I think these two methods could be declared non-virtual, for perf.

Rainyan (approved):

src/game/shared/neo/neo_gamerules.h
AdamTadeusz: and the const keyword is unnecessary too yes?

Rainyan: > and the const keyword is unnecessary too yes?

Yeah, since it's returned by value instead of by reference, I would personally drop it from the retval

sunzenshen (approved):

PR #1208, opened 2025-07-07
Remove references to gamestartup1, use snd_victory_volume for jingle when changing volumes during jingle

AdamTadeusz:


game/neo/sound/playlist/default.txt
```diff
@@ -4,6 +4,6 @@ playlist
   song
   {
     fromgame 1
-    relativepath "ui\gamestartup1.mp3"
+    relativepath "..\soundtrack\112 - out.mp3"
   }
```
Rainyan, line 8: The MP3 player doesn't seem to read this, but instead picks a random song at: https://github.com/NeotokyoRebuild/neo/blob/8c50153ade38979542af9f0e2dfdb90935672809/src/game/client/mp3player.cpp#L907

Maybe we could instead read this file with something like
```cpp
// pseudocode - missing nullptr checks etc

auto kv = new KeyValues("playlist");
kv->LoadFromFile(g_pFullFileSystem, "sound/playlist/default.txt");
auto song = kv->FindKey("song");
auto songPath = song->GetString("relativepath");
kv->deleteThis();
// PlaySongByPath(songPath); // assuming we had some method for this
```

so that the intro song is actually always "Out", and not chosen randomly.

AdamTadeusz: The default playlist is, unintuitively, not the default playlist currently. Selecting playlists is a bit tricky right now and so I made the playlist with all of the songs in ognt's files the default. When you load the playlist named default the list of songs should look something like this. Is this what it looks like on your end? I'm testing it right now and out is the only song that plays

image

AdamTadeusz: to select a playlist you have to right click on the Playlist tab or on the area where the list of songs is, select "Load playlist..." from the options menu that pops up, navigate to the sound/playlist folder and select "default.txt"

Rainyan: What I mean is, with current behaviour in this PR, the song that starts playing when you start the game is chosen at random. Parity would expect the main menu to start with the song "Out". Unless that's something we are looking to change?

I'm not talking about the player behaviour itself, just what the player hears when they launch the game.

AdamTadeusz: Technically the song that starts playing when you start the game currently is a random song from the playlist selected in mp3settings.txt, which is the last playlist that was loaded. I assumed in the same way that you wouldn't want the crosshair to change back to the default parity one on game restart you also wouldn't want the sound played on game start to revert, but since the mp3 player can also be used in game I guess we could add an option to select what song exactly plays when you return to the menu. At least a setting for the index should be easy, but once people start making their own we would need to load out separately to the selected playlist

AdamTadeusz: the addition of the default playlist was for the nt purists, just load that as your last playlist and pause playing of music in game and you've achieved menu music parity

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -149,7 +152,56 @@ ConVar sv_neo_comp("sv_neo_comp", "0", FCVAR_REPLICATED, "Enables competitive ga
...
+			if (Q_strcmp(filename, jingles[j]) == 0)
+			{
+				enginesound->SetVolumeByGuid(sounds[i].m_nGuid, snd_victory_volume.GetFloat());
+				return; // should only ever be playing one jingle at a time; return early
+			}
+		}
+	}
+}
+
+ConVar snd_victory_volume("snd_victory_volume", "0.33", FCVAR_ARCHIVE | FCVAR_DONTRECORD | FCVAR_USERINFO, "Loudness of the victory jingle (0-1).", true, 0.0, true, 1.0, sndVictoryVolumeChangeCallback);
```
Rainyan, line 205: We probably don't need to have this in the userinfo, so could clear that cvar flag while we're at it.

game/neo/sound/playlist/default.txt
Rainyan: I'll resolve this for now, since it's not worth blocking the PR over. Perhaps in the future we could add an options menu item for choosing the menu theme, defaulting to parity at least until we decided otherwise.

src/game/shared/neo/neo_gamerules.cpp
AdamTadeusz: Currently the server plays the victory jingle for all clients, lines 1457 and 3161 in neo_gamerules.cpp the server is getting this client convar value. Without this flag we can't get the correct value

Rainyan (dismissed):

sunzenshen (dismissed):

sunzenshen (approved):

PR #1209, opened 2025-07-07
MP3 Player sounds don't use snd_musicvolume

AdamTadeusz:
## Description

Setting the volume of a music sound to 0 pauses that sound, thus muting in the mp3 player is notpossible in the traditional sense and instead the music is just turned down to the minimum possible value. This value changes depending on the value of snd_musicvolume, this PR uses a neat little workaround so that snd_musicvolume is ignored by sounds played through the mp3 player but the other advantages of the "#" sound character are still present.

Also I noticed the volume slider convar values were not being updated unless the slider that was currently affecting mp3 playback was moved (so changing the in game slider only when in the main menu and then quitting the game would not save that change), so I changed those values to be updated when the slider posts an action signal once the user stops dragging the slider

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #
- related #


src/public/soundflags.h
```diff
@@ -28,6 +28,9 @@ enum
 	CHAN_STATIC		= 6,		// allocate channel from the static area 
 	CHAN_VOICE2		= 7,
 	CHAN_VOICE_BASE	= 8,		// allocate channel for network voice data
+#ifdef NEO
+	CHAN_MP3_PLAYER = 9,
+#endif // NEO
```
Rainyan: The sound engine allocates 5 channels (I think) for network voice data from `CHAN_VOICE_BASE` (8) up to and including channel 8+4=12. So this would be fighting with the second voice channel when multiple players are talking at the same time.

Since this enum is effectively a contract with the closed-source engine sound library, called at: https://github.com/NeotokyoRebuild/neo/blob/8f0fee0b0e4760af3851afb856815de9db2a9250/src/game/shared/SoundEmitterSystem.cpp#L588 I think we should define our custom channels inside the valid `CHAN_USER_BASE` range as commented at: https://github.com/NeotokyoRebuild/neo/blob/8f0fee0b0e4760af3851afb856815de9db2a9250/src/public/soundflags.h#L31

It seems we're already using `CHAN_USER_BASE` for the ghost voice lines channel at: https://github.com/NeotokyoRebuild/neo/blob/8f0fee0b0e4760af3851afb856815de9db2a9250/src/game/shared/neo/weapons/weapon_ghost.cpp#L186 so how about we define `CHAN_MP3_PLAYER` as `CHAN_USER_BASE + 1` instead?

Might not be a bad idea to also refactor that ghost channel definition here in this enum (as `CHAN_GHOST_PICKUP` or something, instead of calling `CHAN_USER_BASE` as a magic value) while we're at it, to keep things tidy.

AdamTadeusz: I read that and understood it to mean that values below CHAN_USER_BASE were free to use and values above were used somewhere in the engine, my bad

Rainyan (approved):

src/public/soundflags.h
Rainyan: Yeah it's a bit confusing. I'm guessing they've reserved those 128 (minus the voice chans) for future use, and that's why the number jumps.

AdamTadeusz:
I added a change callback to the convars to set the slider values

sunzenshen (approved):
I'm not sure that the Mute checkbox is intuitive since on my machine the music just plays at minimum volume, but at least the cvars now update the sliders as expected.

src/game/client/mp3player.cpp
```diff
@@ -2462,6 +2481,34 @@ void CMP3Player::OnSliderMoved()
...
+		return;
+	}
+	const float newValue = pData->GetInt() * 0.01f;
+
+	pData = pData->GetNextKey();
+	if (!pData)
+	{
+		return;
+	}
+	const void* pszName = pData->GetPtr();
```
Rainyan: Unused local variable. I haven't looked at the kv implementation but if we need to do this to walk the tree, then could we mark it `[[maybe_unused]]` to silence the compiler warning?

And actually, since it's *not* marked `nodiscard` in the keyvalues method, we could just call it and discard the return value by not assigning it, as:
```diff
-       const void* pszName = pData->GetPtr();
+       pData->GetPtr();
```

Rainyan (changes requested):

src/game/client/mp3player.cpp
AdamTadeusz: I was comparing names to figure out which convar to use, forgot to remove.

AdamTadeusz: ok i got confused for a second but yeah GetNextKey and GetSubKey for example are used to traverse the tree, this just reads the value without moving

PR #1210, opened 2025-07-08
Start saving DB file again, stop DB file from growing linearly with every game exit

### LINKED ISSUE #1199: restore MP3 db saving functionality
Valve's mp3 player saves a db file with paths to all the songs that were loaded. Currently this functionality has trouble distinguishing between songs that are already in the db file and new songs, causing the entirety of the currently selected playlist to be appended to the db file every time the file is saved, meaning the file will grow in size linearly forever, and as such #947 skips writing of this file.

Note I'm not certain this bug was present in the player originally, could be a problem with using relative filepaths in the playlists or also related to additional semicolons that are now added to the playlist file which were preventing generated playlists from being read correctly on subsequent game launches / playlist loads

AdamTadeusz:
## Description

Not sure if I'm happy with this to be honest, loading the db file loads every single song that was ever played or loaded as a playlist through the mp3 player, perhaps it would make more sense to remove this db functionality altogether. I guess we can open another PR to remove this properly, but it might still be a good idea to have a commit on master where it was working.

## Linked Issues

- fixes #1199


Rainyan (approved):
The UX could brush-up for sure, it's very confusing to navigate. Some items you have to right click but there's no context hint about it etc. And I'm not gonna pretend I understand how these playlists are supposed to be used. But it's good to get this playlist saving logic in as you said, for future iteration.

Rainyan (changes requested):
Actually, after monkeying around with the player settings, I'm seeing staged files in git:
```bash
On branch 250_mp3Database
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   game/neo/resource/mp3settings.txt
        modified:   game/neo/sound/playlist/soundtrack.txt
```
Maybe we should gitignore these, so that devs don't accidentally push their Spice Girls playlist upstream.

Masterkatze (approved):

PR #1212, opened 2025-07-10
Fix main menu maxplayers slider value

Rainyan:
## Description
Since the x64 update 2fdf6fb2, the engine's `MAX_PLAYERS` limit (for x64) has increased from 33 to 101. Update the main menu's server creation panel slider value to use this definition, instead of hardcoding 32 as the upper bound.

We're using `MAX_PLAYERS-1` because as per
https://github.com/NeotokyoRebuild/neo/blob/8f0fee0b0e4760af3851afb856815de9db2a9250/src/game/shared/shareddefs.h#L285-L287 the maximum gamerules allowed players should be <MAX_PLAYERS to make room for the SourceTV/replay bot on the last allocated slot.

## Toolchain
- Windows MSVC VS2022


Rainyan:
Although we might want to revisit this in some form of gamerules-driven maxplayers restrictions, for example restricting CTG mode to parity at 32. Or even lower. But for now, this PR just makes it possible to set the engine-supported maxplayers via the GUI interface, since it could already be done with console commands.

PR #1216, opened 2025-07-10
Add neo_restart_this support

Rainyan:
## Description
Add the `neo_restart_this` cvar for parity.

This cvar differs from `mp_restartgame` in that the latter will give a countdown, whereas `neo_restart_this` is instant.

As a side note, the cvar `mp_restartgame_immediate` seen below is `CTeamplayRoundBasedRules` related, which is not part of our gamerules inheritance chain, and its logic is not quite right, so that's why I'm rolling my own callback for the cvar.

## Toolchain
- Windows MSVC VS2022


AdamTadeusz:
Can't come up with a good way to replicate it it seems very random

sunzenshen (approved):
LGTM.  If anything, the real issue needs to be debugged from RestartGame() if that was related to the bug, rather than from this wrapper.

PR #1218, opened 2025-07-12
Spray handling improvements - Freezetime, disabled spray, cleanups

### LINKED ISSUE #1205: Players cannot use sprays during pre-round freezetime
title

nullsystem:
## Description
* Allow impulse 201 (spray) in freezetime
* Prevent impulse 201 if the client disabled spray
* Delete custom downloaded sprays on both startup and shutdown
    * Both startup and shutdown mostly to workaround filesystem API bug trying to recognize some of the files
* Test on `*.dat` sprays deletion on Windows (rainyan tested)

## Toolchain

## Linked Issues

* fixes #1205

### DIFF
--- a/src/game/client/cdll_client_int.cpp
+++ b/src/game/client/cdll_client_int.cpp
@@ -187,11 +187,13 @@ bool g_bOBSDetected = false;
 #include <tchar.h>
 #include <psapi.h>
 #include <tlhelp32.h>
+#include <direct.h>
 #undef CreateEvent
 #endif
 
 #ifdef LINUX
 #include "neo_fixup_glshaders.h"
+#include <unistd.h>
 #endif
 
 #endif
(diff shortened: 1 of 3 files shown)


src/game/client/cdll_client_int.cpp
```diff
@@ -1303,6 +1303,75 @@ static void NeoClantag_ChangeCallback(IConVar *cvar, [[maybe_unused]] const char
...
+			const bool bIsFile = !filesystem->FindIsDirectory(findHdlSL);
+			if (!bIsExpectedFNameSize || !bIsFile)
+			{
+				continue;
+			}
+
+			bool bFNameIsHexDigit = true;
+			for (int i = 0; i < BASE_FNAMESIZE; ++i)
+			{
+				bFNameIsHexDigit = bFNameIsHexDigit && V_isxdigit(pszFNameSLDat[i]);
```
Rainyan: Nit: could move `bFNameIsHexDigit &&` into the loop condition, so that we break as soon as that stops being true.

src/game/client/cdll_client_int.cpp
Rainyan, line 1: This seems to work fine (tested the .dat removal on Windows), but I wonder about not removing the folders.

Based on: https://github.com/NeotokyoRebuild/neo/blob/23343df7f2033b9bd3a93fa5acfda7b7eae64702/src/game/client/c_baseplayer.cpp#L2477-L2478 it seems the folders are named as pruned 2-char hex values of the content CRC hash, so I guess the worst-case amount of leftover folders (file handles) we'd have to iterate through is 16^2 or 256 folders, so it's not too bad. But it would be nice if we could delete these too, to avoid walking a bunch of empty folders every time the game closes or opens. Since the longer you had the game installed, the more of these folders would begin to accumulate.

I couldn't find any API in the engine to delete folders, but `rmdir` (and `_rmdir` for Windows) seems like it could be relatively safe to call here, since from what I gather both for Linux and Windows it will fail gracefully if attempting to delete a non-empty folder.

https://man7.org/linux/man-pages/man2/rmdir.2.html
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/rmdir-wrmdir?view=msvc-170

Assuming we can safely call libc here.

src/game/server/player.cpp
```diff
@@ -3887,19 +3887,35 @@ void CBasePlayer::PlayerRunCommand(CUserCmd *ucmd, IMoveHelper *moveHelper)
...
 		ucmd->forwardmove = 0;
 		ucmd->sidemove = 0;
 		ucmd->upmove = 0;
-		ucmd->impulse = 0;
-
 #ifdef NEO
-		if (!originalCheck && static_cast<CNEO_Player*>(this)->GetNeoFlags() & NEO_FL_FREEZETIME)
+		if (!originalCheck && bPlayerInFreezeTime)
 		{
+			// Only allow spraying impulse in freeze time
```
Rainyan: Is spraying only supposed to be possible during the freezetime? With this PR, I can keep on pressing my `impulse 201` bind to spray even during the rounds. Tested on `ntre_oilstain_ctg` with some bots spawned to make the round start.

nullsystem: I do intend to allow spraying during rounds. "Only" was referring to impulse rather than freeze time but I guess easily mis-interpret.

src/game/client/cdll_client_int.cpp
nullsystem: I was restricting myself to the SDK API hence not doing directory deletion but guess you're fine with `rmdir`, I'll do that also then.

Rainyan: Yeah, I guess no way around it. As long as the header includes don't overwrite any macros etc needed for the file, it should be fine probably.

nullsystem: Did a quick tests on Linux + Windows and seems rmdir works out fine.

PR #1224, opened 2025-07-23
Bot thermoptic and combat behaviors

sunzenshen:
## Description
Improvements to bot utilization of thermoptic and other related combat behavior.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related https://github.com/NeotokyoRebuild/neo/pull/1185


src/game/server/NextBot/NextBotVisionInterface.h
```diff
@@ -117,6 +117,7 @@ class IVision : public INextBotComponent
 	CUtlVector< CKnownEntity > m_knownEntityVector;		// the set of enemies/friends we are aware of
 	void UpdateKnownEntities( void );
 	bool IsAwareOf( const CKnownEntity &known ) const;	// return true if our reaction time has passed for this entity
+	bool IsPerceptiveOf( const CKnownEntity &known ) const;	// return true if can detect despite obscuring factors
```
Rainyan: Having both `IsAwareOf` and `IsPerceptiveOf` feels (at least a little) like we're overloading the same meaning into two separate methods. Also, I'm not sure if we'd want to write NT-specific implementation inside the base `IVision` interface here; instead we could override `::IsAwareOf` behaviour inside our NT-specific `class CNEOBotVision : public IVision` implementation class to keep things neat.

That being said, in this specific case of handling vision impairment due to stuff like smoke grenades, therm-optic cloaking, etc, I think we might be better off actually implementing that in our derived overrides of these base combat character methods (the `CBaseEntity* target` ones specifically for cloaking, although perhaps the pos vector ones could be useful for determining in-smoke positions?), instead of in the bot code: https://github.com/NeotokyoRebuild/neo/blob/d1112c451c706f72a4b7850a1430c283d2733f0d/src/game/server/basecombatcharacter.h#L176-L184

Admittedly, smoke grenade puffs aren't exactly "fog", and thermoptic cloaking definitely isn't fog, but these virtual base combat character methods essentially determine how visible an enemy should be due to these kinds of environmental factors, so I think it would be a good fit. So maybe it would be cleanest if we refactored the thermoptic (and smoke puff, eventually) detection code into a virtual override of these methods for our `CNEOPlayer::IsHiddenByFog` etc implementations?

These *"is hidden/is obscured by..."* methods appear to be natively used by the bots `IVision` interface, so returning a boolean value from the `IsHiddenByFog` methods, and a unit interval between 0-1 from the `GetFogObscuredRatio` methods, such that the bot cannot see the player when appropriate, will correctly prevent the bot from detecting such a player.

TL;DR: if you could refactor so that instead of implementing a new `IVision::IsPerceptiveOf`, we instead overrode the above-mentioned methods in `CNEOPlayer` and returned the bool/float values accordingly for whether the player should be considered obscured, such that those new overrides then drove the base `IVision` detection logic instead, that'd be great. Or alternatively, if you feel a `CNEOBotVision` override is more appropriate, that would seem fine also. But if you have questions/feel differently, do feel free to ask either here or in Discord of course.

Also thanks for the PR & warm welcome to the dev side of things 😄

sunzenshen: Before proceeding with the code review, I wanted to check if this is what you mean by refactoring IsPerceptiveOf into the fog calculation code:
https://github.com/NeotokyoRebuild/neo/pull/1224/files#diff-c191eb19665428fbcf1e7de1cb583b54a586efe5c432f6eeb2309743c7d96050R1203

I implemented this refactor in 3 phases:
- First I created a direct override copy of the original routines: https://github.com/NeotokyoRebuild/neo/pull/1224/commits/69c42683bcefc660335917f1e5676b46fc8c42e0
- Then I frankly used AI to copy over the IsPerceptiveOf code, by instructing it on the differences, such as how myPlayer is the same is `this` in this new context https://github.com/NeotokyoRebuild/neo/pull/1224/commits/740e7a0b30d63880cbe1e24d3f84823d379b1f97
- Then I deleted the original IsPerceptiveOf code https://github.com/NeotokyoRebuild/neo/pull/1224/commits/0b19822d52e3769b7af3daa522998f6b9f11070d

At least from casual testing, seems like the bot code works the same.

src/game/server/neo/bot/behavior/neo_bot_attack.cpp
```diff
@@ -41,6 +41,14 @@ ActionResult< CNEOBot >	CNEOBotAttack::Update( CNEOBot *me, float interval )
 
 	CNEOBaseCombatWeapon* myWeapon = static_cast<CNEOBaseCombatWeapon* >( me->GetActiveWeapon() );
 	bool isUsingCloseRangeWeapon = me->IsCloseRange( myWeapon );
+
+	bool isThreatSupport = (threat && (static_cast<CNEO_Player*>(threat->GetEntity())->GetClass() == NEO_CLASS_SUPPORT));
```
Rainyan: While currently I think we only recognize players as threats, I imagine that could change at some point with things like the JGR mode, or some other types of threats like turrets. So it might be prudent to use the `ToNEOPlayer(threat->GetEntity())` here instead, since it does a safe conversion by checking whether the entity is actually a player.

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -983,6 +983,35 @@ bool CNEOBot::IsAmmoFull(void) const
 	return isPrimaryFull && isSecondaryFull;
 }
 
+bool CNEOBot::IsCloakEnabled(void) const
+{
+	auto myBody = GetBodyInterface();
+	return myBody ? myBody->IsCloakEnabled() : false;
```
Rainyan: Since this already returns a boolean, perhaps simplify to:
```cpp
return myBody && myBody->IsCloakEnabled();
```

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -983,6 +983,35 @@ bool CNEOBot::IsAmmoFull(void) const
...
+
+float CNEOBot::GetCloakPower(void) const
+{
+	auto myBody = GetBodyInterface();
+	return myBody ? myBody->GetCloakPower() : 0.0f;
+}
+
+void CNEOBot::EnableCloak(float threshold)
+{
+	if ( (GetCloakPower() > threshold)
```
Rainyan: Should we instead compare `>=`, because otherwise someone who wants to enable the cloak only at full charge would never be able to do so?

src/game/server/neo/neo_player.cpp
```diff
@@ -1144,6 +1149,212 @@ void CNEO_Player::CheckThermOpticButtons()
...
+		if ( range >= fog.end )
+			return 1.0f;
+
+		float ratio = (range - fog.start) / (fog.end - fog.start);
+		ratio = MIN( ratio, fog.maxdensity );
+		return ratio;
+	*/
+	return 0.0f;
+}
+
```
Rainyan: This implementation looks identical to `CBaseCombatCharacter::GetFogObscuredRatio`, so preferably we should not declare this specific override at all and let the base class handle it, to keep the code neat.

src/game/server/neo/neo_player.cpp
```diff
@@ -1144,6 +1149,212 @@ void CNEO_Player::CheckThermOpticButtons()
...
+//-----------------------------------------------------------------------------
+// Purpose: return true if given target cant be seen because of fog
+//-----------------------------------------------------------------------------
+bool CNEO_Player::IsHiddenByFog(float range) const
+{
+	if (GetFogObscuredRatio(range) >= 1.0f)
+		return true;
+
+	return false;
+}
```
Rainyan: This has the same content as `CBaseCombatCharacter::IsHiddenByFog`, so I don't think we need this specific override to exist here.

src/game/server/neo/neo_player.cpp
```diff
@@ -1144,6 +1149,212 @@ void CNEO_Player::CheckThermOpticButtons()
...
+
+//-----------------------------------------------------------------------------
+// Purpose: return 0-1 ratio where zero is not obscured, and 1 is completely obscured
+//-----------------------------------------------------------------------------
+float CNEO_Player::GetFogObscuredRatio(const Vector& target) const
+{
+	float range = CBaseCombatCharacter::EyePosition().DistTo(target);
+	return GetFogObscuredRatio(range);
+}
```
Rainyan: Identical to `CBaseCombatCharacter::GetFogObscuredRatio`; could remove from here

src/game/server/neo/neo_player.cpp
```diff
@@ -1144,6 +1149,212 @@ void CNEO_Player::CheckThermOpticButtons()
...
+
+//-----------------------------------------------------------------------------
+// Purpose: return 0-1 ratio where zero is not obscured, and 1 is completely obscured
+//-----------------------------------------------------------------------------
+float CNEO_Player::GetFogObscuredRatio(CBaseEntity* target) const
+{
+	if (!target)
+		return 0.0f; // Changed from false to 0.0f for float return
```
Rainyan: Nit: I'm guessing this comment makes reference to the base class using `false` here. I don't think we necessarily need to comment about it here.

src/game/server/neo/neo_player.cpp
```diff
@@ -1144,6 +1149,212 @@ void CNEO_Player::CheckThermOpticButtons()
...
+
+	auto targetPlayer = static_cast<CNEO_Player*>(target);
```
Rainyan: `ToNEOPlayer(target)` would be safer, so we don't need to worry about the case where target might be something else. Since static_casting a `CBaseEntity*` that was not `CNEO_Player*` as such would be undefined behaviour.

src/game/server/neo/neo_player.cpp
```diff
@@ -1144,6 +1149,212 @@ void CNEO_Player::CheckThermOpticButtons()
...
+	if (!targetIsCloaked)
+	{
+		// Target is not cloaked, so not obscured
+		return 0.0f;
+	}
+
+	// --- Base Detection Chance (Per Tick) ---
+	// This is the baseline chance of detection in ideal conditions (stationary, healthy, class-agnostic bot).
+	// Aiming for ~5% detection per second at 66 ticks/sec, this translates to ~0.078% per tick.
+	float detectionChance = 0.0008f; // Starting very low (0.08% per tick)
```
Rainyan: Prefer `chance * gpGlobals->interval_per_tick`, to decouple the assumption of some specific tickrate from the value.

The `interval_per_tick` returns approximately 1/tickrate, so something like 0.01515 for 66 tick. For example: `0.0528 * gpGlobals->interval_per_tick` would equal about 0.0008 at 66 tick.

src/game/server/neo/neo_player.cpp
```diff
@@ -1144,6 +1149,212 @@ void CNEO_Player::CheckThermOpticButtons()
...
+	// Each health point lost increases detection by 1% of current chance (additive to multiplier)
+	const float MULT_INJURED_PER_HEALTH_POINT_FACTOR = 0.01f;
+
+	// Distance Multipliers (How distance affects detection)
+	// These define ranges where detection scales.
+	const float DISTANCE_MAX_DETECTION_SQ = 10000.0f;  // Max detection effect at 100 units (100^2)
+	const float DISTANCE_MIN_DETECTION_SQ = 30000000.0f; // Min detection effect at 3000 units (3000^2)
+
+	const float DISTANCE_MULT_CLOSE = 5.0f; // Multiplier when very close (e.g., within 100 units)
+	const float DISTANCE_MULT_FAR = 0.01f;    // Multiplier when very far (e.g., beyond 3000 units)
```
Rainyan: Preferably mark these as `constexpr` instead of `const` because they are compile-time constant.

It might also help with readability to write the squares as `value * value` in code rather than as comment, since with constexpr they will be evaluated at compile time with 0 runtime cost.

src/game/server/neo/neo_player.cpp
```diff
@@ -1144,6 +1149,212 @@ void CNEO_Player::CheckThermOpticButtons()
...
+
+	// --- Helper Lambdas for Movement ---
+	auto isMoving = [](const CNEO_Player* player, float tolerance = 10.0f) {
+		return !player->GetAbsVelocity().IsZero(tolerance);
+		};
+	// Defined a clear threshold for 'running' velocity.
+	auto isRunning = [](const CNEO_Player* player, float runSpeedThreshold = 200.0f) {
```
Rainyan: Preferably `constexpr auto` for the lambda declarations.

src/game/server/neo/neo_player.cpp
```diff
@@ -2108,6 +2319,13 @@ bool CNEO_Player::WantsLagCompensationOnEntity( const CBasePlayer *pPlayer,
 void CNEO_Player::FireBullets ( const FireBulletsInfo_t &info )
 {
 	BaseClass::FireBullets(info);
+
+	if (!(static_cast<CNEOBaseCombatWeapon*>(GetActiveWeapon()))->GetNeoWepBits() & NEO_WEP_SUPPRESSED)
```
Rainyan: Problem with operator precedence here; logical NOT (`!`) takes precedence over bitwise AND (`&`), so you need to rearrange the parentheses:
```diff
-       if (!(static_cast<CNEOBaseCombatWeapon*>(GetActiveWeapon()))->GetNeoWepBits() & NEO_WEP_SUPPRESSED)
+       if (!(static_cast<CNEOBaseCombatWeapon*>(GetActiveWeapon())->GetNeoWepBits() & NEO_WEP_SUPPRESSED))
```

because else it would first evaluate the logical NOT into a truth value, and only then evaluate the bitwise AND as `if (boolean & bitmaskB)` (which is always false for bitmasks >1). When in reality, we wanted it to evaluate `if not (bitmaskA & bitmaskB)` instead of `if (not bitmaskA) & bitmaskB`.

For reference: https://en.cppreference.com/w/cpp/language/operator_precedence.html

src/game/server/neo/neo_player.cpp
```diff
@@ -1144,6 +1149,212 @@ void CNEO_Player::CheckThermOpticButtons()
...
+	}
+
+	// Bot Movement Impact
+	if (myPlayerIsMoving)
+	{
+		detectionChance *= MULT_MY_PLAYER_MOVING;
+	}
+
+	// Distance Impact
+	const Vector& myPos = this->GetAbsOrigin(); // TODO: May need GetBot()->GetPosition() equivalent
```
Rainyan: Nit: redundant `this->`, maybe simplify as just `GetAbsOrigin()` because the `this` is implied already.

src/game/server/neo/neo_player.cpp
```diff
@@ -1144,6 +1149,212 @@ void CNEO_Player::CheckThermOpticButtons()
...
+	else // Interpolate between max and min detection effects
+	{
+		// Alpha: 1.0 when at DISTANCE_MAX_DETECTION_SQ, 0.0 when at DISTANCE_MIN_DETECTION_SQ
+		float alpha = 1.0f - ((currentRangeSq - DISTANCE_MAX_DETECTION_SQ) / (DISTANCE_MIN_DETECTION_SQ - DISTANCE_MAX_DETECTION_SQ));
+		distanceMultiplier = DISTANCE_MULT_FAR * (1.0f - alpha) + DISTANCE_MULT_CLOSE * alpha;
+	}
+	detectionChance *= distanceMultiplier;
+
+	// Class-specific Bot Perception
+	if (this->GetClass() == NEO_CLASS_SUPPORT)
```
Rainyan: `this->GetClass()` could be just `GetClass()`

src/game/server/neo/neo_player.cpp
```diff
@@ -1144,6 +1149,212 @@ void CNEO_Player::CheckThermOpticButtons()
...
+	{
+		detectionChance *= MULT_SUPPORT_BOT_VISION;
+	}
+
+	// Injured Target Impact
+	if (targetPlayer->GetHealth() < 100)
```
Rainyan: I think we currently implement class-based health as damage multipliers rather than health multipliers, and indeed our max health is 100 so this is not incorrect. But it might be safer to use `targetPlayer->GetMaxHealth()` instead of hardcoding 100 as the max hp, just in case that logic ever changes, so we don't introduce a sneaky indirect bug here later.

Also note that there's a `->HealthFraction()` utility method available, if you wanna use it instead of manually calculating the deficit.

src/game/server/neo/neo_player.cpp
```diff
@@ -1144,6 +1149,212 @@ void CNEO_Player::CheckThermOpticButtons()
...
+	{
+		float healthDeficit = 100.0f - targetPlayer->GetHealth();
+		detectionChance *= (1.0f + (healthDeficit * MULT_INJURED_PER_HEALTH_POINT_FACTOR));
+	}
+
+	// Assault class motion vision
+	if (this->GetClass() == NEO_CLASS_ASSAULT && targetIsMoving)
+	{
+		detectionChance = fmaxf(detectionChance, MIN_ASSAULT_DETECTION_CHANCE_PER_TICK);
```
Rainyan: Would prefer using the platform-agnostic Min and Max: https://github.com/NeotokyoRebuild/neo/blob/1c47aa8f4258570563dd4fdba6fbf87a141ddddd/src/public/tier0/basetypes.h#L108-L122 for min and max operations instead of CRT here.

src/game/server/neo/neo_player.h
```diff
@@ -168,6 +179,7 @@ class CNEO_Player : public CHL2MP_Player
 	virtual void StopWalking(void) OVERRIDE;
 
 	// Cloak Power Interface
+	float CloakPower_Get(void);
```
Rainyan: This can be declared const: `float CloakPower_Get(void) const;`

And in implementation:
```cpp
float CNEO_Player::CloakPower_Get(void) const
```

src/game/server/neo/neo_player.h
```diff
@@ -152,6 +162,7 @@ class CNEO_Player : public CHL2MP_Player
 
 	int GetSkin() const { return m_iNeoSkin; }
 	int GetClass() const { return m_iNeoClass; }
+	bool GetCloakState() { return m_aggroTimer.IsElapsed() ? m_bInThermOpticCamo : false; }
```
Rainyan: This method name might be a little confusing, since I could imagine someone calling it, assuming it just tells you whether they are cloaked or not, oblivious of the aggro timer thing.

Could we instead have it in CNEOBot, as something like `static bool GetCloakState(const CNEO_Player* other);` or similar utility function, to avoid such accidental usage in the future?

Or alternatively, if we keep it here, then give it a little more descriptive name so that it's clear it does more than merely give you the cloak state. `GetBotPerceivedCloakState`? In which case, it could also be declared `const`, since the call doesn't mutate CNEO_Player.

src/game/server/neo/neo_player.cpp
```diff
@@ -2108,6 +2319,13 @@ bool CNEO_Player::WantsLagCompensationOnEntity( const CBasePlayer *pPlayer,
 void CNEO_Player::FireBullets ( const FireBulletsInfo_t &info )
 {
 	BaseClass::FireBullets(info);
+
+	if (!(static_cast<CNEOBaseCombatWeapon*>(GetActiveWeapon()))->GetNeoWepBits() & NEO_WEP_SUPPRESSED)
+	{
+		// cloak disruption from unsuppressed weapons
+		m_aggroTimer.Invalidate();
+		m_aggroTimer.Start(0.5f);
```
Rainyan: I think the `Invalidate` is redundant if followed by a `Start`, because looking at their implementations: https://github.com/NeotokyoRebuild/neo/blob/1c47aa8f4258570563dd4fdba6fbf87a141ddddd/src/game/shared/util_shared.h#L591-L600

it seems that `Invalidate` is just setting a float type `m_timestamp` to -1 and then immediately calling `Start` is re-setting it to a different value.

src/game/shared/activitylist.cpp
```diff
@@ -2521,6 +2521,10 @@ void ActivityList_RegisterSharedActivities( void )
 	REGISTER_SHARED_ACTIVITY( ACT_MP_CYOA_PDA_INTRO );
 	REGISTER_SHARED_ACTIVITY( ACT_MP_CYOA_PDA_IDLE );
 	REGISTER_SHARED_ACTIVITY( ACT_MP_CYOA_PDA_OUTRO );
+
+#ifdef NEO
+	REGISTER_SHARED_ACTIVITY( ACT_THERMOPTIC );
```
Rainyan: Do we need this activity for anything? Afaik these are mostly related to animations, and at least currently we have no animation associated with cloaking.

src/game/server/neo/neo_player.cpp
```diff
@@ -1080,6 +1081,10 @@ void CNEO_Player::PlayCloakSound(bool removeLocalPlayer)
 		params.m_nChannel = CHAN_VOICE;
 
 		EmitSound(filter, edict()->m_EdictIndex, params);
+
+		// for emulating bot visibility of cloak initiation flash
+		m_aggroTimer.Invalidate();
+		m_aggroTimer.Start(0.5f);
```
Rainyan: Ditto on redundant `Invalidate` when immediately followed by `Start`

src/game/server/neo/neo_player.cpp
```diff
@@ -1144,6 +1149,212 @@ void CNEO_Player::CheckThermOpticButtons()
...
+
+	// Convert ratio to detection chance percentage (invert since obscured ratio is opposite of detection)
+	float detectionChancePercent = (1.0f - obscuredRatio) * 100.0f;
+
+	// Roll the chance to determine if the target is detected
+	bool detected = (RandomInt(1, 100) <= detectionChancePercent);
+
+	// Return true if hidden (not detected)
+	return !detected;
+}
```
Rainyan: This logic could probably simplify to:
```cpp
return GetRandomFloat() < GetFogObscuredRatio(target);
```

src/game/server/neo/neo_player.cpp
```diff
@@ -2997,6 +3215,11 @@ void CNEO_Player::StopWalking(void)
 	m_fIsWalking = false;
 }
 
+float CNEO_Player::CloakPower_Get(void)
+{
+	return m_HL2Local.m_cloakPower;
+}
```
Rainyan: Might be worth moving to the header, so the compiler has the opportunity to inline it.

src/game/server/neo/neo_player.h
```diff
@@ -210,6 +222,7 @@ class CNEO_Player : public CHL2MP_Player
 	void CloakFlash(float time = 0.f);
 private:
 	bool m_bAllowGibbing;
+	CountdownTimer m_aggroTimer;
```
Rainyan: Nit: could we name this `m_botAggroTimer` or similar, so it's a bit more apparent at a glance what it's for. Since "aggro timer" alone might be a little confusing if you don't realize it's to do with bots.

Rainyan (changes requested):

src/game/server/neo/neo_player.cpp
sunzenshen: Maybe this is a code smell, but moving this to the header would require #including c_baseplayer.h* to define m_HL2Local to be available in neo_player.h, which I believe that would add bloat to using that header as a dependency.

\* (and possibly tweaks to CMake since #include "c_baseplayer.h" wasn't immediately available in this context)

I vaguely remember a workaround (something called class prototyping?), but I feel like I should ask before going down that path.

src/game/server/neo/neo_player.h
sunzenshen: I don't think it makes sense to move the timer into CNEOBot,
- because the probability for the bot to see the player is calculated in CNEO_Player::GetFogObscuredRatio, 
- and NextBotVisionInterface calls IsHiddenByFog
- which then calls GetFogObscuredRatio() for the percentage to see.
- You'd have to modify IVision::IsAbleTosee to call other functions than IsHiddenByFog, so you might as well roll the timer check in CNEO_Player

I'll try the GetBotPerceivedCloakState approach

src/game/server/neo/neo_player.cpp
Rainyan: Ah ok, then it's probably indeed better to not move it.

src/game/server/neo/bot/behavior/neo_bot_attack.cpp
sunzenshen: I actually realized that this might be the wrong place for these checks, so moved them near the PressFireButton-related code.

src/game/server/neo/neo_player.cpp
```diff
@@ -1144,6 +1148,143 @@ void CNEO_Player::CheckThermOpticButtons()
...
+	detectionChance *= (1.0f + ((1 - targetPlayer->HealthFraction()) * MULT_INJURED_PER_HEALTH_POINT_FACTOR));
+
+	// Assault class motion vision
+	if (GetClass() == NEO_CLASS_ASSAULT && targetIsMoving)
+	{
+		detectionChance = MAX(detectionChance, MIN_ASSAULT_DETECTION_CHANCE_PER_TICK);
+	}
+
+	// Ensure the final detection chance is within valid bounds [0, 1] (as a ratio)
+	detectionChance = MAX(0.0f, fminf(1.0f, detectionChance));
```
Rainyan: Preferably `Max` and `Min`, and not the C macro versions `MAX`/`MIN` nor `fman`/`fmix`.
```diff
diff --git a/src/game/server/neo/neo_player.cpp b/src/game/server/neo/neo_player.cpp
index 6cd39d4e..f958c494 100644
--- a/src/game/server/neo/neo_player.cpp
+++ b/src/game/server/neo/neo_player.cpp
@@ -1273,11 +1273,11 @@ float CNEO_Player::GetFogObscuredRatio(CBaseEntity* target) const
 	// Assault class motion vision
 	if (GetClass() == NEO_CLASS_ASSAULT && targetIsMoving)
 	{
-		detectionChance = MAX(detectionChance, MIN_ASSAULT_DETECTION_CHANCE_PER_TICK);
+		detectionChance = Max(detectionChance, MIN_ASSAULT_DETECTION_CHANCE_PER_TICK);
 	}
 
 	// Ensure the final detection chance is within valid bounds [0, 1] (as a ratio)
-	detectionChance = MAX(0.0f, fminf(1.0f, detectionChance));
+	detectionChance = Max(0.0f, Min(1.0f, detectionChance));
```

src/game/server/neo/neo_player.cpp
```diff
@@ -1144,6 +1148,143 @@ void CNEO_Player::CheckThermOpticButtons()
...
+	constexpr float MULT_TARGET_RUNNING = 50.0f;   // Running increases detection chance by 50x (very high risk)
+
+	// Bot's State Multipliers (How the observer's state affects its perception)
+	constexpr float MULT_MY_PLAYER_MOVING = 0.5f;  // Observer moving: 50% less chance to detect (detectionChance *= 0.5)
+	constexpr float MULT_SUPPORT_BOT_VISION = 0.6; // Support bot: 40% less chance to detect (detectionChance *= 0.6)
+	constexpr float MIN_ASSAULT_DETECTION_CHANCE_PER_TICK = 0.20f; // 20% detection per tick (very high)
+
+	// Injured Target Multiplier (How target's health affects their stealth)
+	// with 1.0f, someone at 1% health would be almost twice as likely to be seen
+	constexpr float MULT_INJURED_PER_HEALTH_POINT_FACTOR = 1.0f;
```
Rainyan: Would it make sense to instead count the number of damage decals visible on the model, since those are usually what stands out and reveals someone? Something like (pseudocode):
```cpp
virtual char const* CNEO_Player::DamageDecal(int bitsDamageType, int gameMaterial) override
{
	if (IsAlive() && bitsDamageType & (DMG_BULLET | DMG_SLASH))
		++m_numDamageDecals; // sum should be reset for new spawn
	return BaseClass::DamageDecal(bitsDamageType, gameMaterial);
}
```

Since someone merely being low HP for whatever reason may not necessarily mean their model has any of the blood decals drawn on it.

Rainyan (changes requested):

src/game/server/neo/neo_player.cpp
sunzenshen: Admittedly I figured this out after failing to get DamageDecal to trigger from a CNEO_Player override, but this may be better tracked in OnTakeDamage_Alive.
(see: https://github.com/NeotokyoRebuild/neo/pull/1224/commits/398bf2d8ad4cd164dcb11eba1fb5a89e25c105ab)

src/game/server/neo/neo_player.cpp
```diff
@@ -1268,16 +1268,16 @@ float CNEO_Player::GetFogObscuredRatio(CBaseEntity* target) const
...
 	// Assault class motion vision
 	if (GetClass() == NEO_CLASS_ASSAULT && targetIsMoving)
 	{
-		detectionChance = MAX(detectionChance, MIN_ASSAULT_DETECTION_CHANCE_PER_TICK);
+		detectionChance = Max(detectionChance, MIN_ASSAULT_DETECTION_CHANCE_PER_TICK);
 	}
 
 	// Ensure the final detection chance is within valid bounds [0, 1] (as a ratio)
-	detectionChance = MAX(0.0f, fminf(1.0f, detectionChance));
+	detectionChance = Max(0.0f, fminf(1.0f, detectionChance));
```
Rainyan: Sorry to keep harping on this, but could you also change the `fminf` -> `Min` on this line please.

Also on second look, since this is essentially a clamp, it might be neater to use: https://github.com/NeotokyoRebuild/neo/blob/ddaca40a07ea837f97c41570f6f5994d63879bb4/src/public/tier0/basetypes.h#L93-L106 rather than doing the Max(Min(...)) pattern? `detectionChance = Clamp(detectionChance, 0.f, 1.f);` or so. But whichever is fine.

Rainyan (changes requested):

src/game/server/neo/neo_player.cpp
```diff
@@ -1144,6 +1149,142 @@ void CNEO_Player::CheckThermOpticButtons()
...
+	detectionChance *= distanceMultiplier;
+
+	// Class-specific Bot Perception
+	if (GetClass() == NEO_CLASS_SUPPORT)
+	{
+		detectionChance *= MULT_SUPPORT_BOT_VISION;
+	}
+
+	// Injured Target Impact
+	detectionChance *= 1.0f + (MULT_BLEEDING_INJURY_EVENT_FACTOR * m_iBotDetectableBleedingInjuryEvents);
```
Rainyan: Should be `targetPlayer->m_iBotDetectableBleedingInjuryEvents` because else we would be counting the amount of injure decals on the bot itself, and not its target.

sunzenshen: Doh, that's what I get for rushing to cram 2 code review sessions into the same day. Nice catch!

Rainyan (dismissed):

src/game/server/NextBot/Player/NextBotPlayer.h
```diff
@@ -190,6 +195,9 @@ class NextBotPlayer : public PlayerType, public INextBot, public INextBotPlayerI
 	virtual void PressSpecialFireButton( float duration = -1.0f );
 	virtual void ReleaseSpecialFireButton( void );
 
+	virtual void PressThermopticButton( float duration = -1.0f );
```
AdamTadeusz, line 199: missing ifdef NEO

src/game/server/NextBot/Player/NextBotPlayer.h
```diff
@@ -257,6 +265,7 @@ class NextBotPlayer : public PlayerType, public INextBot, public INextBotPlayerI
 	CountdownTimer m_crouchButtonTimer;
 	CountdownTimer m_walkButtonTimer;
 	CountdownTimer m_buttonScaleTimer;
+	CountdownTimer m_thermopticButtonTimer;
```
AdamTadeusz, line 271: missing ifdef NEO

src/game/server/NextBot/Player/NextBotPlayer.h
```diff
@@ -400,6 +409,9 @@ inline void NextBotPlayer< PlayerType >::ReleaseReloadButton( void )
 template < typename PlayerType >
 inline void NextBotPlayer< PlayerType >::PressJumpButton( float duration )
 {
+	// NEO JANK workaround to allow bots to crouch while attempting to preserve crouch jumping
```
AdamTadeusz, line 417: missing ifdef NEO

src/game/server/NextBot/Player/NextBotPlayer.h
```diff
@@ -545,6 +573,7 @@ inline void NextBotPlayer< PlayerType >::Spawn( void )
 	m_buttonScaleTimer.Invalidate();
 	m_forwardScale = m_rightScale = 0.04;
 	m_burningTimer.Invalidate();
+	m_thermopticButtonTimer.Invalidate();
```
AdamTadeusz, line 583: missing ifdef NEO

AdamTadeusz (changes requested):

src/game/server/neo/bot/neo_bot_locomotion.cpp
```diff
@@ -20,7 +20,13 @@ void CNEOBotLocomotion::Update( void )
 	// always 'crouch jump'
 	if ( IsOnGround() )
 	{
+#ifdef NEO
+		// NEO JANK resetting of crouch timer moved to NextBotPlayer::PressJumpButton
+		// so far crouch jump seems to still be working, but watch out for a regression
+		// disabled:
+#else
 		me->ReleaseCrouchButton();
+#endif
```
Rainyan, line 29: Did you mean to ifdef-guard the `ReleaseCrouchButton` to never execute here? Regardless, we are inside a neo-specific file here, so we should not be using `ifdef NEO`. That's only to be used when writing our code inline with Valve's base engine code.

We mostly do this NEO ifdef business so that you can at a glance see that you're looking at NEO code and not Valve code, without needing to embark on a git diff adventure.

src/game/server/NextBot/Player/NextBotPlayer.h
```diff
@@ -411,6 +429,22 @@ inline void NextBotPlayer< PlayerType >::ReleaseJumpButton( void )
...
+}
+
+template < typename PlayerType >
+inline void NextBotPlayer< PlayerType >::ReleaseThermopticButton( void )
+{
+	m_inputButtons &= ~IN_THERMOPTIC;
+	m_thermopticButtonTimer.Invalidate();
+}
+#endif
+
```
Rainyan, line 447: Nit: Ideally, all of these neo-specific methods should live in `class CNEOBot : public NextBotPlayer< CNEO_Player >` instead of being ifdef'd here with the base class. But I'm ok leaving this here if you don't feel like refactoring it as part of this PR.

src/game/server/NextBot/Player/NextBotPlayerLocomotion.h
```diff
@@ -36,6 +36,9 @@ class PlayerLocomotion : public ILocomotion
 	virtual bool ClimbUpToLedge( const Vector &landingGoal, const Vector &landingForward, const CBaseEntity *obstacle );	// initiate a jump to an adjacent high ledge, return false if climb can't start
 	virtual void JumpAcrossGap( const Vector &landingGoal, const Vector &landingForward );	// initiate a jump across an empty volume of space to far side
 	virtual void Jump( void );								// initiate a simple undirected jump in the air
+#ifdef NEO
+	virtual void Thermoptic( void );						// initiate a simple thermoptic toggle
```
Rainyan: Nit: Since we're not really building an interface with the expectation of any other project overriding our base implementation, this could technically be non-virtual to save the overhead of the vtable indirection.

src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
```diff
@@ -70,6 +70,47 @@ ActionResult< CNEOBot >	CNEOBotSeekAndDestroy::Update( CNEOBot *me, float interv
 			return SuspendFor( new CNEOBotAttack, "Going after an enemy" );
 		}
 	}
+	else
+	{
+		// Out of combat
+		me->DisableCloak();
```
Rainyan, line 76: Nit: Uncloaking gives out a sound cue, so if we wanted to be really fancy with the higher skill level bots, they should take into account if they think an enemy might hear their un-cloak when deciding whether to make that noise. But this would probably have to be a whole thing in itself of determining how much noise the bot thinks is acceptable based on current situation, and outside the scope of this PR, to drive decisions like whether to walk or run, etc.

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -1480,25 +1509,36 @@ void CNEOBot::EquipBestWeaponForThreat(const CKnownEntity* threat)
...
-			pChosen = knife;
-		}
-		else if (throwable)
-		{
-			pChosen = throwable;
-		}
+		// passthrough
+	}
+	else if (secondaryWeapon
+		&& primaryWeapon->GetNeoWepBits() & (NEO_WEP_ZR68_L | NEO_WEP_SRS | NEO_WEP_M41_L)
```
Rainyan: Nit: assuming "scoped weapon" was the intention here, we also have the helper bits `NEO_WEP_SCOPEDWEAPON` available.

src/game/server/neo/bot/neo_bot_body.h
```diff
@@ -13,4 +13,7 @@ class CNEOBotBody : public PlayerBody
 	virtual ~CNEOBotBody() { }
 
 	virtual float GetHeadAimTrackingInterval( void ) const;			// return how often we should sample our target's position and velocity to update our aim tracking, to allow realistic slop in tracking
+	virtual float GetCloakPower( void ) const;  		    		// return available thermoptic charge
+	virtual bool IsCloakEnabled( void ) const;           			// return if thermoptic is enabled
```
Rainyan: Nit: Unlikely these will ever be overridden by anyone, so we could make them non-virtual since it doesn't hurt out API usability and has better perf

src/game/server/neo/neo_player.h
```diff
@@ -98,6 +98,12 @@ class CNEO_Player : public CHL2MP_Player
 	virtual const Vector GetPlayerMins(void) const OVERRIDE;
 	virtual const Vector GetPlayerMaxs(void) const OVERRIDE;
 
+	// -----------------------
+	// Fog
+	// -----------------------
+	virtual bool		IsHiddenByFog(CBaseEntity* target) const OVERRIDE;        ///< return true if given target cant be seen because of fog
+	virtual float		GetFogObscuredRatio(CBaseEntity* target) const OVERRIDE;  ///< return 0-1 ratio where zero is not obscured, and 1 is completely obscured
```
Rainyan: Nit: While I originally suggested using this API, it would be nice to comment here what exactly is going on, since it has very little to do with fog in our NT context. Perhaps a task for some future refactor.

Rainyan (dismissed):
LGTM, left some more nits to satisfy my OCD but I don't require changes to merge this

AdamTadeusz (approved):

Rainyan (approved):

PR #1227, opened 2025-07-24
don't spawn if respawns enabled

### LINKED ISSUE #1225: Players are spawned instantly in the tutorial, preventing them from picking a class other than assault
Likely caused by https://github.com/NeotokyoRebuild/neo/pull/1142

AdamTadeusz:
## Linked Issues

- fixes #1225

### DIFF
--- a/src/game/server/neo/neo_player.cpp
+++ b/src/game/server/neo/neo_player.cpp
@@ -2581,7 +2581,7 @@ bool CNEO_Player::ProcessTeamSwitchRequest(int iTeam)
 	{
 		if (!justJoined && GetTeamNumber() != TEAM_SPECTATOR && !IsDead())
 		{
-			if (suicidePlayerIfAlive || NEORules()->CanChangeTeamClassWeaponWhenAlive())
+			if (suicidePlayerIfAlive || NEORules()->CanChangeTeamClassLoadoutWhenAlive())
 			{
 				SoftSuicide();
 			}
--- a/src/game/shared/neo/neo_gamerules.h
+++ b/src/game/shared/neo/neo_gamerules.h
@@ -208,8 +208,8 @@ public:
 	int GetForcedSkin();
 	int GetForcedWeapon();
 	virtual const char* GetGameTypeName(void) OVERRIDE;
-	bool CanChangeTeamClassWeaponWhenAlive();
-	bool CanRespawnAnyTime();
+	virtual const bool CanChangeTeamClassLoadoutWhenAlive();
+	virtual const bool CanRespawnAnyTime();
 
 	void GetDMHighestScorers(
 #ifdef GAME_DLL
--- a/src/game/shared/neo/neo_gamerules.cpp
+++ b/src/game/shared/neo/neo_gamerules.cpp
@@ -3780,17 +3780,17 @@ int CNEORules::GetForcedWeapon(void)
 	return m_iForcedWeapon;
 }
 
-const char* CNEORules::GetGameTypeName(void)
+inline const char* CNEORules::GetGameTypeName(void)
 {
 	return NEO_GAME_TYPE_SETTINGS[GetGameType()].gameTypeName;
 }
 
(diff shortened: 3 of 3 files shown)


Rainyan (approved):

src/game/server/neo/neo_player.cpp
```diff
@@ -2595,12 +2595,19 @@ bool CNEO_Player::ProcessTeamSwitchRequest(int iTeam)
...
-		spawn = NEORules()->FPlayerCanRespawn(this);
-		if (!spawn)
+		// Spawn the player immediately if its a single life game mode
+		spawnImmediately = !NEORules()->CanRespawnAnyTime() && NEORules()->FPlayerCanRespawn(this);
+		if (!spawnImmediately)
 		{
-			// If we're not allowed to be in the current observer mode (because of mp_forcecamera for example), this will give us a new observer mode.
-			SetObserverMode(GetObserverMode());
-			State_Transition(STATE_OBSERVER_MODE);
+			if (NEORules()->CanRespawnAnyTime() || IsFakeClient())
```
sunzenshen, line 2602: It seems you already checked
`NEORules()->CanRespawnAnyTime()`, when you assigned `spawnImmediately =` and then fell in here with `if (!spawnImmediately)`

Rainyan: > It seems you already checked `NEORules()->CanRespawnAnyTime()`, when you assigned `spawnImmediately =` and then fell in here with `if (!spawnImmediately)`

Note that when evaluating the assignment, there's 3 different possibilities that would result in `spawnImmediately = false`:
```cpp
spawnImmediately = false && true; // case 1: NOT CanRespawnAnyTime AND FPlayerCanRespawn
spawnImmediately = true && false; // case 2: CanRespawnAnyTime AND NOT FPlayerCanRespawn
spawnImmediately = false && false; // case 3:  NOT CanRespawnAnyTime AND NOT FPlayerCanRespawn
```

so the truth value of `spawnImmediately` alone can't be used to deduce which side of the AND condition (or both) returned false.

AdamTadeusz: would making CanRespawnAnyTime inline have any effect, since we're calling it from a pointer?

I mean from a performance perspective not a logical perspective

AdamTadeusz:
Ok this works fine, the issue with mp_forcerespawn is present on master as well, most likely since nextbots were first introduced since thats when Actions were added which is where the problem is. Adding some checks whether player is a fake client in CBasePlayer::PlayerDeathThink fixes the problem with mp_forcerespawn set to 0, I want to explore this in more detail, will open a separate pr

src/game/server/neo/neo_player.cpp
sunzenshen: Maybe this is a case where readability of the logic is more important than line by line optimization.  There's also the possibility that his will be optimized for Release builds depending on the build flags in effect.

sunzenshen (approved):

src/game/shared/neo/neo_gamerules.h
```diff
@@ -208,7 +208,8 @@ class CNEORules : public CHL2MPRules, public CGameEventListener
 	int GetForcedSkin();
 	int GetForcedWeapon();
 	virtual const char* GetGameTypeName(void) OVERRIDE;
-	bool CanChangeTeamClassWeaponWhenAlive();
+	virtual const bool CanChangeTeamClassLoadoutWhenAlive();
+	virtual const bool CanRespawnAnyTime();
```
Rainyan, line 212: Nit: The `const` is a bit redundant in the prototypes, because we return bool by value so it doesn't do anything.

Rainyan (approved):

DESTROYGIRL (approved):

src/game/shared/neo/neo_gamerules.h
AdamTadeusz: my bad

PR #1231, opened 2025-07-24
Keybind support for neo_message, and extra line for keybinds

### LINKED ISSUE #1229: Add the ability to display keybinds with neo_message
Currently the tutorial displays keybinds in the bottom left corner of the screen with `env_hudhint` as that can display keybinds by finding an action surrounded by % symbols in the localised text.

For example: `".SPRINT WITH %+sprint%"`

This text is easily missed, and would be better if we had this functionality in our neo_message so players do not need to look at two places on the screen, and recognise that all hints will be in the top left box.

DESTROYGIRL:
## Description
Keybinds can be shown with `neo_message` by surrounding localised text with %, just like with `env_hudhint`
Added an extra line specifically for displaying keybinds/notices that flashes.

NEOASSETS >>>https://github.com/NeotokyoRebuild/neoAssets/pull/62<<<

## Toolchain
- Windows MSVC VS2022

## Linked Issues

- fixes #1229


src/game/server/neo/neo_message.cpp
```diff
@@ -67,5 +69,16 @@ void CNEO_Message::InputShowMessage(inputdata_t& inputData)
...
 void CNEO_Message::InputHideMessage(inputdata_t& inputData)
 {
-    Q_strncpy(m_NetworkedMessageKey.GetForModify(), "", 255);
+	Q_strncpy(m_NetworkedMessageKey.GetForModify(), "", 255);
+	Q_strncpy(m_NetworkedSubMessageKey.GetForModify(), "", 255);
+}
+
+void CNEO_Message::InputShowSubMessage(inputdata_t& inputData)
+{
+	Q_strncpy(m_NetworkedSubMessageKey.GetForModify(), inputData.value.String(), 255);
```
sunzenshen: You may want to replace the 255 constant with sizeof(m_NetworkedSubMessageKey)
e.g.
```cpp
Q_strncpy(m_NetworkedSubMessageKey.GetForModify(), inputData.value.String(), sizeof(m_NetworkedSubMessageKey));
```

sunzenshen: Same for the other occurrences of Q_strncpy... 255

src/game/client/neo/ui/neo_hud_message.cpp
```diff
@@ -138,7 +152,7 @@ void CNEOHud_Message::UpdateStateForNeoHudElementDraw()
 
 void CNEOHud_Message::ShowMessage(wchar_t* message)
 {
-    wcsncpy(m_szMessage, message, sizeof(m_szMessage) / sizeof(wchar_t));
+    wcsncpy(m_szMessage, ProcessKeyBinds(message).c_str(), sizeof(m_szMessage) / sizeof(wchar_t));
```
sunzenshen: I'm not as familiar with the conventions of this engine, but is there a problem if m_szMessage is not null terminated?  If the wstring returned by ProcessKeyBinds is longer than m_szMessage buffer, then wcsncpy would copy up to the limit of the sizeof/sizeof calculations, but the buffer would not be null terminated.

I noticed the codebase doesn't use wcsncpy_s so I assume for game code reasons we don't want to error or crash at all costs. But here's a couple links that explain the difference:
* [StackOverflow](https://stackoverflow.com/questions/14892037/what-is-the-difference-between-wcsncpy-and-wcscpy-s)
* [Spec](https://en.cppreference.com/w/c/string/wide/wcsncpy)

> If count is reached before the entire string src was copied, the resulting wide character array is not null-terminated.

That said, wcscpy_s may not be the right approach as we don't want to crash the game:
https://stackoverflow.com/questions/73866448/changing-the-method-from-wcscpy-to-wcscpy-s

I noticed that the way other parts of the code handle this case is to just manually assign the null, like in:
neo\src\game\client\econ\econ_controls.cpp
```cpp
wcsncpy( wszText, pLocText + startIdx, len );
wszText[len-1] = 0;
```

(same for line 249)

Totally fine if this recommendation is too paranoid for this situation.

sunzenshen: Here's an admittedly goofy way to accomplish safe null termination in a safe way:

```cpp
constexpr size_t kMsgBufferSize = sizeof(m_szMessage) / sizeof(wchar_t);
auto processed = ProcessKeyBinds(message);
wcsncpy(m_szMessage, processed.c_str(), kMsgBufferSize - 1);
m_szMessage[kMsgBufferSize - 1] = L'\0';
```

DESTROYGIRL: > Totally fine if this recommendation is too paranoid for this situation.

Nah I think it's valid. The longest localised text gets close enough to the limit, additionally it has the potential to get even longer after ProcessKeyBinds()

DESTROYGIRL: Hmmm I think `V_wcsncpy` is the sdk solution for this?

Rainyan: Yeah we should preferably opt for the `V_...` str functions when they're available (or `Q_...`, but they're just an alias for `V_...` for compatibility reasons)

sunzenshen (approved):

src/game/server/neo/neo_message.cpp
```diff
@@ -43,12 +53,34 @@ int CNEO_Message::UpdateTransmitState()
...
+void CNEO_Message::Think()
+{
+	DisplayTimer();
+	SetNextThink(gpGlobals->curtime + 0.05f);
+}
+
+void CNEO_Message::DisplayTimer()
+{
+	float elapsed = gpGlobals->curtime - m_flTimerStart;
+	char timer[10];
```
sunzenshen: C++ compilation usually aligns stack allocations at 16 byte boundaries, so this could be a timer[16] with no change in amount of stack space used, as timer[10] will just be alignment padded.  #### minutes is a very long time but might be nice to have some buffer.

DESTROYGIRL: Good to know. Thanks

sunzenshen: Is there a place where this timer is supposed to show up? I didn't notice it when going through the time trial in the weapons tutorial. (Edit: doh, I see "intend to use", so maybe it's not hooked up yet.)

DESTROYGIRL: You should be able to test it by doing the following commands:

1. enter the tutorial
2. `ent_fire neo_message addoutput "timermode 1"` in console
3. `ent_fire neo_message showmessage` in console
4. `ent_fire neo_message stoptimer` in console

sunzenshen (approved):
Timer works on my machine with the provided console commands.

src/game/client/neo/c_neo_message.cpp
```diff
@@ -5,14 +5,18 @@
...
+    RecvPropString(RECVINFO(m_NetworkedMessageKey)),
+    RecvPropString(RECVINFO(m_NetworkedSubMessageKey)),
+    RecvPropBool(RECVINFO(m_bTimerMode))
 END_RECV_TABLE()
 
 C_NEO_Message::C_NEO_Message()
 {
     // DG: Why doesn't GET_HUDELEMENT(CNEOHud_Message) work here
     CHudElement* pPanelName = gHUD.FindElement("neo_message");
     m_pHudMessage = static_cast<CNEOHud_Message*>(pPanelName);
```
Rainyan: The `GET_HUDELEMENT` doesn't work because the class and panel names are different. But you could instead use `GET_NAMED_HUDELEMENT(CNEOHud_Message, neo_message);`

src/game/client/neo/c_neo_message.h
```diff
@@ -13,6 +13,9 @@ class C_NEO_Message : public C_BaseEntity
     void OnDataChanged(DataUpdateType_t updateType) override;
 
 private:
-    char m_NetworkedMessageKey[255];
+    char m_NetworkedMessageKey[256];
+    char m_NetworkedSubMessageKey[256];
+    bool m_bTimerMode;
     CNEOHud_Message* m_pHudMessage;
+    wchar_t *m_LocalizedPrefix;
```
Rainyan: Could we zero-initialize these client-side, just in case we don't receive the initial server value before they are accessed.

For example with value-initialization:
```cpp
char m_NetworkedMessageKey[256]{};
char m_NetworkedSubMessageKey[256]{};
bool m_bTimerMode{};
CNEOHud_Message* m_pHudMessage{};
wchar_t* m_LocalizedPrefix{};
```

src/game/client/neo/ui/neo_hud_message.cpp
```diff
@@ -124,6 +124,20 @@ void CNEOHud_Message::DrawNeoHudElement()
     surface()->DrawSetTextColor(255, 255, 255, 255);
     surface()->DrawSetTextPos(crossInset + (int)(20 * m_fScale), crossInset);
     surface()->DrawPrintText(L"SYSTEM_MSG", 10);
+
+    // Sub message for keybind hints
+    if (m_szSubMessage[0])
+    {
+        surface()->DrawSetTextFont(m_hTitleFont);
+
+        const float period = 2.0f;
```
Rainyan: could be `constexpr float`

src/game/client/neo/ui/neo_hud_message.cpp
```diff
@@ -124,6 +124,20 @@ void CNEOHud_Message::DrawNeoHudElement()
...
+        float phase = fmodf(gpGlobals->curtime, period) / period;
+        int alpha = Lerp(phase, 10, 255);
+
+        surface()->DrawSetTextColor(255, 255, 255, alpha);
+        surface()->DrawSetTextPos(m_Padding, m_BoxHeight - crossSize - crossInset * (1.5 * m_fScale));
+        surface()->DrawPrintText(m_szSubMessage, wcslen(m_szSubMessage));
```
Rainyan: preferably `V_wcslen`

src/game/client/neo/ui/neo_hud_message.cpp
```diff
@@ -148,7 +162,7 @@ void CNEOHud_Message::ShowMessage(wchar_t* message)
     int maxWidth = 0;
     int totalHeight = 0;
 
-    wchar_t* messageCopy = wcsdup(message);
+    wchar_t* messageCopy = wcsdup(m_szMessage);
     wchar_t* context = nullptr;
     wchar_t* token = wcstok(messageCopy, L"\n", &context); // Any \n in the localised text will declare a new line
```
Rainyan: Could we avoid the heap allocation here with for example:
```diff
--- a/src/game/client/neo/ui/neo_hud_message.cpp
+++ b/src/game/client/neo/ui/neo_hud_message.cpp
@@ -162,7 +162,8 @@ void CNEOHud_Message::ShowMessage(wchar_t* message)
     int maxWidth = 0;
     int totalHeight = 0;

-    wchar_t* messageCopy = wcsdup(m_szMessage);
+    wchar_t messageCopy[sizeof(m_szMessage)];
+    V_memcpy(messageCopy, m_szMessage, sizeof(messageCopy));
     wchar_t* context = nullptr;
     wchar_t* token = wcstok(messageCopy, L"\n", &context); // Any \n in the localised text will declare a new line

@@ -181,7 +182,6 @@ void CNEOHud_Message::ShowMessage(wchar_t* message)
         totalHeight += textHeight + 2;
         token = wcstok(nullptr, L"\n", &context);
     }
-    free(messageCopy);
```

src/game/client/neo/ui/neo_hud_message.cpp
```diff
@@ -178,3 +192,64 @@ void CNEOHud_Message::HideMessage()
     m_szMessage[0] = L'\0';
     m_bShouldDraw = false;
 }
+
+std::wstring CNEOHud_Message::ProcessKeyBinds(wchar_t* rawmessage)
+{
```
Rainyan: Instead of making our own parser, we could probably use `UTIL_ReplaceKeyBindings` here. Simple example:
```diff
--- a/src/game/client/neo/ui/neo_hud_message.cpp
+++ b/src/game/client/neo/ui/neo_hud_message.cpp
@@ -195,6 +195,11 @@ void CNEOHud_Message::HideMessage()

 std::wstring CNEOHud_Message::ProcessKeyBinds(wchar_t* rawmessage)
 {
+    std::wstring res;
+    res.reserve(1024);
+    UTIL_ReplaceKeyBindings(rawmessage, 0, res.data(), res.capacity());
+    return res;
+
```

At least assuming its output formatting is acceptable for this use case. If we do make this change, then it's probably worth removing `::ProcessKeyBinds` since we could just call `UTIL_ReplaceKeyBindings` directly.

src/game/server/neo/neo_message.h
```diff
@@ -9,15 +9,24 @@ class CNEO_Message : public CBaseEntity
     DECLARE_DATADESC();
 
     void Spawn() override;
-    void Precache(void);
+    void Precache();
+    void Think();
     int UpdateTransmitState();
```
Rainyan: Preferably annotate these four methods as `virtual` and `override`

src/game/server/neo/neo_message.cpp
```diff
@@ -67,5 +99,29 @@ void CNEO_Message::InputShowMessage(inputdata_t& inputData)
...
+}
+
+void CNEO_Message::InputShowSubMessage(inputdata_t& inputData)
+{
+	V_strncpy(m_NetworkedSubMessageKey.GetForModify(), inputData.value.String(), sizeof(m_NetworkedSubMessageKey));
+}
+
+void CNEO_Message::InputHideSubMessage(inputdata_t& inputData)
+{
+	V_strncpy(m_NetworkedSubMessageKey.GetForModify(), "", sizeof(m_NetworkedSubMessageKey));
```
Rainyan: Since `strncpy` would zero the entire length of the leftover memory block, this could be optimized to just null-terminate the first character instead: `*m_NetworkedSubMessageKey.GetForModify() = 0;`

This doesn't really matter for performance, but since this is networked, zeroing the whole block might cause a spike in the netcode delta that must be sent to each client, so it'd be nicer to just send the delta of the first element being zeroed since the early null terminator is all we need.

src/game/server/neo/neo_message.cpp
```diff
@@ -43,12 +53,34 @@ int CNEO_Message::UpdateTransmitState()
...
+	SetNextThink(gpGlobals->curtime + 0.05f);
+}
+
+void CNEO_Message::DisplayTimer()
+{
+	float elapsed = gpGlobals->curtime - m_flTimerStart;
+	char timer[16];
+	V_snprintf(timer, sizeof(timer), "%02d:%02d:%03d", int(elapsed / 60.0f), int(elapsed) % 60, int((elapsed - int(elapsed)) * 1000.0f)); // minutes, seconds, milliseconds
+
+	V_strncpy(m_NetworkedMessageKey.GetForModify(), timer, sizeof(m_NetworkedMessageKey));
```
Rainyan: The timer allocation is a bit redundant, since we could format directly to the target buffer:
```cpp
float elapsed = gpGlobals->curtime - m_flTimerStart;
V_snprintf(m_NetworkedMessageKey.GetForModify(), sizeof(m_NetworkedMessageKey),
	"%02d:%02d:%03d", int(elapsed / 60.0f),
	int(elapsed) % 60,
	int((elapsed - int(elapsed)) * 1000.0f)); // minutes, seconds, milliseconds
```

src/game/client/neo/ui/neo_hud_message.h
```diff
@@ -19,6 +19,8 @@ class CNEOHud_Message : public CNEOHud_ChildElement, public CHudElement, public
 
     void ShowMessage(wchar_t* message);
     void HideMessage();
+    void ShowSubMessage(wchar_t* message);
+    void HideSubMessage();
```
Rainyan: `ShowMessage` and `ShowSubMessage` should preferably have their argument as type `const wchar_t*`

src/game/client/neo/ui/neo_hud_message.h
```diff
@@ -27,12 +29,15 @@ class CNEOHud_Message : public CNEOHud_ChildElement, public CHudElement, public
     virtual ConVar* GetUpdateFrequencyConVar() const override;
 
 private:
+    std::wstring ProcessKeyBinds(wchar_t* rawmessage);
+
```
Rainyan: This can be const-qualified as
```cpp
std::wstring ProcessKeyBinds(const wchar_t* rawmessage) const;
```
But do consider using `UTIL_ReplaceKeyBindings` instead.

src/game/client/neo/c_neo_message.cpp
```diff
@@ -23,17 +27,42 @@ void C_NEO_Message::OnDataChanged(DataUpdateType_t updateType)
         return;
     }
 
-    if (Q_strlen(STRING(m_NetworkedMessageKey)) > 0)
+    if (V_strlen(m_NetworkedMessageKey) > 0)
```
Rainyan: Instead of calculating the string length, you can check if the first character is non-zero (the null terminator), since this is faster: `if (*m_NetworkedMessageKey)`

src/game/client/neo/c_neo_message.cpp
```diff
@@ -23,17 +27,42 @@ void C_NEO_Message::OnDataChanged(DataUpdateType_t updateType)
...
+                m_pHudMessage->ShowMessage(localizedText);
+            }
+        }
+    }
+    else
+    {
+        m_pHudMessage->HideMessage();
+    }
+
+    if (V_strlen(m_NetworkedSubMessageKey) > 0)
```
Rainyan: `if (*m_NetworkedSubMessageKey)`

game/neo/resource/neo_english.txt
Rainyan, line 1: `%+crouch%` should be `%+duck%` for NT

src/game/server/neo/neo_message.cpp
```diff
@@ -43,12 +53,34 @@ int CNEO_Message::UpdateTransmitState()
     return SetTransmitState(FL_EDICT_ALWAYS);
 }
 
+void CNEO_Message::Think()
+{
+	DisplayTimer();
+	SetNextThink(gpGlobals->curtime + 0.05f);
```
Rainyan, line 59: If we care about accurate completion time, then could this think delay introduce some inaccuracy? Maybe lower the think delay to `+ 0.01f` and then display only up to hundreths of a second for the time instead of milliseconds?

Rainyan (changes requested):

src/game/server/neo/neo_message.cpp
DESTROYGIRL: When the timer is stopped DisplayTimer() is called one final time so it should be accurate regardless of the think delay

src/game/client/neo/ui/neo_hud_message.cpp
DESTROYGIRL: I can't say it's not upsetting to hear that this exists but the ProcessKeyBinds() will need to stay regardless to keep the capitalisation I assume

src/game/client/neo/c_neo_message.h
DESTROYGIRL: Only the top 3 are networked, so I shouldn't need to do it for `m_pHudMessage` and `m_LocalizedPrefix` correct?

Rainyan: > Only the top 3 are networked, so I shouldn't need to do it for `m_pHudMessage` and `m_LocalizedPrefix` correct?

Oh yeah my bad, those are already indeed initialized in the ctor.

src/game/client/neo/ui/neo_hud_message.cpp
```diff
@@ -178,3 +192,64 @@ void CNEOHud_Message::HideMessage()
     m_szMessage[0] = L'\0';
     m_bShouldDraw = false;
 }
+
+std::wstring CNEOHud_Message::ProcessKeyBinds(const wchar_t* rawmessage) const
```
Rainyan, line 196: We should ideally be able to handle the standard `"%%"` percent sign escape case, so that we have support for inputs like:
```
"Tutorial_Bind_Use"			"Drink your 100%% orange juice with %+use%"
```
so something like
```diff
--- a/src/game/client/neo/ui/neo_hud_message.cpp
+++ b/src/game/client/neo/ui/neo_hud_message.cpp
@@ -217,6 +217,13 @@ std::wstring CNEOHud_Message::ProcessKeyBinds(const wchar_t* rawmessage) const
             break;
         }

+        if (start + 1 == end)
+        {
+            outputMessage.append(L"%");
+            pos = end + 1;
+            continue;
+        }
+
         std::wstring unicodeBinding = inputMessage.substr(start + 1, end - start - 1); // Extract the command

         char binding[64];
```

Rainyan (approved):

Issue #1166, opened 2025-05-30
Legacy menus / console / mp3 player don't regain focus on mouse input


AdamTadeusz:
When a legacy ui element looses focus you cannot bring it back in focus by clicking on it with the mouse

AdamTadeusz:
@nullsystem any ideas?

AdamTadeusz:
Probably caused by the neoroot panel input now being a popup of the same height and width as the screen, so any other popup that's not in the foreground can never be interacted with. Not sure why OnKeyCodePressed doesn't trigger when neorootinput is no longer a popup

PR #1233, opened 2025-07-27
Fix UI bind bugs

### LINKED ISSUE #1235: Unbound keys should not activate unbound UI elements
### Build Info

0658b63

### Description

If the user inputs a key that has no mapping in the engine, that key will result in `BUTTON_CODE_NONE`. Any UI element which the user has not associated with a key will also map to `BUTTON_CODE_NONE`. Therefore, pressing these kinds of keys (such as virtual multimedia keys like "Volume Up") can incorrectly trigger UI elements that are un-bound, for example the MP3 player UI, to unwantedly pop up.

Here's a player report from Discord with a video that likely is demonstrating this same problem: https://discord.com/channels/1235346473827434517/1235346647618420876/1391482692377513985

### To Reproduce

* Have no keybind set for `neo_mp3` MP3 player UI element (or unbind them from your autoexec's, etc)
* Launch the game and enter the main menu
* Close the console if it's open
* Use the "Volume Up" media key from a keyboard with such a button
  * If your keyboard has no such button, other similar multimedia keys/virtual keys may also work
  * Or you could emulate this input in software like [AutoHotkey](https://www.autohotkey.com/docs/v1/KeyList.htm#multimedia)

### Expected behavior

Pressing the "Volume Up" media key does not display the MP3 menu.

### Actual behavior

Pressing the media key does display the MP3 menu.

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

11

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

Possibly related:
* https://learn.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes
* https://github.com/NeotokyoRebuild/neo/blob/0658b63c194ccb5d923dcc04168faa9879cd1191/src/vgui2/src/vgui_key_translation.h

Rainyan:
## Description
Fix #1235, #1236

We still have the problem #1237  here that the user can't assign multiple keybinds for the same command, because the latest bind will always take precedence. But that's outside the scope of this PR.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes: #1235 
- fixes: #1236 
- related: #1237 
- possibly(?) related: #1166

### DIFF
--- a/src/game/client/mp3player.cpp
+++ b/src/game/client/mp3player.cpp
@@ -1126,6 +1126,11 @@ ConCommand neo_mp3("neo_mp3", &neo_mp3_callback, "Toggle the mp3 player", FCVAR_
 
 void CMP3Player::OnKeyCodePressed(vgui::KeyCode code)
 {
+	if (code == BUTTON_CODE_NONE || code == BUTTON_CODE_INVALID)
+	{
+		return;
+	}
+
 	const auto toggleMP3Bind = gameuifuncs->GetButtonCodeForBind("neo_mp3");
 	if (toggleMP3Bind != code || !g_pPlayer)
 	{
--- a/src/game/client/mp3player.cpp
+++ b/src/game/client/mp3player.cpp
@@ -1126,6 +1126,8 @@ ConCommand neo_mp3("neo_mp3", &neo_mp3_callback, "Toggle the mp3 player", FCVAR_
 
 void CMP3Player::OnKeyCodePressed(vgui::KeyCode code)
 {
+	// This can happen if the user presses a key which has no corresponding mapping,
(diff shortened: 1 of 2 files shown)


sunzenshen (approved):

AdamTadeusz:
Now that these are refreshed every time, and each of them is used in one place only, shouldn't we instead just remove the ButtonCode_t vars from the Keys struct?

Rainyan:
> Now that these are refreshed every time, and each of them is used in one place only, shouldn't we instead just remove the ButtonCode_t vars from the Keys struct?

Probably. But this is really just a hotfix and not a refactor.

PR #1234, opened 2025-07-27
Input to refill a player's clips

DESTROYGIRL:
## Description
For mapping & debugging
Test with `ent_fire player refillammo`

## Toolchain

- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/neo_player.cpp
+++ b/src/game/server/neo/neo_player.cpp
@@ -119,6 +119,7 @@ DEFINE_FIELD(m_bClientWantNeoName, FIELD_BOOLEAN),
 
 // Inputs
 DEFINE_INPUTFUNC(FIELD_STRING, "SetPlayerModel", InputSetPlayerModel),
+DEFINE_INPUTFUNC(FIELD_VOID, "RefillAmmo", InputRefillAmmo),
 
 END_DATADESC()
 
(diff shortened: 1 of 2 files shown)


src/game/server/neo/neo_player.cpp
```diff
@@ -2280,6 +2281,20 @@ void CNEO_Player::InputSetPlayerModel( inputdata_t & inputdata )
 	}
 }
 
+void CNEO_Player::InputRefillAmmo( inputdata_t & inputdata)
+{
+	CBaseCombatWeapon* pWeapon = GetActiveWeapon();
+
+	if (pWeapon)
+	{
+		pWeapon->SetPrimaryAmmoCount(pWeapon->GetDefaultClip1());
```
sunzenshen, line 2290: This functionality is for situations like the specific weapon tutorials right?

I'm not a mapper, but my impression from this code is that this would only refill the total ammo for a player to the max size of the clip/magazine.

I wonder if something along the lines of the neo_player.cpp spawn code would be of interest, such as this call stack:
- CNEO_Player::GiveDefaultItems(void)
- CNEO_Player::GiveLoadoutWeapon(void)
- CNEO_Player::Spawn(void)

DESTROYGIRL: I only want to refill clips, as I'm using this to create pseudo-infinite ammo for the firing range, but still have the player reload when the mag is empty.
Giving weapons and so fourth is already possible, but not filling/modifying ammo of owned weapons

sunzenshen (approved):
Okay, makes sense that it's being used in that context.

AdamTadeusz (approved):

AdamTadeusz:
Now that I said that I realise InputRefillAmmo doesn't specify which weapon it refills, Maybe this input you have here now should be called InputRefillActiveAmmo

DESTROYGIRL:
I'd prefer to make a specialist entity that targets the !activator for managing a player's weapons & ammo rather than trying to cram it all in player inputs because it would be easier for mappers but for now this works

PR #1238, opened 2025-07-27
NeoUI - Refactor + add vertical per row-cell layout

### LINKED ISSUE #1246: Non-rendered rows are still highlighted
https://github.com/user-attachments/assets/514d4ba0-1b71-42ab-91cf-1432dfbc37d3

nullsystem:
## Description
* Refactor layouting related section mostly so it's more workable with the vertical layouting and more clear start-end point during alterting the layouting
* New vertical per row-cell layouting which allows multiple sub-rows used, mostly for label-only decorated visuals
* `xWide` change from multiple ternary to if statements for clearer readability
* Split off from the WIP killer damage HUD and only benefits that PR for now
* Fixed row highlights

## Toolchain

## Linked Issues

- related #1190
- fixes #1246

### DIFF
--- a/src/game/client/neo/ui/neo_ui.cpp
+++ b/src/game/client/neo/ui/neo_ui.cpp
@@ -37,75 +37,98 @@ void SwapColorNormal(const Color &color)
 
 void MultiWidgetHighlighter(const int iTotalWidgets)
 {
-	if (c->eMode == MODE_PAINT)
+	// NEO NOTE (nullsystem): c->iLayoutY instead of c->irWidgetLayoutY
+	// as this happens before rather than during widget Begin/End
+	if (c->eMode != MODE_PAINT || !IN_BETWEEN_AR(0, c->iLayoutY, c->dPanel.tall))
 	{
-		// Peek-forward what the area of the multiple widgets will cover without modifying the context
-		int iMulLayoutX = c->iLayoutX;
-		int iMulLayoutY = c->iLayoutY;
-		int iFirstLayoutX = iMulLayoutX;
-		int iFirstLayoutY = iMulLayoutY;
-		int iMulIdxRowParts = c->iIdxRowParts;
-		int iXWide = 0;
-		for (int i = 0; i < iTotalWidgets; ++i)
+		return;
+	}
+
+	Assert(c->layout.iRowPartsTotal > 0);
+	const int iRowPartsTotal = Max(1, c->layout.iRowPartsTotal);
+
+	// Peek-forward what the area of the multiple widgets will cover without modifying the context
+	int iMulLayoutX = c->iLayoutX;
+	int iMulLayoutY = c->iLayoutY;
+	int iAllXWide = 0;
+	int iMulIdxRowParts = c->iIdxRowParts;
+
+	if (iMulIdxRowParts < 0)
+	{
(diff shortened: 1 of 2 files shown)


src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -41,18 +41,45 @@ void MultiWidgetHighlighter(const int iTotalWidgets)
...
+			int xWide;
+			if (bNextIsLast)
+			{
+				xWide = c->dPanel.wide - iMulLayoutX;
+			}
+			else
+			{
+				if (c->layout.iRowParts)
+				{
+					xWide = (c->layout.iRowParts[iMulIdxRowParts] / 100.0f) * c->dPanel.wide;
```
sunzenshen: This array access into iRowParts seems a bit raw without bounds checking.  Is there any context that could be used to sanity check the array size? (Is it worth sanity check against iRowPartsTotal? e.g. [line 506](https://github.com/NeotokyoRebuild/neo/pull/1238/files#diff-4b8343ff499c3b26df2ff819dec5e3ee84c2ae123eed5fcac49d33ec41119ccdR506))

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -41,18 +41,45 @@ void MultiWidgetHighlighter(const int iTotalWidgets)
...
+				}
+				else
+				{
+					xWide = (1.0f / static_cast<float>(c->layout.iRowPartsTotal)) * c->dPanel.wide;
```
sunzenshen: Would it be worth sanity checking that iRowPartsTotal is more than 0?

Something like the sanity check here?
[line 506](https://github.com/NeotokyoRebuild/neo/pull/1238/files#diff-4b8343ff499c3b26df2ff819dec5e3ee84c2ae123eed5fcac49d33ec41119ccdR506)

src/game/client/neo/ui/neo_ui.cpp
sunzenshen: This cramming of a float calculation into an int might involve subtle precision issues with float.  Is it possible to avoid the float division and make make a comparison with just ints?  Or a rounding to known good integer values?  Depends on the exact intention of precision requirements of xWide.

src/game/client/neo/ui/neo_ui.cpp
nullsystem: I think `iRowPartsTotal <= 0` shouldn't actually be thing, but guess just in-case instead will do an early return in this function and some assert checking on when it gets set.

src/game/client/neo/ui/neo_ui.cpp
nullsystem: `bNextIsLast` is the sanity checking against `iRowPartsTotal`, final item and above won't hit this. `< 0` already covered before the loop.

I don't think precision matters too much in this case, it looks and layouts good enough already and so far the UI haven't got a need for very precise controls yet.

AdamTadeusz (approved):

nullsystem:
Added sanity checks + asserts for `iRowPartsTotal`, fixed row highlights

sunzenshen (approved):

AdamTadeusz (approved):

PR #1248, opened 2025-07-31
NeoNextBots shoot at visible parts of other clients

AdamTadeusz:
## Description

Previously the bots would always shoot center mass, but IsLineOfSightClearToEntity checks center mass, eye pos and origin. This pr adds a map to NextBotVisionInterface that stores the ideal position for the bot to aim at when shooting at entities.

### DIFF
--- a/src/game/server/NextBot/NextBotVisionInterface.cpp
+++ b/src/game/server/NextBot/NextBotVisionInterface.cpp
@@ -775,8 +775,10 @@ bool IVision::IsLineOfSightClearToEntity( const CBaseEntity *subject, Vector *vi
 	{
 		*visibleSpot = result.endpos;
 	}
+#ifdef NEO
 
 	idealTargetPoint[subject->entindex()] = result.endpos;
+#endif // NEO
 
 	return ( result.fraction >= 1.0f && !result.startsolid );
 
--- a/src/game/server/NextBot/NextBotVisionInterface.cpp
+++ b/src/game/server/NextBot/NextBotVisionInterface.cpp
@@ -565,7 +565,11 @@ void IVision::Update( void )
 
 
 //------------------------------------------------------------------------------------------
+#ifdef NEO
+bool IVision::IsAbleToSee( CBaseEntity *subject, FieldOfViewCheckType checkFOV, Vector *visibleSpot )
+#else
 bool IVision::IsAbleToSee( CBaseEntity *subject, FieldOfViewCheckType checkFOV, Vector *visibleSpot ) const
+#endif // NEO
 {
 	VPROF_BUDGET( "IVision::IsAbleToSee", "NextBotExpensive" );
 
@@ -708,7 +712,11 @@ bool IVision::IsLineOfSightClear( const Vector &pos ) const
 
 
 //------------------------------------------------------------------------------------------
+#ifdef NEO
+bool IVision::IsLineOfSightClearToEntity( const CBaseEntity *subject, Vector *visibleSpot )
+#else
 bool IVision::IsLineOfSightClearToEntity( const CBaseEntity *subject, Vector *visibleSpot ) const
+#endif // NEO
 {
 #ifdef TERROR
 	// TODO: Integration querycache & its dependencies
@@ -767,6 +775,8 @@ bool IVision::IsLineOfSightClearToEntity( const CBaseEntity *subject, Vector *vi
 	{
 		*visibleSpot = result.endpos;
 	}
+
+	idealTargetPoint[subject->entindex()] = result.endpos;
 
 	return ( result.fraction >= 1.0f && !result.startsolid );
 
--- a/src/game/server/NextBot/NextBotVisionInterface.h
+++ b/src/game/server/NextBot/NextBotVisionInterface.h
@@ -8,6 +8,9 @@
 
 #include "NextBotComponentInterface.h"
 #include "NextBotKnownEntity.h"
+#ifdef NEO
+#include <map>
+#endif // NEO
 
 class IBody;
 class INextBotEntityFilter;
@@ -81,7 +84,11 @@ public:
 	 * If 'visibleSpot' is non-NULL, the highest priority spot on the subject that is visible is returned.
 	 */ 
 	enum FieldOfViewCheckType { USE_FOV, DISREGARD_FOV };
+#ifdef NEO
+	virtual bool IsAbleToSee( CBaseEntity *subject, FieldOfViewCheckType checkFOV, Vector *visibleSpot = NULL );
+#else
 	virtual bool IsAbleToSee( CBaseEntity *subject, FieldOfViewCheckType checkFOV, Vector *visibleSpot = NULL ) const;
+#endif // NEO
 	virtual bool IsAbleToSee( const Vector &pos, FieldOfViewCheckType checkFOV ) const;
 
 	virtual bool IsIgnored( CBaseEntity *subject ) const;		// return true to completely ignore this entity (may not be in sight when this is called)
@@ -102,7 +109,12 @@ public:
 	 * Returns true if the ray between the position and the subject is unobstructed.
 	 * A visible spot on the subject is returned in 'visibleSpot'.
 	 */
+#ifdef NEO // May as well update idealTargetPoint here to not do extra tracelines
+	virtual bool IsLineOfSightClearToEntity( const CBaseEntity *subject, Vector *visibleSpot = NULL );
+	std::map<int, Vector> idealTargetPoint;
+#else
 	virtual bool IsLineOfSightClearToEntity( const CBaseEntity *subject, Vector *visibleSpot = NULL ) const;
+#endif // NEO
 
 	/// @todo: Implement LookAt system
 	virtual bool IsLookingAt( const Vector &pos, float cosTolerance = 0.95f ) const;					// are we looking at the given position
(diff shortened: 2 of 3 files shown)


src/game/server/NextBot/NextBotVisionInterface.h
```diff
@@ -102,7 +109,12 @@ class IVision : public INextBotComponent
 	 * Returns true if the ray between the position and the subject is unobstructed.
 	 * A visible spot on the subject is returned in 'visibleSpot'.
 	 */
+#ifdef NEO // May as well update idealTargetPoint here to not do extra tracelines
+	virtual bool IsLineOfSightClearToEntity( const CBaseEntity *subject, Vector *visibleSpot = NULL );
+	std::map<int, Vector> idealTargetPoint;
```
sunzenshen: Would it be simpler to mark idealTargetPoint as mutable and keep the existing const version of IsAbleToSee()?

sunzenshen: Supposedly Valve has a C++ CUtlMap type that is a real-time friendly version of map, but don't have any experience using it, so up to you if trying it out makes sense:

https://developer.valvesoftware.com/wiki/CUtlMap

src/game/server/NextBot/NextBotVisionInterface.cpp
```diff
@@ -767,6 +775,10 @@ bool IVision::IsLineOfSightClearToEntity( const CBaseEntity *subject, Vector *vi
 	{
 		*visibleSpot = result.endpos;
 	}
+#ifdef NEO
+
+	idealTargetPoint[subject->entindex()] = result.endpos;
```
sunzenshen: Does this map keep growing? I don't notice a clearing of this map in this PR.

Would it be possible to expose a map clearing function in here, and clear the map every time a bot spawns or dies?  I guess the tricky aspect is that I recall that the spawn code is in CNEO_Player, but maybe there's another place that makes sense to regularly clear this data structure.

For example, maybe in `neo\src\game\server\neo\bot\behavior\neo_bot_seek_and_destroy.cpp` when a bot has lost track of an enemy and they are roaming again.

src/game/server/NextBot/NextBotVisionInterface.cpp
```diff
@@ -708,7 +712,11 @@ bool IVision::IsLineOfSightClear( const Vector &pos ) const
 
 
 //------------------------------------------------------------------------------------------
+#ifdef NEO
+bool IVision::IsLineOfSightClearToEntity( const CBaseEntity *subject, Vector *visibleSpot )
+#else
 bool IVision::IsLineOfSightClearToEntity( const CBaseEntity *subject, Vector *visibleSpot ) const
```
sunzenshen: I guess this is a convention since the case is so rare and the pattern is present in the other macro cases, but what is the likelihood that subject will ever be a null value due to something happening before this function?  If it's a known convention that this is unlikely I guess that's okay, but wanted to make sure.

src/game/server/NextBot/NextBotVisionInterface.cpp
```diff
@@ -767,6 +775,10 @@ bool IVision::IsLineOfSightClearToEntity( const CBaseEntity *subject, Vector *vi
 	{
 		*visibleSpot = result.endpos;
 	}
+#ifdef NEO
+
+	idealTargetPoint[subject->entindex()] = result.endpos;
+#endif // NEO
 
 	return ( result.fraction >= 1.0f && !result.startsolid );
```
sunzenshen, line 821: Would it make sense to check if there was a successful trace, before writing a value into idealTargetPoint?  My impression so far is that the update to valid target points happens even if all the traces fail.

src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
```diff
@@ -232,6 +232,11 @@ Vector CNEOBotMainAction::SelectTargetPoint( const INextBot *meBot, const CBaseC
 			}
 		}
 
+		if (me->GetVisionInterface()->idealTargetPoint.contains(subject->entindex()))
+		{
+			return me->GetVisionInterface()->idealTargetPoint[subject->entindex()];
```
sunzenshen: It seems like the idealTargetPoint map is searched twice by this point.  Maybe it's possible to use a C++ iterator?  I don't exactly know what the handling characteristics are with Valve's `CUtlMap`, but maybe you could just query the value of the iterator to check for success. e.g.:

```cpp
auto &map = me->GetVisionInterface()->GetIdealTargetMap();
auto it = map.Find( subject->entindex() );
if ( it != map.InvalidIndex() )
    return map[ it ];
```

src/game/server/NextBot/NextBotVisionInterface.h
AdamTadeusz: https://tenor.com/view/the-burnt-peanut-excellent-work-nice-surprised-wow-gif-12418556657724337676

src/game/server/NextBot/NextBotVisionInterface.cpp
AdamTadeusz: INextBotComponent::Reset is called on bot spawn for all components, so clearing it on spawn should be easy

src/game/server/NextBot/NextBotVisionInterface.cpp
AdamTadeusz: so far every functions that calls IsLineOfSightClearToEntity checks that subject isn't null before reaching this point, except in the case of IsAbleToSee where GetBot()->IsRangeGreaterThan( subject, ... returns true since subject is null and thus isAbleToSee also returns early

src/game/server/NextBot/NextBotVisionInterface.cpp
AdamTadeusz: doesn't affect bot behavior currently but added a check anyway as unsure how expensive InsertOrReplace is

nullsystem:
This can be changed to give headshot aiming a higher precedence. Somewhere down the line when there's ability to customize bots can be used to differentiate "experts" and "normal" bots for example and add more logics like weapon type and distance.

```diff
diff --git a/src/game/server/NextBot/NextBotVisionInterface.cpp b/src/game/server/NextBot/NextBotVisionInterface.cpp
index 5b9474fc..d316e072 100644
--- a/src/game/server/NextBot/NextBotVisionInterface.cpp
+++ b/src/game/server/NextBot/NextBotVisionInterface.cpp
@@ -773,6 +773,18 @@ bool IVision::IsLineOfSightClearToEntity( const CBaseEntity *subject, Vector *vi
        trace_t result;
        NextBotTraceFilterIgnoreActors filter( subject, COLLISION_GROUP_NONE );

+#ifdef NEO
+       UTIL_TraceLine( GetBot()->GetBodyInterface()->GetEyePosition(), subject->EyePosition(), MASK_BLOCKLOS_AND_NPCS|CONTENTS_IGNORE_NODRAW_OPAQUE, &filter, &result );
+       if ( result.DidHit() )
+       {
+               UTIL_TraceLine( GetBot()->GetBodyInterface()->GetEyePosition(), subject->WorldSpaceCenter(), MASK_BLOCKLOS_AND_NPCS|CONTENTS_IGNORE_NODRAW_OPAQUE, &filter, &result );
+
+               if ( result.DidHit() )
+               {
+                       UTIL_TraceLine( GetBot()->GetBodyInterface()->GetEyePosition(), subject->GetAbsOrigin(), MASK_BLOCKLOS_AND_NPCS|CONTENTS_IGNORE_NODRAW_OPAQUE, &filter, &result );
+               }
+       }
+#else
        UTIL_TraceLine( GetBot()->GetBodyInterface()->GetEyePosition(), subject->WorldSpaceCenter(), MASK_BLOCKLOS_AND_NPCS|CONTENTS_IGNORE_NODRAW_OPAQUE, &filter, &result );
        if ( result.DidHit() )
        {
@@ -783,6 +795,7 @@ bool IVision::IsLineOfSightClearToEntity( const CBaseEntity *subject, Vector *vi
                        UTIL_TraceLine( GetBot()->GetBodyInterface()->GetEyePosition(), subject->GetAbsOrigin(), MASK_BLOCKLOS_AND_NPCS|CONTENTS_IGNORE_NODRAW_OPAQUE, &filter, &result );
                }
        }
+#endif

        if ( visibleSpot )
        {
```

sunzenshen (approved):
I was curious to know how https://github.com/NeotokyoRebuild/neo/commit/4e44af93601e9c353f455f2482fe52aa8bb78a88 would affect the behavior of this PR, since there didn't seem to be a merge conflict.

Locally I rebased this branch on top of upstream/master https://github.com/NeotokyoRebuild/neo/commit/e80e9b0e98ca1f9410a05f477f4ae8a903df5e22 and observed some bot matches.  Overall the behavior seems smooth, though I'm not sure if there's any redundancies in routines now.  Up to you if you'd like to sanity test your expected behavior based on top of the latest master branch.

src/game/server/NextBot/NextBotVisionInterface.h
```diff
@@ -102,6 +102,9 @@ class IVision : public INextBotComponent
 	 * Returns true if the ray between the position and the subject is unobstructed.
 	 * A visible spot on the subject is returned in 'visibleSpot'.
 	 */
+#ifdef NEO // May as well update idealTargetPoint here to not do extra tracelines
+	mutable CUtlMap<int, Vector> idealTargetPoint;
```
nullsystem: Nit: Add `m_` prefix to the name like how other names in this class is.

nullsystem (approved):

PR #1249, opened 2025-07-31
Accurate RTT shadow angles relative to world lights

DESTROYGIRL:
## Description
A backported feature from the Alien Swarm SDK by [Saul Rennison](https://developer.valvesoftware.com/wiki/User:Saul)

Current issues:

- [ ] Transitioning between two lights at near opposite angles can make the object's shadow do a very jarring sweep

- [ ] When an object stops moving it makes a very obvious flick to the high fidelity shadow at certain angles

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/client/CMakeLists.txt
+++ b/src/game/client/CMakeLists.txt
@@ -1517,6 +1517,8 @@ target_sources_grouped(
     neo/neo_fixup_glshaders.cpp
     neo/neo_fixup_glshaders.h
     neo/c_neo_bloom_controller.cpp
+    neo/worldlight.cpp
+    neo/worldlight.h
 )
 
 target_sources_grouped(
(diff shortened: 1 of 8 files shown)


Rainyan (changes requested):
It seems that currently, the approach angle and player velocity affect the resolving of the shadow direction.

I'm worried this would have a negative effect on gameplay, because it's unintuitive where exactly your shadow will be displayed for the opponents—even when standing in the same exact location, it's possible to have a different shadow cast outcome depending on the route you took there and your speed:

[lerp.webm](https://github.com/user-attachments/assets/163888dd-3e0c-4711-936a-41513b86611d)

---

There's also this strange issue where fidgeting around will make the shadow slowly tend towards what I assume is the neutral (FLT_MAX?) shadow position:

[shadow.webm](https://github.com/user-attachments/assets/d1c3f700-fcda-423b-b866-c34ca1ac2d54)

which also would be a poor experience for the competitive meta, I fear.

PR #1250, opened 2025-08-01
Background map option readable

AdamTadeusz:
## Description

Makes the background map option readable

https://github.com/user-attachments/assets/d8c729ec-7869-4f67-b608-609e54334f05

## Toolchain

- Windows MSVC VS2022


src/game/client/neo/ui/neo_root_settings.cpp
```diff
@@ -690,7 +821,8 @@ void NeoSettings_General(NeoSettings *ns)
 	NeoUI::RingBox(L"Show FPS", SHOWFPS_LABELS, ARRAYSIZE(SHOWFPS_LABELS), &pGeneral->iShowFps);
 	NeoUI::RingBoxBool(L"Show rangefinder", &pGeneral->bEnableRangeFinder);
 	NeoUI::RingBoxBool(L"Extended Killfeed", &pGeneral->bExtendedKillfeed);
-	NeoUI::SliderInt(L"Selected Background", &pGeneral->iBackground, 1, 4); // NEO TODO (Adam) switch to RingBox with values read from ChapterBackgrounds.txt
+	
+	NeoUI::RingBox(L"Selected Background", const_cast<const wchar_t **>(ns->p2WszCBList), ns->iCBListSize, &pGeneral->iBackground);
```
AdamTadeusz, line 856: It would be cool if ringbox could take a KeyValues object to read from

.gitignore
```diff
@@ -43,8 +43,12 @@ out
 /game/neo/mapsrc/*.prt
 /game/neo/mapsrc/*.vmx
 
-# MP3 Databalse
+# MP3 Player
 /game/neo/resource/mp3player_db.txt
+/game/neo/resource/mp3settings.txt
+
+# Selected Background
+/game/neo/scripts/chapterbackgrounds.txt
```
Rainyan: The capitalization should match `/game/neo/scripts/ChapterBackgrounds.txt` exactly here, since AFAIK gitignore is case-sensitive. So this wouldn't correctly ignore for filesystems that are case-sensitive (Linux).

src/game/client/neo/ui/neo_root_settings.cpp
```diff
@@ -234,12 +235,140 @@ void NeoSettingsInit(NeoSettings *ns)
 			vgui::surface()->DrawGetTextureSize(pTex->iTexId, pTex->iWide, pTex->iTall);
 		}
 	}
+
+	NeoSettingsBackgroundsInit(ns);
+}
+
+// sv_unlockedchapters is not a consistent way of guaranteeing what background shows up as the menu background.
+// Instead, we read all available backgrounds from a second file neo_backgrounds.txt, and overwrite chapterbackgrounds.txt
+// to only include the one background that the user is insterested in.
```
Rainyan: Nit: typo `insterested` -> `interested` (since we gotta fixup the gitignore case anyways so might as well point it out 😄)

src/game/client/neo/ui/neo_root_settings.cpp
```diff
@@ -234,12 +235,140 @@ void NeoSettingsInit(NeoSettings *ns)
...
+#define NEO_FALLBACK_BACKGROUND_FILENAME "background01"
+
+	if (ns->backgrounds)
+	{
+		ns->backgrounds->deleteThis();
+	}
+	ns->backgrounds = new KeyValues( "neo_backgrounds" );
+	ns->iCBListSize = 0;
+
+	const auto allocate = [](NeoSettings *ns, int size) {
```
Rainyan: Nit: this can be `constexpr` instead of `const`

src/game/client/neo/ui/neo_root_settings.cpp
```diff
@@ -234,12 +235,140 @@ void NeoSettingsInit(NeoSettings *ns)
...
+		allocate(ns, dispSize);
+		g_pVGuiLocalize->ConvertANSIToUnicode(NEO_FALLBACK_BACKGROUND_DISPLAYNAME, ns->p2WszCBList[0], sizeof(wchar_t) * dispSize);
+		NeoSettingsBackgroundWrite(ns, NEO_FALLBACK_BACKGROUND_FILENAME);
+		return;
+	}
+
+	for ( KeyValues* background = ns->backgrounds->GetFirstSubKey(); background != NULL; /*background = background->GetNextKey()*/)
+	{ // Iterate once to get the number of options and longest background map name
+		const char* displayName = background->GetName();
+		if (displayName == "") // NEO NOTE (Adam) If name missing read will fail
```
Rainyan: This is comparing the `const char*` numeric ptr address of `displayName` to the address of the statically stored string literal `""`. Even if they're both equal in contents, we don't have guarantees that they're contained at the same pointer address, so I think this is probably not correct.

Maybe you meant `if (!*displayName)` or similar, to dereference the charptr and check whether its first character was the null terminator? Or if there's a chance of it returning nullptr, `if (!displayName || !*displayName)` to prevent the nullptr dereference.

We also have the util function: https://github.com/NeotokyoRebuild/neo/blob/30ef642d932c308c7fe533de665d6c621f70c904/src/game/client/cdll_util.h#L161-L165 if you prefer writing: `if (FStrEq(displayName, "")` instead.

src/game/client/neo/ui/neo_root_settings.cpp
```diff
@@ -234,12 +235,140 @@ void NeoSettingsInit(NeoSettings *ns)
...
+		{ // no display name, skip
+			KeyValues* thisKey = background;
+			background = background->GetNextKey();
+			ns->backgrounds->RemoveSubKey(thisKey);
+			continue;
+		}
+
+		const char* fileName = background->GetString("fileName");
+		if (fileName == "")
```
Rainyan: Ditto on pointer compare logic problem.

src/game/client/neo/ui/neo_root_settings.cpp
```diff
@@ -235,12 +236,140 @@ void NeoSettingsInit(NeoSettings *ns)
...
+	ns->backgrounds = new KeyValues( "neo_backgrounds" );
+	ns->iCBListSize = 0;
+
+	constexpr auto allocate = [](NeoSettings *ns, int size) {
+		ns->p2WszCBList = (wchar_t **)calloc(sizeof(wchar_t *), ns->iCBListSize);
+		ns->p2WszCBList[0] = (wchar_t *)calloc(sizeof(wchar_t) * size, ns->iCBListSize);
+	};
+
+	// Setup Background Map options
+	int dispSize = max(sizeof(NEO_FALLBACK_BACKGROUND_DISPLAYNAME) + 1, sizeof(NEO_RANDOM_BACKGROUND_NAME) + 1);
```
Rainyan: Nit: `sizeof` of a string literal already includes the null terminator, so the `+ 1` is probably redundant here. Also, this assignment can be declared `constexpr`.

src/game/client/neo/ui/neo_root_settings.cpp
```diff
@@ -235,12 +236,140 @@ void NeoSettingsInit(NeoSettings *ns)
...
+		if (FStrEq(fileName, ""))
+		{ // no file name, skip
+			KeyValues* thisKey = background;
+			background = background->GetNextKey();
+			ns->backgrounds->RemoveSubKey(thisKey);
+			continue;
+		}
+
+		ns->iCBListSize++;
+		dispSize = max(dispSize, (sizeof(wchar_t) / sizeof(char)) * (sizeof(displayName) + 1));
```
Rainyan: Unless I'm misunderstanding what this is calculating, this logic might be incorrect: `sizeof(displayName)` would return the sizeof the ptr at compile time, consider for example:
```cpp
constexpr size_t ptrSize = sizeof(const char*);
constexpr const char* someString = "Just some example string.";
static_assert(ptrSize == sizeof(someString)); // this is true!
```
So this is probably not what we want, but `V_strlen(displayName) + 1` instead? Also the `sizeof(wchar_t) / sizeof(char)` seems unuseful here since we do the `sizeof(wchar_t) * dispSize` math below at line 301.

Would also prefer `max` -> `Max` to avoid the "max" C macro usage in preference of the "Max" function usage.

src/game/client/neo/ui/neo_root_settings.cpp
AdamTadeusz: I'm updating the value of dispSize later if not returning early

Rainyan:
@AdamTadeusz Is there a chance your version of `ChapterBackgrounds.txt` doesn't follow what exact capitalization in the gitignore? As a follow-up on the gitignore problem.

Rainyan (approved):

sunzenshen (approved):
Verified that when look at the menu, the names of the maps are more readable than the integer numbers that were previously there in other builds.  Also changed the background map, and checked that the associated map is loaded after restarting the game.

PR #1256, opened 2025-08-03
NeoUI - Clipboard + selection + free-cursor for TextEdit

### LINKED ISSUE #1255: [FEATURE] NeoUI - Cut, Copy, and Paste TextEdit
### Description

Add a right click menu and Ctrl-XCV to do cut, copy, and paste

### Proposed Solution

* Right-click menu that have cut, copy, and paste option
* Ctrl X - Cut, Ctrl C - Copy, Ctrl V - Paste
* TextEdit altered to have text selectable and cursor in different place

### Alternatives Considered

_No response_

### Additional Context

_No response_

nullsystem:
## Description
* Add right-click menu for NeoUI, only used by TextEdit for clipboard functionality
* Clipboard actions envoked both by keybind and right-click menu
* Add selection support for TextEdit
* Cursor set on any position other than final + paste in-between texts
* Text selection with shift modifier + HOME/END/LEFT/RIGHT
* Prev/next word with ctrl modifier + LEFT/RIGHT (vgui ctrl modifier behavior)
* Password TextEdit toggle now a flag in function rather than context bool variable
* TODO later?: Multi-lined text editor, read-only mode, slider's text editor. Those can be done later as this PR is already long.
* Refactored NeoUI mouse handling, added ability to override foreground/text color
* Mark create server password field as password if streamer mode is on

## Toolchain

## Linked Issues

* fixes #1255

### DIFF
--- a/src/game/client/neo/ui/neo_root_settings.cpp
+++ b/src/game/client/neo/ui/neo_root_settings.cpp
@@ -672,8 +672,8 @@ static_assert(ARRAYSIZE(DLFILTER_STRMAP) == ARRAYSIZE(DLFILTER_LABELS));
 void NeoSettings_General(NeoSettings *ns)
 {
 	NeoSettings::General *pGeneral = &ns->general;
-	NeoUI::TextEdit(L"Name", pGeneral->wszNeoName, MAX_PLAYER_NAME_LENGTH);
-	NeoUI::TextEdit(L"Clan tag", pGeneral->wszNeoClantag, NEO_MAX_CLANTAG_LENGTH);
+	NeoUI::TextEdit(L"Name", pGeneral->wszNeoName, MAX_PLAYER_NAME_LENGTH - 1);
+	NeoUI::TextEdit(L"Clan tag", pGeneral->wszNeoClantag, NEO_MAX_CLANTAG_LENGTH - 1);
 	NeoUI::RingBoxBool(L"Show only steam name", &pGeneral->bOnlySteamNick);
 	NeoUI::RingBoxBool(L"Friendly marker spectator only clantags", &pGeneral->bMarkerSpecOnlyClantag);
 
--- a/src/game/client/neo/ui/neo_root_settings.cpp
+++ b/src/game/client/neo/ui/neo_root_settings.cpp
@@ -245,6 +245,7 @@ void NeoSettingsDeinit(NeoSettings *ns)
 void NeoSettingsRestore(NeoSettings *ns, const NeoSettings::Keys::Flags flagsKeys)
 {
 	ns->bModified = false;
+	ns->bIsValid = true;
 	NeoSettings::CVR *cvr = &ns->cvr;
 	{
 		NeoSettings::General *pGeneral = &ns->general;
@@ -671,14 +672,27 @@ static_assert(ARRAYSIZE(DLFILTER_STRMAP) == ARRAYSIZE(DLFILTER_LABELS));
 void NeoSettings_General(NeoSettings *ns)
 {
 	NeoSettings::General *pGeneral = &ns->general;
-	NeoUI::TextEdit(L"Name", pGeneral->wszNeoName, SZWSZ_LEN(pGeneral->wszNeoName));
-	NeoUI::TextEdit(L"Clan tag", pGeneral->wszNeoClantag, SZWSZ_LEN(pGeneral->wszNeoClantag));
+	NeoUI::TextEdit(L"Name", pGeneral->wszNeoName, MAX_PLAYER_NAME_LENGTH);
+	NeoUI::TextEdit(L"Clan tag", pGeneral->wszNeoClantag, NEO_MAX_CLANTAG_LENGTH);
 	NeoUI::RingBoxBool(L"Show only steam name", &pGeneral->bOnlySteamNick);
 	NeoUI::RingBoxBool(L"Friendly marker spectator only clantags", &pGeneral->bMarkerSpecOnlyClantag);
 
 	wchar_t wszTotalClanAndName[NEO_MAX_DISPLAYNAME];
-	GetClNeoDisplayName(wszTotalClanAndName, pGeneral->wszNeoName, pGeneral->wszNeoClantag, pGeneral->bOnlySteamNick);
-	NeoUI::Label(L"Display name", wszTotalClanAndName);
+	ns->bIsValid = GetClNeoDisplayName(wszTotalClanAndName,
+			pGeneral->wszNeoName, pGeneral->wszNeoClantag,
+			((pGeneral->bOnlySteamNick) ?
+			 	CL_NEODISPLAYNAME_FLAG_CHECK | CL_NEODISPLAYNAME_FLAG_ONLYSTEAMNICK :
+				CL_NEODISPLAYNAME_FLAG_CHECK));
+	if (ns->bIsValid)
+	{
+		NeoUI::Label(L"Display name", wszTotalClanAndName);
+	}
+	else
+	{
+		NeoUI::BeginOverrideFgColor(COLOR_RED);
+		NeoUI::Label(L"Display name (invalid)", wszTotalClanAndName);
+		NeoUI::EndOverrideFgColor();
+	}
 
 	NeoUI::SliderInt(L"FOV", &pGeneral->iFov, 75, 110);
 	NeoUI::SliderInt(L"Viewmodel FOV Offset", &pGeneral->iViewmodelFov, -20, 40);
--- a/src/game/client/neo/ui/neo_ui.cpp
+++ b/src/game/client/neo/ui/neo_ui.cpp
@@ -2,12 +2,16 @@
 
 #include <vgui/ILocalize.h>
 #include <vgui/IInput.h>
+#include <vgui/ISystem.h>
 #include <vgui_controls/Controls.h>
 #include <filesystem.h>
 #include <stb_image.h>
 #include <materialsystem/imaterial.h>
 
 #include "neo_misc.h"
+
+// memdbgon must be the last include file in a .cpp file!!!
+#include "tier0/memdbgon.h"
 
 static const wchar_t *ENABLED_LABELS[] = {
 	L"Disabled",
@@ -22,6 +26,11 @@ static Context *c = &g_emptyCtx;
 const int ROWLAYOUT_TWOSPLIT[] = { 40, -1 };
 static constexpr int WDGINFO_ALLOC_STEPS = 64;
 
+#define DEBUG_NEOUI 0 // NEO NOTE (nullsystem): !!! Always flip to 0 on master + PR !!!
+#ifndef DEBUG
+static_assert(DEBUG_NEOUI == 0);
+#endif
+
 void SwapFont(const EFont eFont, const bool bForce)
 {
 	if (c->eMode != MODE_PAINT || (!bForce && (c->eFont == eFont))) return;
@@ -33,6 +42,17 @@ void SwapColorNormal(const Color &color)
 {
 	c->normalBgColor = color;
 	vgui::surface()->DrawSetColor(c->normalBgColor);
+}
+
+void BeginOverrideFgColor(const Color &fgColor)
+{
+	c->cOverrideFgColor = fgColor;
+	c->bIsOverrideFgColor = true;
+}
+
+void EndOverrideFgColor()
+{
+	c->bIsOverrideFgColor = false;
 }
 
 void MultiWidgetHighlighter(const int iTotalWidgets)
@@ -160,7 +180,6 @@ void BeginContext(NeoUI::Context *pNextCtx, const NeoUI::Mode eMode, const wchar
 	c->bValueEdited = false;
 	c->eButtonTextStyle = TEXTSTYLE_CENTER;
 	c->eLabelTextStyle = TEXTSTYLE_LEFT;
-	c->bTextEditIsPassword = false;
 	c->selectBgColor = COLOR_NEOPANELSELECTBG;
 	c->normalBgColor = COLOR_NEOPANELACCENTBG;
 	// Different pointer, change context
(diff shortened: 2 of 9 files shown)


src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -228,6 +230,100 @@ void BeginContext(NeoUI::Context *pNextCtx, const NeoUI::Mode eMode, const wchar
...
+		case MODE_KEYTYPED:
+		{
+			bDeInitRightClick = true;
+		} break;
+		case MODE_MOUSEPRESSED:
+		{
+			bDeInitRightClick = true;
+			if (c->dimRightClick.wide == 0 || c->dimRightClick.tall == 0)
+			{
+				// Wide/tall set to 0 indicates it's a newly initialized right-click menu
```
Rainyan: Nit: the comment should probably go above line 276 that it is commenting about, for readability. Since at least personally I find it a bit confusing to see a comment about a condition we already entered, inside the condition scope.

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -228,6 +230,100 @@ void BeginContext(NeoUI::Context *pNextCtx, const NeoUI::Mode eMode, const wchar
...
+							iNextY + c->layout.iRowTall);
+				}
+				const wchar *pwszItemLabel = c->pwszRightClickList[i];
+				vgui::surface()->DrawSetTextPos(dim.x + c->iMarginX,
+						iNextY + c->iMarginY);
+				vgui::surface()->DrawPrintText(pwszItemLabel, V_wcslen(pwszItemLabel));
+			}
+			vgui::surface()->DrawSetColor(c->normalBgColor);
+		} break;
+		case MODE_KEYPRESSED:
```
Rainyan, line 285: Nit: would prefer annotating intentional fallthrough with `case MODE_KEYPRESSED: [[fallthrough]];` attribute

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -228,6 +230,100 @@ void BeginContext(NeoUI::Context *pNextCtx, const NeoUI::Mode eMode, const wchar
...
+			}
+		} break;
+		case MODE_MOUSEMOVED:
+		{
+			const Dim &dim = c->dimRightClick;
+			if (IN_BETWEEN_EQ(dim.x, c->iMouseAbsX, dim.x + dim.wide) &&
+					IN_BETWEEN_EQ(dim.y, c->iMouseAbsY, dim.y + dim.tall))
+			{
+				const int iRelMenuPosY = c->iMouseAbsY - dim.y;
+				c->iRightClickHotItem = (iRelMenuPosY / static_cast<float>(dim.tall)) * c->ipwszRightClickListSize;
```
Rainyan, line 323: Can `dim.tall` ever equal 0 here? Since in the `c->iRightClickHotItem` assignment, we're implicitly casting from float to int, and as per spec:
> 7.3.10 Floating-integral conversions [conv.fpint]
1 A prvalue of a floating-point type can be converted to a prvalue of an integer type. The conversion truncates;
that is, the fractional part is discarded. The behavior is undefined if the truncated value cannot be represented
in the destination type.

And since the IEEE 754 defined div-by-zero could result in +infinity, NaN, etc, which wouldn't be representable as int, that could trigger UB for a case where `dim.tall == 0`.

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include <vgui/ILocalize.h>
 #include <vgui/IInput.h>
+#include <vgui/ISystem.h>
```
Rainyan, line 5: Preferably add the:
```cpp
#include "tier0/memdbgon.h"
```
as the last include of a cpp file.

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -22,6 +23,8 @@ static Context *c = &g_emptyCtx;
 const int ROWLAYOUT_TWOSPLIT[] = { 40, -1 };
 static constexpr int WDGINFO_ALLOC_STEPS = 64;
 
+#define DEBUG_NEOUI 0 // NEO NOTE (nullsystem): !!! Always flip to 0 on master + PR !!!
+
```
Rainyan: Maybe add:
```cpp
#ifndef DEBUG
static_assert(!DEBUG_NEOUI);
#endif
```
below this line, to catch the problem in CI builds if this ever was accidentally flipped?

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -1539,71 +1698,264 @@ void TextEdit(wchar_t *wszText, const int iMaxBytes)
...
+				case KEY_RIGHT:
+				{
+					c->iTextSelCur = clamp(c->iTextSelCur + ((c->eCode == KEY_LEFT) ? -1 : +1), 0, iWszTextSize);
+					if (!bIsShiftDown)
+					{
+						c->iTextSelStart = c->iTextSelCur;
+					}
+				} break;
+				case KEY_BACKSPACE:
+				case KEY_DELETE:
```
Rainyan: Nit: preferably add `[[fallthrough]]` attribute for the switch fallthrough cases: `KEY_HOME`, `KEY_LEFT`, `KEY_BACKSPACE`

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -1516,17 +1612,80 @@ void SliderU8(const wchar_t *wszLeftLabel, uint8 *ucValue, const uint8 iMin, con
...
+		const wchar_t wchTemp = wszMutVisText[i + 1];
+		wszMutVisText[i + 1] = '\0';
+		int iFontWide, iFontTall;
+		vgui::surface()->GetTextSize(pFontI->hdl, wszMutVisText, iFontWide, iFontTall);
+		c->irTextWidths[i] = iFontWide;
+		wszMutVisText[i + 1] = wchTemp;
+	}
+}
+
+void TextEdit(const wchar_t *wszLeftLabel, wchar_t *wszText, const int iMaxBytes, const ETextEditFlags flags)
```
Rainyan: The "iMaxBytes" variable name is a little confusing. It seems that at the call sites to this function, we are passing the string length (using `SZWSZ_LEN`) to `iMaxBytes`, and not the max bytes count.

This gets extra confusing later on, when we do:
```cpp
V_wcsncpy(wszText, wszStaticTmpText, sizeof(wchar_t) * iMaxBytes);
```
where based on the variable name "iMaxBytes", it looks like we're multiplying the max byte count by the byte size of the underlying data type, which would be incorrect. But this is of course not what's actually happening, since in reality we're multiplying the string max length by the data type size.

Maybe `iMaxBytes` should be renamed to something like `iMaxLength`, to avoid this confusion? Same thing with the other `TextEdit` overloaded function below at line 1685, and its `iMaxLength` parameter name.

src/game/client/neo/ui/neo_ui.cpp
nullsystem: They're empty cases though, so they're never warned by the compilers, even in gcc `-Wimplicit-fallthrough=5` at max level fallthrough warning doesn't flag `KEY_HOME`, `KEY_LEFT`, and `KEY_BACKSPACE` up at all. I think it's similar with MSVC also: https://learn.microsoft.com/en-us/cpp/code-quality/c26819?view=msvc-170 they even says it's fine to not mark them as fallthrough. I'd rather not mark fallthrough attribute for empty cases as they're not like the non-empty ones and looks like code noise if putting the fallthrough attribute on them.

Also see: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-break

nullsystem: Nevertheless I think we should at least turn those warnings on (they seems off on Linux) and similar warnings like indent warns on non-bracket if statements. Can also turn them to errors also but errors at most limit to NT;RE part of the code rather than the SDK itself.

src/game/client/neo/ui/neo_ui.cpp
nullsystem: I don't think it'll hit this case since mouse pressed which is always the case for initializing the menu always sorts out wide-tall 0 (either sets wide/tall or deinit on failure) before mouse move gets it. However guess out of caution can just add extra wide/tall > 0 checks anyway.

src/game/client/neo/ui/neo_ui.cpp
nullsystem: Same comment: https://github.com/NeotokyoRebuild/neo/pull/1256#discussion_r2270555654

src/game/client/neo/ui/neo_ui.cpp
nullsystem: The reason why `iMaxBytes` was a thing was due to trying to refit `wchar_t` characters into `char` mainly for `neo_name` and `neo_clantag` which both are `char`. I think I may will just simply change `NeoUI` back to only deal with the length of the `wchar_t` and deal with the issue externally in the settings.

nullsystem: I've changed the name and clantag to be validated separately while TextEdit now simply deals in the length of `wchar_t` as `iMaxWszTextSize`.

Masterkatze:
Just tested in on Linux - works without any issues, nice!

> Text selection with shift modifier + HOME/END/LEFT/RIGHT

Can you also add Ctrl key modifier (moving without text selection)? That would be great.

src/game/client/neo/ui/neo_root_settings.cpp
```diff
@@ -671,14 +672,27 @@ static_assert(ARRAYSIZE(DLFILTER_STRMAP) == ARRAYSIZE(DLFILTER_LABELS));
 void NeoSettings_General(NeoSettings *ns)
 {
 	NeoSettings::General *pGeneral = &ns->general;
-	NeoUI::TextEdit(L"Name", pGeneral->wszNeoName, SZWSZ_LEN(pGeneral->wszNeoName));
-	NeoUI::TextEdit(L"Clan tag", pGeneral->wszNeoClantag, SZWSZ_LEN(pGeneral->wszNeoClantag));
+	NeoUI::TextEdit(L"Name", pGeneral->wszNeoName, MAX_PLAYER_NAME_LENGTH);
```
Rainyan: Unless we're doing something different from the base engine here, I believe max playername length of `MAX_PLAYER_NAME_LENGTH` includes the null terminator: https://github.com/NeotokyoRebuild/neo/blob/006174a685e563557a4bec6136146420b2786c39/src/public/const.h#L39-L41

So you actually only get `MAX_PLAYER_NAME_LENGTH-1` text characters. So probably the `wszNeoName` allocation should be of size `MAX_PLAYER_NAME_LENGTH` instead of `MAX_PLAYER_NAME_LENGTH+1`, and the `iMaxWszTextSize` value should be one less than `MAX_PLAYER_NAME_LENGTH`.

Probably also relevant for the name overflow check in `src/game/shared/neo/neo_player_shared.cpp` line 236.

nullsystem:
> Just tested in on Linux - works without any issues, nice!
> 
> > Text selection with shift modifier + HOME/END/LEFT/RIGHT
> 
> Can you also add Ctrl key modifier (moving without text selection)? That would be great.

@Masterkatze Added it in now, it's like the vgui behavior where it always moves to the first character of word even for ctrl+right and just uses spaces as deliminator, which is different to how EX: a web browser deals with it.

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -1535,98 +1719,429 @@ void TextEdit(wchar_t *wszText, const int iMaxBytes)
...
+			{
+				wchar_t wszClipboard[MAX_TEXTINPUT_U8BYTES_LIMIT] = {};
+				const int iClipboardWSZBytes = vgui::system()->GetClipboardText(0, wszClipboard, ARRAYSIZE(wszClipboard));
+				if (iClipboardWSZBytes > 0)
+				{
+					// Write to temporary first otherwise wszText -> wszText can corrupt
+					static wchar_t wszStaticTmpText[MAX_TEXTINPUT_U8BYTES_LIMIT];
+					if (bFromEnd)
+					{
+						V_wcsncat(wszText, wszClipboard, iMaxWszTextSize);
```
Rainyan: There seems to be kind of a  nasty inconsistency here between `V_wcsncat` and `wcsncat` behaviour, wherein the Valve version drops the final character for max-sized input.

To reproduce:
* Copy text to clipboard:
  * `0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDE`
  * This is a 63 character input, so it should fit into a 64 allocated buffer where the final character is to hold the null terminator
* Launch the game
* Go fullscreen 1920x1080 or similar, so the paste output fits in the screen visually
* Go to main menu
* Press "Create Server"
* Delete the existing Hostname value (it has a 63+1 allocated buffer, so the paste should fit)
* Try to paste your clipboard contents

What happens:
* Clipboard contents are pasted to Hostname up to `...ABCD`, but are missing the final 63th character `E`

What should happen:
* Contents end in `...ABCDE`, i.e. the whole clipboard fits

You can manually type in the final `E` to see that it indeed should fit.

Switching this cat func to non `V_` version: `wcsncat(wszText, wszClipboard, iMaxWszTextSize);`
makes it work fine (at least for MSVC).

Not sure what the best fix would be. I guess we could either call this with `iMaxWszTextSize+1` to account for the different behaviour, or just use the standard `wcsncat`.

It's probably unsafe to change the Valve format function behaviour itself, since that could introduce a bug if we ever used SDK code (or other imported thirdparty code) that relied on it, for example the `V_wcscat_safe` function does call it and is used by the SDK code.

Rainyan (changes requested):

src/game/client/neo/ui/neo_ui.cpp
nullsystem: `wcslen` on Linux/GCC seemed to managed to bypass `iMaxWszTextSize` buffer/size and crash the game somewhat, I'll call it with `iMaxWszTextSize+1` instead then. This is also not the only one that seemed to differ, `V_wcsncpy` seems to take the length as bytes length rather than `wcsncpy` which states "n amount of wchar_t".

src/game/client/neo/ui/neo_root_settings.cpp
nullsystem: I've adjusted the `MAX_PLAYER_NAME_LENGTH` usages now, should be 31 as observed with both convar string limiter and NeoUI TextEdit. I've also applied it the the clantag also so that one is really 11 chars + 0.

Rainyan (approved):
LGTM 👍

PR #1261, opened 2025-08-05
Bots - Weapon aim on target/fire

### LINKED ISSUE #1260: Bots - Make them aim when shooting
Just simply have them aim while they shoot or in sniper, and not aim otherwise like running around.

nullsystem:
## Description
* Non-sniper - Bot just holds aim when firing
* Sniper - Bot just holds on stationary and/or firing
* Both - Also aim when seeing target and small delay after directly not "should aim"
* Move default crosshair define to shared, set for all m_szCrosshair
* Running/walking behavior altered again, but for dealing with aim and fixed `m_bOnTarget` not set to false when it needs to
* Aim towards the head rather than center

## Toolchain

## Linked Issues

* fixes #1260

### DIFF
--- a/src/game/server/neo/bot/neo_bot.cpp
+++ b/src/game/server/neo/bot/neo_bot.cpp
@@ -628,7 +628,7 @@ CNEOBot::CNEOBot()
 		engine->SetFakeClientConVarValue(edict, "neo_name", "");
 		engine->SetFakeClientConVarValue(edict, "cl_neo_streamermode", "0");
 		engine->SetFakeClientConVarValue(edict, "neo_clantag", "");
-		engine->SetFakeClientConVarValue(edict, "cl_neo_crosshair", "2;0;-1;0;6;0.000;2;4;0;0;1;0;0;"); // update with CL_NEO_CROSSHAIR_DEFAULT
+		engine->SetFakeClientConVarValue(edict, "cl_neo_crosshair", NEO_CROSSHAIR_DEFAULT);
 		engine->SetFakeClientConVarValue(edict, "hap_HasDevice", "0");
 	}
 }
@@ -674,6 +674,9 @@ void CNEOBot::Spawn()
 	SetBrokenFormation(false);
 
 	GetVisionInterface()->ForgetAllKnownEntities();
+
+	m_qPrevShouldAim = ANSWER_NO;
+	m_flLastShouldAimTime = 0.0f;
 }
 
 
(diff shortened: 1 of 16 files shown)


src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -2341,12 +2344,48 @@ INextBotEventResponder *CNEOBotIntention::NextContainedResponder(INextBotEventRe
...
+QueryResultType CNEOBotBehavior::ShouldWalk(const CNEOBot *me, const QueryResultType qShouldAimQuery) const
+{
+	QueryResultType result = ANSWER_UNDEFINED;
+
+	auto *neoAction = static_cast<CNEOBotMainAction *>(m_action);
+	if ( neoAction )
+	{
+		// find innermost child action
+		CNEOBotMainAction *action;
+		for( action = neoAction; action->m_child; action = static_cast<CNEOBotMainAction *>(action->m_child) )
```
sunzenshen, line 2361: Not sure if the pointers could become a cycle, but it may be worth defining a maximum traversal depth constant where this `for` loop short circuits if an unreasonable number of traversals happen.

nullsystem: This is the same kind of code that can be found in `NextBotBehavior.h`, if the inf. traversal depth an issue then so are those. Not sure if I'll bother with this PR for now.

Could be looked at separately and apply the depth on all `Should` functions? But unsure how the maximum be determined.

AdamTadeusz:
we call Weapon_SetZoom and Weapon_Switch very often in bots even when the current value they change is the same as the value being changed to. Returning early in those functions if that is the case, not updating m_flFOVTime if the new value is the same as the old or 0, and also not calling me->Weapon_AimToggle in neo_bot_behavior (perhaps another early return here instead?) if the bot is already at the correct aim state results in a much smoother viewing experience

nullsystem:
I've put some weapon toggle related changes, but I've also now made it aim towards the head rather than center, but put a TODO for later (not this PR) for more specialized head vs center aim behavior.

sunzenshen:
Is that head aiming approach a potential duplication/conflict with
https://github.com/NeotokyoRebuild/neo/pull/1248 ?

nullsystem:
Had no idea about that PR, well seems more complicated but whatever I'll concede

nullsystem:
Well both are slightly different, that one is just visible area so it'll always shoots center when there's nothing blocking the way, where-as this approach is just aim for the head always. Not too sure really, probably to determine to try to do head aim on higher difficulty bots or something.

PR #1266, opened 2025-08-11
Bots - Utilize auto lean

### LINKED ISSUE #1265: Bots - Utilize (auto) lean
* Probably could just utilize the already existing auto lean feature and that'll make the bot utilize leaning

nullsystem:
## Description

* Supports only lean on aim, other leans on walk + aim

## Toolchain

## Linked Issues

* fixes #1265


src/game/server/NextBot/Player/NextBotPlayer.h
```diff
@@ -136,6 +136,12 @@ class INextBotPlayerInput
 	// This is just an "alias" to PressWalkButton
 	virtual void PressRunButton( float duration = -1.0f ) = 0;
 	virtual void ReleaseRunButton( void ) = 0;
+
+	virtual void PressLeanLeftButton( float duration = -1.0f ) = 0;
+	virtual void ReleaseLeanLeftButton( void ) = 0;
+
+	virtual void PressLeanRightButton( float duration = -1.0f ) = 0;
+	virtual void ReleaseLeanRightButton( void ) = 0;
```
sunzenshen, line 144: Just based on previous code reviews, I wonder if these actions qualify for an `#ifdef NEO` block since most Valve Source Engine games don't have lean.

sunzenshen: Wonder if these can just be non-virtual APIs since these are already under an `#ifdef NEO` block and leaning is exclusive to NT.

src/game/server/NextBot/Player/NextBotPlayer.h
```diff
@@ -497,6 +511,34 @@ inline void NextBotPlayer< PlayerType >::ReleaseRunButton( void )
 {
 	return NextBotPlayer<PlayerType>::ReleaseWalkButton();
 }
+
+template < typename PlayerType >
+inline void NextBotPlayer< PlayerType >::PressLeanLeftButton( float duration )
+{
+	m_inputButtons |= IN_LEAN_LEFT;
+	m_leanLeftButtonTimer.Start( duration );
```
sunzenshen, line 513: I wonder if it makes sense to break with the usual convention here, and to also invalidate the timer for the opposite lean direction when the other lean button is pressed.

src/game/shared/neo/neo_predicted_viewmodel.cpp
```diff
@@ -309,64 +320,93 @@ float CNEOPredictedViewModel::lean(CNEO_Player *player){
...
+						float offset = (i / (tracelineAngleChange - i));
+						endPos.x += cos(DEG2RAD(viewAng.y) + offset) * distance;
+						endPos.y += sin(DEG2RAD(viewAng.y) + offset) * distance;
+						UTIL_TraceLine(startPos, endPos, MASK_SHOT_HULL, player, COLLISION_GROUP_NONE, &tr);
+						if (tr.fraction != 1.0)
+						{
+							total -= 1;
+						}
+						distance = Max(20, distance - 5);
+#if defined(CLENT_DLL) && defined(DEBUG)
```
sunzenshen: Is there a typo here with
`CLENT_DLL`
vs potentially
`CLIENT_DLL` ?

src/game/shared/neo/neo_predicted_viewmodel.cpp
```diff
@@ -309,64 +320,93 @@ float CNEOPredictedViewModel::lean(CNEO_Player *player){
...
+						if (cl_neo_lean_automatic_debug.GetBool())
+						{
+							DebugDrawLine(startPos, endPos, 0, tr.fraction == 1.0 ? 255 : 0, 255, 0, 0);
+						}
+#endif // defined(CLENT_DLL) && defined(DEBUG)
```
sunzenshen: CLIENT_DLL?

src/game/shared/neo/neo_predicted_viewmodel.cpp
```diff
@@ -309,64 +320,93 @@ float CNEOPredictedViewModel::lean(CNEO_Player *player){
...
+						float offset = (i / (tracelineAngleChange + i));
+						endPos.x += cos(DEG2RAD(viewAng.y) + offset) * distance;
+						endPos.y += sin(DEG2RAD(viewAng.y) + offset) * distance;
+						UTIL_TraceLine(startPos, endPos, MASK_SHOT_HULL, player, COLLISION_GROUP_NONE, &tr);
+						if (tr.fraction != 1.0)
+						{
+							total += 1;
+						}
+						distance = Max(20, distance - 5);
+#if defined(CLENT_DLL) && defined(DEBUG)
```
sunzenshen: CLIENT_DLL?

src/game/shared/neo/neo_predicted_viewmodel.cpp
```diff
@@ -309,64 +320,93 @@ float CNEOPredictedViewModel::lean(CNEO_Player *player){
...
+						if (cl_neo_lean_automatic_debug.GetBool())
+						{
+							DebugDrawLine(startPos, endPos, 255, tr.fraction == 1.0 ? 255 : 0, 0, 0, 0);
+						}
+#endif // defined(CLENT_DLL) && defined(DEBUG)
```
sunzenshen: CLIENT_DLL?

src/game/shared/neo/neo_predicted_viewmodel.cpp
```diff
@@ -309,64 +320,93 @@ float CNEOPredictedViewModel::lean(CNEO_Player *player){
...
-				distance = startingDistance;
-				for (int i = -1; i >= -10; i--)
+#ifdef CLIENT_DLL
+				if (total < 0)	{ IN_LeanRight(); }
+				else if (total > 0) { IN_LeanLeft(); }
+				else { IN_LeanReset(); }
+#else
+				// server/GAME_DLL already have this only used for IsBot, so it
+				// should be fine
+				auto *botPlayer = static_cast<CNEOBot *>(player);
```
sunzenshen: Is `player` always castable to CNEOBot in this context?  If not, static_cast here might be undefined behavior.  Would it make sense to have a dynamic cast like in ToNEOPlayer (neo_player.h) that checks for a nullptr result?

src/game/shared/neo/neo_predicted_viewmodel.cpp
```diff
@@ -400,10 +440,17 @@ float CNEOPredictedViewModel::lean(CNEO_Player *player){
...
+	if (player->IsBot())
+#endif
+	{
+		viewAng.z = -m_flLeanRatio * neo_lean_fp_angle.GetFloat();
 #ifdef CLIENT_DLL
-	viewAng.z = -m_flLeanRatio * neo_lean_fp_angle.GetFloat();
-	engine->SetViewAngles(viewAng);
+		engine->SetViewAngles(viewAng);
+#else
+		player->PlayerData()->v_angle = viewAng;
```
sunzenshen, line 450: Would it make sense to normalize assigned angles like this one using `AngleNormalize`?

https://developer.valvesoftware.com/wiki/Mapbase/Scripting/Script_Functions

> Clamps an angle to be in between -360 and 360.

src/game/shared/neo/neo_predicted_viewmodel.cpp
nullsystem: It generally should be CNEOBot since the IsBot check covers it, but I guess just in-case can do a dynamic_cast instead.

src/game/server/NextBot/Player/NextBotPlayer.h
nullsystem: I think I'd rather keep press/release stuff 1-to-1 and just handle cancelling out buttons outside of this class.

sunzenshen (approved):
Overall impression from watching bot matches is that while sometimes the auto-lean triggers in some unintuitive directions, it doesn't seem to impact bot survivability negatively, and I think this behavior accidentally helps the situation where bots are in a single file and all want to open fire.  At this point waiting for others to take a look at this build and share their feelings on this change.

sunzenshen (approved):

Issue #650, opened 2024-09-25
[FEATURE] Root NeoUI - Server list auto-join + blacklist


nullsystem:
### Description

While the serverlist has generally been implemented, it's missing auto-join and server blacklisting.

### Proposed Solution

* Auto-join for when the server is full
* Server blacklisting so blacklisted servers don't show up in the normal list
* Improve player list as it's rather basic at the moment

nullsystem:
Related: #994

PR #1270, opened 2025-08-13
NeoUI - Blacklist servers

### LINKED ISSUE #1268: NeoUI serverlist - Server blacklisting
Can set and utilize the pre-existing blacklist: `cfg/server_blacklist.txt`

```
"serverblacklist"
{
        "Server"
        {
                "name"          "Some server"
                "date"          "1754949812"
                "addr"          "1.2.3.4:27015"
        }
        "Server"
        {
                "name"          "foobar"
                "date"          "1754949905"
                "addr"          "0.0.0.0:0"
        }
}
```

nullsystem:
## Description
* Uses our own blacklist: ntre_server_blacklist.txt
* Format is similar to SDK's blacklist, except it can deal with by "name contains" blacklisting and uses uint64 serialization for time instead
* New blacklist tab which have its own dedicated list
* Import which can also work with SDK's blacklist, but export is strictly NT;RE blacklist only
* Usual actions like SDK's: Add server to blacklist, add custom hostname/IP to blacklist

## Toolchain

## Linked Issues

- fixes #1268
- related #650


src/game/client/neo/ui/neo_root.cpp
```diff
@@ -2267,10 +2436,12 @@ LightmappedGeneric
...
 {
-	static void (CNeoRoot::*FILESELMODEFNS[FILEIODLGMODE__TOTAL])(const char *) = {
-		&CNeoRoot::OnFileSelectedMode_Spray,		// FILEIODLGMODE_SPRAY
+	static void (*FILESELMODEFNS[FILEIODLGMODE__TOTAL])(const char *) = {
+		&OnFileSelectedMode_Spray,		// FILEIODLGMODE_SPRAY
+		&ServerBlacklistRead,						// FILEIODLGMODE_BLACKLIST_IMPORT
+		&ServerBlacklistWrite,						// FILEIODLGMODE_BLACKLIST_EXPORT
 	};
-	(this->*FILESELMODEFNS[m_eFileIOMode])(szFullpath);
+	(*FILESELMODEFNS[m_eFileIOMode])(szFullpath);
```
sunzenshen: It may be worth adding a quick check that m_eFileIOMode is within the array bounds of FILESELMODEFNS[FILEIODLGMODE__TOTAL]

src/game/client/neo/ui/neo_root.cpp
```diff
@@ -1297,73 +1362,119 @@ void CNeoRoot::MainLoopServerBrowser(const MainLoopParam param)
...
+						m_eFileIOMode = CNeoRoot::FILEIODLGMODE_BLACKLIST_EXPORT;
+						m_pFileIODialog->AddFilter("*.txt", "TXT file blacklist", true);
+						m_pFileIODialog->AddFilter("*", "ALL file", false);
+						m_pFileIODialog->DoModal();
+					}
+					if (m_iSelectedServer >= 0)
+					{
+						if (NeoUI::Button(L"Remove").bPressed)
+						{
+							g_blacklistedServers.Remove(m_iSelectedServer);
```
sunzenshen, line 1421: Some kind of sanity error might be helpful here if m_iSelectedServer is unexpectedly not in the global blacklist of servers?

src/tier1/CMakeLists.txt
```diff
@@ -20,7 +20,7 @@ target_include_directories(tier1
 )
 
 if(OS_WINDOWS)
-    target_link_libraries(tier1 PRIVATE Rpcrt4.lib)
+    target_link_libraries(tier1 PRIVATE Rpcrt4.lib Ws2_32.lib)
```
Masterkatze, line 23: Is Ws2_32.lib a direct dependency of tier1 or it just a convenience to add it here?

nullsystem: `tier1/NetAdr.cpp` I think the only direct dependency of `Ws2_32.lib`/winsock, in the other places that uses it doesn't seem to be referenced by CMake or are just `utils` source files.

src/game/client/neo/ui/neo_root.cpp
nullsystem: This button only appears if: it's in the blacklist tab which is just the global blacklist, and if `m_iSelectedServer` set which gets reset when you switch tab. The only thing can be done is to sanity check to the count but that's it.

src/game/client/neo/ui/neo_root_serverbrowser.cpp
```diff
@@ -1,22 +1,219 @@
...
+
+		return (sortCtx.bDescending) ? (V_wcscmp(wszRight, wszLeft) > 0) : (V_wcscmp(wszLeft, wszRight) > 0);
+	});
+
+	if (IN_BETWEEN_AR(0, g_pNeoRoot->m_iSelectedServer, g_blacklistedServers.Count()))
+	{
+		g_pNeoRoot->m_iSelectedServer = -1;
+		for (int i = 0; i < g_blacklistedServers.Count(); ++i)
+		{
+			if (V_memcmp(&curSbi, &g_blacklistedServers[i], sizeof(ServerBlacklistInfo)) == 0)
```
sunzenshen, line 189: Just to sanity check, what are the chances that we would want this comparison to match even if `timeVal` was different?  Is it possible that we would consider these structs to be a a close enough match even if the times are not the same, but other fields matched?

To recap above:

```cpp
struct ServerBlacklistInfo
{
	// Data section
	wchar_t wszName[256];
	time_t timeVal;
	netadr_s netAdr;
	EServerBlacklistType eType;

	// Visual cache only
	wchar_t wszDateTimeAdded[48];
};
```

nullsystem: `timeVal` is used within table sorting so I think it makes sense to match it. I think it might be more of a concern about duplicate entries (of `wszName, netAdr, eType`) which neither custom add or import bother to deal with at the moment, although not too concern about duplicates myself. Also, all the values are just shifted around the vector but not actually changed so again should be fine to just memcmp the whole struct even `wszDateTimeAdded` and suffice enough to keep the index on sorting changes.

AdamTadeusz:
I notice that adding a server to the blacklist from the server browser works fine, but when adding an item to the blacklist either via hostname/ip or containing string that the game freezes for a solid second, most noticeable if the menu background music is playing in the background. This is regardless of the pause music on window unfocus so the game going out of focus is probably not the problem

AdamTadeusz:
Also items added to the ntre server blacklist obviously don't show up in the legacy menu, but the reason why might not be obvious to players, though I guess once we hide the legacy menu completely that won't be a problem. I wonder if it would be worthwhile adding ~29 files to edit the tab name in the legacy menu to make it more obvious that its not the same as the one in the normal menu

image

nullsystem:
> I notice that adding a server to the blacklist from the server browser works fine, but when adding an item to the blacklist either via hostname/ip or containing string that the game freezes for a solid second, most noticeable if the menu background music is playing in the background. This is regardless of the pause music on window unfocus so the game going out of focus is probably not the problem

@AdamTadeusz I don't really have the issue with freezing for a second, it's instant for me.

AdamTadeusz:
perhaps its a windows only issue, ill check if I can replicate it again

AdamTadeusz (approved):

nullsystem:
> just out of curiosity, what do we gain from resolving the hostname? Since we don't display whether the hostname is valid or not anywhere would it make sense to just skip this altogether?

Mostly cause it's stored and compared as IP+port so hostnames have to resolve it to be usable for filtering. I think it's unlikely hostname will be used anyway as it's usually IP+port when people are dealing with servers and kind of on the player to provide a connectable hostname anyway.

PR #1271, opened 2025-08-15
Ghost Boundary

### LINKED ISSUE #1175: There should be a smarter way of blocking the ghost
Currently in OG and Rebuild, the only (simple) way of blocking a ghost into an area is by using a ghost clip, which is just a vphysics clip with a filter. This only stops the ghost from passing through when dropped. If the player is holding the ghost it will pass through without an issue.

The issue is with this is that players can fall to their deaths with the ghost held, and therefore the ghost can still get stuck out of bounds.
It also poses the problem that a clip is too restrictive. On Rise, there is no optimal way of clipping off the ghost that prevents it from being dropped off of the edge, allows it to be thrown down to lower levels, and also ensures it is still obtainable.

It's a complex issue that was briefly discussed here: https://discord.com/channels/1235346473827434517/1244281729036845200/1344244377345720401

DESTROYGIRL:
## Description
See "Last-Held-Stationary-Position Teleport" in the [design doc](https://docs.google.com/document/d/1016YJVcHvvkZ03Zln91sISmczkDIceBbGBroW17LSCU/edit?usp=sharing)

A brush entity that can prevent the ghost from becoming unplayable. When touched by a ghost or a player carrying the ghost, the ghost will be teleported back to where the player was last touching the ground.

For reviewers: Download [this map](https://cdn.discordapp.com/attachments/940737618843410562/1406256936311328838/testingtransferable.bsp?ex=68a1ce54&is=68a07cd4&hm=6e2537cbc45c09cca4592ebd2c9aed059b86cbf6f6e3c8a891d479666501fa00&)
Spawn a bot, try getting the ghost into the pit. You cannot

[neoAssets](https://github.com/NeotokyoRebuild/neoAssets/pull/65)

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1175

### DIFF
--- a/src/game/server/neo/neo_ghost_boundry.cpp
+++ b/src/game/server/neo/neo_ghost_boundry.cpp
@@ -2,6 +2,9 @@
 #include "neo_gamerules.h"
 #include "weapon_ghost.h"
 
+#define THINK_INTERVAL 0.05f
+#define GHOST_ANGLES QAngle(15, 0, 270) // Angles the ghost will reset to (laid on its back)
+
 //##############################
 //  Trigger Weapon
 //##############################
@@ -11,9 +14,6 @@ LINK_ENTITY_TO_CLASS(neo_trigger_weapon, CNEO_TriggerWeapon);
 BEGIN_DATADESC(CNEO_TriggerWeapon)
     DEFINE_THINKFUNC(Think),
 END_DATADESC()
-
-#define THINK_INTERVAL 0.05f
-#define GHOST_ANGLES QAngle(15, 0, 270)
 
 void CNEO_TriggerWeapon::Spawn()
 {
@@ -41,7 +41,7 @@ void CNEO_TriggerWeapon::CheckForWeapon()
     CollisionProp()->WorldSpaceAABB(&mins, &maxs);
 
     CBaseEntity* pEntList[128];
-    int count = UTIL_EntitiesInBox(pEntList, sizeof(pEntList), mins, maxs, null); // Low accuracy
+    int count = UTIL_EntitiesInBox(pEntList, ARRAYSIZE(pEntList), mins, maxs, null); // Low accuracy
 
     for (int i = 0; i < count; ++i)
     {
--- a/src/game/server/neo/neo_ghost_boundry.cpp
+++ b/src/game/server/neo/neo_ghost_boundry.cpp
@@ -1,0 +1,151 @@
+#include "neo_ghost_boundry.h"
+#include "neo_gamerules.h"
+#include "weapon_ghost.h"
+
+//##############################
+//  Trigger Weapon
+//##############################
+
+LINK_ENTITY_TO_CLASS(neo_trigger_weapon, CNEO_TriggerWeapon);
+
+BEGIN_DATADESC(CNEO_TriggerWeapon)
+    DEFINE_THINKFUNC(Think),
+END_DATADESC()
+
+#define THINK_INTERVAL 0.05f
+#define GHOST_ANGLES QAngle(15, 0, 270)
+
+void CNEO_TriggerWeapon::Spawn()
+{
+    BaseClass::Spawn();
+    InitTrigger();
+
+    SetNextThink(gpGlobals->curtime + THINK_INTERVAL);
+}
+
+void CNEO_TriggerWeapon::Think()
+{
+    CheckForWeapon();
+
+    SetNextThink(gpGlobals->curtime + THINK_INTERVAL);
+}
+
+// Weapons have the FSOLID_TRIGGER solidflag which is why they can't be used in triggers.
+// If we remove it weapons can't be picked up by walking over them.
+
+// NEO TODO DG: When there is a use for it improve this so it functions as expected- like other triggers.
+// What's here is fine for the ghost boundry but using this trigger for weapons will spam StartTouch every think
+void CNEO_TriggerWeapon::CheckForWeapon()
+{
+    Vector mins, maxs;
+    CollisionProp()->WorldSpaceAABB(&mins, &maxs);
+
+    CBaseEntity* pEntList[128];
+    int count = UTIL_EntitiesInBox(pEntList, sizeof(pEntList), mins, maxs, null); // Low accuracy
+
+    for (int i = 0; i < count; ++i)
+    {
+        CBaseEntity* pEnt = pEntList[i];
+
+        if (!pEnt->IsBaseCombatWeapon())
+        {
+            continue;
+        }
+
+        if (pEnt->GetOwnerEntity())
+        {
+            continue;
+        }
+
+        if (IsEntityInside(pEnt)) // High(er) accuracy
+        {
+            StartTouch(pEnt);
+        }
+    }
+}
+
+bool CNEO_TriggerWeapon::IsEntityInside(CBaseEntity* pEnt)
+{
+    trace_t tr;
+    Vector point = pEnt->WorldSpaceCenter();
+    UTIL_TraceModel(point, point, pEnt->CollisionProp()->OBBMins(), pEnt->CollisionProp()->OBBMaxs(), this, COLLISION_GROUP_NONE, &tr);
+
+    return tr.startsolid; // If the trace started inside the trigger or not
+}
+
+
+//##############################
+//  Ghost Boundry
+//##############################
+
+LINK_ENTITY_TO_CLASS(neo_ghost_boundry, CNEO_GhostBoundry);
+
+BEGIN_DATADESC(CNEO_GhostBoundry)
+    DEFINE_THINKFUNC(Think),
+END_DATADESC()
+
+void CNEO_GhostBoundry::Think()
+{
+    CheckForWeapon();
+
+    if (!NEORules()->GhostExists())
+    {
+        SetNextThink(gpGlobals->curtime + THINK_INTERVAL);
+        return;
+    }
+
+    int iGhosterIndex = NEORules()->GetGhosterPlayer();
+    if (iGhosterIndex == 0)
+    {
+        SetNextThink(gpGlobals->curtime + THINK_INTERVAL);
+        return;
+    }
+
+    CBasePlayer* pPlayer = UTIL_PlayerByIndex(iGhosterIndex);
+    if (pPlayer->GetFlags() & FL_ONGROUND)
+    {
+        m_vecLastGhosterPos = pPlayer->GetAbsOrigin();
+    }
+
+    SetNextThink(gpGlobals->curtime + THINK_INTERVAL);
+}
+
+void CNEO_GhostBoundry::StartTouch(CBaseEntity *pOther)
+{
+    BaseClass::StartTouch(pOther);
+
+    Vector vecGhostSpawn = NEORules()->m_vecPreviousGhostSpawn;
+    if ((vecGhostSpawn == vec3_origin) && (m_vecLastGhosterPos == vec3_origin))
+    {
+        return; // Don't bother teleporting if there's no ghost spawn and no ghoster pos
+    }
+
+    CWeaponGhost *pDroppedGhost = dynamic_cast<CWeaponGhost*>(pOther);
+    if (pDroppedGhost)
+    {
+        if (m_vecLastGhosterPos == vec3_origin)
+        {
+            pDroppedGhost->VPhysicsGetObject()->SetPosition(vecGhostSpawn, GHOST_ANGLES, true);
+            // Don't freeze it, ghosts aren't frozen when they spawn
+        }
+        else
+        {
+            pDroppedGhost->VPhysicsGetObject()->SetPosition(m_vecLastGhosterPos, GHOST_ANGLES, true);
+            pDroppedGhost->VPhysicsGetObject()->EnableMotion(false);
+        }
+
+        return;
+    }
+
+    CNEO_Player *pNEOPlayer = dynamic_cast<CNEO_Player*>(pOther);
+    if (pNEOPlayer && pNEOPlayer->IsCarryingGhost())
+    {
+        CBaseCombatWeapon* pHeldGhost = pNEOPlayer->Weapon_OwnsThisType("weapon_ghost");
+        pNEOPlayer->Weapon_Drop(pHeldGhost);
+
+        pHeldGhost->VPhysicsGetObject()->SetPosition(m_vecLastGhosterPos, GHOST_ANGLES, true);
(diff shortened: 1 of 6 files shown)


src/game/server/neo/neo_ghost_boundry.h
```diff
@@ -0,0 +1,28 @@
...
+	void CheckForWeapon();
+
+private:
+	bool IsEntityInside(CBaseEntity* pEnt);
+};
+
+class CNEO_GhostBoundry : public CNEO_TriggerWeapon
+{
+public:
+	DECLARE_CLASS(CNEO_GhostBoundry, CBaseTrigger);
```
sunzenshen: Just to sanity check: is the parameter `CBaseTrigger` supposed to match the class inheritance above on line 18?  There it inherits from CNEO_TriggerWeapon.  Not 100% sure if this is intentional because of the properties of CBaseTrigger, or a copy and paste discrepancy. I'll defer to your judgement about this difference.

src/game/server/neo/neo_ghost_boundry.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+// NEO TODO DG: When there is a use for it improve this so it functions as expected- like other triggers.
+// What's here is fine for the ghost boundry but using this trigger for weapons will spam StartTouch every think
+void CNEO_TriggerWeapon::CheckForWeapon()
+{
+    Vector mins, maxs;
+    CollisionProp()->WorldSpaceAABB(&mins, &maxs);
+
+    CBaseEntity* pEntList[128];
+    int count = UTIL_EntitiesInBox(pEntList, sizeof(pEntList), mins, maxs, null); // Low accuracy
```
sunzenshen: Perhaps you meant `sizeof(pEntList)` to be `ARRAYSIZE(pEntList)`?

From my understanding of the API spec, the second argument is the max size of the list:
https://developer.valvesoftware.com/wiki/UTIL_EntitiesInBox

`sizeof` might potentially be counting the size in bytes of the pEntList (which depends on the individual size of CBaseEntity pointers), and not the array size.  That's my reasoning for why maybe ARRAYSIZE would be a better call here, even if probably the size in bytes of each pointer ends up working okay in this case.

src/game/server/neo/neo_ghost_boundry.h
```diff
@@ -0,0 +1,28 @@
+#include "cbase.h"
```
sunzenshen: Remember to include both `#pragma once` and a C style header macro guard at the very top of header files.

src/game/server/neo/neo_ghost_boundry.cpp
```diff
@@ -0,0 +1,151 @@
...
+//##############################
+
+LINK_ENTITY_TO_CLASS(neo_trigger_weapon, CNEO_TriggerWeapon);
+
+BEGIN_DATADESC(CNEO_TriggerWeapon)
+    DEFINE_THINKFUNC(Think),
+END_DATADESC()
+
+#define THINK_INTERVAL 0.05f
+#define GHOST_ANGLES QAngle(15, 0, 270)
```
sunzenshen: I honestly have no idea what these numbers mean, but I'm guessing this is setting the default orientation of the reset ghost postion?  Maybe a comment might help justify the reasoning for these numbers?

src/game/server/neo/neo_ghost_boundry.cpp
DESTROYGIRL: Yes thank you

src/game/server/neo/neo_ghost_boundry.cpp
```diff
@@ -0,0 +1,151 @@
...
+        return;
+    }
+
+    CBasePlayer* pPlayer = UTIL_PlayerByIndex(iGhosterIndex);
+    if (pPlayer->GetFlags() & FL_ONGROUND)
+    {
+        m_vecLastGhosterPos = pPlayer->GetAbsOrigin();
+    }
+
+    SetNextThink(gpGlobals->curtime + THINK_INTERVAL);
```
Rainyan: All 3 exits from this function call `SetNextThink(gpGlobals->curtime + THINK_INTERVAL);`, so those could be merged into just one statement at the top of the function.

src/game/server/neo/neo_ghost_boundry.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+// NEO TODO DG: When there is a use for it improve this so it functions as expected- like other triggers.
+// What's here is fine for the ghost boundry but using this trigger for weapons will spam StartTouch every think
+void CNEO_TriggerWeapon::CheckForWeapon()
+{
+    Vector mins, maxs;
+    CollisionProp()->WorldSpaceAABB(&mins, &maxs);
+
+    CBaseEntity* pEntList[128];
+    int count = UTIL_EntitiesInBox(pEntList, ARRAYSIZE(pEntList), mins, maxs, null); // Low accuracy
```
Rainyan: Nit: `null` is a seemingly unrelated definition that happens to evaluate to a `long` with value 0, which is implicitly castable to int, so it works here somewhat by coincidence. But ideally we should just pass in `0` to the `int flagMask` parameter, so we're not coupling the value to an unrelated type from a VGUI header that happens to be included in this scope:

```diff
-    int count = UTIL_EntitiesInBox(pEntList, ARRAYSIZE(pEntList), mins, maxs, null); // Low accuracy
+    int count = UTIL_EntitiesInBox(pEntList, ARRAYSIZE(pEntList), mins, maxs, 0); // Low accuracy
```

DESTROYGIRL:
Magical linux build fail that wasn't my fault?

src/game/server/neo/neo_ghost_boundry.cpp
```diff
@@ -0,0 +1,151 @@
...
+{
+    BaseClass::StartTouch(pOther);
+
+    Vector vecGhostSpawn = NEORules()->m_vecPreviousGhostSpawn;
+    if ((vecGhostSpawn == vec3_origin) && (m_vecLastGhosterPos == vec3_origin))
+    {
+        return; // Don't bother teleporting if there's no ghost spawn and no ghoster pos
+    }
+
+    CWeaponGhost *pDroppedGhost = dynamic_cast<CWeaponGhost*>(pOther);
```
Rainyan: Nit: this is bit of an awkward syntax and I'd prefer we had nicer API for it, but you can do:
```diff
-    CWeaponGhost *pDroppedGhost = dynamic_cast<CWeaponGhost*>(pOther);
+    auto pDroppedGhost = (pOther && pOther->IsBaseCombatWeapon() && static_cast<CNEOBaseCombatWeapon*>(pOther)->IsGhost())
+        ? static_cast<CWeaponGhost*>(pOther) : nullptr;
```
which will utilize dynamic dispatch and avoid RTTI overhead from calling `dynamic_cast`.

This works because we know that anything which `IsBaseCombatWeapon` must also be `CNEOBaseCombatWeapon` for our code base.

src/game/server/neo/neo_ghost_boundry.cpp
```diff
@@ -0,0 +1,151 @@
...
+        else
+        {
+            pDroppedGhost->VPhysicsGetObject()->SetPosition(m_vecLastGhosterPos, GHOST_ANGLES, true);
+            pDroppedGhost->VPhysicsGetObject()->EnableMotion(false);
+        }
+
+        return;
+    }
+
+    CNEO_Player *pNEOPlayer = dynamic_cast<CNEO_Player*>(pOther);
```
Rainyan: For this one, we have a fast & safe helper method, which is the preferable way to do the conversion:
```diff
-    CNEO_Player *pNEOPlayer = dynamic_cast<CNEO_Player*>(pOther);
+    CNEO_Player* pNEOPlayer = ToNEOPlayer(pOther);
```
This is faster than doing a dynamic cast.

src/game/server/neo/neo_ghost_boundry.h
```diff
@@ -0,0 +1,29 @@
...
+class CNEO_GhostBoundry : public CNEO_TriggerWeapon
+{
+public:
+	DECLARE_CLASS(CNEO_GhostBoundry, CNEO_TriggerWeapon);
+	DECLARE_DATADESC();
+
+	virtual void Think() override;
+	virtual void StartTouch(CBaseEntity *pOther) override;
+
+	Vector		m_vecLastGhosterPos = vec3_origin;
```
Rainyan: We could make this member variable `private`.

src/game/shared/neo/neo_gamerules.h
```diff
@@ -423,7 +424,6 @@ class CNEORules : public CHL2MPRules, public CGameEventListener
 	int m_iEntPrevCapSize = 0;
 	int m_iPrintHelpCounter = 0;
 	bool m_bGamemodeTypeBeenInitialized = false;
-	Vector m_vecPreviousGhostSpawn = vec3_origin;
```
Rainyan, line 426: Instead of making `m_vecPreviousGhostSpawn` public, we could instead declare `CNEO_GhostBoundry` as a friend, so something like:

```diff
--- a/src/game/shared/neo/neo_gamerules.h
+++ b/src/game/shared/neo/neo_gamerules.h
@@ -401,7 +401,6 @@ public:
        bool m_bPausedByPreRoundFreeze = false;
        bool m_bPausingTeamRequestedUnpause = false;
        bool m_bThinkCheckClantags = false;
-       Vector m_vecPreviousGhostSpawn = vec3_origin;
 #endif
        CNetworkVar(float, m_flPauseEnd);

@@ -424,6 +423,8 @@ private:
        int m_iEntPrevCapSize = 0;
        int m_iPrintHelpCounter = 0;
        bool m_bGamemodeTypeBeenInitialized = false;
+       friend class CNEO_GhostBoundry;
+       Vector m_vecPreviousGhostSpawn = vec3_origin;
 #endif
        CNetworkVar(int, m_nRoundStatus);
        CNetworkVar(int, m_iHiddenHudElements);
```

So this way we don't need to break the encapsulation globally.

Rainyan (changes requested):

Rainyan:
> Magical linux build fail that wasn't my fault?

Looks like the CI machine internet just ate dust for whatever reason. Probably not code related 😄

Rainyan:
Also, this is a complete nit, but could we rename this thing "boundry" -> "boundary"? Because I think it's a typo(?)

src/game/server/neo/neo_ghost_boundry.cpp
DESTROYGIRL: Yeah, I noticed I `SetNextThink()` inside the `Spawn()` function, should I just call `Think()` directly?

Rainyan: > Yeah, I noticed I `SetNextThink()` inside the `Spawn()` function, should I just call `Think()` directly?

I couldn't tell off the top of my head, but I think you probably just wanna queue the next think with `SetNextThink` like you are already doing, and let the engine figure out when it wants to handle that, rather than call it directly.

Something like this:
```diff
--- a/src/game/server/neo/neo_ghost_boundry.cpp
+++ b/src/game/server/neo/neo_ghost_boundry.cpp
@@ -87,17 +87,16 @@ END_DATADESC()
 void CNEO_GhostBoundry::Think()
 {
     CheckForWeapon();
+    SetNextThink(gpGlobals->curtime + THINK_INTERVAL);

     if (!NEORules()->GhostExists())
     {
-        SetNextThink(gpGlobals->curtime + THINK_INTERVAL);
         return;
     }

     int iGhosterIndex = NEORules()->GetGhosterPlayer();
     if (iGhosterIndex == 0)
     {
-        SetNextThink(gpGlobals->curtime + THINK_INTERVAL);
         return;
     }

@@ -106,8 +105,6 @@ void CNEO_GhostBoundry::Think()
     {
         m_vecLastGhosterPos = pPlayer->GetAbsOrigin();
     }
-
-    SetNextThink(gpGlobals->curtime + THINK_INTERVAL);
 }
 ```

So this just refactors the repeated code into a single statement without changing behavior.

Rainyan (approved):
LGTM.

One thing I noticed is that there's a slight difference in ghost teleport location depending on how it goes OOB:

* If carried by a ghoster, it seems to:
  * Teleport to last valid standing pos
  * Have physics disabled
* If shot OOB:
  * Teleports back to the ghost spawn
  * Does *not* have physics disabled

It would be nice if there's a way to simplify this down to one set of OOB behaviours, so it's more predictable to the player. But maybe this is something to iterate on for some future PR.

Rainyan:
> I considered this to be a very rare scenario where this would happen, because it would mean players could manipulate the ghost off an edge from where it spawned without ever holding it. I don't think any maps have ghost spawns this close to an edge, maybe saitama(?), because when a ghost spawns it isn't physics disabled, so I thought the behaviour should be the same. Otherwise the ghost would probably be floating in mid air, or wherever the mapper placed the ghost spawn

Good points. Mostly thinking in terms of generalising the behaviour, not just for vanilla maps, but for any hypothetical map X with whatever geometry they might come up with. There's also things like a grenade hitting the ghost to make it fly into some pit. And it's also noteworthy that ghost won't only remain at spawnpoints, but it can be dropped at any point in the map either intentionally or by the carrier dying.

DESTROYGIRL:
> There's also things like a grenade hitting the ghost to make it fly into some pit. And it's also noteworthy that ghost won't only remain at spawnpoints, but it can be dropped at any point in the map either intentionally or by the carrier dying.

Not sure I'm following your logic, but if the carrier dies and it falls into a pit, it's just going to teleport back to where the carrier was before they died
The difference here shouldn't be noticeable during regular gameplay, as it will only ever happen if the ghost was pitted before ever being picked up, which means none of the players have actually taken intrest in using it yet. In this scenario, it's as if the ghost had never been knocked into the pit in the first place. If we disable physics on it, players are more likley to notice that

I guess there is the discrepency of how the ghost might land on the ground when recalled back to the spawn VS how it landed originally at the start of the round but that sounds like the mapper placed the spawn too high if it's causing an issue
And if the ghost is spawning and falling into a pit on it's own, I consider that a map issue too

Rainyan:
I meant this in reference to:
> I don't think any maps have ghost spawns this close to an edge, maybe saitama(?)

since even if there's no mapper-designed ghostspawns close to a cliff, player actions could result in a situation like that nonetheless.

sunzenshen (approved):
Updates LGTM. Downloaded the test map and checked that the behavior seems reasonable.

PR #1281, opened 2025-08-18
Don't draw smoke fog overlay when xraying

AdamTadeusz:
## Description

The value of g_SmokeFogOverlayAlpha is calculated somewhere in the engine and prevents the xray overlay from working, g_SmokeFogOverlayThermalOverride isn't enough to fix it since we don't have access to that code so have to set the value of g_SmokeFogOverlayAlpha to 0 to get it working again.

To test this start a match with bots, spawn as a support, throw a smoke, kill yourself, go into free roam spectate mode and fly into the smoke, enable xray, and look towards the bots.

Will still draw the smoke fog overlay in the case where we are on a team and mp_forcecamera is set to anything other than OBS_ALLOW_ALL

### DIFF
--- a/src/game/client/smoke_fog_overlay.cpp
+++ b/src/game/client/smoke_fog_overlay.cpp
@@ -151,7 +151,6 @@ void UpdateThermalOverride()
 	}
 	else if (localPlayer->IsObserver() && glow_outline_effect_enable.GetBool() && (localPlayer->GetTeamNumber() == TEAM_SPECTATOR || mp_forcecamera.GetInt() == OBS_ALLOW_ALL))
 	{
-		g_SmokeFogOverlayThermalOverride = true;
 		g_SmokeFogOverlayAlpha = 0;
 	}
 	g_SmokeFogOverlayThermalOverride = false;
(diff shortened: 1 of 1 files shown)


src/game/client/smoke_fog_overlay.cpp
```diff
@@ -141,9 +144,15 @@ void UpdateThermalOverride()
...
 				g_SmokeFogOverlayThermalOverride = true;
+				g_SmokeFogOverlayAlpha = 0;
 				return;
 			}
 		}
 	}
+	else if (localPlayer->IsObserver() && glow_outline_effect_enable.GetBool() && (localPlayer->GetTeamNumber() == TEAM_SPECTATOR || mp_forcecamera.GetInt() == OBS_ALLOW_ALL))
+	{
+		g_SmokeFogOverlayThermalOverride = true;
+		g_SmokeFogOverlayAlpha = 0;
```
sunzenshen, line 154: Did you mean to add an early `return;` here similar to those of lines 136 and 148?

It seems like you set 
```cpp
g_SmokeFogOverlayThermalOverride = true
````
in this block, but then after this `else if` block finishes
```cpp
g_SmokeFogOverlayThermalOverride = false;
```
on line 157 overwrites your prior setting.

AdamTadeusz: doh, but actually this looks better ingame

DESTROYGIRL (approved):

PR #1283, opened 2025-08-18
Bots - Shoot through glass + penetrateables

### LINKED ISSUE #1282: Bots - Shoot through glass (when not using shotguns)
Currently the bots just doesn't shoot through glass at all. It should try to shoot through glass when it's not using a shotgun. However, need to figure out dealing with non-penetrable glass vs penetrable glass. Possibly also to switch weapon usually secondary if needing to fire.

nullsystem:
## Description
* Bots should now be able to shoot through glass if not using a shotgun
* If using a shotgun, try to change to a weapon that can utilize the glass if possible
* Change back to a shotgun if not behind glass?
* Tested/check if it doesn't straight up shoot through any penetrable like wallhack ability, non-glass penetrable only shoot through if it's recently seen enemy
* Enemy player that can shoot towards the ghoster behind penetrable should be able to try shooting at it
* Test/check for non-penetrable glass (tarmac non-penetrable vs oilstain penetrable)
* Slight scale between weapon penetration on when the bots will attempt to do penetration, so like SRS will do longer distance than SRM.
* Limited wallbang at ghoster ability to hard + expert bots

## Toolchain

## Linked Issues

* fixes #1282

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
@@ -321,7 +321,7 @@ bool CNEOBotMainAction::IsImmediateThreat( const CBaseCombatCharacter *subject, 
 		return false;
 
 	// if they can't hurt me, they aren't an immediate threat
-	if ( !me->IsLineOfFireClear( threat->GetEntity() ) )
+	if ( !me->IsLineOfFireClear( threat->GetEntity(), CNEOBot::LINE_OF_FIRE_FLAGS_DEFAULT ) )
 		return false;
 
 	Vector to = me->GetAbsOrigin() - threat->GetLastKnownPosition();
@@ -587,13 +587,64 @@ void CNEOBotMainAction::FireWeaponAtEnemy( CNEOBot *me )
 	if ( threat == NULL || !threat->GetEntity() || !threat->IsVisibleRecently() )
 		return;
 
-	// don't shoot through windows/etc
-	if ( !me->IsLineOfFireClear( threat->GetEntity()->EyePosition() ) )
+	CNEOBot::LineOfFireFlags lofFlags = CNEOBot::LINE_OF_FIRE_FLAGS_DEFAULT;
+	auto *neoThreat = ToNEOPlayer(threat->GetEntity());
+	const bool bThreatIsGhoster = neoThreat && neoThreat->IsCarryingGhost();
+	if (bThreatIsGhoster)
 	{
-		if ( !me->IsLineOfFireClear( threat->GetEntity()->WorldSpaceCenter() ) )
+		lofFlags |= CNEOBot::LINE_OF_FIRE_FLAGS_PENETRATION;
+	}
+
+	// don't shoot through non-shootables, check if shootable by any non-shotguns weapons
+	Vector vShootablePos = threat->GetEntity()->EyePosition();
+	if ( !me->IsLineOfFireClear( vShootablePos, lofFlags ) )
+	{
+		vShootablePos = threat->GetEntity()->WorldSpaceCenter();
+		if ( !me->IsLineOfFireClear( vShootablePos, lofFlags ) )
 		{
-			if ( !me->IsLineOfFireClear( threat->GetEntity()->GetAbsOrigin() ) )
+			vShootablePos = threat->GetEntity()->GetAbsOrigin();
+			if ( !me->IsLineOfFireClear( vShootablePos, lofFlags ) )
+			{
 				return;
+			}
+		}
+	}
+
+	// Check again if shotgun from last "any" shootable pos
+	bool bNotPrimary = false;
+	bool bShotgunSituationHandled = false;
+	// If holding non-primary, have shotgun in primary, and sight clear for shotgun, try to switch to shotgun
+	// Otherwise if holding shotgun and sight not clear for shotgun, try to switch to secondary
+	if (myWeapon && (myWeapon->GetNeoWepBits() & (NEO_WEP_MILSO | NEO_WEP_TACHI | NEO_WEP_KYLA | NEO_WEP_KNIFE | NEO_WEP_THROWABLE)))
+	{
+		auto *primaryWeapon = static_cast<CNEOBaseCombatWeapon *>(me->Weapon_GetSlot(0));
+		if (primaryWeapon && (primaryWeapon->GetNeoWepBits() & (NEO_WEP_AA13 | NEO_WEP_SUPA7)))
+		{
+			const bool bClearForShotgun = me->IsLineOfFireClear(vShootablePos, CNEOBot::LINE_OF_FIRE_FLAGS_SHOTGUN);
+			if (bClearForShotgun)
+			{
+				bNotPrimary = false;
+				me->EquipBestWeaponForThreat(threat, bNotPrimary);
+				myWeapon = static_cast<CNEOBaseCombatWeapon *>(me->GetActiveWeapon());
+			}
+			bShotgunSituationHandled = true;
+		}
+	}
+	else if (myWeapon && (myWeapon->GetNeoWepBits() & (NEO_WEP_AA13 | NEO_WEP_SUPA7)))
+	{
+		const bool bClearForShotgun = me->IsLineOfFireClear(vShootablePos, CNEOBot::LINE_OF_FIRE_FLAGS_SHOTGUN);
+		if (!bClearForShotgun)
+		{
+			// Try to switch over to non-shotgun weapon
+			bNotPrimary = true;
+			me->EquipBestWeaponForThreat(threat, bNotPrimary);
+			myWeapon = static_cast<CNEOBaseCombatWeapon *>(me->GetActiveWeapon());
+			// If it's still Supa7, then it should try to dodge the threat instead
+			if (myWeapon && (myWeapon->GetNeoWepBits() & (NEO_WEP_AA13 | NEO_WEP_SUPA7)))
+			{
+				return;
+			}
+			bShotgunSituationHandled = true;
 		}
 	}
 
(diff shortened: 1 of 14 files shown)


sunzenshen (dismissed):
LGTM, bots now can shoot through glass when they couldn't do so before.


https://github.com/user-attachments/assets/03acaac2-80b6-488f-84f8-b0e414067e9b



https://github.com/user-attachments/assets/367b67c2-1b21-4064-aa2d-e781d9af4bbb

src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
```diff
@@ -587,13 +587,64 @@ void CNEOBotMainAction::FireWeaponAtEnemy( CNEOBot *me )
...
+	Vector vShootablePos = threat->GetEntity()->EyePosition();
+	if ( !me->IsLineOfFireClear( vShootablePos, lofFlags ) )
+	{
+		vShootablePos = threat->GetEntity()->WorldSpaceCenter();
+		if ( !me->IsLineOfFireClear( vShootablePos, lofFlags ) )
+		{
+			vShootablePos = threat->GetEntity()->GetAbsOrigin();
+			if ( !me->IsLineOfFireClear( vShootablePos, lofFlags ) )
+			{
+				return;
```
Rainyan: Why do we do LoF checks from the worldspace center & abs origin here as fallbacks? Don't the bullets always originate from the eye? Or is this some kind of special case we need to handle?

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -1693,50 +1701,131 @@ bool CNEOBot::IsQuietWeapon(CNEOBaseCombatWeapon* weapon) const
...
+		return true;
+	}
+	return BaseClass::IsLineOfSightClear(entity, checkType);
+}
+
+// Had to override this function otherwise it'll not find it within CNEOBot
+bool CNEOBot::IsLineOfSightClear(const Vector &pos, LineOfSightCheckType checkType, CBaseEntity *entityToIgnore) const
+{
+	return BaseClass::IsLineOfSightClear(pos, checkType, entityToIgnore);
+}
```
Rainyan: Nit: Admittedly I'm not too familiar with this syntax, but as suggested in https://stackoverflow.com/a/3678284 perhaps we could use the `using` directive below the header's class method declaration, to bring the base class overload declaration back into scope:

```diff
diff --git a/src/game/server/neo/bot/neo_bot.cpp b/src/game/server/neo/bot/neo_bot.cpp
index 82a4906a..b81f695c 100644
--- a/src/game/server/neo/bot/neo_bot.cpp
+++ b/src/game/server/neo/bot/neo_bot.cpp
@@ -1763,12 +1763,6 @@ bool CNEOBot::IsLineOfSightClear(CBaseEntity *entity, LineOfSightCheckType check
        return BaseClass::IsLineOfSightClear(entity, checkType);
 }

-// Had to override this function otherwise it'll not find it within CNEOBot
-bool CNEOBot::IsLineOfSightClear(const Vector &pos, LineOfSightCheckType checkType, CBaseEntity *entityToIgnore) const
-{
-       return BaseClass::IsLineOfSightClear(pos, checkType, entityToIgnore);
-}
-
 //-----------------------------------------------------------------------------------------------------
 // Return true if a weapon has no obstructions along the line between the given points
 bool CNEOBot::IsLineOfFireClear(const Vector& from, const Vector& to, const LineOfFireFlags flags) const
diff --git a/src/game/server/neo/bot/neo_bot.h b/src/game/server/neo/bot/neo_bot.h
index 1a1745b1..51938c50 100644
--- a/src/game/server/neo/bot/neo_bot.h
+++ b/src/game/server/neo/bot/neo_bot.h
@@ -199,7 +199,7 @@ public:
        bool IsLineOfFirePenetrationClear(const trace_t &tr, const Vector &from, const Vector &to) const;

        bool IsLineOfSightClear(CBaseEntity *entity, LineOfSightCheckType checkType = IGNORE_NOTHING) const override;
-       bool IsLineOfSightClear(const Vector &pos, LineOfSightCheckType checkType = IGNORE_NOTHING, CBaseEntity *entityToIgnore = NULL) const override;
+       using BaseClass::IsLineOfSightClear;

        bool IsLineOfFireClear(const Vector& where, const LineOfFireFlags flags) const;                 // return true if a weapon has no obstructions along the line from our eye to the given position
        bool IsLineOfFireClear(CBaseEntity* who, const LineOfFireFlags flags) const;                            // return true if a weapon has no obstructions along the line from our eye to the given entity
```

So that we didn't need the extra method that simply calls the BaseClass method without doing anything else, and avoid needing to traverse through multiple pointers in the vtable.

Rainyan (approved):
> TODO (LATER): Test/check for non-penetrable glass?

Tested against the tarmac non-penetrable glass, and indeed it seems the bots will frutilessly empty their magazine against the bulletproof windows.

To repro:
```
neo_bot_quota 2;
map ntre_tarmac_ctg;

// Spawn to the garage side, so the bot spawns on the tarmac side.

// Then, teleport near the bulletproof window.
setang 7.062000 13.297527 0.000000;
setpos 477.418365 -252.499146 -210.468750;
```

But we can leave it for some future PR if you don't feel like it belongs here.

src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
Rainyan: Nevermind, I misread this assuming `shootablepos` was the bullet-originating position, not the target.

nullsystem:
@Rainyan I've added in glass penetration vs non-penetration check, just does an extra trace then penetration check without distance limitation.

@sunzenshen I've added in distance restriction scaling by weapon's penetration, so for example: a bot using an SRS will attempt a further distance than MX which will attempt further than SRM.

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -1693,50 +1701,173 @@ bool CNEOBot::IsQuietWeapon(CNEOBaseCombatWeapon* weapon) const
...
+		if (!neoWeapon)
+		{
+			return false;
+		}
+
+		// Glass mode can just skip the distance check
+		bool bAttemptPenTest = (eMode == LINE_OF_FIRE_PENETRATION_MODE_GLASS);
+		if (!bAttemptPenTest)
+		{
+			const float flMaxTryDist = clamp(
```
sunzenshen: nit:  apparently there's a Valve-specific Clamp in \neo\src\public\tier0\basetypes.h

```cpp
// This is the preferred clamp operator. Using the clamp macro can lead to
// unexpected side-effects or more expensive code. Even the clamp (all
// lower-case) function can generate more expensive code because of the
// mixed types involved.
template< class T >
T Clamp( T const &val, T const &minVal, T const &maxVal )
{
	if( val < minVal )
		return minVal;
	else if( val > maxVal )
		return maxVal;
	else
		return val;
}
```

Up to you if you think the Valve one is worth using over std::clamp

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -1693,50 +1701,173 @@ bool CNEOBot::IsQuietWeapon(CNEOBaseCombatWeapon* weapon) const
...
+		const ELineOfFirePenetrationMode eMode) const
+{
+	// Min is making sure at least the distance under it is considered
+	// Max is scaling up to penetration value of 100.0f which no weapon have, but
+	// scaling works out in general
+	static constexpr const float FL_METERS_TRY_PENETRATE_MIN = 9.0f;
+	static constexpr const float FL_METERS_TRY_PENETRATE_MAX = 36.0f;
+
+	int material = physprops->GetSurfaceData(tr.surface.surfaceProps)->game.material;
+	if (material != 'J')
```
sunzenshen: nit: maybe a comment explaining what 'J' refers to may be helpful.

Rainyan:
> Bots appear to open fire on a player that is holding the ghost behind a paper wall near the house spawn on Isolation

Maybe you meant this already as a positive, but this is probably the desired behaviour, since good players should absolutely wallbang the ghoster when they have the chance.

It'd be really cool if we had some kind of "confidence level" that the bots had on the enemy location, and if they felt confident enough (seen player recently in area, sound cues, etc), they could outright start wallbanging thin surfaces to do preemptive damage, like a real high-level human player would. But that's out of scope for this PR surely.

src/game/server/neo/bot/neo_bot.cpp
Rainyan: ^ Agreed that this magic value is a bit hard to grep/google for, so maybe adding a comment with link to https://developer.valvesoftware.com/wiki/Material_type wouldn't hurt.

Or well, having it refactored to some user-friendlier enumeration wouldn't hurt either but probably out of scope / difficult to make compatible across the whole engine... probably not worth it to make any drastic change to the underlying data type.

Rainyan (changes requested):
For the tarmac bulletproof glass case mentioned before, it seems that if you are holding the ghost, the bots will still attempt to wallbang.

If it's not too much work, could you make it such that even if the bots have vision of the ghoster (possibly through a bullet-proof window), if they know their bullets cannot penetrate (the wall thickness check), they won't waste bullets trying to shoot through the impenetrable areas, but instead consider it as impossible and choose whatever is their alternative action to shooting instantly.

src/game/server/neo/bot/neo_bot.cpp
nullsystem: It seems to not even been defined in `decals.h` (and that VPC article) properly, but assuming the original penetration code states `'J'` is blockbullets, I've `ifdef NEO` in `CHAR_TEX_BLOCKBULLETS` as `'J'` and used that define.

AdamTadeusz: https://tenor.com/view/burnt-peanut-the-burnt-peanut-side-eye-gif-14097258958682501049

sunzenshen:
Ghost behind ballistic glass now works as expected:

https://github.com/user-attachments/assets/898fbb74-284a-4b34-84be-3716d42fb4ce

PR #1285, opened 2025-08-18
More bot names

DESTROYGIRL:
## Description
More bot names, and there should never be any duplicate names on a 32 player server now

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/neo_bot.cpp
+++ b/src/game/server/neo/bot/neo_bot.cpp
@@ -107,18 +107,72 @@ const char* GetRandomBotName(void)
 	static const char* nameList[] =
 	{
 		"Deej",
-		"Ed",
+		"0edit",
 		"Filter Decay",
 		"Gato",
 		"Grey",
 		"Glasseater",
-		"Ivy",
+		"Operation Ivy",
 		"KillahMo",
 		"pushBAK",
 		"Tatsur0",
+		"rain",
+		"The Major",
+		"Wolf",
+		"Sanjuro Executive",
+		"Grumpy Old Man",
+		"Biodroid Prototype",
+		"TGR",
+		"Jinball",
+		"Pequod",
+		"The Chimpanzee",
+		"Marchenko",
+		"Jeff",
+		"M.A.W.S.",
+		"Tin Soldier",
+		"SUKKIT",
+		"Supaman",
+		"Hadaly",
+		"Komatsu Lifthelper",
+		"HE Specialist",
+		"Left Leaner",
+		"Super Class A Hacker",
+		"and You",
 	};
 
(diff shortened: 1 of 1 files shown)


src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -107,18 +107,72 @@ const char* GetRandomBotName(void)
...
+
+	for (int j = 1; j <= gpGlobals->maxClients && availableList.Count() > 0; ++j)
+	{
+		auto player = UTIL_PlayerByIndex(j);
+		if (!player)
+		{
+			continue;
+		}
+
+		for (int k = availableList.Count() - 1; k >= 0; --k)
```
sunzenshen, line 159: This might be an optimization nit, but it might be cleaner to use a hashmap to check if a name was used, instead of using a double nest of for loops.

Something like:

```cpp
CUtlStringMap<bool> usedNames;
for (int j = 1; j <= gpGlobals->maxClients; ++j) {
    auto player = UTIL_PlayerByIndex(j);
    if (player) {
        usedNames.Insert(player->GetPlayerName(), true);
    }
}
```

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -107,18 +107,72 @@ const char* GetRandomBotName(void)
...
+		{
+			if (V_stristr(player->GetPlayerName(), nameList[availableList[k]]))
```
sunzenshen, line 161: This is probably fine for the current list, but if we ever add more short names like "Ed" to the list, a partial match could potentially lead to an accidental collision.  Changing "Ed" to "0edit" made this case even more rare, but up to you if you think this would be worth being more explicit about.

A more explicit string comparison would be along the lines of:
```cpp
if (V_stricmp(playerName, nameList[availableList[k]]) == 0)
```

An exact match could also be faster than a substring match, but I'd be surprised if it makes a difference here.

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -107,18 +107,72 @@ const char* GetRandomBotName(void)
...
+	}
+
+	if (availableList.Count() == 0)
+	{
+		return nameList[RandomInt(0, nameCount - 1)];
+	}
+
+	return nameList[availableList[RandomInt(0, availableList.Count() - 1)]];
+
+	//return nameList[RandomInt(0, ARRAYSIZE(nameList) - 1)];
```
sunzenshen, line 175: If you come back with another commit, might be worth cleaning up this previous line.

DESTROYGIRL:
This will probably be closed in favour of https://github.com/NeotokyoRebuild/neo/pull/1289 but will leave it up until it gets merged.

PR #1288, opened 2025-08-19
Refactor loadout source code and header usages

### LINKED ISSUE #1287: Refactor neo_weapon_loadout and some headers
The API of the loadout can be simplified and reduce repetitive codes. Also some header are re-arrange so there's less likely to hit some compiler+linker issues when including headers.

nullsystem:
## Description
* New headers and re-arranged around for less of a chance of annoying compiler+linker errors when trying to work on adding new source codes
* Loadout completely refactored to reduce repetative defined weapons and simplified to just using two arrays
* Required for the WIP bot profiles feature
* Got rid of `m_szAmmoType` as that's never used

## Toolchain

## Linked Issues

* fixes #1287

### DIFF
--- a/src/game/shared/neo/neo_weapon_loadout.h
+++ b/src/game/shared/neo/neo_weapon_loadout.h
@@ -1,249 +1,43 @@
-#ifndef NEO_WEAPON_LOADOUT_H
-#define NEO_WEAPON_LOADOUT_H
-#ifdef _WIN32
 #pragma once
-#endif
 
-#define DEV_WEAPON_LOADOUTS 12
-#define RECON_WEAPON_LOADOUTS 9
-#define ASSAULT_WEAPON_LOADOUTS 12
-#define SUPPORT_WEAPON_LOADOUTS 9
-#define VIP_WEAPON_LOADOUTS 9
+#include "neo_enums.h"
 
-class CLoadoutWeaponClass
+// NEO NOTE (nullsystem): If there's NEO_BUILD_WEAPON_PBK56S / INCLUDE_WEP_PBK
+// inclusion or new weapons, may want to change/alter the loadouts for it.
+static const constexpr int MAX_WEAPON_LOADOUTS = 12;
+
+enum ELoadoutCount
 {
-public:
-	const char* m_szWeaponName;
+	NEO_LOADOUT_DEV = NEO_CLASS__ENUM_COUNT,
+	NEO_LOADOUT_INVALID,
+	NEO_LOADOUT__COUNT,
+};
+
+struct WeaponInfo
+{
+	const char *m_szWeaponName = "";
+	const char *m_szVguiImage = "loadout/loadout_none";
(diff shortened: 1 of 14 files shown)


src/game/shared/neo/neo_weapon_loadout.h
```diff
@@ -1,249 +1,35 @@
-#ifndef NEO_WEAPON_LOADOUT_H
-#define NEO_WEAPON_LOADOUT_H
-#ifdef _WIN32
 #pragma once
-#endif
```
sunzenshen, line 5: Out of curiosity, from this point on, are we only using `#pragma once` for any new code from this point on?  Wondering if there's any historical reason why we use old style macro-based header guards.

src/game/shared/neo/neo_weapon_loadout.h
```diff
@@ -1,249 +1,35 @@
...
-	{
-		m_szWeaponName = weaponName;
-		m_iWeaponPrice = weaponPrice;
-		m_szVguiImage = vguiImage;
-		m_szVguiImageNo = vguiImageNo;
-		m_szWeaponEntityName = weaponEntityName;
-		m_szAmmoType = ammoType;
-	}
-};
+static const constexpr int MAX_WEAPON_LOADOUTS = 12;
```
sunzenshen, line 7: Is it possible that this constant might change in the future?  For example, whether the PBK56 ever gets added in the future.  I wonder if it makes sense to call out the related macro `NEO_BUILD_WEAPON_PBK56S` just in case we need a known string to a constant to be easily searchable.

src/game/shared/neo/neo_weapon_loadout.h
nullsystem: It was just following the SDK's convention which I think the SDK's work on Linux pre-dates GCC 3.4 before it had pragma once hence it was like this. Similar to this we was also using stuff like `OVERRIDE` macro as oppose to `override` keyword in the past due to using VS2013 + GCC 4, but of course since the toolchain was updated newer code moved on but older code haven't changed over similar to the header guards not changed over either.

Can just simply update all `neo` source codes to all use `#pragma once` and `override` keyword anyway.

src/game/shared/neo/neo_weapon_loadout.h
nullsystem: It feels kind of up in the air whether PBK56 or whatever extra new weapons going to change the load-outs total or not, both recons and support have 9 of 12 total slots. Probably can just comment about `NEO_BUILD_WEAPON_PBK56S` here for the time being anyway.

src/game/shared/neo/neo_weapon_loadout.h
sunzenshen: > Can just simply update all `neo` source codes to all use `#pragma once` and `override` keyword anyway.

Maybe in a different PR. 👍

sunzenshen (approved):

Rainyan (approved):

PR #1292, opened 2025-08-20
Zoom sensitivity ratio in settings

AdamTadeusz:
## Description

exposes the zoom_sensitivity_ratio convar.

negative values invert controls when aimed in ( can only be set via console/config file )
a value of 0 removes the scaling of mouse sensitivity with fov
a value less than 1 scales down sensitivity when aimed in
a value greater than 1 scales up sensitivity when aimed in ( values greater than 10 only via console/config file )


sunzenshen (commented):
Checked that fractions less than 1.00 result in reduced sensitivity when zoomed, and more than 1.00 results in higher sensitivity.

There's a bit of an edge case where a 0.00 feels... off but not consistently slower or faster 1.00?

One potential nit is that I'm actually not sure what the utility is of values more than 1.00 or less than 0.01.  Perhaps it makes sense to have a min and max configurable in the config GUI.

sunzenshen (approved):

DESTROYGIRL (approved):

Issue #1254, opened 2025-08-02
[FEATURE] Take over bots


Rainyan:
### Description

Add ability to take over bots in your team while in dead-spec.

### Proposed Solution

As a player in Jinrai or NSF team, after I've died for the ongoing round, I want to be able to take control over any existing alive bots in my team, because it's more fun to me than spectating the bots.

### Alternatives Considered

While the bots can be fun to play with/against, passively spectating the bots play out the round doesn't feel as enticing, since the human factor is missing. In contrast, spectating human players is more engaging to me, because I can speculate on their decision-making process—and there's the practical limitation of taking over human players from spec being unreasonable.

### Additional Context

This feature exists in a lot of other games with bots, such as Counter-Strike, where pressing the use key allows controlling a bot.

PR #1294, opened 2025-08-21
Spectator ability to take over bots and AFK players

sunzenshen:
## Description
Proof of concept for taking over AFK players and bots as a spectator

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes https://github.com/NeotokyoRebuild/neo/issues/1254

### DIFF
--- a/src/game/client/neo/c_neo_player.cpp
+++ b/src/game/client/neo/c_neo_player.cpp
@@ -752,6 +752,9 @@ bool C_NEO_Player::ShouldDraw( void )
 void C_NEO_Player::OnDataChanged( DataUpdateType_t type )
 {
 	BaseClass::OnDataChanged(type);
+
+	// Had to delay client sync until player model was ready
+	CSpectatorTakeoverPlayerUpdateOnDataChanged();
 }
 
 float C_NEO_Player::GetFOV( void )
@@ -1868,3 +1871,74 @@ const char *C_NEO_Player::GetOverrideStepSound(const char *pBaseStepSound)
 
 	return BaseClass::GetOverrideStepSound(pBaseStepSound);
 }
+
+// Receive server message to smooth out player takeover
+void __MsgFunc_CSpectatorTakeoverPlayer(bf_read &msg)
+{
+	int spectatorTakingOverUserID = msg.ReadLong();
+	int playerTakeoverTargetUserID = msg.ReadLong();
+
+	C_NEO_Player* pSpectatorTakingOver = USERID2NEOPLAYER(spectatorTakingOverUserID);
+	C_NEO_Player* pPlayerTakeoverTarget = USERID2NEOPLAYER(playerTakeoverTargetUserID);
+
+
+	if (!pPlayerTakeoverTarget || !pSpectatorTakingOver)
+	{
+		return;
+	}
+
+	// Save for later in C_NEO_Player::OnDataChanged
+	pSpectatorTakingOver->m_hSpectatorTakeoverTarget = pPlayerTakeoverTarget;
+}
+
+CUserMessageRegister CSpectatorTakeoverPlayerRegistration("CSpectatorTakeoverPlayer", __MsgFunc_CSpectatorTakeoverPlayer);
+
+void C_NEO_Player::CSpectatorTakeoverPlayerUpdateOnDataChanged()
+{
+    if ( C_NEO_Player* pTarget = m_hSpectatorTakeoverTarget.Get() )
+	{
+        // Check if our model is stable from server & matches the target
+        if (GetModel() && (GetModel() == pTarget->GetModel()))
+        {
+            CSpectatorTakeoverPlayerUpdate(pTarget);
+            m_hSpectatorTakeoverTarget = nullptr;
+        }
+    }
+}
+
+void C_NEO_Player::CSpectatorTakeoverPlayerUpdate(C_NEO_Player* pPlayerTakeoverTarget)
+{
+	if (!pPlayerTakeoverTarget)
+	{
+		return;
+	}
+
+	m_nSkin = pPlayerTakeoverTarget->m_iNeoSkin;
+	m_iNeoClass = pPlayerTakeoverTarget->m_iNeoClass;
+	m_iLoadoutWepChoice = pPlayerTakeoverTarget->m_iLoadoutWepChoice;
+	m_iNextSpawnClassChoice = pPlayerTakeoverTarget->m_iNextSpawnClassChoice;
+
+	m_bInThermOpticCamo = pPlayerTakeoverTarget->m_bInThermOpticCamo;
+	m_bInVision = pPlayerTakeoverTarget->m_bInVision;
+	m_bInAim = pPlayerTakeoverTarget->m_bInAim;
+	m_bCarryingGhost = pPlayerTakeoverTarget->m_bCarryingGhost;
+	m_bIneligibleForLoadoutPick = pPlayerTakeoverTarget->m_bIneligibleForLoadoutPick;
+	m_bInLean = pPlayerTakeoverTarget->m_bInLean;
+
+	m_flCamoAuxLastTime = pPlayerTakeoverTarget->m_flCamoAuxLastTime;
+	m_nVisionLastTick = pPlayerTakeoverTarget->m_nVisionLastTick;
+	m_flLastAirborneJumpOkTime = pPlayerTakeoverTarget->m_flLastAirborneJumpOkTime;
+	m_flLastSuperJumpTime = pPlayerTakeoverTarget->m_flLastSuperJumpTime;
+	m_bPreviouslyReloading = pPlayerTakeoverTarget->m_bPreviouslyReloading;
+	m_bLastTickInThermOpticCamo = pPlayerTakeoverTarget->m_bLastTickInThermOpticCamo;
+	m_bIsAllowedToToggleVision = pPlayerTakeoverTarget->m_bIsAllowedToToggleVision;
+	m_flTocFactor = pPlayerTakeoverTarget->m_flTocFactor;
+
+	pPlayerTakeoverTarget->SnatchModelInstance(this);
+	SetAbsOrigin(pPlayerTakeoverTarget->GetAbsOrigin());
+	SetAbsAngles(pPlayerTakeoverTarget->GetAbsAngles());
+	SetAbsVelocity(pPlayerTakeoverTarget->GetAbsVelocity());
+	SetLocalAngles(pPlayerTakeoverTarget->GetLocalAngles());
+	pl.v_angle = pPlayerTakeoverTarget->pl.v_angle; // mimic SnapEyeAngles in neo_player
+}
+
--- a/src/game/server/neo/neo_player.cpp
+++ b/src/game/server/neo/neo_player.cpp
@@ -3449,9 +3449,9 @@ const char *CNEO_Player::GetOverrideStepSound(const char *pBaseStepSound)
 // Start spectator takeover of player related code:
 ConVar sv_neo_spec_replace_player_loadout_enable("sv_neo_spec_replace_player_loadout_enable", "0", FCVAR_NONE, "Allow loadout change after spectator takeover.", true, 0, true, 1);
 ConVar sv_neo_spec_replace_player_bot_enable("sv_neo_spec_replace_player_bot_enable", "1", FCVAR_NONE, "Allow spectators to take over bots.", true, 0, true, 1);
-ConVar sv_neo_spec_replace_player_afk_enable("sv_neo_spec_replace_player_afk_enable", "1", FCVAR_NONE, "Allow spectators to take over AFK players.", true, 0, true, 1);
+ConVar sv_neo_spec_replace_player_afk_enable("sv_neo_spec_replace_player_afk_enable", "0", FCVAR_NONE, "Allow spectators to take over AFK players.", true, 0, true, 1);
 ConVar sv_neo_spec_replace_player_afk_time_sec( "sv_neo_spec_replace_player_afk_time_sec",
-	"120", FCVAR_NONE,
+	"180", FCVAR_NONE,
 	"Seconds of inactivity before a player is considered AFK for spectator takeover.",
 	true, -1, true, 999);
 ConVar sv_neo_spec_replace_player_min_exp("sv_neo_spec_replace_player_min_exp",
@@ -3459,7 +3459,8 @@ ConVar sv_neo_spec_replace_player_min_exp("sv_neo_spec_replace_player_min_exp",
 	"Minimum experience allowed to takeover players ",
 	true, -999, true, 999);
 
-bool CNEO_Player::IsAFK() const {
+bool CNEO_Player::IsAFK() const
+{
     // NEO JANK GetTimeSinceLastUserCommand seems to return 0 as long as the player is connected, so use an alternative timer
     return GetTimeSinceWeaponFired() > sv_neo_spec_replace_player_afk_time_sec.GetInt();
 }
--- a/src/game/server/neo/neo_player.cpp
+++ b/src/game/server/neo/neo_player.cpp
@@ -934,6 +934,8 @@ void CNEO_Player::HandleSpeedChanges( CMoveData *mv )
 
 void CNEO_Player::PreThink(void)
 {
+	SpectatorTakeoverPlayerPreThink();
+
 	BaseClass::PreThink();
 
 	if (m_HL2Local.m_flSuitPower <= 0.0f && IsSprinting())
@@ -1728,6 +1730,28 @@ bool CNEO_Player::ClientCommand( const CCommand &args )
 
 		return true;
 	}
+	else if (FStrEq(args[0], "spectatortakeoverplayer"))
+	{
+		if (args.ArgC() < 2)
+		{
+			return false;
+		}
+
+		int targetIndex = V_atoi(args[1]);
+		CBaseEntity* pTarget = UTIL_PlayerByUserId(targetIndex);
+		CNEO_Player* pPlayerToTakeover = ToNEOPlayer(pTarget);
+
+		if (pPlayerToTakeover)
+		{
+			SpectatorTryReplacePlayer(pPlayerToTakeover);
+		}
+		else
+		{
+			return false;
+		}
+
+		return true;
+	}
 
 	return BaseClass::ClientCommand(args);
 }
@@ -1842,8 +1866,19 @@ void CNEO_Player::StartShowDmgStats(const CTakeDamageInfo *info)
 	MessageEnd();
 }
 
-void CNEO_Player::AddPoints(int score, bool bAllowNegativeScore)
+void CNEO_Player::AddPoints(int score, bool bAllowNegativeScore, bool bIgnorePlayerTakeover)
 {
+	if (!bIgnorePlayerTakeover && m_hSpectatorTakeoverPlayerTarget.Get())
+	{
+		if (score >= 0)
+		{
+			// Don't penalize possessed/bot for takeover player's actions
+			m_hSpectatorTakeoverPlayerTarget->AddPoints(score, false);
+			return;
+		}
+		// If a player made a mistake while taking over another player, continue to penalize them
+	}
+
 	// Positive score always adds
 	if (score < 0)
 	{
@@ -1912,6 +1947,8 @@ void CNEO_Player::Event_Killed( const CTakeDamageInfo &info )
 	{
 		SetDeadModel(info);
 	}
+
+	SpectatorTakeoverPlayerRevert(false); // soft reset: may still have live impostor
 }
 
 void CNEO_Player::Weapon_DropAllOnDeath( const CTakeDamageInfo &info )
@@ -2707,6 +2744,8 @@ bool CNEO_Player::ProcessTeamSwitchRequest(int iTeam)
 
 	NEORules()->m_bThinkCheckClantags = true;
 
+	SpectatorTakeoverPlayerRevert(); // hard reset: no longer on the same team
+
 	return true;
 }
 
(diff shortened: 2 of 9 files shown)


src/game/server/neo/neo_player.cpp
```diff
@@ -3231,3 +3261,233 @@ void CNEO_Player::ModifyFireBulletsDamage(CTakeDamageInfo* dmgInfo)
...
+	m_bSpectatorTakeoverPlayerPending = true;
+	m_hSpectatorTakeoverPlayerTarget = pNeoPlayerToReplace;
+	ForceRespawn();
+
+	// Send client confirmation to smooth out transition
+	CBroadcastRecipientFilter updatefilter;
+    updatefilter.MakeReliable();
+    UserMessageBegin(updatefilter, "CSpectatorTakeoverPlayer");
+		WRITE_SHORT(this->entindex());
+		WRITE_SHORT(pNeoPlayerToReplace->entindex());
```
Rainyan: Entindex is not guaranteed unique, so we should send userid over network instead of entindex, because we have no time guarantee on when the message arrives. So technically that entindex could become invalid, or be occupied by someone else by the time the message arrives.

src/game/server/neo/neo_player.cpp
```diff
@@ -3231,3 +3261,233 @@ void CNEO_Player::ModifyFireBulletsDamage(CTakeDamageInfo* dmgInfo)
 {
 	dmgInfo->SetDamage(dmgInfo->GetDamage() * sv_neo_wep_dmg_modifier.GetFloat());
 }
+
+// Start spectator takeover of player related code:
+ConVar sv_neo_spec_replace_player_bot_enable("sv_neo_spec_replace_player_bot_enable", "1", FCVAR_CHEAT, "Allow spectators to take over bots.");
+ConVar sv_neo_spec_replace_player_afk_enable("sv_neo_spec_replace_player_afk_enable", "1", FCVAR_CHEAT, "Allow spectators to take over AFK players.");
+ConVar sv_neo_spec_replace_player_afk_time_sec("sv_neo_spec_replace_player_afk_time_sec", "300", FCVAR_CHEAT, "Seconds of inactivity before a player is considered AFK.");
```
Rainyan: These probably don't need to be marked as cheats, so servers can disable/edit the feature if they want to, without requiring turning on `sv_cheats`

src/game/server/neo/neo_player.cpp
```diff
@@ -3231,3 +3261,233 @@ void CNEO_Player::ModifyFireBulletsDamage(CTakeDamageInfo* dmgInfo)
...
+
+	if (m_bSpectatorTakeoverPlayerPending)
+	{
+		if (!IsAlive() || IsObserver())
+		{
+			// Something is wrong, abort the takeover.
+			m_bSpectatorTakeoverPlayerPending = false;
+			m_hSpectatorTakeoverPlayerTarget = nullptr;
+			return;
+		}
```
Rainyan, line 3577: Probably should fail more loudly if this is unexpected, so at least an `Assert(false)` or `DevWarning` would be useful so it doesn't go unnoticed

Rainyan (changes requested):
Screenshot 2025-08-26 164040

A possessed bot can still see its own IFF HUD icon in the place where it got taken over.

src/game/client/neo/c_neo_player.cpp
```diff
@@ -1804,3 +1807,73 @@ void C_NEO_Player::ModifyFireBulletsDamage(CTakeDamageInfo* dmgInfo)
...
+	pSpectatorTakingOver->m_hSpectatorTakeoverTarget = pPlayerTakeoverTarget;
+}
+
+CUserMessageRegister CSpectatorTakeoverPlayerRegistration("CSpectatorTakeoverPlayer", __MsgFunc_CSpectatorTakeoverPlayer);
+
+void C_NEO_Player::CSpectatorTakeoverPlayerUpdateOnDataChanged()
+{
+    if (m_hSpectatorTakeoverTarget.Get())
+    {
+        C_NEO_Player* pTarget = m_hSpectatorTakeoverTarget.Get();
```
Rainyan: Nit: This double lookup most likely is optimized away, but you could simplify it to:
```diff
 void C_NEO_Player::CSpectatorTakeoverPlayerUpdateOnDataChanged()
 {
-    if (m_hSpectatorTakeoverTarget.Get())
+    if (C_NEO_Player* pTarget = m_hSpectatorTakeoverTarget.Get())
```
where the if-scoped assignment is also the condition.

The `GetModel` check could also simplify to:
```diff
         // Check if our model is stable from server & matches the target
-        if (GetModel() && pTarget->GetModel() && GetModel() == pTarget->GetModel())
+        if (GetModel() && GetModel() == pTarget->GetModel())
```
because the equality check together with `if (GetModel())` already implies that `pTarget->GetModel()` must not equal nullptr.

Rainyan:
It seems that players who possess a bot, and then disconnect, will cause the restored bot to A-pose:

Apose

To reproduce (tested with native Windows 11 as Client 1 & VirtualBox popOS Linux as Client 2):
* Host a server on Client 1
* Join the server on Client 2
* Join the same team on both clients
* Wait for round to start
* Suicide on client 2
* Possess a bot with client 2
* Disconnect with client 2

What happens:
* The disconnected bot goes to the default "A-pose"

What should happen:
* The bot appears normal

sunzenshen:
> A possessed bot can still see its own IFF HUD icon in the place where it got taken over.

> It seems that players who possess a bot, and then disconnect, will cause the restored bot to A-pose:

At this point, I believe marking the bot as dead/invisible/non-solid is the cleanest way to handle these two cases.

Here's a video that demonstrates a player taking over a bot and disconnecting:

https://github.com/user-attachments/assets/757c842a-7b49-4381-a1bd-bc3c6545e10d

In order to add some options to mitigate griefing with this feature, I have added a configurable minimum XP threshold under which players cannot use the takeover feature anymore.  This is aimed specifically for a self `kill` takeover another bot repetition.

Known bug? I wasn't able to get the AFK feature working on my test setup, because GetTimeSinceLastUserCommand() always returns 0 with my local test setup. For now I have left the feature convar disabled by default.

sunzenshen:
The latest changes involve changing the criteria for which players are counted as AFK to use a different timer based on the last time they fired their weapon.  This is admittedly kind of a jank workaround for the previous timer always being 0, but at least it allowed me to test takeovers of AFK players. That led to removing the freezing flag from the taken over player, because there was a bug where an AFK player could not move around as a spectator after being taken over, until the round reset.

As far as I can tell, I have addressed the feedback from earlier threads, so I will mark those as resolved, but feel free to revisit them to see if I have addressed the issues in the way that is satisfactory.

src/game/client/neo/c_neo_player.cpp
```diff
@@ -1867,4 +1870,95 @@ const char *C_NEO_Player::GetOverrideStepSound(const char *pBaseStepSound)
...
+	m_bPreviouslyReloading = pPlayerTakeoverTarget->m_bPreviouslyReloading;
+	m_bLastTickInThermOpticCamo = pPlayerTakeoverTarget->m_bLastTickInThermOpticCamo;
+	m_bIsAllowedToToggleVision = pPlayerTakeoverTarget->m_bIsAllowedToToggleVision;
+	m_flTocFactor = pPlayerTakeoverTarget->m_flTocFactor;
+
+	pPlayerTakeoverTarget->SnatchModelInstance(this);
+	SetAbsOrigin(pPlayerTakeoverTarget->GetAbsOrigin());
+	SetAbsAngles(pPlayerTakeoverTarget->GetAbsAngles());
+	SetAbsVelocity(pPlayerTakeoverTarget->GetAbsVelocity());
+	SetLocalAngles(pPlayerTakeoverTarget->EyeAngles());
```
Rainyan: It seems `LocalAngles` maps to `m_angRotation` whereas `EyeAngles` maps to `m_angEyeAngles`, so maybe it'd be safer to to assign `SetLocalAngles(pPlayerTakeoverTarget->GetLocalAngles())`, just in case a leaned player can somehow carry over their eye roll angle to the angRotation variable?

Seems we're assigning the eyeangles in the base clas `::PostThink` anyways, so it should hopefully keep up to sync without an explicit assignment here(?)

src/game/client/neo/c_neo_player.cpp
```diff
@@ -1867,4 +1870,95 @@ const char *C_NEO_Player::GetOverrideStepSound(const char *pBaseStepSound)
...
+            player_info_t info;
+            if (engine->GetPlayerInfo(i, &info) && info.userID == userID)
+            {
+                return ToNEOPlayer(pPlayer);
+            }
+        }
+    }
+
+    return nullptr;
+}
```
Rainyan: Nit: this function body could simplify to `return ToNEOPlayer(ClientEntityList().GetEnt(engine->GetPlayerForUserID(userID)));`, similar to this macro oneliner: https://github.com/NeotokyoRebuild/neo/blob/d2b54b536c8b290c82efb95a0637d30eac74c6c0/src/game/client/clientmode_shared.h#L49

Maybe also worth moving the function to the `neo_player_shared.h` header, so it can be reused in NEO code elsewhere?

src/game/server/neo/neo_player.cpp
```diff
@@ -1733,6 +1735,24 @@ bool CNEO_Player::ClientCommand( const CCommand &args )
...
+		int targetIndex = V_atoi(args[1]);
+		CBaseEntity* pTarget = UTIL_PlayerByUserId(targetIndex);
+		CNEO_Player* pPlayerToTakeover = ToNEOPlayer(pTarget);
+
+		if (pPlayerToTakeover)
+		{
+			SpectatorTryReplacePlayer(pPlayerToTakeover);
+		}
+
+		return true;
```
Rainyan, line 1753: Nit: should probably return false for the `!pPlayerToTakeover` failure case here(?)

src/game/server/neo/neo_player.cpp
```diff
@@ -1849,6 +1869,12 @@ void CNEO_Player::StartShowDmgStats(const CTakeDamageInfo *info)
 
 void CNEO_Player::AddPoints(int score, bool bAllowNegativeScore)
 {
+	if (m_hSpectatorTakeoverPlayerTarget.Get())
+	{
+		m_hSpectatorTakeoverPlayerTarget->AddPoints(score, bAllowNegativeScore);
+		return;
+	}
```
Rainyan: Since it appears we're doing some kind of I assume anti-griefing logic with the `sv_neo_spec_replace_player_min_exp`, would it make sense to not allow negative score assignments here, since any negative scores would be assigned to the bot, and not to the trolling player?

Also, it looks like suiciding as a player-possessed bot gives the bot twice the XP penalty, somewhere in the game logic. Here's the bot score after:
* Dying to the enemy normally as a human player (name "rain")
* Taking over the bot (name "BOT Gato")
* Suiciding with a grenade as the bot
* Losing the round

Where I'd expect to see a -1XP penalty, but the bot has a -2XP score. And also myself as the "trolling" player still have a clean slate of 0XP, so the `..._min_exp` cvar limit would not consider me unfit to do the same trolling again the next round - this is probably something we would want to prevent for game quality.

Screenshot 2025-09-05 182508

It also seems my player's death counter was incremented twice for the round, but I'm not sure whether that's the intended behaviour here or not (I guess taking over a bot & dying means one did *technically* die twice).

Rainyan:
Also just speculation on my part, but the double ragdoll case could be happening because of the incorrect suicide + knifekill event recording, assuming the ragdoll logic has an event listener somewhere that's responding to both.

src/game/server/neo/neo_player.cpp
```diff
@@ -3386,4 +3414,264 @@ const char *CNEO_Player::GetOverrideStepSound(const char *pBaseStepSound)
...
+
+		// If the replaced player was a human, they were AFK so kick them to spectator
+		if (pPlayerTakenOver && !pPlayerTakenOver->IsBot())
+		{
+			pPlayerTakenOver->ChangeTeam(TEAM_SPECTATOR);
+		}
+	}
+}
+
+void CNEO_Player::RestorePlayerFromSpectatorTakeover(const CTakeDamageInfo &pInfo)
```
sunzenshen: I suspect this version of the RestorePlayerFromSpectatorTakeover is what is causing a double count, but I'll need to check.  Perhaps at this point we agree that the bot being taken over should not be penalized for any mistakes or demerits the player is performing in its stead.

src/game/client/neo/c_neo_player.cpp
sunzenshen: > Seems we're assigning the eyeangles in the base clas `::PostThink` anyways, so it should hopefully keep up to sync without an explicit assignment here(?)

I was thinking it would be safer to assign too many things in one commit and then possibly see what we can get away with cutting after we have a baseline working feature.

src/game/server/neo/neo_player.cpp
```diff
@@ -1917,6 +1943,8 @@ void CNEO_Player::Event_Killed( const CTakeDamageInfo &info )
 	{
 		SetDeadModel(info);
```
sunzenshen, line 1948: Calling the Event_Killed event on the bot too probably was spawning the 2nd body here.

src/game/server/neo/neo_player.cpp
```diff
@@ -3386,4 +3414,264 @@ const char *CNEO_Player::GetOverrideStepSound(const char *pBaseStepSound)
...
+			pPlayerTakenOver->ChangeTeam(TEAM_SPECTATOR);
+		}
+	}
+}
+
+void CNEO_Player::RestorePlayerFromSpectatorTakeover(const CTakeDamageInfo &pInfo)
+{
+	if (m_hSpectatorTakeoverPlayerTarget.Get())
+	{
+		m_hSpectatorTakeoverPlayerTarget->Event_Killed(pInfo);
```
sunzenshen: Specifically, passing Event_Killed(pInfo) to the bot was causing it to get killed a 2nd time, with another ragdoll, along with the indignity of getting assigned negative points from the takeover player's scoring.

sunzenshen:
@Rainyan I made changes with respect to the latest round of feedback, but there's one known bug that is technical debt from the current approach of respawning the current player in place of a taken over player.  It turns out that while I (hopefully) resolved the second ragdoll being generated, this whole time it turns out that each player only gets one ragdoll to create in the world at a time.  This is possibly true for TDM mode too, as when a player dies, their existing ragdoll from a previous death gets cleaned up just before creating another ragdoll.

The relevant code can be found in `\neo\src\game\server\hl2mp\hl2mp_player.cpp`:

```cpp
void CHL2MP_Player::CreateRagdollEntity( void )
{
	if ( m_hRagdoll )
	{
		UTIL_RemoveImmediate( m_hRagdoll );
		m_hRagdoll = NULL;
	}

	// If we already have a ragdoll, don't make another one.
	CHL2MPRagdoll *pRagdoll = dynamic_cast< CHL2MPRagdoll* >( m_hRagdoll.Get() );
```

Here's video of a possessor player's original ragdoll disappearing as the ragdoll of the newer incarnation is created:

https://github.com/user-attachments/assets/6c1d836c-aa6b-46df-9e8f-2abb9a15eca6

(Also demonstrates that the bot FilterDecay isn't penalized for the takeover player's self `kill`)

Here's another video of sanity checking the scoring system after the code review changes:

https://github.com/user-attachments/assets/cac9d8e1-cfda-4697-9116-0af75318fd5c

So to recap, the known suboptimal issues are:
- To prevent editing too many unrelated functions in this PR, AFK checking is overly simple, only checking the last time a player fired a weapon.  I upped the time threshold to 2 minutes given that simplistic check.
- When a player that is taking over another player dies, their original ragdoll disappears, as each player only gets one ragdoll at a given time.

Rainyan:
Thanks. Yeah, I think as long as we use the possessed bot's ragdoll to present their death visualization it should be fine for the 99% case (ignoring deathmatch outliers). Because it's the possessed bot "dying", and not the controlling player. But I think focusing on good CTG mode experience is perfectly fine here, and if it produces a bug in alternative modes, then we can focus on those problems if they arise.

src/game/server/neo/neo_player.cpp
```diff
@@ -3386,4 +3423,255 @@ const char *CNEO_Player::GetOverrideStepSound(const char *pBaseStepSound)
...
+    return GetTimeSinceWeaponFired() > sv_neo_spec_replace_player_afk_time_sec.GetInt();
+}
+
+void CNEO_Player::SpectatorTryReplacePlayer(CNEO_Player* pNeoPlayerToReplace)
+{
+	CSingleUserRecipientFilter filter(this);
+
+	if (!IsObserver() && IsAlive())
+	{
+		DevWarning("A client initiating player takeover without being in observer mode might indicate server command bugs or tampering.");
```
Rainyan: The warning line should end with a `\n` newline char so that the next console print isn't concatenated on the same line.

src/game/server/neo/neo_player.cpp
```diff
@@ -3386,4 +3423,255 @@ const char *CNEO_Player::GetOverrideStepSound(const char *pBaseStepSound)
...
+		return;
+	}
+
+	if (!pNeoPlayerToReplace->IsAlive())
+	{
+		UTIL_ClientPrintFilter(filter, HUD_PRINTCONSOLE, "Shell takeover failed: The target is dead.");
+		return;
+	}
+
+	if (pNeoPlayerToReplace->GetTeamNumber() != GetTeamNumber())
```
Rainyan: Nit: possibly reads better as `if (!InSameTeam(pNeoPlayerToReplace))`

src/game/server/neo/neo_player.cpp
```diff
@@ -3386,4 +3423,255 @@ const char *CNEO_Player::GetOverrideStepSound(const char *pBaseStepSound)
...
+	const bool bIsTargetAFK = pNeoPlayerToReplace->IsAFK();
+	const bool bAllowBotTakeover = sv_neo_spec_replace_player_bot_enable.GetBool();
+	const bool bAllowAfkTakeover = sv_neo_spec_replace_player_afk_enable.GetBool();
+
+	// If no valid condition is met, determine the specific reason and inform the user.
+	if (bIsTargetBot)
+	{
+		if (!bAllowBotTakeover)
+		{
+			UTIL_ClientPrintFilter(filter, HUD_PRINTCONSOLE, "Shell takeover failed: Taking over drones is disabled.");
```
Rainyan: Language nit: should this be "shells" instead of "drones", if we're using that terminology here?

src/game/server/neo/neo_player.cpp
```diff
@@ -3386,4 +3423,255 @@ const char *CNEO_Player::GetOverrideStepSound(const char *pBaseStepSound)
...
+	{
+		char timeBuf[4]; // assuming max sv_neo_spec_replace_player_afk_time_sec is 999
+		int  secondsLeft = sv_neo_spec_replace_player_afk_time_sec.GetInt()
+			- static_cast<int>(pNeoPlayerToReplace->GetTimeSinceWeaponFired());
+		V_snprintf(timeBuf, sizeof(timeBuf), "%d", secondsLeft);
+		UTIL_ClientPrintFilter(
+			filter,
+			HUD_PRINTCONSOLE,
+			"Shell takeover failed: Shell is not considered inactive until %s1 seconds.",
+			timeBuf);
```
Rainyan: If we're coupling the cvar max value to this alloc size here, then it'd be prudent to assert it in debug, to catch a bug where the min/max was changed without this alloc changing. For example:
```diff
        else
        {
-               char timeBuf[4]; // assuming max sv_neo_spec_replace_player_afk_time_sec is 999
+               constexpr int bufsize = 4; // assuming max sv_neo_spec_replace_player_afk_time_sec is 999
+#ifdef DEBUG
+               float flmax{};
+               bool hasMax = sv_neo_spec_replace_player_afk_time_sec.GetMax(flmax);
+               Assert(hasMax);
+               int nDigits = Ceil2Int(log10(flmax));
+               Assert(nDigits == bufsize-1);
+#endif
+               char timeBuf[bufsize];
+
                int  secondsLeft = sv_neo_spec_replace_player_afk_time_sec.GetInt()
                        - static_cast<int>(pNeoPlayerToReplace->GetTimeSinceWeaponFired());
                V_snprintf(timeBuf, sizeof(timeBuf), "%d", secondsLeft);
```

But I think preferrably we should simply use the `UTIL_VarArgs` helper here, since that way we can decouple the cvar max value from the alloc entirely, and just use the statically allocated buffer from the helper func without worrying about the buffer size:
```diff
@@ -3510,15 +3510,12 @@ void CNEO_Player::SpectatorTryReplacePlayer(CNEO_Player* pNeoPlayerToReplace)
        }
        else
        {
-               char timeBuf[4]; // assuming max sv_neo_spec_replace_player_afk_time_sec is 999
                int  secondsLeft = sv_neo_spec_replace_player_afk_time_sec.GetInt()
                        - static_cast<int>(pNeoPlayerToReplace->GetTimeSinceWeaponFired());
-               V_snprintf(timeBuf, sizeof(timeBuf), "%d", secondsLeft);
                UTIL_ClientPrintFilter(
                        filter,
                        HUD_PRINTCONSOLE,
-                       "Shell takeover failed: Shell is not considered inactive until %s1 seconds.",
-                       timeBuf);
+                       UTIL_VarArgs("Shell takeover failed: Shell is not considered inactive until %d seconds.", secondsLeft));
                return;
```

Rainyan (changes requested):
It looks like the player is able to re-choose their loadout after taking over a bot, this is probably unintended?

To repro:
* Spawn in a team with `neo_bot_quota 10`
* Wait for the round to begin
* Suicide
* Take over a bot
* Press F2 (or `loadoutmenu` console command) to open the loadout menu
* Choose some other loadout

What happens:
* You're able to choose a loadout mid-round

What should happen:
* You shouldn't be able to choose a new loadout mid-round

src/game/server/neo/neo_player.cpp
sunzenshen: I was trying to be cute about this case specifically being a bot, but I can see the lack of clarity here.  Maybe "bot" still works in this context?

sunzenshen:
> It looks like the player is able to re-choose their loadout after taking over a bot, this is probably unintended?

After [another update](https://github.com/NeotokyoRebuild/neo/pull/1294/commits/ed1730f3591c446a8ee4b6f26be9716bb1f40b77), here is video of testing that a takeover player cannot swap loadouts mid-round:

https://github.com/user-attachments/assets/09a112f6-c8b0-440b-9357-33f9787dea36

Incidentally, after testing ghost caps, and reviewing older video, came up with a fix for the originally deceased player taking over another bot/player, not getting a single point when their team ghost captures:

https://github.com/user-attachments/assets/90ea7381-6bae-48a8-a07a-61acce1d1ce0

src/game/server/neo/neo_player.h
```diff
@@ -210,7 +210,8 @@ class CNEO_Player : public CHL2MP_Player
 	AttackersTotals GetAttackersTotals() const;
 	void StartShowDmgStats(const CTakeDamageInfo *info);
 
-	void AddPoints(int score, bool bAllowNegativeScore);
+	void AddPoints(int score, bool bAllowNegativeScore) { AddPoints(score, bAllowNegativeScore, false); }
+	void AddPoints(int score, bool bAllowNegativeScore, bool bIgnorePlayerTakeover);
```
Rainyan: Nit: this could simplify to one definition with a default boolean value:
```diff
-       void AddPoints(int score, bool bAllowNegativeScore) { AddPoints(score, bAllowNegativeScore, false); }
-       void AddPoints(int score, bool bAllowNegativeScore, bool bIgnorePlayerTakeover);
+       void AddPoints(int score, bool bAllowNegativeScore, bool bIgnorePlayerTakeover=false);
```

src/game/server/neo/neo_player.cpp
```diff
@@ -3386,4 +3423,253 @@ const char *CNEO_Player::GetOverrideStepSound(const char *pBaseStepSound)
...
+			// After giving weapons, ensure the correct active weapon is set.
+			if (pTakeoverTargetActiveWeapon)
+			{
+				CBaseCombatWeapon* pPlayerActiveWeapon = Weapon_OwnsThisType(pTakeoverTargetActiveWeapon->GetClassname());
+				if (pPlayerActiveWeapon)
+				{
+					Weapon_Switch(pPlayerActiveWeapon);
+				}
+			}
+			m_bIneligibleForLoadoutPick = true; // Prevent loadout change after takeover.
```
Rainyan: I know I was requesting this change before, but after thinking about this some more, maybe we would want to do something like `m_bIneligibleForLoadoutPick = pPlayerTakeoverTarget->m_bIneligibleForLoadoutPick;` after all?

Since the case might be that the possessee hasn't yet made the loadout option ineligible (eg. they are an AFK player, or they are a bot that hasn't yet shot/dropped their gun).

So in these types of cases, it might be reasonable for the possessor to be allowed to re-choose the loadout, just as real players are allowed to do after spawning in the world, as long as `m_bIneligibleForLoadoutPick` doesn't forbid it. If you think this makes sense?

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -3895,6 +3918,13 @@ bool CNEORules::GetTeamPlayEnabled() const
 #ifdef GAME_DLL
 bool CNEORules::FPlayerCanRespawn(CBasePlayer* pPlayer)
 {
+	// Special case for spectator player takeover
+	CNEO_Player* pNeoPlayer = ToNEOPlayer(pPlayer);
+	if (pNeoPlayer && pNeoPlayer->GetSpectatorTakeoverPlayerPending())
```
Rainyan: Nit: we're shadowing this `pNeoPlayer` variable with a redeclaration at line 3962, so maybe remove that redundant bottom declaration in favor of this one here. It doesn't really have an effect on the code but is maybe clearer without the extra variable.

Also I believe we're already assuming that `pPlayer` must be a valid player ptr here, so the `if pNeoPlayer` nullptr check is probably redundant. Or it could at least be demoted to an `Assert` if you wanna keep the check, so that we only do it for debug builds and not for release.

src/game/server/neo/neo_player.cpp
```diff
@@ -3386,4 +3425,251 @@ const char *CNEO_Player::GetOverrideStepSound(const char *pBaseStepSound)
...
+	if (m_hSpectatorTakeoverPlayerTarget.Get())
+	{
+		CNEO_Player* pPlayerTakenOver = m_hSpectatorTakeoverPlayerTarget.Get();
+		m_hSpectatorTakeoverPlayerTarget->m_hSpectatorTakeoverPlayerImpersonatingMe = nullptr;
+		m_hSpectatorTakeoverPlayerTarget = nullptr;
+
+		// Determine if we should kick taken over player to spectator team
+		if ( pPlayerTakenOver && !pPlayerTakenOver->IsBot()  // player is human
+			&& !pPlayerTakenOver->IsAlive() // player has not respawned by takeover
+			&& pPlayerTakenOver->IsAFK() ) // player is still AFK
```
sunzenshen, line 3692: These new checks were to handle the case where an AFK player notices they are no longer alive, so they take over another bot/player. If the player that was taking them over dies/disconnects, the recovered AFK player would get kicked to spectator team, like in this bug reproduction video:

https://github.com/user-attachments/assets/01fd6afa-f1dc-423e-ab4d-98cda6f6063f

src/game/server/neo/neo_player.cpp
sunzenshen: I tried this in the new commit, and this seems to work okay.  The one quirk of allowing loadout changes at this point is that the loadout selection screen will actually be at the rank the takeover player is.  So if you take over a bot that is a lower rank than you, you can select weapons at your rank level, even when the bot wouldn't have that selection.  This also happens in reverse, if you take over another bot/player as the lowest rank, you can swap to an MPN but nothing else.

sunzenshen:
Video of fixed PreThink takeover logic for previously overridden AFK players:

https://github.com/user-attachments/assets/ae6beab6-1d90-4869-a52d-413bd951fe19

src/game/server/neo/neo_player.cpp
Rainyan: One abuse scenario I can come up with is, let's say you're a player with 25 XP and there's 3 bots in the team. The player could suicide, then take over each of the bots in turn, spawn as them, choose a lieut weapon, then suicide and pick another bot. So this way they could spawn several restricted weapons per life (3 extra SRS/AA13 in this example) for their human players to pick up. I don't know if this is actually abusable in real game, but maybe something worth thinking about. Probably not a dealbreaker for merging this PR, but it might be safer if we can restrict the weapons to the possessed player's XP, unless it's a huge code rewrite.

src/game/server/neo/neo_player.cpp
```diff
@@ -3386,4 +3425,253 @@ const char *CNEO_Player::GetOverrideStepSound(const char *pBaseStepSound)
...
+				}
+			}
+
+			// TODO: Limit loadout options based on takeover target's XP rank.
+			// Currently, loadout options are based on former spectator's XP rank.
+			// To evaluate loadout options on takeover target's XP rank, need to send networked takeover state to client.
+			// In an active game, bots will likely already be ineligible for loadout pick, by the time an average takeover happens.
+			m_bIneligibleForLoadoutPick = sv_neo_spec_replace_player_loadout_enable.GetBool() ?
+				pPlayerTakeoverTarget->m_bIneligibleForLoadoutPick
+				: true;
```
sunzenshen, line 3645: I decided to make loadout changes after spectator takeover configurable for reasons I will comment in the related discussion.

src/game/server/neo/neo_player.cpp
sunzenshen: I decided to add another convar to choose whether players can change loadout after spectator takeover with the current limited implementation, with the default that they cannot.

The main reason is that in order to get the loadout menu to properly update, I would need to send over more network variables to provide client context of whether the c_neo_player is taking over another player at a given time. (If anything, I've tried to limit as much logic as I can get away with on the server side, with the client side blood decals being the most notable exception.)  I don't think adding the extra state tracking over the network is worth it for the following reason:

In a game where players are actually trying to compete and not just standing around testing things, by the time a player is killed, becomes a spectator, and chooses a bot to replace, most bots have already fired their weapon or are so far away from spawn that a loadout change is questionable, assuming the spectator was not trying to off themselves as early as possible.  In the case of an AFK player, ideally that situation is rare.

I set the convar default to disabled, to prioritize fairness, but honestly even in your cheating example, it will be tough to replace enough bots fast enough as they run to their doom.  And if someone was really AFK, that team is already going to be at enough of a disadvantage that having an LT player take over is not really an overwhelming advantage.

The one thing that could change that dynamic is whether in the future we add support for commanding bots to follow you, but if the bots haven't engaged anyone with their weapon by the time a takeover happens, that team is playing way too passively.

Also if later we go with the approach of transferring client commands to bot mimicry, this kind of option wouldn't be available anyway.

src/game/server/neo/neo_player.cpp
```diff
@@ -3386,4 +3425,269 @@ const char *CNEO_Player::GetOverrideStepSound(const char *pBaseStepSound)
...
+		// For new round reset where all players will be reset
+		// For a player that disconnects, so they can no longer receive points
+		// When a possessEE changes teams, it doesn't make sense to penalize possessOR for doing well
+
+		// Example for when not to enter here:
+		// Player A is AFK, while on same team as Player B and Bot C
+		// Player B is a spectator that takes over Player A
+		// Player A returns to keyboard during round and takes over Bot C
+		// Player A dies: should reset target Bot C, but not impersonator Player B
+		// because Player B is still scoring points for Player A, despite Player A being dead again
```
sunzenshen, line 3712: Realized that last night I accidentally broke this multi-possessor edge case, but the comments above also explain why I added this additional reset section.

src/game/server/neo/neo_player.cpp
```diff
@@ -3386,4 +3425,269 @@ const char *CNEO_Player::GetOverrideStepSound(const char *pBaseStepSound)
...
+	}
+	m_hSpectatorTakeoverPlayerTarget = nullptr;
+
+	// Hard Reset Criteria: Does it make sense to route points to this player anymore?
+	if (bHardReset)
+	{
+		// Examples for when to hard reset:
+		// For new round reset where all players will be reset
+		// For a player that disconnects, so they can no longer receive points
+		// When a possessEE changes teams, it doesn't make sense to penalize possessOR for doing well
```
sunzenshen, line 3705: This one case where a possessee changes teams might be controversial, since a spectator taking over will benefit from a hard reset if the player they are taking over swaps sides.

I at this time think that a hard reset is warranted in this case, because the mechanics of routing takeover points to a teammate that joined either spectator or the other side is odd, and I don't think it makes sense to incentivize a takeover player to suddenly stop trying to win because their awards would go to a now enemy.

Rainyan (approved):
LGTM, tested again with 2 Steam clients (1 Windows, 1 Linux). Sorry about the delay with the review.

sunzenshen:
Done resolving the merge conflict, it was just an additional newline at the neo_player.cpp that caused a collision.  Squashed to simplify the rebase, rebuilt, and sanity checked that the feature still works on a single machine.

src/game/server/neo/neo_player.cpp
```diff
@@ -3406,3 +3445,267 @@ const char *CNEO_Player::GetOverrideStepSound(const char *pBaseStepSound)
...
+ConVar sv_neo_spec_replace_player_min_exp("sv_neo_spec_replace_player_min_exp",
+	"0", FCVAR_NONE,
+	"Minimum experience allowed to takeover players ",
+	true, -999, true, 999);
+
+bool CNEO_Player::IsAFK() const
+{
+    // NEO JANK GetTimeSinceLastUserCommand seems to return 0 as long as the player is connected, so use an alternative timer
+    return GetTimeSinceWeaponFired() > sv_neo_spec_replace_player_afk_time_sec.GetInt();
+}
```
Rainyan, line 3475: Might be preferrable to return `GetIdleTime` (unsure whether its implementation works for us currently though...) instead of time last shot - because there's cases where a player could well be active in the round but just not shooting, for example for rounds where the ghost spawns close to them, or if they're just playing sneaky. Especially for lower afk counter amounts, like 30 seconds, etc, where it's quite plausible the player just might not have shot in that amount of time.

src/game/server/neo/neo_player.cpp
```diff
@@ -3406,3 +3445,267 @@ const char *CNEO_Player::GetOverrideStepSound(const char *pBaseStepSound)
...
+		if (!bAllowAfkTakeover)
+		{
+			UTIL_ClientPrintFilter(filter, HUD_PRINTCONSOLE, "Shell takeover failed: Taking over inactive shells is disabled.");
+			return;
+		}
+	}
+	else
+	{
+		int  secondsLeft = sv_neo_spec_replace_player_afk_time_sec.GetInt()
+			- static_cast<int>(pNeoPlayerToReplace->GetTimeSinceWeaponFired());
```
Rainyan: Also it's a bit questionable to manually calculate from `GetTimeSinceWeaponFired` here - would prefer if we instead called `IsAFK` here, and that returned the time for us, so that we would be coupled to the `IsAFK` logic.

src/game/server/neo/neo_player.cpp
Rainyan: Actually nevermind, seems like the time counter of `GetIdleTime` is capped at 4 seconds for whatever it's used for. Well, maybe this `GetTimeSinceWeaponFired` business is good enough of an approximation in that case, for now. Perhaps we should roll our own buttons-pressed timer eventually for this (unless one already exists)?

sunzenshen: Yeah, I looked at all of the available timers, and a few of them did not work intuitively from what I would have expected.

sunzenshen: In the Discord, @AdamTadeusz had an interesting idea of parsing the client commands or checking the size to see if there were any interesting commands to reset an AFK timer.  That might be one idea to try in another PR, but for now probably safer to disable AFK takeovers by default until we can figure out a clean and intuitive solution.

src/game/server/neo/neo_player.cpp
sunzenshen: Good idea, decided to last minute refactor this.

Rainyan (dismissed):

src/game/server/neo/neo_player.cpp
```diff
@@ -1842,8 +1866,19 @@ void CNEO_Player::StartShowDmgStats(const CTakeDamageInfo *info)
...
 }
 
-void CNEO_Player::AddPoints(int score, bool bAllowNegativeScore)
+void CNEO_Player::AddPoints(int score, bool bAllowNegativeScore, bool bIgnorePlayerTakeover)
 {
+	if (!bIgnorePlayerTakeover && m_hSpectatorTakeoverPlayerTarget.Get())
+	{
+		if (score >= 0)
+		{
+			// Don't penalize possessed/bot for takeover player's actions
```
AdamTadeusz: would it be more clear if this comment said "reward possessed/bot for takeover player's actions"

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -1180,6 +1180,44 @@ void CNEORules::Think(void)
...
+				{
+					auto* neoPlayer = static_cast<CNEO_Player*>(player);
+
+
+					// Player takeover logic
+					auto playerControllingMe = neoPlayer->m_hSpectatorTakeoverPlayerImpersonatingMe.Get();
+					if (playerControllingMe)
+					{
+						// Will get XP from player possessing me based on their win conditions
+						return;
```
AdamTadeusz: should this be "continue" ?

AdamTadeusz (approved):

PR #1295, opened 2025-08-21
zero out ragdoll velocity if ragdoll is settled, refactor ragdoll velocity stuff out of c_baseanimating into cragdoll

AdamTadeusz:
## Description

There is an issue on master thats kinda difficult to replicate, but sometimes you will notice some ragdolls will freeze while still being slightly highlighted in motion vision, with the velocity no longer updated as the ragdoll goes to sleep. This PR sets the value of the ragdoll velocity to 0 when a ragdoll goes to sleep

## Toolchain

- Windows MSVC VS2022

### DIFF
--- a/src/game/client/ragdoll.cpp
+++ b/src/game/client/ragdoll.cpp
@@ -265,6 +265,9 @@ bool CRagdoll::TransformVectorToWorld(int iBoneIndex, const Vector *vPosition, V
 //-----------------------------------------------------------------------------
 void CRagdoll::PhysForceRagdollToSleep()
 {
+#ifdef NEO
+	m_vecLastVelocity = vec3_origin;
+#endif // NEO
 	for ( int i = 0; i < m_ragdoll.listCount; i++ )
 	{
 		if ( m_ragdoll.list[i].pObject )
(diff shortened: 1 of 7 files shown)


src/game/client/ragdoll.cpp
```diff
@@ -273,6 +284,9 @@ static ConVar ragdoll_sleepaftertime( "ragdoll_sleepaftertime", "5.0f", 0, "Afte
 void CRagdoll::CheckSettleStationaryRagdoll()
 {
 	Vector delta = GetRagdollOrigin() - m_vecLastOrigin;
+#ifdef NEO
+	m_vecLastVelocity = delta / (gpGlobals->frametime);
```
sunzenshen: sanity check: frametime will never ever be zero in normal circumstances, right?

sunzenshen: only other feedback I could think of at the moment is maybe add a debug only sanity assert on frametime, and maybe m_pRagdoll earlier.

AdamTadeusz: looking at how frametime is used elsewhere in the code this is absolutely a valid point

Rainyan (approved):

PR #1296, opened 2025-08-22
Weapons drop refactor

AdamTadeusz:
## Description

Hopefully fixes the problem with spectators having guns. Copies CBaseCombatCharacter::Weapon_DropAll 's mechanic of dropping the active weapon last, also removes any weapons that cannot be dropped (currently only the knife, actually can be dropped in the original, perhaps should revert this change?)

## Toolchain

- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/neo_player.cpp
+++ b/src/game/server/neo/neo_player.cpp
@@ -1925,18 +1925,14 @@ void CNEO_Player::Weapon_DropOnDeath(CNEOBaseCombatWeapon* pNeoWeapon, Vector da
 
 	Vector playerVelocity = vec3_origin;
 	GetVelocity(&playerVelocity, nullptr);
-	constexpr float DAMAGE_FORCE_SCALE = 0.0005f;
-	if (VPhysicsGetObject())
-	{
-		playerVelocity += (damageForce * DAMAGE_FORCE_SCALE) / VPhysicsGetObject()->GetInvMass();
-	}
-	else {
-		constexpr float SRM_WEAPON_MASS = 0.012f;
-		playerVelocity += (damageForce * DAMAGE_FORCE_SCALE) / SRM_WEAPON_MASS;
-	}
-	
 	pNeoWeapon->Drop(playerVelocity);
 	Weapon_Detach(pNeoWeapon);
+
+	if (pNeoWeapon->VPhysicsGetObject())
+	{
+		constexpr float DAMAGE_FORCE_SCALE = 0.1f;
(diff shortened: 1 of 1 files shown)


src/game/server/neo/neo_player.h
```diff
@@ -74,7 +74,7 @@ class CNEO_Player : public CHL2MP_Player
 	virtual void PickupObject(CBaseEntity *pObject, bool bLimitMassAndSize) OVERRIDE;
 	virtual void PlayStepSound(Vector &vecOrigin, surfacedata_t *psurface, float fvol, bool force) OVERRIDE;
 	virtual void Weapon_Drop(CBaseCombatWeapon *pWeapon, const Vector *pvecTarget = NULL, const Vector *pVelocity = NULL) OVERRIDE;
-	virtual void Weapon_DropOnDeath(CBaseCombatWeapon *pWeapon, Vector pVelocity, CBaseEntity *pAttacker = NULL);
+	virtual void Weapon_DropOnDeath(CNEOBaseCombatWeapon *pWeapon, Vector damageForce);
```
sunzenshen: Is this function actually overriding a base virtual interface ?  Seems like the contract is changing with these adjustments.  I searched the code base, and I didn't see other hits for this function name.  Perhaps this doesn't need to be virtual since no one else is inheriting from CNEO_Player?

AdamTadeusz: 👍

src/game/server/neo/neo_player.cpp
```diff
@@ -1892,56 +1904,39 @@ void CNEO_Player::Event_Killed( const CTakeDamageInfo &info )
...
 	Vector playerVelocity = vec3_origin;
-	GetVelocity(&playerVelocity, NULL);
-	
+	GetVelocity(&playerVelocity, nullptr);
+	constexpr float DAMAGE_FORCE_SCALE = 0.0005f;
 	if (VPhysicsGetObject())
 	{
-		playerVelocity += velocity * VPhysicsGetObject()->GetInvMass();
+		playerVelocity += (damageForce * DAMAGE_FORCE_SCALE) / VPhysicsGetObject()->GetInvMass();
 	}
```
DESTROYGIRL: Do we want `pNeoWeapon->VPhysicsGetObject()` here instead? Because in the else we divide by a default weapon mass not a default player mass

DESTROYGIRL (approved):

nullsystem (approved):

PR #1311, opened 2025-08-25
Show loading screen if unsure what map we're loading into

AdamTadeusz:
## Description

When creating a listen server, the host_map convar gets updated almost immediately with the name of the map we're loading into, if connecting to a dedicated server however this only happens when we're almost finished loading. 

Since the whole reason we're doing this is to stop the loading screen from showing up when loading into a background map, and we should only really be loading into a background map on a listen server, we can assume that if host_map doesn't return a value, that we are connecting to a dedicated server at which point we should be showing a loading screen. In reality we still see the loading screen for a couple frames but that's much better than the current behavior.

As a side note, engine->GetLevelName() will return the map on a dedicated server when connecting quicker than host_map, but it's still quite far into the loading process.

Under the notes section https://developer.valvesoftware.com/wiki/Custom_loading_screen there is mention that some mods were able to obtain this value almost instantly, presumably when connecting to a dedicated server, but the author explains that this is probably impossible without engine access and I'm not sure what mods exactly were able to do this. The linked github repo also uses host_map for near instantaneous access to the map name when connecting to a listen server.

This only applies to using the "connect" command from the console. When looking at dedicated servers in the server browser we can see the map that was loaded when the server list was refreshed, so keeping track of what map we're most likely loading into should be easy if we ever want to extend the loading screen functionality to change the background to an image appropriate to the map we're loading into regardless of whether we're connecting to a listen server or dedicated server.

### DIFF
--- a/src/game/client/neo/ui/neo_loading.cpp
+++ b/src/game/client/neo/ui/neo_loading.cpp
@@ -187,7 +187,7 @@ void CNeoLoading::OnMainLoop(const NeoUI::Mode eMode)
 
 	static bool bStaticInitNeoUI = false;
 	bool bSkipRender = false;
-	if (iStrIdx == m_aStrIdxMap[LOADINGSTATE_LOADING])
+	if (iStrIdx == m_aStrIdxMap[LOADINGSTATE_LOADING] && m_pHostMap)
 	{
 		auto hostMapName = m_pHostMap->GetString();
 		if (Q_stristr(hostMapName, "background_"))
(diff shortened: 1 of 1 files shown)


AdamTadeusz:
I mean we can make this be the case. By default you just see the blurred image on disconnect

DESTROYGIRL:
I imagined it was an intentional thing, cuz people who want to switch between maps/servers quickly might not like the extended load time/freeze, but it's whatever really. Anyway I think this solution is fine for now

sunzenshen (approved):

PR #1324, opened 2025-08-28
Use class name for killer info, unify name for all neo_npc_targetsystem

DESTROYGIRL:
## Description
`GetClassname` instead of `GetDebugName` as mappers can set the targetname to pretty much anything

## Toolchain
- Windows MSVC VS2022


sunzenshen (approved):
Not sure which map has a generic turret, but the new name for the tank looks good to me:
image

Deaths by players still looks okay too.
image

PR #1326, opened 2025-08-28
Hide squad list and killfeed if the scoreboard is showing

DESTROYGIRL:
## Description
This is to prevent readability issues on the scoreboard.
Much of the information provided by these two HUD elements is on the scoreboard, and the scoreboard can be quickly turned on/off as it has to be held open

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/client/neo/ui/neo_hud_deathnotice.cpp
+++ b/src/game/client/neo/ui/neo_hud_deathnotice.cpp
@@ -23,6 +23,7 @@
 #include "spectatorgui.h"
 #include "takedamageinfo.h"
 #include "c_neo_killer_damage_infos.h"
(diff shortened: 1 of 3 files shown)


sunzenshen (dismissed):
Depending on the resolution, maybe this hiding of UI elements is not needed (e.g. 4K), but at least looks to be working:


https://github.com/user-attachments/assets/46125480-9ae2-4fbf-aadf-c56206f8b69c

sunzenshen:
> Would you prefer a toggle be added for this behaviour?

Probably a good idea, with the default being enabled.

Rainyan (approved):

Rainyan (approved):

PR #1333, opened 2025-09-01
Option to make the round info and scoreboard HUD not swap team sides

DESTROYGIRL:
## Description
Playtesters expressed confusion regarding the teams icon/score swapping sides, as currently the local player's team is always left.
This can cause it to become misaligned with the scoreboard and not appear where the player expected their team's score to be

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/client/neo/ui/neo_hud_round_state.cpp
+++ b/src/game/client/neo/ui/neo_hud_round_state.cpp
@@ -30,7 +30,7 @@ DECLARE_NAMED_HUDELEMENT(CNEOHud_RoundState, NRoundState);
 
 NEO_HUD_ELEMENT_DECLARE_FREQ_CVAR(RoundState, 0.1)
 
-ConVar cl_neo_round_hud_swap_sides("cl_neo_round_hud_swap_sides", "0", FCVAR_ARCHIVE, "Make the team of the local player always appear on the left side of the round info", true, 0.0, true, 1.0);
+ConVar cl_neo_round_hud_swap_sides("cl_neo_round_hud_swap_sides", "1", FCVAR_ARCHIVE, "Make the team of the local player always appear on the left side of the round info", true, 0.0, true, 1.0);
 ConVar cl_neo_squad_hud_original("cl_neo_squad_hud_original", "1", FCVAR_ARCHIVE, "Use the old squad HUD", true, 0.0, true, 1.0);
 ConVar cl_neo_squad_hud_star_scale("cl_neo_squad_hud_star_scale", "0", FCVAR_ARCHIVE, "Scaling to apply from 1080p, 0 disables scaling");
 extern ConVar sv_neo_dm_win_xp;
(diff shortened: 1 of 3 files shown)


DESTROYGIRL:
Is a menu option justified for this, if so how to describe it

sunzenshen (approved):

Agiel:
Worth noting that the "X vs Y" text in the og comp plugin also has your team always on the left.

DESTROYGIRL:
> I think the current behavior makes more sense, I would rather instead switch the teams in the scoreboard to align with the top element.

Sure, we can switch the default to enabled if you plan on implementing that but I think we should still keep this as an option regardless

src/game/client/neo/ui/neo_hud_round_state.cpp
```diff
@@ -30,6 +30,12 @@ DECLARE_NAMED_HUDELEMENT(CNEOHud_RoundState, NRoundState);
 
 NEO_HUD_ELEMENT_DECLARE_FREQ_CVAR(RoundState, 0.1)
 
+static void teamSwapSidesCallback(IConVar* var, const char* pOldValue, float flOldValue)
+{
+	g_pNeoScoreBoard->UpdateTeamColumnsPosition(GetLocalPlayerTeam());
```
AdamTadeusz: elsewhere where g_pNeoScoreBoard is used we check whether its a nullptr or not but that's probably unnecessary, even changing the value of cl_neo_hud_team_swap_sides in the autoconfig doesn't cause any issues. It might be possible that the callback is called if the value is changed while the client is being shut down?

src/game/client/neo/ui/neo_hud_round_state.cpp
```diff
@@ -30,6 +30,12 @@ DECLARE_NAMED_HUDELEMENT(CNEOHud_RoundState, NRoundState);
...
+}
+
+ConVar cl_neo_hud_team_swap_sides("cl_neo_hud_team_swap_sides", "1", FCVAR_ARCHIVE, "Make the team of the local player always appear on the left side of the round info and scoreboard", true, 0.0, true, 1.0, teamSwapSidesCallback);
```
AdamTadeusz: since teamSwapSidesCallback is only used in one place, you could use a lambda function here, something like

```
ConVar cl_neo_hud_team_swap_sides("cl_neo_hud_team_swap_sides", "1", FCVAR_ARCHIVE, "Make the team of the local player always appear on the left side of the round info and scoreboard", true, 0.0, true, 1.0,
	[]([[maybe_unused]] IConVar* var, [[maybe_unused]] const char* pOldValue, [[maybe_unused]] float flOldValue){
		g_pNeoScoreBoard->UpdateTeamColumnsPosition(GetLocalPlayerTeam());
});
```

src/game/client/neo/ui/neo_hud_round_state.cpp
DESTROYGIRL: I didn't perform this check in #1326 and haven't had any issues since its been merged but could do

sunzenshen (approved):
lgtm:

Image

Image

`cl_neo_hud_team_swap_sides 0`

Image

Image

PR #1108, opened 2025-04-28
Ghost overtime

AdamTadeusz:
## Description

Ghost overtime

## Toolchain

- Windows MSVC VS2022
- Linux GCC Distro Native [Specify distro + GCC version]
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #1018

PR #1342, opened 2025-09-11
CTG & JGR Overtime

### LINKED ISSUE #1018: Ghost overtime is implemented natively
title

DESTROYGIRL:
## Description
Continuation of #1108 by @AdamTadeusz which has been slimmed down.

In CTG:
- Overtime is displayed on the round timer. The round timer will always be the thing showing you how much time is left
- The grace duration is static, when the ghost is dropped is will always start counting down from the set grace duration UNLESS overtime is less than the grace duration, in which case it will just stick to the overtime left (to prevent any extension of overtime)
- Round timer turns red when in grace / overtime is less than set grace duration
- Grace decay remains as an option for testing with and without

In JGR:
- If the Juggernaut is being controlled by the losing team when the timer hits zero, they can continue to score points until the JGR dies, or they gain more points than the opposing team

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1018 
- related #1108

### DIFF
--- a/src/game/shared/neo/neo_gamerules.cpp
+++ b/src/game/shared/neo/neo_gamerules.cpp
@@ -232,7 +232,6 @@ BEGIN_NETWORK_TABLE_NOBASE( CNEORules, DT_NEORules )
 	RecvPropInt(RECVINFO(m_iEscortingTeam)),
 	RecvPropBool(RECVINFO(m_bGhostExists)),
 	RecvPropFloat(RECVINFO(m_flGhostLastHeld)),
-	RecvPropFloat(RECVINFO(m_flAccumulatedOvertime)),
 	RecvPropVector(RECVINFO(m_vecGhostMarkerPos)),
 	RecvPropInt(RECVINFO(m_iJuggernautPlayerIndex)),
 	RecvPropBool(RECVINFO(m_bJuggernautItemExists)),
@@ -256,7 +255,6 @@ BEGIN_NETWORK_TABLE_NOBASE( CNEORules, DT_NEORules )
 	SendPropInt(SENDINFO(m_iEscortingTeam)),
 	SendPropBool(SENDINFO(m_bGhostExists)),
 	SendPropFloat(SENDINFO(m_flGhostLastHeld)),
-	SendPropFloat(SENDINFO(m_flAccumulatedOvertime)),
 	SendPropVector(SENDINFO(m_vecGhostMarkerPos), -1, SPROP_COORD_MP_LOWPRECISION | SPROP_CHANGES_OFTEN, MIN_COORD_FLOAT, MAX_COORD_FLOAT),
 	SendPropInt(SENDINFO(m_iJuggernautPlayerIndex)),
 	SendPropBool(SENDINFO(m_bJuggernautItemExists)),
@@ -686,7 +684,6 @@ void CNEORules::ResetMapSessionCommon()
 	m_flNeoRoundStartTime = 0.0f;
 	m_flNeoNextRoundStartTime = 0.0f;
 	m_flGhostLastHeld = 0.0f;
-	m_flAccumulatedOvertime = 0.0f;
 #ifdef GAME_DLL
 	m_pRestoredInfos.Purge();
 	m_readyAccIDs.Purge();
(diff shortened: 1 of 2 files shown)


src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -1173,6 +1212,47 @@ void CNEORules::Think(void)
...
+			{
+				if (GetGlobalTeam(TEAM_JINRAI)->GetScore() > GetGlobalTeam(TEAM_NSF)->GetScore())
+				{
+					SetWinningTeam(TEAM_JINRAI, NEO_VICTORY_POINTS, false, true, false, false);
+					return;
+				}
+
+				if (GetGlobalTeam(TEAM_NSF)->GetScore() > GetGlobalTeam(TEAM_JINRAI)->GetScore())
+				{
+					SetWinningTeam(TEAM_NSF, NEO_VICTORY_POINTS, false, true, false, false);
```
sunzenshen, line 1210: Should we sanity check that the team scores are not exactly equal, for the NEO_VICTORY_STALEMATE case?

DESTROYGIRL: Not sure I understand, if the juggernaut dies when the points are equal it should stalemate as per line 1259?

sunzenshen: Doh, didn't see that line, it's literally outside of window showed to me on GitHub.

sunzenshen (approved):

src/game/server/neo/neo_player.cpp
```diff
@@ -2788,7 +2788,7 @@ int	CNEO_Player::OnTakeDamage_Alive(const CTakeDamageInfo& info)
 				m_rfAttackersAccumlator.Set(attackerIdx, flDmgAccumlator);
 				m_rfAttackersHits.GetForModify(attackerIdx) += 1;
 
-				if (bIsTeamDmg && sv_neo_teamdamage_kick.GetBool() && NEORules()->GetRoundStatus() == NeoRoundStatus::RoundLive)
+				if (bIsTeamDmg && sv_neo_teamdamage_kick.GetBool() && NEORules()->GetRoundStatus() == NeoRoundStatus::RoundLive || NEORules()->GetRoundStatus() == NeoRoundStatus::Overtime)
```
Agiel: Missing parentheses around the round status check.

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -1055,6 +1066,34 @@ void CNEORules::Think(void)
...
+			}
+		}
+	}
+	else
+	{
+		if (m_nGameTypeSelected == NEO_GAME_TYPE_CTG && m_nRoundStatus == NeoRoundStatus::Overtime && m_iGhosterPlayer)
+		{
+			m_flGhostLastHeld = gpGlobals->curtime;
+		}
+	}
```
Agiel: This part might be more readable if the entire thing is wrapped in a `if (m_nGameTypeSelected == NEO_GAME_TYPE_CTG)`.

In general this file could probably be cleaned up with some switch statements on the game type, but that's outside the scope of this issue obviously.

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -1620,6 +1702,36 @@ float CNEORules::GetRoundRemainingTime() const
...
+
+	if (sv_neo_ctg_ghost_overtime_grace_decay.GetBool())
+	{
+		if (m_iGhosterPlayer)
+		{
+			return overtime;
+		}
+		else
+		{
+			return (sv_neo_ctg_ghost_overtime.GetFloat() + sv_neo_ctg_ghost_overtime_grace.GetFloat() - m_flAccumulatedOvertime) / ((sv_neo_ctg_ghost_overtime.GetFloat() + sv_neo_ctg_ghost_overtime_grace.GetFloat()) / sv_neo_ctg_ghost_overtime_grace.GetFloat());
```
Agiel: This could be something like
```
float overtimeAtGhostDrop = (m_flNeoRoundStartTime + roundTimeLimit + sv_neo_ctg_ghost_overtime.GetFloat()) - m_flGhostLastHeld;
return overtimeAtGhostDrop / sv_neo_ctg_ghost_overtime.GetFloat() * sv_neo_ctg_ghost_overtime_grace.GetFloat() - (gpGlobals->curtime - m_flGhostLastHeld);
```
which should let you get rid of `m_flAccumulatedOvertime`.

Agiel (approved):

sunzenshen (approved):

PR #1345, opened 2025-09-14
Bots follow you when you press +use on them

sunzenshen:
## Description
Simple snake formation while copying thermoptic and crouch actions. If more than one bot is following, they will choose a neighbor to follow to reduce collisions.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
@@ -81,7 +81,66 @@ ActionResult< CNEOBot >	CNEOBotSeekAndDestroy::Update( CNEOBot *me, float interv
 	else
 	{
 		// Out of combat
-		me->DisableCloak();
+		if ( me->m_hLeadingPlayer.Get() )
+		{
+			RecomputeSeekPath( me );
+
+			CUtlVector<CHandle<CNEO_Player>> visitedPlayers;
+			CNEO_Player *pPlayerToMirror = me->m_hLeadingPlayer.Get();
+			bool bShouldCloak = false;
+			bool bShouldCrouch = false;
+
+			while ( pPlayerToMirror )
+			{
+				// Cycle breaker
+				if ( visitedPlayers.Find( pPlayerToMirror ) != visitedPlayers.InvalidIndex() )
+				{
+					break;
+				}
+				visitedPlayers.AddToTail( pPlayerToMirror );
+
+				if ( !bShouldCloak && pPlayerToMirror->GetInThermOpticCamo() )
+				{
+					bShouldCloak = true;
+				}
+
+				if ( !bShouldCrouch && pPlayerToMirror->IsDucking() )
+				{
+					bShouldCrouch = true;
+				}
+
+				if ( bShouldCloak && bShouldCrouch )
+				{
+					break;
+				}
+
+				// Move to the next leader in the chain
+				pPlayerToMirror = pPlayerToMirror->m_hLeadingPlayer.Get();
+			}
+
+			if ( bShouldCloak )
+			{
+				me->EnableCloak( 4.0f );
+			}
+			else
+			{
+				me->DisableCloak();
+			}
+
+			if ( bShouldCrouch )
+			{
+				me->PressCrouchButton(0.5f);
+			}
+			else
+			{
+				me->ReleaseCrouchButton();
+			}
+		}
+		else
+		{
+			me->DisableCloak();
+			me->ReleaseCrouchButton();
+		}
 
 		// Reload when safe
 		CNEOBaseCombatWeapon* myWeapon = static_cast<CNEOBaseCombatWeapon*>(me->GetActiveWeapon());
@@ -317,12 +376,58 @@ private:
 };
 
 
+static ConVar sv_neo_bot_follow_stop_distance_sq("sv_neo_bot_follow_stop_distance_sq", "5000");
 //---------------------------------------------------------------------------------------------
 void CNEOBotSeekAndDestroy::RecomputeSeekPath( CNEOBot *me )
 {
 	if ( m_bOverrideApproach )
 	{
 		return;
+	}
+
+	// Check if following a player
+	if (me->m_hLeadingPlayer.Get())
+	{
+		CNEO_Player* pCommander = me->m_hLeadingPlayer.Get();
+		if (pCommander && pCommander->IsAlive())
+		{
+			if (me->GetAbsOrigin().DistToSqr(pCommander->GetAbsOrigin()) < sv_neo_bot_follow_stop_distance_sq.GetFloat())
+			{
+				// Anti-collision: follow neighbor in snake chain
+				for (int idx = 1; idx <= gpGlobals->maxClients; ++idx)
+				{
+					CNEO_Player* pOther = static_cast<CNEO_Player*>(UTIL_PlayerByIndex(idx));
+					if (!pOther || !pOther->IsBot() || pOther == me || pOther->m_hLeadingPlayer != me->m_hLeadingPlayer)
+						continue;
+
+					if (me->GetAbsOrigin().DistToSqr(pOther->GetAbsOrigin()) < sv_neo_bot_follow_stop_distance_sq.GetFloat() / 4)
+					{
+						me->PressBackwardButton();
+						// Link up to snake chain of followers
+						me->m_hLeadingPlayer = pOther;
+					}
+				}
+
+				m_hTargetEntity = NULL;
+				m_bGoingToTargetEntity = false;
+				m_path.Invalidate();
+				return;
+			}
+
+			// Set the bot's goal to the commander's position.
+			m_vGoalPos = pCommander->GetAbsOrigin();
+			CNEOBotPathCost cost(me, SAFEST_ROUTE);
+			if (m_path.Compute(me, m_vGoalPos, cost, 0.0f, true, true) && m_path.IsValid() && m_path.GetResult() == Path::COMPLETE_PATH)
+			{
+				// Path to commander found, so return.
+				return;
+			}
+		}
+		else
+		{
+			// Commander is no longer valid or alive, stop following.
+			me->m_hLeadingPlayer = nullptr;
+		}
 	}
 
 	m_hTargetEntity = NULL;
--- a/src/game/server/neo/neo_player.cpp
+++ b/src/game/server/neo/neo_player.cpp
@@ -550,6 +550,7 @@ void CNEO_Player::Spawn(void)
 
 	m_flRanOutSprintTime = 0.0f;
 	m_flNextPingTime = 0.0f;
+	m_hLeadingPlayer = nullptr; // Reset leading player on spawn
 
 	Weapon_SetZoom(false);
 
@@ -2995,6 +2996,50 @@ void CNEO_Player::SendTestMessage(const char *message)
 void CNEO_Player::SetTestMessageVisible(bool visible)
 {
 	m_bShowTestMessage = visible;
+}
+
+void CNEO_Player::ToggleFollowPlayer( CNEO_Player *pCommander )
+{
+	if ( m_hLeadingPlayer.Get() == pCommander )
+	{
+		// Bot is already following this player, so toggle off.
+		m_hLeadingPlayer = nullptr;
+	}
+	// Without other checks, players can steal others' bots.
+	else
+	{
+		// Bot starts following this player.
+		m_hLeadingPlayer = pCommander;
+	}
+}
+
+void CNEO_Player::PlayerUse( void )
+{
+	BaseClass::PlayerUse();
+
+	if ( (m_afButtonPressed & IN_USE) && !FindUseEntity() )
+	{
+		// Select bot under cursor to follow/unfollow.
+		Vector eyePos = EyePosition();
+		Vector forward;
+		EyeVectors( &forward );
+		Vector traceEnd = eyePos + forward * MAX_COORD_RANGE;
+
+		trace_t tr;
+		// Use MASK_SHOT to hit players/NPCs, with CommanderFindGoal as an example.
+		UTIL_TraceLine( eyePos, traceEnd, MASK_SHOT, this, COLLISION_GROUP_NONE, &tr );
+
+		if ( tr.DidHit() && tr.m_pEnt )
+		{
+			CNEO_Player* pTargetPlayer = ToNEOPlayer(tr.m_pEnt);
+			if ( pTargetPlayer && pTargetPlayer->IsBot())
+			{
+				// The hit entity is a bot! Now, toggle its follow state.
+				pTargetPlayer->ToggleFollowPlayer( this );
+				// TODO: Do we want to allow using players for some kind of communication?
+			}
+		}
+	}
 }
 
 void CNEO_Player::StartAutoSprint(void)
--- a/src/game/server/neo/neo_player.h
+++ b/src/game/server/neo/neo_player.h
@@ -57,6 +57,7 @@ public:
 	virtual void CalculateSpeed(void);
 	virtual void PreThink(void) OVERRIDE;
 	virtual void PlayerDeathThink(void) OVERRIDE;
+	virtual void PlayerUse(void) OVERRIDE;
 	virtual bool HandleCommand_JoinTeam(int team) OVERRIDE;
 	virtual bool ClientCommand(const CCommand &args) OVERRIDE;
 	virtual void CreateViewModel(int viewmodelindex = 0) OVERRIDE;
@@ -291,6 +292,10 @@ public:
 
 	// Bot-only usage
 	float m_flRanOutSprintTime = 0.0f;
+	CHandle<CNEO_Player> m_hLeadingPlayer; // The player this bot is following
+
+public:
+	void ToggleFollowPlayer( CNEO_Player *pCommander );
 
 private:
 	bool m_bFirstDeathTick;


sunzenshen:
https://github.com/user-attachments/assets/bea3dff0-5e1a-45ab-ac95-fbd76d3bc560

https://github.com/user-attachments/assets/5579993f-bd89-49ba-8748-36793716aa78

AdamTadeusz:
https://tenor.com/view/heavy-time-to-dance-conga-tf2-gif-19530172

src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
```diff
@@ -81,7 +81,66 @@ ActionResult< CNEOBot >	CNEOBotSeekAndDestroy::Update( CNEOBot *me, float interv
...
+					bShouldCrouch = true;
+				}
+
+				if ( bShouldCloak && bShouldCrouch )
+				{
+					break;
+				}
+
+				// Move to the next leader in the chain
+				pPlayerToMirror = pPlayerToMirror->m_hLeadingPlayer.Get();
```
sunzenshen: Along the lines of whether it makes sense for the full chain to mimic cloak and crouch behavior, maybe it makes sense only to check the person you are following immediately. (This also reduces the complexity of searching for chain cycles.)

This has the effect that bots behind a Support will not cloak, but that actually might make sense if the Support will always be visible.

Rainyan:
(I'm temporarily removing the review request since it's a draft, and the pushes will keep pinging the team members otherwise)

sunzenshen:
https://github.com/user-attachments/assets/fc1262f3-9269-47f6-938c-d71478fb81b8

- collecting bots for your current squad star by pressing use on them
- if a bot is in a different squad star, they will change their squad star to match your current one
- highlights which bots are in your squad star with color in the top-left squad star menu
- placing a ping sends bots in your currently selected squad star to that waypoint

https://github.com/user-attachments/assets/a331808d-ad88-4464-bd02-3a77e2f41788

- bots in a squad star attempt to cover surrounding angles using a rudimentary vector repulsion calculation where bots are repulsed from looking at close walls and nearby squad star teammates
- (known fixed issue in the video: the bots had a tendency to look randomly up or down, which was fixed by setting the z dimension of their final look vector to 0)
- collect bots waiting at a squad star waypoint by walking near to have them follow you

Not shown in the videos:
- calibrating how close bots follow you by setting a ping near your feet. distance clamped between a cvar minimum following distance and a maximum following distance
- releasing bots to follow default behaviors by pressing use on them while your squad star is the same.

sunzenshen:
Looked into improving the bot spacing and angle coverage when they arrive at the ghost objective:
- increased arrival zone to prevent bots from re-navigating on top of ghost
- proximity based player repulsion instead of squad based repulsion before, in order to allow non squad members to adjust fanout next to each other
- recalculate path on commander ping to improve responsiveness of repositioning
- reduced ping cooldown for bot following to make bots more responsive to following

The behavior inside of enclosed spaces with windows is admittedly be bit odd in the latter part of this clip:

https://github.com/user-attachments/assets/f39d9c15-3352-4102-ad4a-a5f1c1451bfc

Spaces that are more open with clear sightlines tend to work a bit better:

https://github.com/user-attachments/assets/ce26e9a7-24c1-4a47-af3f-64b86c97f75c

AdamTadeusz:
Also I guess this should be a separate PR, but it would be cool if the squad leader was sneaking, for the bots to sneak as well as long as they're fairly close,

AdamTadeusz:
Once the game crashing in deathmatch with bots is fixed, the bots will react to the player's request to follow them and move to the player's pings in deathmatch

image

AdamTadeusz:
also can the cs2 squad hud get the unique colour for bots following the player too 👉 👈

src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
```diff
@@ -722,7 +728,8 @@ void CNEOBotMainAction::FireWeaponAtEnemy( CNEOBot *me )
 			if ( me->IsContinuousFireWeapon( myWeapon ) )
 			{
 				// spray for a bit
-				me->PressFireButton( neo_bot_fire_weapon_min_time.GetFloat() );
+				// NEO Jank: Spraying wastes bullets into friends and walls, so just tap fire
+				me->PressFireButton( /*neo_bot_fire_weapon_min_time.GetFloat()*/ );
```
AdamTadeusz: I would suggest either removing the convar entirely if it's not used to not confuse players, or reduce the default value so the bots behave the same as if we didn't pass a value in

src/game/server/neo/neo_player.cpp
```diff
@@ -3063,6 +3084,152 @@ void CNEO_Player::SetTestMessageVisible(bool visible)
...
+	{
+		// Select bot under cursor to follow/unfollow.
+		Vector eyePos = EyePosition();
+		Vector forward;
+		EyeVectors( &forward );
+		Vector traceEnd = eyePos + forward * MAX_COORD_RANGE;
+
+		trace_t tr;
+		// Use MASK_SHOT to hit players/NPCs, with CommanderFindGoal as an example.
+		UTIL_TraceLine( eyePos, traceEnd, MASK_SHOT, this, COLLISION_GROUP_NONE, &tr );
```
AdamTadeusz: MASK_SHOT will check whether the traceline hits any of the bots hitbox. Our friendly fire warning instead uses a player's bounding box. This leads to some confusing behavior where we look at a bot and the friendly fire warning pops up indicating that we are indeed hovering over a bot, but pressing use doesn't do anything.

Can we change MASK_SHOT to MASK_SHOT_HULL?

src/game/shared/neo/neo_player_shared.cpp
```diff
@@ -165,7 +165,39 @@ void CheckPingButton(CNEO_Player* player)
 		{
 			player->m_flNextPingTime = gpGlobals->curtime;
 		}
+
+		UpdatePingCommands(player, tr.endpos);
+	}
+}
+
+static ConVar sv_neo_bot_cmdr_enable("sv_neo_bot_cmdr_enable", "1",
```
AdamTadeusz: Can we surround these convars in a ifdef GAME_DLL to get rid of the warning in the console on launch?

src/game/shared/neo/neo_player_shared.cpp
```diff
@@ -165,7 +165,39 @@ void CheckPingButton(CNEO_Player* player)
...
+	FCVAR_NONE, "Allow bots to follow you after you press use on them", true, 0, true, 1);
+static ConVar sv_neo_bot_cmdr_stop_distance_sq_max("sv_neo_bot_cmdr_stop_distance_sq_max", "50000",
+	FCVAR_NONE, "Maximum distance bot following gap interval can be set by player pings", true, 5000, true, 500000);
+static ConVar sv_neo_bot_cmdr_ping_ignore_delay_min_sec("sv_neo_bot_cmdr_ping_ignore_delay_min_sec", "3",
+	FCVAR_NONE, "Minimum time bots ignore pings for new waypoint settings", true, 0, true, 1000);
+static ConVar sv_neo_bot_cmdr_travel_time_factor("sv_neo_bot_cmdr_travel_time_factor", "300",
```
AdamTadeusz: Is this convar used anywhere?

src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
```diff
@@ -316,7 +358,7 @@ class CNotOwnedWeaponFilter : public IEntityFindFilter
 	EHANDLE		m_hEntityFound;
 };
 
-
+extern ConVar sv_neo_bot_cmdr_stop_distance_sq;
```
AdamTadeusz: Is this a standard? This convar is defined in this file already and it seems unnecessary to put this here

src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
```diff
@@ -472,9 +520,10 @@ void CNEOBotSeekAndDestroy::RecomputeSeekPath( CNEOBot *me )
 			}
 			else
 			{
-				constexpr int DISTANCE_CONSIDERED_ARRIVED_SQUARED = 10000;
+				constexpr int DISTANCE_CONSIDERED_ARRIVED_SQUARED = 100000;
```
AdamTadeusz: Kinda tricky to understand how many hammer units this is by glancing at it, would it be better to write this is as x * x, something like

Image

Surely the compiler is smart enough to simplify that right?

src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
```diff
@@ -15,6 +17,17 @@ extern ConVar neo_bot_defense_must_defend_time;
...
+ConVar sv_neo_bot_cmdr_stop_distance_sq("sv_neo_bot_cmdr_stop_distance_sq", "5000",
+	FCVAR_NONE, "Minimum distance gap between following bots", true, 3000, true, 100000);
+ConVar sv_neo_bot_cmdr_look_weights_friendly_repulsion("sv_neo_bot_cmdr_look_weights_friendly_repulsion", "20",
+	FCVAR_NONE, "Weight for friendly bot repulsion force", true, 1, true, 9999);
+ConVar sv_neo_bot_cmdr_look_weights_wall_repulsion("sv_neo_bot_cmdr_look_weights_wall_repulsion", "10",
+	FCVAR_NONE, "Weight for wall repulsion force", true, 1, true, 9999);
+ConVar sv_neo_bot_cmdr_look_weights_explosives_repulsion("sv_neo_bot_cmdr_look_weights_explosives_repulsion", "20",
+	FCVAR_NONE, "Weight for explosive repulsion force", true, 1, true, 9999);
+
+ConVar sv_neo_grenade_check_radius("sv_neo_grenade_check_radius", "1500",
```
AdamTadeusz: Ideally some of these values would be in hammer units and not squared, maybe a neotodo to update in the future so its easier for the user to change and also doesn't square the value every time its used.

Admittedly its a bit of faffing about and it doesn't work if the convar is used in multiple files so maybe it's a bad idea, but here is how you could theoretically do it.

Image

src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
sunzenshen: This is the only significant use case as far as I can tell. I ended up removing this convar.

src/game/server/neo/neo_player.cpp
sunzenshen: The contrast between the two is the MASK_SHOT is less likely to get blocked by the hull so you can kind of do a trick shot between a teammate's limbs, but MASK_SHOT_HULL is indeed a lot easier to catch running teammates with, so despite the increased blocking I think MASK_SHOT_HULL is fine.

src/game/shared/neo/neo_player_shared.cpp
sunzenshen: I originally was fiddling around with adding more time in which collection of bots by walking over to them was enabled, but the original problem was solved by keeping a smaller radius in which bots are collected.  So I forgot to clean up the unused convar.

src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
sunzenshen: I went ahead and changed this example, but admittedly with the other convars I've been tweaking their values till the last minute.  At this point though I'll let other people try out the updated numbers in my last update.

src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
sunzenshen: I think I'd rather keep experiments with lambdas in a separate PR just to let the changes in the PR bake, so we are not questioning whether something went awry with the lambda calculation.  I added a //NEOTODO comment pointing out the readability issues of these pre-squared convars.

AdamTadeusz (dismissed):

src/game/server/neo/neo_player.cpp
```diff
@@ -1941,7 +1946,22 @@ void CNEO_Player::Event_Killed( const CTakeDamageInfo &info )
 		SetDeadModel(info);
 	}
 
+	// If teamkilled by commander, the commander loses all subordinates.
```
AdamTadeusz: *looses

sunzenshen: I ended up deciding to reword this to focus on what the subordinates do (stop following commander).

src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
```diff
@@ -20,13 +20,19 @@ ConVar neo_bot_disable_seek_and_destroy( "neo_bot_disable_seek_and_destroy", "0"
...
+ConVar sv_neo_bot_cmdr_look_weights_wall_repulsion("sv_neo_bot_cmdr_look_weights_wall_repulsion", "3",
 	FCVAR_NONE, "Weight for wall repulsion force", true, 1, true, 9999);
-ConVar sv_neo_bot_cmdr_look_weights_explosives_repulsion("sv_neo_bot_cmdr_look_weights_explosives_repulsion", "2",
+ConVar sv_neo_bot_cmdr_look_weights_explosives_repulsion("sv_neo_bot_cmdr_look_weights_explosives_repulsion", "4",
 	FCVAR_NONE, "Weight for explosive repulsion force", true, 1, true, 9999);
 
+ConVar sv_neo_bot_cmdr_look_weights_friendly_max_dist("sv_neo_bot_cmdr_look_weights_friendly_max_dist", "5000",
+	FCVAR_NONE, "Distance to compare friendly repulsion forces", true, 1, true, 100000);
+ConVar sv_neo_bot_cmdr_look_weights_wall_repulsion_whisker_dist("sv_neo_bot_cmdr_look_weights_wall_repulsion_whisker_dist", "500",
+	FCVAR_NONE, "Distance to extend whiskers", true, 1, true, 100000);
```
sunzenshen: Because I ended up tweaking the effective range force repulsions, these convars are probably useful for anyone who wants to experiment with this thresholds.

src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
```diff
@@ -20,13 +20,19 @@ ConVar neo_bot_disable_seek_and_destroy( "neo_bot_disable_seek_and_destroy", "0"
 // NEOTODO: Figure out a clean way to represent config distances as Hammer units, without needing to square on use
 ConVar sv_neo_bot_cmdr_stop_distance_sq("sv_neo_bot_cmdr_stop_distance_sq", "5000",
 	FCVAR_NONE, "Minimum distance gap between following bots", true, 3000, true, 100000);
-ConVar sv_neo_bot_cmdr_look_weights_friendly_repulsion("sv_neo_bot_cmdr_look_weights_friendly_repulsion", "3",
+
+ConVar sv_neo_bot_cmdr_look_weights_friendly_repulsion("sv_neo_bot_cmdr_look_weights_friendly_repulsion", "2",
 	FCVAR_NONE, "Weight for friendly bot repulsion force", true, 1, true, 9999);
-ConVar sv_neo_bot_cmdr_look_weights_wall_repulsion("sv_neo_bot_cmdr_look_weights_wall_repulsion", "1",
+ConVar sv_neo_bot_cmdr_look_weights_wall_repulsion("sv_neo_bot_cmdr_look_weights_wall_repulsion", "3",
```
sunzenshen: In combination with the decreased wall whisker range, wall whisker checks have shorter range, but count more when the whiskers do touch.

src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
```diff
@@ -607,8 +614,9 @@ bool CNEOBotSeekAndDestroy::FollowCommandChain(CNEOBot* me)
 	if (pCommander->IsAlive())
 	{
 		// Follow commander if they are close enough to collect you (and ping cooldown elapsed)
-		if (!me->m_hLeadingPlayer && pCommander->m_tBotPlayerPingCooldown.IsElapsed()
-			&& me->GetAbsOrigin().DistToSqr(pCommander->GetAbsOrigin()) < sv_neo_bot_cmdr_stop_distance_sq.GetFloat())
+		if (pCommander->m_tBotPlayerPingCooldown.IsElapsed()
+			&& (me->GetStar() == pCommander->GetStar()) // only follow when commander is focused on your squad
```
sunzenshen: My hunch now is that commanders may only intend to have bots follow them if they are currently focused on that squad.  There may be cases where a commander wants to leave a squad at a certain position, and doesn't want them to follow if they happen to be moving close by while focused on another squad.

src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
```diff
@@ -668,7 +678,14 @@ bool CNEOBotSeekAndDestroy::FollowCommandChain(CNEOBot* me)
 	CNEO_Player* pLeader = me->m_hLeadingPlayer.Get();
 	if (pLeader && pLeader->IsAlive())
 	{
-		if (me->GetAbsOrigin().DistToSqr(pLeader->GetAbsOrigin()) < follow_stop_distance_sq)
+		// Commander can swap and order around different squads while having followers
+		// but otherwise bots only follow squadmates in the same star or their commander
```
sunzenshen: After some consideration, I decided the initial squad splitting behavior was unintuitive for a player trying to form a new squad, so decided that resetting the leader when the previous changes squads and is not the commander was a least getting closer to the right direction.

AdamTadeusz:
we're gonna have to refactor this at some point down the line so following a leader and their commands is a separate behavior

AdamTadeusz (approved):

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -2470,6 +2475,57 @@ bool CNEOBot::IsFiring() const
...
+		{
+			ClearAttribute(CNEOBot::RELOADING);
+			return false;
+		}
+		return true;
+	}
+	return false;
+}
+
 NeoClass CNEOBot::ChooseRandomClass() const
```
sunzenshen, line 2608: In case anyone was wondering what the merge conflict was, the above code block conflicted with https://github.com/NeotokyoRebuild/neo/commit/6351094f7d5a46837b6abff0dd67de344afb2177 at the same location of the below ChooseRandomClass implementation.

src/game/server/neo/bot/behavior/neo_bot_attack.cpp
```diff
@@ -58,10 +58,10 @@ ActionResult< CNEOBot >	CNEOBotAttack::Update( CNEOBot *me, float interval )
 	bool bHasRangedWeapon = me->IsRanged( myWeapon );
 
 	if ( bHasRangedWeapon
-		&& (me->IsThreatFiringAtMe(threat->GetEntity()) || threat->IsVisibleInFOVNow())
+		&& me->IsThreatFiringAtMe(threat->GetEntity())
```
sunzenshen: After checking that my branch still works okay after a series of rebases, I decided that the bots were a bit too timid (they would run for cover when they see an enemy, but in effect this meant that they would cower around corners).

So I tweaked the logic to have bots only take cover if someone they notice is directly shooting at them.  I feel like this is more analogous to how people actually play.

As part of demoing this change, I recorded some footage for review:
- https://streamable.com/93kzvn
- https://streamable.com/s9yfuy

src/game/client/neo/ui/neo_hud_round_state.cpp
```diff
@@ -671,6 +667,12 @@ void CNEOHud_RoundState::DrawPlayerList()
 				}
 			}
 
+			// Don't include local player or enemies in squad/friendly lists
+			if ((i == localPlayerIndex) || !ArePlayersOnSameTeam(i, localPlayerIndex))
```
sunzenshen: Top team status fix: I accidentally forgot to count the player as part of the  # vs #" live counters when refactoring the squad lists to show who was selected.

src/game/server/neo/bot/behavior/neo_bot_attack.cpp
sunzenshen: The bots retreating back to a corner when they were given a waypoint command was annoying during testing, as usually the commander's intent is to attack a hostile room.  I decided to skip timid turtling behavior when bots are following a commander, as it may also be annoy for bots to run away for cover if one is leading a line of bots into a danger zone.

sunzenshen:
Turning this into a draft so I can create smaller PRs based on smaller component parts of this large PR.

sunzenshen:
Updated this PR to allow using the command feature as a NavMesh debugging tool:

https://github.com/user-attachments/assets/06970d19-a1d8-4f31-b2fc-721f967388f4

New ConVar `sv_neo_bot_cmdr_debug_pause_uncommanded 1;` to allow commanding bots, while pausing every other bot that doesn't have a commander.

sunzenshen:
Refactored this PR to remove the changes not related to commanding bots, and moved most of the logic into a new behavior class definition, and moved the transition from CNEOSeekAndDestroy into instead the tactical monitor class.  I think this PR is ready for a second look, but it's also of lower priority in my mind to the other bot PRs.

One side effect of this refactor that reviewers may notice is that squad followers will still take cover if they are under fire, but they no longer follow enemies in the Attack behavior.  This was intentional as I noticed it was easier to manually set up ambushes or to micromanage which cover bots should run for when the bots don't ignore your ping waypoints to run after an enemy.

src/game/client/neo/ui/neo_hud_round_state.cpp
```diff
@@ -615,6 +615,13 @@ void CNEOHud_RoundState::DrawNeoHudElement()
 
 void CNEOHud_RoundState::DrawPlayerList()
 {
+	ConVarRef cl_neo_bot_cmdr_enable_ref("sv_neo_bot_cmdr_enable");
+	if (cl_neo_bot_cmdr_enable_ref.IsValid() && cl_neo_bot_cmdr_enable_ref.GetBool())
+	{
+		DrawPlayerList_BotCmdr();
```
sunzenshen, line 622: Decided that the best way to ensure that there are no regressions possible for the spring tournament, while allowing the use of this feature for use in navmesh debugging, was to hide the related changes behind a feature flag.  The idea is that if sv_neo_bot_cmdr_enable is not enabled, the behavior should be indistinguishable from before this feature is introduced.

sunzenshen: The relevant console commands for testing are:

```
sv_neo_bot_cmdr_enable 1;
```

src/game/shared/neo/neo_player_shared.cpp
```diff
@@ -157,9 +161,43 @@ void CheckPingButton(CNEO_Player* player)
 		{
 			player->m_flNextPingTime = gpGlobals->curtime;
 		}
+
+		UpdatePingCommands(player, tr.endpos);
 	}
 }
 
+ConVar sv_neo_bot_cmdr_enable("sv_neo_bot_cmdr_enable", "0",
```
sunzenshen, line 169: Disabled this feature by default so that we can use this feature for navmesh debugging while not impacting the player experience if this feature is disabled.

src/game/client/neo/ui/neo_hud_round_state.cpp
Rainyan: We should probably either return guard the `!cl_neo_bot_cmdr_enable_ref.IsValid()` path too, or simply Assert .IsValid() for debug rather than fail for it silently

src/game/client/neo/ui/neo_hud_round_state.cpp
```diff
@@ -712,8 +719,124 @@ void CNEOHud_RoundState::DrawPlayerList()
 	}
 }
 
+void CNEOHud_RoundState::DrawPlayerList_BotCmdr()
+{
+	if (g_PR)
+	{
```
Rainyan: Nit: we could flip this to `if (!g_PR) { return; }`, so we can unindent the entire function to the left by one tab for readability

src/game/client/neo/ui/neo_hud_round_state.cpp
```diff
@@ -712,8 +719,124 @@ void CNEOHud_RoundState::DrawPlayerList()
...
+		m_nonCommandedList.RemoveAll();
+		m_nonSquadList.RemoveAll();
+
+		bool squadMateFound = false;
+		m_iLeftPlayersAlive = 0;
+		m_iRightPlayersAlive = 0;
+		const int localStar = g_PR->GetStar(localPlayerIndex);
+
+		// Single pass to collect and categorize players
+		for (int i = 0; i < (MAX_PLAYERS + 1); i++)
```
Rainyan: Player entity indices are one-indexed, so we could loop starting from `i = 1` to skip the redundant 0 index case.

Also, could probably loop up to `<= gpGlobals->maxClients` as the upper bound, which can be less than the absolute hard-coded `MAX_PLAYERS`.

src/game/client/neo/ui/neo_hud_round_state.cpp
```diff
@@ -712,8 +719,124 @@ void CNEOHud_RoundState::DrawPlayerList()
...
+				squadMateFound = true;
+			}
+			else
+			{
+				m_nonSquadList.AddToTail(i);
+			}
+		}
+
+		// Draw commanded players first
+		const auto player = ToNEOPlayer(UTIL_PlayerByIndex(localPlayerIndex));
```
Rainyan: Nit: could be `C_NEO_Player::GetLocalNEOPlayer` to avoid the entindex lookup

src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
```diff
@@ -0,0 +1,374 @@
+#include "cbase.h"
+#include "neo_player.h"
+#include "bot/neo_bot.h"
+#include "bot/behavior/neo_bot_command_follow.h"
+#include "nav_mesh.h"
+
```
Rainyan, line 6: Perhaps we should add the mem debug header, for consistency:
```cpp
#include "tier0/memdbgon.h"
```

src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
```diff
@@ -0,0 +1,374 @@
...
+ConVar sv_neo_bot_cmdr_look_weights_friendly_repulsion("sv_neo_bot_cmdr_look_weights_friendly_repulsion", "2",
+	FCVAR_CHEAT, "Weight for friendly bot repulsion force", true, 1, true, 9999);
+ConVar sv_neo_bot_cmdr_look_weights_wall_repulsion("sv_neo_bot_cmdr_look_weights_wall_repulsion", "3",
+	FCVAR_CHEAT, "Weight for wall repulsion force", true, 1, true, 9999);
+ConVar sv_neo_bot_cmdr_look_weights_explosives_repulsion("sv_neo_bot_cmdr_look_weights_explosives_repulsion", "4",
+	FCVAR_CHEAT, "Weight for explosive repulsion force", true, 1, true, 9999);
+
+ConVar sv_neo_bot_cmdr_look_weights_friendly_max_dist("sv_neo_bot_cmdr_look_weights_friendly_max_dist", "5000",
+	FCVAR_CHEAT, "Distance to compare friendly repulsion forces", true, 1, true, 100000);
```
Rainyan: Nit: squared? Perhaps rename as `..._sq` for clarity

src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
```diff
@@ -0,0 +1,374 @@
...
+	if (!pCommander)
+	{
+		return false;
+	}
+
+	// Mirror behavior of leader if we have one
+	if (me->m_hLeadingPlayer.Get())
+	{
+		CNEO_Player *pPlayerToMirror = me->m_hLeadingPlayer.Get();
+		if (pPlayerToMirror)
```
Rainyan: This looks like the inner and outer if-conditions on lines 92 and 95 are the same; could we perhaps simplify this to something like:
```cpp
if (CNEO_Player *pPlayerToMirror = me->m_hLeadingPlayer.Get())
{
```

src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
```diff
@@ -0,0 +1,374 @@
...
+		// Follow commander if they are close enough to collect you (and ping cooldown elapsed)
+		if (pCommander->m_tBotPlayerPingCooldown.IsElapsed()
+			&& (me->GetStar() == pCommander->GetStar()) // only follow when commander is focused on your squad
+			&& (me->GetAbsOrigin().DistToSqr(pCommander->GetAbsOrigin()) < sv_neo_bot_cmdr_stop_distance_sq.GetFloat()))
+		{
+			// Use sv_neo_bot_cmdr_stop_distance_sq for consistent bot collection range
+			// follow_stop_distance_sq would be confusing if player doesn't know about distance tuning
+			me->m_hLeadingPlayer = pCommander;
+			m_vGoalPos = vec3_origin;
+			pCommander->m_vLastPingByStar.GetForModify(me->GetStar()) = vec3_origin;
```
Rainyan: Maybe out of scope for this PR, but perhaps we should consider refactoring the waypoint logic to use a different magic value from (such a `vec3_invalid`) for marking the "invalid/no location". Since it could be conceivable that a bot would want to reach coordinates 0,0,0 (`vec3_origin`) in some map under some circumstances.

Also, it would be nice to define this as a distinct invalid waypoint value constant instead of relying on the magic value itself, so the intent is more clear.

src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
```diff
@@ -0,0 +1,374 @@
...
+
+	// Calibrate dynamic follow distance
+	float follow_stop_distance_sq = sv_neo_bot_cmdr_stop_distance_sq.GetFloat();
+	if (pCommander->m_flBotDynamicFollowDistanceSq > follow_stop_distance_sq)
+	{
+		follow_stop_distance_sq = pCommander->m_flBotDynamicFollowDistanceSq;
+	}
+
+	if (pCommander->IsAlive())
+	{
```
Rainyan: Nit: this condition goes 4 levels deep (if -> else if -> if -> if ...) which can make it a bit confusing to keep track of the scope. I'd personally prefer if we reversed this base case condition to an early return:
```cpp
if (!pCommander->IsAlive())
{
	// Commander died
	return false; // This triggers OnEnd which clears vars
}
```
so that we can collapse the lowest indent away entirely.

src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
```diff
@@ -0,0 +1,374 @@
...
+		// Leader is no longer valid or alive
+		me->m_hLeadingPlayer = nullptr;
+		return true;
+	}
+}
+
+// ---------------------------------------------------------------------------------------------
+// Process commander ping waypoint commands for bots
+// The movementTarget is mutable to accomodate spreading out at goal zone 
+bool CNEOBotCommandFollow::FanOutAndCover(CNEOBot* me, Vector& movementTarget, bool bMoveToSeparate /*= false*/, float flArrivalZoneSizeSq /*= -1.0f*/)
```
Rainyan: Nit: The `bMoveToSeparate` default value `/*= false*/` commented here doesn't match the header declaration default value `true`. Should probably be fixed to avoid confusion for someone going by this line alone.

src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
```diff
@@ -0,0 +1,374 @@
...
+	{
+		// Combined loop for player forces
+		for (int idx = 1; idx <= gpGlobals->maxClients; ++idx)
+		{
+			CBasePlayer* pPlayer = UTIL_PlayerByIndex(idx);
+			if (!pPlayer || pPlayer == me || !pPlayer->IsAlive())
+				continue;
+
+			if (pPlayer->GetTeamNumber() == me->GetTeamNumber())
+			{
```
Rainyan: Nit: since we're nesting quite deep here, would prefer flipping this condition to an early `continue` for readability. Likewise the `if (pOther)` at line 280 could be flipped. Because else, the most complex case here is a: `if->for->if->if->if->if` which personally I find a little visually confusing when trying to quickly reason about the control flow.

src/game/server/neo/bot/behavior/neo_bot_command_follow.h
```diff
@@ -0,0 +1,34 @@
+#ifndef NEO_BOT_COMMAND_FOLLOW_H
+#define NEO_BOT_COMMAND_FOLLOW_H
+#ifdef _WIN32
+#pragma once
+#endif
+
+#include "Path/NextBotChasePath.h"
```
Rainyan: Nit: do we need this include for anything here? Also, perhaps we should include NextBotBehavior.h for the `class Action` definition, for good measure.

src/game/server/neo/bot/behavior/neo_bot_command_follow.h
```diff
@@ -0,0 +1,34 @@
...
+{
+public:
+	CNEOBotCommandFollow();
+
+	virtual ActionResult< CNEOBot >	OnStart( CNEOBot *me, Action< CNEOBot > *priorAction );
+	virtual ActionResult< CNEOBot >	Update( CNEOBot *me, float interval );
+	virtual ActionResult< CNEOBot >	OnResume( CNEOBot *me, Action< CNEOBot > *interruptingAction );
+	virtual void					OnEnd( CNEOBot *me, Action< CNEOBot > *nextAction );
+
+	virtual const char *GetName( void ) const	{ return "CommandFollow"; };
```
Rainyan: Nit: would prefer if we explicitly declared these `override`

src/game/server/neo/bot/behavior/neo_bot_pause.cpp
```diff
@@ -0,0 +1,61 @@
+#include "cbase.h"
+#include "neo_player.h"
+#include "bot/neo_bot.h"
+#include "bot/behavior/neo_bot_pause.h"
+#include "bot/behavior/neo_bot_command_follow.h"
+
```
Rainyan, line 6: Perhaps we should add the mem debug header, for consistency:
```cpp
#include "tier0/memdbgon.h"
```

src/game/server/neo/bot/behavior/neo_bot_pause.cpp
```diff
@@ -0,0 +1,61 @@
...
+			me->StopLookingAroundForEnemies();
+		}
+	}
+
+	// Stop moving
+	const PathFollower *path = me->GetCurrentPath();
+	if (path)
+	{
+		// Stop existing path
+		const_cast<PathFollower *>(path)->Invalidate();
```
Rainyan: This looks like it might be undefined behaviour, since we're mutating the object through what was a const pointer. Should we instead declare a non-const version of the `GetCurrentPath` getter, and mutate through that?

src/game/server/neo/bot/behavior/neo_bot_pause.h
```diff
@@ -0,0 +1,19 @@
...
+#include "bot/neo_bot.h"
+
+class CNEOBotPause : public Action< CNEOBot >
+{
+public:
+	virtual ActionResult< CNEOBot >	OnStart( CNEOBot *me, Action< CNEOBot > *priorAction );
+	virtual ActionResult< CNEOBot >	Update( CNEOBot* me, float interval );
+	virtual void					OnEnd( CNEOBot *me, Action< CNEOBot > *nextAction );
+
+	virtual const char* GetName( void ) const { return "Pause"; };
```
Rainyan: Nit: preferably mark virtual overrides explicitly with the `override` specifier

src/game/server/neo/neo_player.cpp
```diff
@@ -3177,6 +3201,120 @@ void CNEO_Player::SetTestMessageVisible(bool visible)
...
+void CNEO_Player::ToggleBotFollowCommander(CNEO_Player* pCommander)
+{
+	if (!sv_neo_bot_cmdr_enable.GetBool())
+	{
+		return;
+	}
+
+	if (!pCommander)
+	{
+		DevMsg("ToggleBotFollowCommander called without valid player handle!\n");
```
Rainyan: Nit: Maybe should be `DevWarning` to get the more prominent red warning print color for the console output

src/game/shared/neo/neo_player_shared.cpp
```diff
@@ -157,9 +161,43 @@ void CheckPingButton(CNEO_Player* player)
...
+{
+#ifdef GAME_DLL
+	if (sv_neo_bot_cmdr_enable.GetBool())
+	{
+		player->m_vLastPingByStar.GetForModify(player->GetStar()) = pingPos;
+		float distSqrToPing = player->GetAbsOrigin().DistToSqr(pingPos);
+		if (distSqrToPing < sv_neo_bot_cmdr_stop_distance_sq_max.GetFloat())
+		{
+			// If pinging close to self, calibrate follow distance of commanded bots based on distance to ping
+			player->m_flBotDynamicFollowDistanceSq = Clamp(distSqrToPing, 5000.0f, sv_neo_bot_cmdr_stop_distance_sq_max.GetFloat());
```
Rainyan: Nit: maybe it'd be good to get the clamp min value from the cvar directly, so these cannot go out of sync if someone updates the cvar min value code but forgets about this hard-coded 5000.f

src/game/client/neo/ui/neo_hud_round_state.cpp
sunzenshen: went with an Assert

src/game/client/neo/ui/neo_hud_round_state.cpp
sunzenshen: I applied this suggestion to the other 2 places that had this logic (even if they were the original player list logic and not related to this PR) and it looks fine even when this PR's feature flag is disabled.

src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
sunzenshen: Added a new constant VECTOR_INVALID_WAYPOINT

src/game/server/neo/bot/behavior/neo_bot_pause.cpp
sunzenshen: I was kind of overthinking this, nulled out the path in a simpler way.

src/game/shared/neo/neo_player_shared.cpp
sunzenshen: FCVAR_REPLICATED - I need this flag to be shared with the client so that their playerlist UI can switch between the original player/avatar list behavior or the new highlighted commanded bots version.

FCVAR_ARCHIVE - Allows server operators or single players to decide once if they want to try out the feature, but this feature is still off by default so that parity gameplay is not affected for those not opting into using the feature.

src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
```diff
@@ -0,0 +1,369 @@
...
+	FCVAR_CHEAT, "Weight for friendly bot repulsion force", true, 1, true, 9999);
+ConVar sv_neo_bot_cmdr_look_weights_wall_repulsion("sv_neo_bot_cmdr_look_weights_wall_repulsion", "3",
+	FCVAR_CHEAT, "Weight for wall repulsion force", true, 1, true, 9999);
+ConVar sv_neo_bot_cmdr_look_weights_explosives_repulsion("sv_neo_bot_cmdr_look_weights_explosives_repulsion", "4",
+	FCVAR_CHEAT, "Weight for explosive repulsion force", true, 1, true, 9999);
+
+ConVar sv_neo_bot_cmdr_look_weights_friendly_max_dist_sq("sv_neo_bot_cmdr_look_weights_friendly_max_dist_sq", "5000",
+	FCVAR_CHEAT, "Distance to compare friendly repulsion forces", true, 1, true, 100000);
+ConVar sv_neo_bot_cmdr_look_weights_wall_repulsion_whisker_dist("sv_neo_bot_cmdr_look_weights_wall_repulsion_whisker_dist", "500",
+	FCVAR_CHEAT, "Distance to extend whiskers", true, 1, true, 100000);
```
AdamTadeusz: perhaps the max value here should be MAX_TRACE_LENGTH

src/game/client/neo/ui/neo_hud_round_state.cpp
```diff
@@ -767,6 +894,62 @@ int CNEOHud_RoundState::DrawPlayerRow(int playerIndex, const int yOffset, bool s
...
+	}
+	g_pVGuiLocalize->ConvertANSIToUnicode(squadMateText, wSquadMateText, sizeof(wSquadMateText));
+
+	int fontWidth, fontHeight;
+	surface()->DrawSetTextFont(small ? m_hOCRSmallerFont : m_hOCRSmallFont);
+	surface()->GetTextSize(m_hOCRSmallFont, m_wszPlayersAliveUnicode, fontWidth, fontHeight);
+
+	surface()->DrawSetTextColor(colorOverride ? *colorOverride : (isAlive ? COLOR_FADED_WHITE : COLOR_DARK_FADED_WHITE));
+	surface()->DrawSetTextPos(8, yOffset);
+	surface()->DrawPrintText(wSquadMateText, V_wcslen(wSquadMateText));
```
Rainyan: Perf nit, since the HUD draw loop is inherently somewhat of a hot path: you can get the string length "for free", if you store the result of `g_pVGuiLocalize->ConvertANSIToUnicode` and subtract 1 (the null terminator)

That being said, I imagine this is negligible (and we already do this exact pattern elsewhere), so I'm not requiring any changes here necessarily.

src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
```diff
@@ -0,0 +1,369 @@
...
+				vWallRepulsion += tr.plane.normal * flRepulsion;
+			}
+		}
+
+		// Combine forces and look at final direction
+		float friendlyRepulsionWeight = sv_neo_bot_cmdr_look_weights_friendly_repulsion.GetFloat();
+		float wallRepulsionWeight = sv_neo_bot_cmdr_look_weights_wall_repulsion.GetFloat();
+		vFinalForce = (vBotRepulsion * friendlyRepulsionWeight) + (vWallRepulsion * wallRepulsionWeight);
+		vFinalForce.NormalizeInPlace();
+		vFinalForce.z = 0; // avoid tilting awkwardly up or down
```
Rainyan: This may be fine, but since it seems like we're doing the "unit vector times a scale" pattern, would it make more sense to first zero the Z axis, and only then normalize (since zeroing after the normalization means its length no longer equals one, i.e. the magnitude of the final X and Y depend on what Z was before zeroing)? I'm not really sure so this is more of a question than anything. If it produces good behaviour then I imagine it hardly matters.

Rainyan (approved):
This is an awesome feature and I wish we can reach consensus to default-enable it eventually.

sunzenshen:
> This is an awesome feature and I wish we can reach consensus to default-enable it eventually.

My thinking was that we could wait to enable this feature by default until after the spring tournament build cutoff is released.  By enabling the feature by default after the spring tournament, we can avoid any possible unintended issues cropping up during the the competition, and also benefit from some potential increased attention to advertise the feature.
Then after a few months of baking or even just before anyone wants to make updates to neo_hud_round_state, we could have a PR ready to make the bot version of the round state code the primary path that always happens, and then later on we just completely delete the old version of the round state code when we feel confident about the bot-command-aware approach.

src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
AdamTadeusz: @sunzenshen I was referring to sv_neo_bot_cmdr_look_weights_wall_repulsion_whisker_dist only. I'm looking at how sv_neo_bot_cmdr_look_weights_friendly_max_dist_sq is being used and perhaps it should have the same max value like you just implemented it, but then maybe the name of the convar is misleading? Since sv_neo_bot_cmdr_look_weights_friendly_max_dist_sq is being used to calculate a scale value between 0 and 1, and flDist in that calculation is the square root of the distance between two vectors, is sv_neo_bot_cmdr_look_weights_friendly_max_dist_sq actually a normal distance instead of a distance squared?

Alternatively perhaps flDistSqr should be used to calculate flRepulsionScale instead of flDist, at which point it would make sense for sv_neo_bot_cmdr_look_weights_friendly_max_dist_sq to keep _sq at the end of its name but then maybe the max value of MAX_TRACE_LENGTH doesn't make as much sense anymore, since the maximum value of flDistSqr would be MAX_TRACE_LENGTH ^ 2

sunzenshen: In retrospect, I think I must have been embarassingly conflating _sq and flDistSqr, which as you point out the latter was already being handled with a FastSqrt.  I think removing the _sq is the more intuitive choice, as if we went with using MAX_TRACE_LENGTH ^ 2 as the max range, then the default values and configurable range would involve very large numbers, which could be much harder for console users to intuit based on Hammer units.

src/game/client/neo/ui/neo_hud_round_state.cpp
```diff
@@ -712,8 +720,127 @@ void CNEOHud_RoundState::DrawPlayerList()
...
+			if (g_PR->IsAlive(i))
+			{
+				m_iLeftPlayersAlive++;
+			}
+		}
+		if (playerTeam != localPlayerTeam)
+		{
+			continue;
+		}
+		if (i == localPlayerIndex || localPlayerSpec || hideDueToScoreboard || !ArePlayersOnSameTeam(i, localPlayerIndex))
```
Rainyan, line 781: Nit: It seems like `!ArePlayersOnSameTeam` is redundant, since we already checked `if (playerTeam != localPlayerTeam)` on line 777 above. Maybe it should be removed, since it's going to invoke the virtual call `g_Teams[i]->ContainsPlayer` twice per each iteration, so worst case it's `gpGlobals->maxClients * 2 - 1` extra virtual method calls per each frame this draw call happens.

Or if the `ArePlayersOnSameTeam` case is a concern, perhaps we should assert on it for debug builds, rather than silently guard against it with the continue. But I imagine it should realistically never occur(?)

sunzenshen: I'll address this in a separate PR.

sunzenshen: Addressed in #1617

PR #1348, opened 2025-09-14
Reduce likelihood that bots fire through friendlies

sunzenshen:
## Description
Check for friendlies between bot and target before opening fire.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/neo_bot.cpp
+++ b/src/game/server/neo/bot/neo_bot.cpp
@@ -1803,9 +1803,40 @@ bool CNEOBot::IsLineOfFireClear(const Vector& where, const LineOfFireFlags flags
 
 
 //-----------------------------------------------------------------------------------------------------
+// Return whether there is a friendly player blocking the line of fire
+bool CNEOBot::IsLineOfFireClearOfFriendlies(const Vector& from, CBaseEntity* who) const
+{
+	if (NEORules()->IsTeamplay())
+	{
+		trace_t playerTrace;
+		CTraceFilterSimple playerFilter(this, COLLISION_GROUP_NONE);
+		const Vector to = who->WorldSpaceCenter();
+		UTIL_TraceLine(from, to, MASK_SHOT_HULL, &playerFilter, &playerTrace);
+
+		if (playerTrace.DidHit())
+		{
+			if (playerTrace.m_pEnt && playerTrace.m_pEnt->IsPlayer())
+			{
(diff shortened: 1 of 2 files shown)


src/game/server/neo/bot/neo_bot.cpp
sunzenshen, line 1843: My understanding from these existing filters is that friendly teammates will not block these tracelines (so IsLineOfFireClear will return true even if there is a friendly blocking a clean shot.  That said, there were some impressions from Discord that there should be a cleaner way to check for friendly fire scenarios.  For example, perhaps these tracelines and filters could be configured in a way that does not require an additional traceline.

Opening this PR up for review mostly to surface such discussion.

AdamTadeusz: a tracefilterchain ensures a traceline hits an entity only if both filters passed to it hit that entity, so yes this IsLineOfFireClear check would not prevent incidents of friendly fire. I imagine the filter to ignore all actors was added so that the target itself doesn't block the trace.

We could replace the NextBotTraceFilterIgnoreActors with a custom NextBotTraceFilterIgnoreTeamActors instead. This would implement its own new filter function similar to IgnoreActorsTraceFilterFunction. To avoid making a NextBotTraceFilterIgnoreJinrai and NextBotTraceFilterIgnoreNSF, we could add CONTENTS_TEAM1 and CONTENTS_TEAM2 content flags to the lofMask depending on the team of the shooter, and then check those against the team of the hit entity in the filter function.

I think the IgnoreActorsOnTeamTraceFilterFunction would look something like this

Image

In this example we would want to set CONTENTS_TEAM1 if shooter is NSF, and CONTENTS_TEAM2 if shooter is Jinrai.

I want to stress that I haven't tested this yet, I plan to do so but wanted to just give an alternative that avoids the extra traceline

sunzenshen: Feel free to push directly to this branch. 👍 

I'll follow up later when I decide to pick up this branch again.

AdamTadeusz (approved):
if memory serves CTraceFilterChain requires that an entity is hit by both filters that make up the filter chain, so a custom CTraceFilterChain would have to be added where the shouldhit checks whether either of the filters that make it up would hit instead of only if both should hit, so im just gonna approve this for now, maybe add a TODO to simplify this in the future

PR #1352, opened 2025-09-16
Automatic clientside demo recording

### LINKED ISSUE #223: Match demo recording
map demos are automatically recorded on the server on match start. Players can toggle recording of demos on match start

DESTROYGIRL:
## Description
Server cvar to make clients automatically record demos like in the comp plugin for OG

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #223

### DIFF
--- a/src/game/shared/neo/neo_gamerules.cpp
+++ b/src/game/shared/neo/neo_gamerules.cpp
@@ -1,6 +1,7 @@
 #include "cbase.h"
 #include "neo_gamerules.h"
 #include "neo_version_info.h"
+#include "neo_misc.h"
 #include "in_buttons.h"
 #include "ammodef.h"
 
@@ -28,7 +29,6 @@
 #include "player_resource.h"
 #include "inetchannelinfo.h"
 #include "neo_dm_spawn.h"
-#include "neo_misc.h"
 #include "neo_game_config.h"
 #include "nav_mesh.h"
 
@@ -135,6 +135,7 @@ ConVar sv_neo_readyup_lobby("sv_neo_readyup_lobby", "0", FCVAR_REPLICATED, "If e
 ConVar sv_neo_pausematch_enabled("sv_neo_pausematch_enabled", "0", FCVAR_REPLICATED, "If enabled, players will be able to pause the match mid-game.", true, 0.0f, true, 1.0f);
 ConVar sv_neo_pausematch_unpauseimmediate("sv_neo_pausematch_unpauseimmediate", "0", FCVAR_REPLICATED | FCVAR_CHEAT, "Testing only - If enabled, unpause will be immediate.", true, 0.0f, true, 1.0f);
 ConVar sv_neo_readyup_countdown("sv_neo_readyup_countdown", "5", FCVAR_REPLICATED, "Set the countdown from fully ready to start of match in seconds.", true, 0.0f, true, 120.0f);
+ConVar sv_neo_client_autorecord("sv_neo_client_autorecord", "0", FCVAR_REPLICATED | FCVAR_DONTRECORD, "Record demos clientside", true, 0, true, 1);
 ConVar sv_neo_ghost_spawn_bias("sv_neo_ghost_spawn_bias", "0", FCVAR_REPLICATED, "Spawn ghost in the same location as the previous round on odd-indexed rounds (Round 1 = index 0)", true, 0, true, 1);
 ConVar sv_neo_juggernaut_spawn_bias("sv_neo_juggernaut_spawn_bias", "0", FCVAR_REPLICATED, "Spawn juggernaut in the same location as the previous round on odd-indexed rounds (Round 1 = index 0)", true, 0, true, 1);
 
@@ -147,6 +148,7 @@ static void neoSvCompCallback(IConVar* var, const char* pOldValue, float flOldVa
 	sv_neo_pausematch_enabled.SetValue(bCurrentValue);
 	sv_neo_ghost_spawn_bias.SetValue(bCurrentValue);
 	sv_neo_juggernaut_spawn_bias.SetValue(bCurrentValue);
+	sv_neo_client_autorecord.SetValue(bCurrentValue);
 }
 
 ConVar sv_neo_comp("sv_neo_comp", "0", FCVAR_REPLICATED, "Enables competitive gamerules", true, 0.f, true, 1.f
@@ -154,6 +156,8 @@ ConVar sv_neo_comp("sv_neo_comp", "0", FCVAR_REPLICATED, "Enables competitive ga
 	, neoSvCompCallback
 	#endif // GAME_DLL
 );
+
+ConVar sv_neo_comp_name("sv_neo_comp_name", "", FCVAR_REPLICATED, "Name of the competition. Leave blank if unused.");
 
 #ifdef CLIENT_DLL
 extern ConVar snd_victory_volume;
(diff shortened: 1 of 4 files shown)


AdamTadeusz:
Should clients be able to control whether the server can start and stop recordings? If I'm playing on a device short on storage I imagine I don't want every random server I join to force me to record a demo. Also If I do want a demo but I have my own naming convention, this would stop my recording in favor of the new recording.

DESTROYGIRL:
I thought about this but I wasn't really sure, in current tournament play you have to record clientside demos or else you're breaking a rule.
Servers don't use this outside of pugs / comp, if a pub server has this switched on permanently then its probably a bad server and you should stop playing on it.
It's also worth noting that this doesn't actually restrict control from the client, they can still stop a recording at any time, and if they have their own naming convention its easy to go back and rename as the names have map, timestamp, etc

So having the option probably wouldn't hurt outside of people unintentionally breaking tournament rules. Maybe it's a question for orga

src/game/shared/neo/neo_misc.cpp
```diff
@@ -22,3 +30,47 @@
...
+	int year, month, dayOfWeek, day, hour, minute, second;
+	g_pVGuiSystem->GetCurrentTimeAndDate(&year, &month, &dayOfWeek, &day, &hour, &minute, &second);
+	V_snprintf(timeSection, sizeof(timeSection), "%04d%02d%02d-%02d%02d", year, month, day, hour, minute);
+
+	// Competition name
+	char compSection[32];
+	V_strcpy(compSection, sv_neo_comp_name.GetString());
+
+	// Build the filename
+	char replayName[sizeof(steamSection) + sizeof(mapSection) + sizeof(timeSection) + sizeof(compSection) + 2];
```
sunzenshen: Might be simpler to just declare this as replayName[MAX_PATH].

src/game/shared/neo/neo_misc.cpp
```diff
@@ -22,3 +30,47 @@
...
+	const CSteamID steamID = GetSteamIDForPlayerIndex(GetLocalPlayerIndex());
+	if (steamID.IsValid())
+	{
+		auto accountID = steamID.GetAccountID();
+		V_snprintf(steamSection, sizeof(steamSection), "%d", accountID);
+	}
+
+	// Map name
+	char mapSection[256];
+	V_strcpy(mapSection, GameRules()->MapName());
```
sunzenshen: Just to be proactive, might be safer to use V_strncpy:

```cpp
V_strncpy(mapSection, GameRules()->MapName(), sizeof(mapSection));
```

src/game/shared/neo/neo_misc.cpp
```diff
@@ -22,3 +30,47 @@
...
+
+	// Time and date
+	char timeSection[16];
+	int year, month, dayOfWeek, day, hour, minute, second;
+	g_pVGuiSystem->GetCurrentTimeAndDate(&year, &month, &dayOfWeek, &day, &hour, &minute, &second);
+	V_snprintf(timeSection, sizeof(timeSection), "%04d%02d%02d-%02d%02d", year, month, day, hour, minute);
+
+	// Competition name
+	char compSection[32];
+	V_strcpy(compSection, sv_neo_comp_name.GetString());
```
sunzenshen: Might be safer to use V_strncpy:
```cpp
V_strncpy(compSection, sv_neo_comp_name.GetString(), sizeof(compSection));
```

src/game/shared/neo/neo_misc.cpp
```diff
@@ -22,3 +30,47 @@
 {
 	return LoopAroundMinMax(iValue, 0, iSize - 1);
 }
+
+#ifdef CLIENT_DLL
+void StartAutoClientRecording()
+{
+	// SteamID
+	char steamSection[64];
```
sunzenshen, line 41: There's a branch below where this is never initialized, and then might have garbage data for the copies later.

Could initialize as declaration on this same line:

```cpp
char steamSection[64] = { 0 };
```

Or add an else block below if you wanted a placeholder, e.g.:
```cpp
else
{
    V_strncpy(steamSection, "nosteamid", sizeof(steamSection));
}
```

sunzenshen (approved):

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -138,6 +138,10 @@ ConVar sv_neo_pausematch_unpauseimmediate("sv_neo_pausematch_unpauseimmediate",
 ConVar sv_neo_readyup_countdown("sv_neo_readyup_countdown", "5", FCVAR_REPLICATED, "Set the countdown from fully ready to start of match in seconds.", true, 0.0f, true, 120.0f);
 ConVar sv_neo_ghost_spawn_bias("sv_neo_ghost_spawn_bias", "0", FCVAR_REPLICATED, "Spawn ghost in the same location as the previous round on odd-indexed rounds (Round 1 = index 0)", true, 0, true, 1);
 ConVar sv_neo_juggernaut_spawn_bias("sv_neo_juggernaut_spawn_bias", "0", FCVAR_REPLICATED, "Spawn juggernaut in the same location as the previous round on odd-indexed rounds (Round 1 = index 0)", true, 0, true, 1);
+ConVar sv_neo_client_autorecord("sv_neo_client_autorecord", "0", FCVAR_REPLICATED | FCVAR_DONTRECORD, "Record demos clientside", true, 0, true, 1);
+#ifdef CLIENT_DLL
+ConVar cl_neo_client_autorecord_allow("cl_neo_client_autorecord_allow", "1", FCVAR_ARCHIVE, "Allow servers to automatically record demos on the client", true, 0, true, 1);
```
AdamTadeusz, line 143: hot take this should default to 0 xD

DESTROYGIRL: It's pretty pointless if 0 by default, we wanna avoid cases of players forgetting to turn it on or not knowing they had to

src/game/shared/neo/neo_misc.cpp
```diff
@@ -73,6 +74,11 @@ void StartAutoClientRecording()
 		V_snprintf(replayName, sizeof(replayName), "%s_%s_%s", timeSection, mapSection, steamSection);
 	}
 
-	engine->StartDemoRecording(replayName); // Start recording
+	if (!g_pFullFileSystem->IsDirectory("demos"))
+	{
+		g_pFullFileSystem->CreateDirHierarchy("demos");
+	}
+
+	engine->StartDemoRecording(replayName, "demos"); // Start recording
```
AdamTadeusz: Could you define a DEMOS_DIRECTORY_NAME and use that here instead?

DESTROYGIRL: I was considering doing that when adding it but then i was like "naaaaaaah why would we ever need that" lol

AdamTadeusz: https://tenor.com/view/do-it-star-wars-emperor-palpatine-palpatine-gif-799657800635657398

src/game/shared/neo/neo_misc.cpp
```diff
@@ -1,5 +1,16 @@
...
 
+#ifdef CLIENT_DLL
+#include "steamclientpublic.h"
+#include "vgui/ISystem.h"
+#include "tier3.h"
+#include <filesystem.h>
+
+extern ConVar sv_neo_comp_name;
+
+#define DEMOS_DIRECTORY_NAME "demos"
```
sunzenshen, line 11: Maybe this could be an FCVAR_ARCHIVE ConVar, but up to you whether we want to allow the user to define a demo directory.  One argument against is what if they try weird path traversal shenanigans and try to overwrite a different folder location.

DESTROYGIRL: I feel like if users were that particular with how their demos were organised they wouldn't be storing them in their install folder anyway but i dunno

PR #1354, opened 2025-09-16
Always draw match status over gamemode description

DESTROYGIRL:
## Description
Changes made to hide the warmup text was not enough, overlap would still happen in other instances (e.g. match point)
Match status should take priority over the gamemode description

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #1337

### DIFF
--- a/src/game/client/neo/ui/neo_hud_round_state.cpp
+++ b/src/game/client/neo/ui/neo_hud_round_state.cpp
@@ -244,6 +244,52 @@ void CNEOHud_RoundState::UpdateStateForNeoHudElementDraw()
 	{
 		m_pWszStatusUnicode = L"Match point";
 	}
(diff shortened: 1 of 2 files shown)


AdamTadeusz:
If we want to go down the route of showing either the objective or the round status, lets just treat them as the same element instead.

image

DESTROYGIRL:
Do you mean treat them as the same element or literally make them the same element

AdamTadeusz:
> Do you mean treat them as the same element or literally make them the same element

I think we can literally make them the same element

AdamTadeusz:
Well they're both part of neo_hud_roundstate so its already the same element no? I just mean write whatever we want to display under the timer to one wchar so thats the "under the timer wchar" wchar

DESTROYGIRL:
Is this what you had in mind

AdamTadeusz (approved):

sunzenshen (approved):

PR #1361, opened 2025-09-19
NeoImageButton

AdamTadeusz:
## Description

Replaces the ImagePanel Button couples in the class and loadout selection menus with a single neoImageButton object, removes the incorrect border and adds an on hover border like in the original

### DIFF
--- a/src/game/client/neo/game_controls/neo_image_button.h
+++ b/src/game/client/neo/game_controls/neo_image_button.h
@@ -26,6 +26,9 @@ protected:
 	void Paint(void) override;
 
 private:
+	CPanelAnimationVarAliasType(int, wide, "wide", "256", "int");
+	CPanelAnimationVarAliasType(int, tall, "tall", "256", "int");
+
 	int	m_iTextureID = -1;
 };
 
--- a/src/game/client/neo/game_controls/neo_image_button.cpp
+++ b/src/game/client/neo/game_controls/neo_image_button.cpp
@@ -31,9 +31,6 @@ void CNeoImageButton::Paint()
 		return;
 	}
 
-	int wide, tall;
-	GetSize(wide, tall);
-
 	if (m_iTextureID != -1)
 	{
 		surface()->DrawSetColor(255, 255, 255, 255);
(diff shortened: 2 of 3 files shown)


src/game/client/neo/game_controls/neo_classmenu.cpp
```diff
@@ -141,12 +138,24 @@ void CNeoClassMenu::FireGameEvent(IGameEvent* event)
...
+	m_pSkin2_Button = FindControl<CNeoImageButton>("Skin2_Button");
+	if (m_pSkin2_Button)
+	{
+		m_pSkin2_Button->SetButtonTexture("vgui/cm/jinrai_assault02");
+	}
+
+	m_pSkin3_Button = FindControl<CNeoImageButton>("Skin3_Button");
+	if (m_pSkin3_Button)
+	{
+		m_pSkin1_Button->SetButtonTexture("vgui/cm/jinrai_assault03");
```
sunzenshen: Is `m_pSkin1_Button` a potential copy/pasta typo here?  I think this might accidentally set model 1 rather than 3.

src/game/client/neo/game_controls/neo_image_button.h
```diff
@@ -0,0 +1,34 @@
...
+
+	void SetButtonTextureID(int textureID);
+	void SetButtonTexture(const char* texture);
+
+protected:
+	void ApplySchemeSettings(IScheme* pScheme) override;
+	void Paint(void) override;
+
+private:
+	int	m_iTextureID;
```
sunzenshen: Maybe I'm missing some context, but I don't see a place where the `CNeoImageButton` constructors actually set this value to -1.  Maybe logically it works out okay on the code paths, but it might be simple enough to set a {-1} initialization here just in case.

AdamTadeusz: The assumption that SetButtonTexture or SetButtonTextureID is always ran after the constructor is probably one we shouldn't make you're right

src/game/client/neo/game_controls/neo_classmenu.cpp
AdamTadeusz: good catch, reminder to not open PRs at 10 in the evening

sunzenshen (approved):
LGTM: Loaded up a few clients and verified that the correct models are displayed after selection. (I figure weapons not being selected correctly will be a bit more obvious to detect during playtests.)

https://github.com/user-attachments/assets/7de86d75-18c5-40a8-a0ca-a56f3f66a569

https://github.com/user-attachments/assets/e2a61778-ec2e-4c07-a5ae-45fd8124c387

DESTROYGIRL (approved):

PR #1392, opened 2025-09-22
Don't draw crosshair for spectators when hip fire crosshair is enabled

AdamTadeusz:


Rainyan (approved):
LGTM, and thanks for refactoring to just 1 player ptr, much neater now.

PR #1393, opened 2025-09-22
Draw dynamic crosshair inaccuracy for the balc

AdamTadeusz:


AdamTadeusz:
this is for when you have both the hip fire and dynamic crosshair enabled, you can find it at the very bottom of the crosshair section of the settings

Rainyan (approved):

PR #1394, opened 2025-09-22
Fix floating guns around players

### LINKED ISSUE #1274: Floating weapons
### Build Info

v17-alpha

### Description

Second-hand reporting this bug from Discord:

Sometimes when the Assault class dies during the switch from an empty primary to the secondary weapon, the pistol gets stuck floating in spectator mode and you can fly around as a pistol.

Image

### To Reproduce

* Be assault
* Die during switch from empty primary to secondary pistol

### Expected behavior

No spooky floating guns

### Actual behavior

Spooky floating guns

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

11

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

Occurred during streamed gameplay of v17 build for: https://www.twitch.tv/senkaku

Rainyan:
## Description
There was an issue where players would appear to have weapons like pistols or grenades floating around them, as reported in #1274 For this particular report with the demo replay, this turned out to be a case where we were rendering the weapons of the player spectating the living player, and for whatever reason that spectator still had items in their inventory, and those items were then rendered hovering/flying around the world at the spectator's position.

This PR addresses the bug by preventing dead players from "bumping" weapons, and should they still for whatever reason have a weapon, prevents that weapon from being drawn in the world.

## Steps to reproduce
* Download the demo file from the issue https://github.com/NeotokyoRebuild/neo/issues/1274#issuecomment-3193836000
* Checkout to git tag `v17.0-alpha` to have a build compatible with the demo
* Load the demo with `demoui`, or hotkey Shift+F2
* Jump to tick `20000`
* Watch the demo until tick `33000` or so

### What happens
* Floating guns are seen

### What should happen with this patch applied
* Floating guns are not seen
* If you want to verify this, note that you'll have to cherry-pick the patch on top of v17 to be compatible with the demo file, most likely!

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1274

### DIFF
--- a/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
+++ b/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
@@ -1141,7 +1141,10 @@ bool CNEOBaseCombatWeapon::ShouldDraw(void)
 
 	// No supernatural gunowners allowed here
 	if (!pOwner->IsAlive())
+	{
+		Assert(false);
 		return false;
+	}
 
 	C_BasePlayer* pLocalPlayer = C_BasePlayer::GetLocalPlayer();
 


src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
```diff
@@ -1139,6 +1139,10 @@ bool CNEOBaseCombatWeapon::ShouldDraw(void)
 	if (!pOwner)
 		return true;
 
+	// No supernatural gunowners allowed here
+	if (!pOwner->IsAlive())
+		return false;
```
AdamTadeusz, line 1146: Would an assert here be a good idea?

Rainyan: Seems like a good idea, I've no idea how commonplace this is but it's probably never what we want to happen so worth asserting.

AdamTadeusz (approved):

PR #1395, opened 2025-09-23
Display takeover player name in scoreboard/feed

sunzenshen:
## Description
When a spectator takes over another player, display their name with the format of: "target_name (spectator_name)"
Also rearranges the scoreboard such that it looks like a takeover player is taking the place of the possessed player/bot on the scoreboard.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/client/neo/ui/neo_scoreboard.cpp
+++ b/src/game/client/neo/ui/neo_scoreboard.cpp
@@ -829,19 +829,8 @@ void CNEOScoreBoard::UpdatePlayerAvatar( int playerIndex, KeyValues *kv )
 	// Update their avatar
 	if (UpdateAvatars() && steamapicontext->SteamFriends() && steamapicontext->SteamUtils() )
 	{
-		int playerIndexForAvatar = playerIndex;
-
-		C_NEO_Player *pPlayer = ToNEOPlayer(UTIL_PlayerByIndex(playerIndex));
-		C_NEO_Player *pImpersonator = pPlayer ? ToNEOPlayer(pPlayer->m_hSpectatorTakeoverPlayerImpersonatingMe.Get()) : nullptr;
-
-		if (pImpersonator)
-		{
-			// Show the avatar of the impersonator
-			playerIndexForAvatar = pImpersonator->entindex();
-		}
-
 		player_info_t pi;
-		if ( engine->GetPlayerInfo( playerIndexForAvatar, &pi ) )
+		if ( engine->GetPlayerInfo( playerIndex, &pi ) )
 		{
 			if ( pi.friendsID )
 			{


sunzenshen:
With this iteration that modifies the scoreboard and killfeed, this is what the scoreboard looks like:

image

These were my settings for setting an excess amount of characters:

image

Note that with this current iteration of the PR, I am displaying the clan tag in the former-spectator's name, but not in the taken-over bot name.  Let me know of your opinions on the formatting.

This is what the killfeed looks like with takeover name:

image

Note that the clan tag is not displayed.  I believe this is similar to the clan tag not being displayed with normal names.

This is what ghost cap looks like:

image

I guess the ghost cap display was not part of the killfeed name code, but I'd like to get more feedback on this approach of individually formatting the takeover name per UI element before tracking down that ghost cap name code.

(This ghost cap credit without the bot takeover may be unintuitive because the former-spectator will not get credit for being alive during the ghost cap.)

Same for self `kill`:
image
(I guess this one actually might be okay, since in this case, the death penalty will apply to the spectator, so this result of this `kill` result is actually intuitive)

Same for if someone else kills you while you are taking over another player:

image

(Likewise, since currently deaths are NOT transferred to the taken over player, maybe this result is actually okay too.)

sunzenshen:
Based on feedback from Discord, rearranged the scoreboard such that it looks like you are assuming the identity of the taken over player:

https://github.com/user-attachments/assets/3afd76fe-47b6-4f62-a76e-33dcd231beee

Under the hood, your scoreboard entry imitates the player you are hijacking, while that hijacked player's entry imitates your stats.

Known name change coverage gaps:
- squad team name list (a.k.a stars)
- spectator focus on a player

The reason I have omitted changing those names is because I think it's debatable whether such replacements are more useful than confusing.  The argument for changing the scoreboard and killfeed names is to highlight how scores are transferred from hijacker to hijacked.  (To recap earlier discussion, this is a balancing decision to mitigate the scenario of a high skill player using bots to farm experience through multiple lives in the same round, such that they consolidate experience into a single player to purchase high cost loadouts early in a match.)

It may actually be confusing to override the squad team names, because then it adds noise with picking out a teammate's name in the heat of action.
For spectator focus on a player, it's potentially confusing for a competition announcer to parse out the name of a player they are watching and commentating on, especially when the player has a name long enough to hit the max name length edge case.

If there is a desire for name overrides for those cases, I suggest implementing them in a separate PR after discussion.

sunzenshen:
> What does "SAT" mean? Or rather what does it stand for, if anything

SATellite

src/game/client/neo/ui/neo_hud_deathnotice.cpp
```diff
@@ -698,8 +701,39 @@ void CNEOHud_DeathNotice::RetireExpiredDeathNotices( void )
 }
 
 //-----------------------------------------------------------------------------
-// Purpose: Server's told us that someone's died
```
AdamTadeusz, line 701: Should this old comment be moved down to around line 736?

src/game/client/game_controls/SpectatorGUI.cpp
```diff
@@ -864,7 +864,9 @@ void CSpectatorGUI::UpdatePlayerLabel()
 		V_wcsncpy(playerText, L"Unable to find #Spec_PlayerItem*", sizeof(playerText));
 		memset(playerName, 0x0, sizeof(playerName));
 
-		g_pVGuiLocalize->ConvertANSIToUnicode(UTIL_SafeName(gr->GetPlayerName(playernum)), playerName, sizeof(playerName));
+		C_NEO_Player* pNeoPlayer = ToNEOPlayer(UTIL_PlayerByIndex(playernum));
+		const char* pPlayerDisplayName = pNeoPlayer ? pNeoPlayer->GetPlayerNameWithTakeoverContext(playernum) : gr->GetPlayerName(playernum);
+		g_pVGuiLocalize->ConvertANSIToUnicode(UTIL_SafeName(pPlayerDisplayName), playerName, sizeof(playerName));
```
sunzenshen, line 870: Correlates to spectator name:

image

src/game/client/neo/ui/neo_hud_deathnotice.cpp
```diff
@@ -810,9 +777,16 @@ void CNEOHud_DeathNotice::AddPlayerDeath(IGameEvent* event)
...
+	C_NEO_Player* pKiller = ToNEOPlayer(UTIL_PlayerByIndex(killer));
+	C_NEO_Player* pVictim = ToNEOPlayer(UTIL_PlayerByIndex(victim));
+	C_NEO_Player* pAssist = ToNEOPlayer(UTIL_PlayerByIndex(assist));
+
+	if (pKiller)
+		killer_name = pKiller->GetPlayerNameWithTakeoverContext(killer);
+	if (pVictim)
+		victim_name = pVictim->GetPlayerNameWithTakeoverContext(victim);
+	if (pAssist)
+		assists_name = pAssist->GetPlayerNameWithTakeoverContext(assist);
```
sunzenshen, line 789: Correlates to kill feed:

image

src/game/client/neo/ui/neo_hud_friendly_marker.cpp
```diff
@@ -175,7 +175,7 @@ void CNEOHud_FriendlyMarker::DrawPlayer(Color teamColor, C_NEO_Player *player, c
 			};
 
 			// Draw player's name and health
-			auto playerName = player->GetNeoPlayerName();
+			auto playerName = player->GetPlayerNameWithTakeoverContext(player->entindex());
```
sunzenshen: Correlates to friendly marker:

image

src/game/client/neo/ui/neo_hud_round_state.cpp
```diff
@@ -722,13 +722,35 @@ int CNEOHud_RoundState::DrawPlayerRow(int playerIndex, const int yOffset, bool s
...
 	{
-		V_snprintf(squadMateText, SQUAD_MATE_TEXT_LENGTH, "%s  [%s]  DEAD", g_PR->GetPlayerName(playerIndex), squadMateClass);
+		auto* pImpersonator = pNeoPlayer ? pNeoPlayer->m_hSpectatorTakeoverPlayerImpersonatingMe.Get() : nullptr;
+
+		const char* pPlayerDisplayName = pImpersonator ?
+		    pImpersonator->GetPlayerName()
+			: g_PR->GetPlayerName(playerIndex);
+
+		const char* displayClass = pImpersonator ? GetNeoClassName(pImpersonator->m_iClassAtTimeOfDeath) : squadMateClass;
+		V_snprintf(squadMateText, SQUAD_MATE_TEXT_LENGTH, "%s  [%s]  DEAD", pPlayerDisplayName, displayClass);
```
sunzenshen, line 756: Correlates to squad list:

Screenshot 2025-10-14 234305

src/game/client/neo/ui/neo_scoreboard.cpp
```diff
@@ -874,8 +829,19 @@ void CNEOScoreBoard::UpdatePlayerAvatar( int playerIndex, KeyValues *kv )
...
+
+		if (pImpersonator)
+		{
+			// Show the avatar of the impersonator
+			playerIndexForAvatar = pImpersonator->entindex();
+		}
+
 		player_info_t pi;
-		if ( engine->GetPlayerInfo( playerIndex, &pi ) )
+		if ( engine->GetPlayerInfo( playerIndexForAvatar, &pi ) )
```
sunzenshen: Correlates to scoreboard:

Screenshot 2025-10-14 215348

This version of the scoreboard is intentionally busy to show where are all the places we could make changes. Let me know for example if you want me to display only one avatar per impersonator and where should the avatar be.

src/game/server/neo/neo_player.cpp
```diff
@@ -1907,6 +1909,11 @@ void CNEO_Player::AddPoints(int score, bool bAllowNegativeScore, bool bIgnorePla
 
 void CNEO_Player::Event_Killed( const CTakeDamageInfo &info )
 {
+	if (m_hSpectatorTakeoverPlayerTarget == nullptr)
+	{
+		m_iClassAtTimeOfDeath = GetClass();
```
sunzenshen: This is used for both the squad list and the scoreboard:
(Otherwise when the spectator changes class under the hood to match the impersonated player, that class choice is lost.)

Screenshot 2025-10-14 234305

Screenshot 2025-10-14 215348

src/game/client/neo/ui/neo_hud_deathnotice.cpp
AdamTadeusz: Since this saves the killers name when the event is fired, it will even retain the name of the player controlling the bot after that player relinquishes control due to death for example, very cool 👍

AdamTadeusz (dismissed):

DESTROYGIRL:
Conflicts

src/game/server/neo/neo_player.cpp
```diff
@@ -3692,6 +3702,20 @@ void CNEO_Player::SpectatorTakeoverPlayerRevert(bool bHardReset)
...
 			pPlayerTakenOver->ChangeTeam(TEAM_SPECTATOR);
 		}
+
+		// Reset spectator's class to their original selection
+		switch (m_iClassAtTimeOfDeath)
+		{
+		case NEO_CLASS_RECON:
+		case NEO_CLASS_ASSAULT:
+		case NEO_CLASS_SUPPORT:
+			m_iNeoClass = m_iClassAtTimeOfDeath;
```
sunzenshen: Pardon for slipping in this change, but now that we track m_iClassAtTimeOfDeath for various display reasons, we can resolve a playtester complaint about being disoriented with taking over a bot as spectator, but then their class is different from their usual preference when the round resets.

Here's a truncated demo for restoring an original class selection for Assault, Recon, and finally Support:

https://github.com/user-attachments/assets/59013c35-bf28-47e7-b09c-e1c984590d62

Though I suppose that what the default case actually highlights is that perhaps we should store the class at time of spectator takeover, not the class at time of death.  (see: https://github.com/NeotokyoRebuild/neo/pull/1395#discussion_r2468220115)

src/game/server/neo/neo_player.cpp
```diff
@@ -3550,6 +3555,7 @@ void CNEO_Player::SpectatorTryReplacePlayer(CNEO_Player* pNeoPlayerToReplace)
 	}
 
 	m_bSpectatorTakeoverPlayerPending = true;
+	m_iClassAtTimeOfDeath = GetClass();
```
sunzenshen: > Though I suppose that what this actually highlights is that perhaps we should store the class at time of spectator takeover, not the class at time of death. https://github.com/NeotokyoRebuild/neo/pull/1395#discussion_r2468095801

That is, update the original class here instead of on death.  I think this is cleaner than updating on death (see: https://github.com/NeotokyoRebuild/neo/pull/1395/commits/26d5fe2b945fc98e98894edcc949bb83136f4032), as by the time a spectator has chosen a bot to take over, the class state of the deceased player should have stabilized at that point. 

(e.g. JGR is already respawned, and a former VIP spectator probably doesn't have time to take over a bot as the round is already over.)

src/game/server/neo/neo_player.cpp
sunzenshen: (Also feel free to suggest a different name than m_iClassAtTimeOfDeath.  I kept it the same to make the diff comparison easier to read, but open to suggestions.)

AdamTadeusz (approved):

sunzenshen:
After fixing the merge conflict, performed a quick sanity test that the player UI elements are still working as intended:

image

^
(break: different round between the two screenshots)
v

Screenshot 2025-12-04 223401

That said, I'm not 100% sure if the non-host not having a Steam avatar is a quirk of host vs non-host (or of Steam logins), or if that's some kind of issue with displaying the Steam avatar?

For the record I'm still using the -insecure multiple game client bat script:

```bat
hl2_win64.exe -game neo -insecure -allowdebug -sw -dev -multirun +sv_lan 1
```

src/game/client/neo/c_neo_player.h
```diff
@@ -175,14 +175,19 @@ class C_NEO_Player : public C_HL2MP_Player
 	virtual void TeamChange(int iNewTeam) override;
 
 	// Spectator takeover
+	bool m_bCopyOverTakeoverPlayerDetails{ false };
+	CNetworkHandle(C_NEO_Player, m_hSpectatorTakeoverPlayerTarget);
+	CNetworkHandle(C_NEO_Player, m_hSpectatorTakeoverPlayerImpersonatingMe);
 	void CSpectatorTakeoverPlayerUpdateOnDataChanged();
 	void CSpectatorTakeoverPlayerUpdate(C_NEO_Player* pPlayerTakeoverTarget);
-	CHandle<C_NEO_Player> m_hSpectatorTakeoverTarget;
```
sunzenshen, line 180: The last merge conflict was around here, where the the above CHandle was not yet removed in upstream/master.

src/game/client/neo/ui/neo_hud_round_state.cpp
```diff
@@ -722,16 +722,38 @@ int CNEOHud_RoundState::DrawPlayerRow(int playerIndex, const int yOffset, bool s
...
+			if (pTakeoverTarget)
+			{
+				displayRankName = GetRankName(g_PR->GetXP(pTakeoverTarget->entindex()), true);
+			}
+		}
+
 		const int healthMode = cl_neo_hud_health_mode.GetInt();
 		char playerHealth[7]; // 4 digits + 2 letters
 		V_snprintf(playerHealth, sizeof(playerHealth), healthMode ? "%dhp" : "%d%%", g_PR->GetDisplayedHealth(playerIndex, healthMode));
-		V_snprintf(squadMateText, SQUAD_MATE_TEXT_LENGTH, "%s %s  [%s]  %s", g_PR->GetPlayerName(playerIndex), squadMateRankName, squadMateClass, playerHealth);
```
sunzenshen, line 730: The second merge conflict was a result of the updated health bar display conflicting with these changes, so as part of the rebase I reinstated these player display name tweaks.

src/game/server/neo/neo_player.cpp
DESTROYGIRL: I'd do `m_iClassBeforeTakeover`

DESTROYGIRL (dismissed):
Seems to work fine, and is easy to understand for me. I personally would not copy the profile picture over but we are far past the bikesheddery at this point
Also see my comment about `m_iClassAtTimeOfDeath` naming suggestion

src/game/client/neo/ui/neo_scoreboard.cpp
```diff
@@ -829,19 +829,8 @@ void CNEOScoreBoard::UpdatePlayerAvatar( int playerIndex, KeyValues *kv )
...
-
-		if (pImpersonator)
-		{
-			// Show the avatar of the impersonator
-			playerIndexForAvatar = pImpersonator->entindex();
-		}
-
 		player_info_t pi;
-		if ( engine->GetPlayerInfo( playerIndexForAvatar, &pi ) )
+		if ( engine->GetPlayerInfo( playerIndex, &pi ) )
```
sunzenshen, line 833: The screenshot with the non-host duplicate player not seeing avatars was a little disconcerting to me, so I decided it was safer to revert this avatar duplication change.  The avatar will now stay with your still deceased scoreboard entry, though both the bot you are controlling and your entry are still highlighted:

image

DESTROYGIRL (approved):

Issue #1381, opened 2025-09-21
Taking control of Juggernaut should not fail studio assertions


Rainyan:
### Build Info

ed5a266ff2f6f6805

### Description

In JGR mode maps, if you take control of the JGR whilst you also have a primary weapon out, the studiomdl code throws on several assertions (and likely the model sequence numbers go out of whack, also).

### To Reproduce

* Load `map ntre_terminal_jgr`
* Join a team
* Wait for match to start
* Choose a weapon and spawn in the map
* Go to the JGR
* Make sure your primary weapon is equipped
* Take control of the JGR
  * Recommended to use `+use` via console instead of holding a key, because it's easy to accidentally dismiss the debugger break otherwise

### Expected behavior

No assertions throw

### Actual behavior

Two studiomdl assertions fail sequentially:

```
studio.cpp (934) : Assertion Failed: ( i >= 0 && i < GetNumSeq() ) || ( i == 1 && GetNumSeq() <= 1 )
studio.h (1783) : Assertion Failed: 0
```

along with the side effects of the sequence number being incorrect (could lead to animation bugs, although I couldn't immediately spot any).

Full assertion details:

```
studio.cpp (934) : Assertion Failed: ( i >= 0 && i < GetNumSeq() ) || ( i == 1 && GetNumSeq() <= 1 )

`i` was `125`, `CStudioHdr::GetNumSeq` returned `21`

Call stack:
>	server.dll!CStudioHdr::pSeqdesc(int i) Line 934	C++
 	server.dll!CBoneSetup::AccumulatePose(Vector * pos, Quaternion * q, int sequence, float cycle, float flWeight, float flTime, CIKContext * pIKContext) Line 2421	C++
 	server.dll!IBoneSetup::AccumulatePose(Vector * pos, Quaternion * q, int sequence, float cycle, float flWeight, float flTime, CIKContext * pIKContext) Line 2312	C++
 	server.dll!CBaseAnimatingOverlay::GetSkeleton(CStudioHdr * pStudioHdr, Vector * pos, Quaternion * q, int boneMask) Line 478	C++
 	server.dll!CHL2MP_Player::SetupBones(matrix3x4_t * pBoneToWorld, int boneMask) Line 1890	C++
 	server.dll!CBaseAnimating::GetBoneCache() Line 2789	C++
 	server.dll!CBaseAnimating::GetBoneTransform(int iBone, matrix3x4_t & pBoneToWorld) Line 1568	C++
[...]

PR #1398, opened 2025-09-25
JGR Tweaks + squad friendly markers

DESTROYGIRL:
## Description
- Fixes not being able to spawn in JGR
- Spawns any already dead teammates instantly when the JGR is picked up by a team member
- Gives the JGR & VIP a unique friendly marker to distinguish them, also gives non-squad members their correct marker back like in OG
- Always show the JGR position on the compass, regardless of if it's being controlled or not
- Fix looping boot sound continuing into next round after reset
- Avoid the first two asserts described in #1381 but not all of them

neoAssets >>> https://github.com/NeotokyoRebuild/neoAssets/pull/72

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #1381

### DIFF
--- a/src/game/shared/neo/neo_gamerules.cpp
+++ b/src/game/shared/neo/neo_gamerules.cpp
@@ -1371,9 +1371,9 @@ void CNEORules::Think(void)
 
 		m_bJuggernautItemExists = true;
 	}
-	else
+	else if (m_pJuggernautPlayer)
 	{
-		if (m_pJuggernautPlayer && (m_pJuggernautPlayer->GetAbsOrigin().IsValid()))
+		if (m_pJuggernautPlayer->GetAbsOrigin().IsValid())
 		{
 			m_vecJuggernautMarkerPos = m_pJuggernautPlayer->WorldSpaceCenter();
 		}
@@ -1382,6 +1382,10 @@ void CNEORules::Think(void)
 			Assert(false);
 		}
 
+		m_bJuggernautItemExists = false;
+	}
+	else
+	{
 		m_bJuggernautItemExists = false;
 	}
 
@@ -1929,7 +1933,10 @@ void CNEORules::SpawnTheJuggernaut(const Vector* origin)
 		}
 		else if (auto* jgr = dynamic_cast<CNEO_Juggernaut*>(pEnt))
 		{
-			m_pJuggernautItem = jgr;
+			if (!jgr->IsMarkedForDeletion())
+			{
+				m_pJuggernautItem = jgr;
+			}
 		}
 
 		pEnt = gEntList.NextEnt(pEnt);
--- a/src/game/shared/neo/neo_juggernaut.cpp
+++ b/src/game/shared/neo/neo_juggernaut.cpp
@@ -7,6 +7,9 @@
 #include "model_types.h"
 #endif
 
+// memdbgon must be the last include file in a .cpp file!!!
+#include "tier0/memdbgon.h"
+
 #define USE_DURATION 5.0f
 #define USE_DISTANCE_SQUARED 22500.0f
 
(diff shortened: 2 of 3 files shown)


src/game/client/neo/ui/neo_hud_compass.cpp
```diff
@@ -97,8 +97,8 @@ void CNEOHud_Compass::UpdateStateForNeoHudElementDraw()
 	auto pLocalPlayer = C_NEO_Player::GetLocalNEOPlayer();
 	Assert(pLocalPlayer);
 
-	// Point the objective arrow to the ghost, if it exists
-	if (NEORules()->GhostExists() || NEORules()->JuggernautItemExists())
+	// Point the objective arrow to the relevant objective, if it exists
+	if (NEORules()->GhostExists() || NEORules()->GetJuggernautMarkerPos() != vec3_origin)
```
Rainyan, line 101: This is good enough for me and I'm not asking for changes to this PR regarding it, but in the future we might wanna figure out some way to signal this that doesn't overlap with `vec3_origin`, i.e. xyz position 0,0,0. Because the objective could plausibly be located at position 0,0,0 especially if it was mapper-placed.

IIRC the ghost logic written by me some years earlier is also guilty of this...

src/game/shared/neo/neo_juggernaut.h
Rainyan, line 1: This file is missing the `#pragma once` directive from its first line, and this can cause compile errors for circular includes.

src/game/shared/neo/neo_juggernaut.cpp
Rainyan, line 1: It would be preferrable to add the
```cpp
#include "tier0/memdbgon.h"
```
memory debug include as the last in the cpp file.

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -1369,11 +1369,19 @@ void CNEORules::Think(void)
...
 	{
+		if (m_pJuggernautPlayer && (m_pJuggernautPlayer->GetAbsOrigin().IsValid()))
+		{
+			m_vecJuggernautMarkerPos = m_pJuggernautPlayer->WorldSpaceCenter();
+		}
+		else
+		{
+			Assert(false);
+		}
+
```
Rainyan, line 1384: Maybe it'd be better to assert the Vector validity specifically, so:
```cpp
else if (m_pJuggernautItem)
{
	if (m_pJuggernautItem->GetAbsOrigin().IsValid())
	{
		m_vecJuggernautMarkerPos = m_pJuggernautItem->WorldSpaceCenter();
	}
	else
	{
		Assert(false);
	}
```
since I assume we don't really want to assert for the pointer value here.

src/game/shared/neo/neo_gamerules.cpp
Rainyan, line 1: At around `@@ -1929,12 +1932,22 @@ void CNEORules::SpawnTheJuggernaut(const Vector* origin)`, it would be good to check:

```diff
@@ -1929,12 +1932,22 @@ void CNEORules::SpawnTheJuggernaut(const Vector* origin)
                }
                else if (auto* jgr = dynamic_cast<CNEO_Juggernaut*>(pEnt))
                {
-                       m_pJuggernautItem = jgr;
+                       if (!jgr->IsMarkedForDeletion())
+                       {
+                               m_pJuggernautItem = jgr;
+                       }
                }

                pEnt = gEntList.NextEnt(pEnt);
```
because else there is a chance to cache into `m_pJuggernautItem` a dangling pointer that's about to be deleted soon. Which would then throw a memory violation when dereferenced in `CNEORules::Think` at a later time after the object pointed to was already gone.

Rainyan (changes requested):

src/game/client/neo/ui/neo_hud_compass.cpp
Rainyan: Nevermind, outdated comment

Rainyan (approved):

sunzenshen (approved):

PR #1407, opened 2025-09-29
Introduce air friction as a replacement for the hard speed cap to prevent ghost hopping

Agiel:
## Description
This is intended to replace the current hard speed cap that comes with the neo_ghost_bhopping cvar. Air friction is applied when going faster than the current max walking speed (based on class and active weapon), and scales linearly with how much faster the player is going.

The friction can be controlled with the following new cvars:
* **sv_neo_airfriction** _(Default: 0)_: The amount of air friction applied to everyone.
* **sv_neo_airfriction_objective** _(Default: 0.5)_: The amount of air friction applied to the ghost carrier, VIP and juggernaut.
* **sv_neo_airstopspeed** _(Default: 0)_: The minimum stopping speed when in the air. Doesn't seem necessary with the default friction value of 0.5, but it's added for parity with ground friction.

For testing purposes, a cheat protected cvar has been added to allow jump buffering.
* **sv_neo_jump_buffer** _(Default: 0)_

For reference, the max speeds set by the nt-anti-ghosthop plugin are:
#define MAX_SPEED_RECON 255.59
#define MAX_SPEED_ASSAULT 204.47
#define MAX_SPEED_SUPPORT 204.47

The default friction value of 0.5 gives us speeds in the same ballpark (tested with jump buffer enabled). Further testing will let us know if the value needs to be tweaked further.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes
- related

### DIFF
--- a/src/game/client/neo/c_neo_player.cpp
+++ b/src/game/client/neo/c_neo_player.cpp
@@ -833,7 +833,6 @@ void C_NEO_Player::PlayStepSound( Vector &vecOrigin,
 	BaseClass::PlayStepSound(vecOrigin, psurface, fvol, force);
 }
 
-extern ConVar neo_ghost_bhopping;
 void C_NEO_Player::CalculateSpeed(void)
 {
 	float speed = GetNormSpeed();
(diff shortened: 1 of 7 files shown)


DESTROYGIRL:
> For testing purposes, a cheat protected cvar has been added to allow jump buffering.
image

I know where this is going, skip to the day long bikeshedding debate

sunzenshen (approved):

PR #1408, opened 2025-09-30
Turn off walk toggle when sprinting

### LINKED ISSUE #1305: Sprinting should automatically toggle off walking if it was on before
### Build Info

20250427_55304de

### Description

In OGNT, if walking is toggled on and you start sprinting, walking is automatically toggled off when you cancel your sprint. In this build, walking remains toggled on.

### To Reproduce

1. Enable toggled walking in Options.
2. Enter a game/server and initiate walking.
3. Sprint in any direction and then cancel your sprint.
4. When you've cancelled your sprinting, you will return to a walking state.

### Expected behavior

Walking should be automatically toggled off when the player starts sprinting. This means that when sprinting is cancelled the player should return to a state where the movement speed is at a standard pace (neither walking nor sprinting).

### Actual behavior

_No response_

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

sunzenshen:
## Description
Walking is automatically toggled off when the player starts sprinting.
This means that when sprinting is cancelled the player should return to a state where the movement speed is at a standard pace (neither walking nor sprinting).

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1305

### DIFF
--- a/src/game/client/in_main.cpp
+++ b/src/game/client/in_main.cpp
@@ -1579,7 +1579,8 @@ int CInput::GetButtonBits( int bResetState )
 	CalcButtonBits( bits, IN_VISION, s_ClearInputState, &in_vision, bResetState);
 	if (KeyState(&in_speed))
 	{
-		bits &= ~(IN_WALK);
+		// Cancel walk toggle if sprinting
+		KeyUp(&in_walk, nullptr);
 	}
 #endif
 


DESTROYGIRL (approved):
Side effect of this is that people who use hold walk rather than toggle will have to press the walk key again after leaving sprint to return to walking state.
It's not a huge issue I guess but I think the expected behaviour would be to return to the walking state if its still held down.

PR #1412, opened 2025-10-04
IsOfficialMap

DESTROYGIRL:
## Description
This changes nothing user facing but makes the function actually useful for future stuff with our maps...
The fog controller hack was not needed

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/shared/hl2mp/hl2mp_gamerules.cpp
+++ b/src/game/shared/hl2mp/hl2mp_gamerules.cpp
@@ -891,38 +891,6 @@ bool CHL2MPRules::IsOfficialMap( void )
 { 
 	static const char *s_OfficialMaps[] =
 	{
-#ifdef NEO
-		"background_alley",
-		"background_door",
-		"ntre_ballistrade_ctg",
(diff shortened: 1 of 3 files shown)


src/game/server/fogcontroller.cpp
```diff
@@ -371,12 +371,13 @@ void CFogSystem::LevelInitPostEntity( void )
 		pFogController = static_cast<CFogController*>( gEntList.FindEntityByClassname( pFogController, "env_fog_controller" ) );
 		if ( pFogController )
 		{
+#ifndef NEO
```
sunzenshen, line 374: Should this be #ifdef because this is specific to NT?  Not sure if you meant "if not defined" here, to disable this section of code.

DESTROYGIRL: This was a hack for TF2 because it wouldve required them to recompile every single map probably, so I've chosen to remove it since we can handle that ourselves in the maps

sunzenshen (approved):

PR #1414, opened 2025-10-05
Detpacks can be destroyed with explosives

sunzenshen:
## Description
Add server option to allow detpacks to be destroyed by other explosives.
Disabled by default to follow original game parity behavior. (Cannot be damaged, and must be triggered by player.)

## Toolchain
- Windows MSVC VS2022


sunzenshen:
Demo of behavior when optional convar is enabled:

https://github.com/user-attachments/assets/15335ff6-bdee-4e36-8978-bcbef17f70d9

AdamTadeusz:
https://github.com/user-attachments/assets/1d89c766-5eec-4fd5-a232-4b9c49ce8ecf

Regardless of how the detpack is set off by another player I imagine you would want the player credited for the kill to be whoever set off the detpack

PR #1421, opened 2025-10-07
Fix bad assert

Rainyan:
## Description
Fix a case where a debug assert at client frame draw would fail before a weapon's dead owner was cleared from the weapon at the following tick. This assert would then keep spamming several times, depending on frame rate.

The faulty assert was originally introduced as part of #1394

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
+++ b/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
@@ -1142,7 +1142,19 @@ bool CNEOBaseCombatWeapon::ShouldDraw(void)
 	// No supernatural gunowners allowed here
 	if (!pOwner->IsAlive())
 	{
-		Assert(false);
+#if DEBUG
+		static int deathTicks[MAX_PLAYERS]{};
+		if (!*deathTicks)
+			for (int i = 0; i < ARRAYSIZE(deathTicks); ++i) deathTicks[i] = -1;
+		const int ownerIdx = pOwner->entindex();
+		const int deathTick = deathTicks[ownerIdx];
+		if (deathTick != -1)
+		{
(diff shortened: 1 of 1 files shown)


src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
```diff
@@ -1142,7 +1142,19 @@ bool CNEOBaseCombatWeapon::ShouldDraw(void)
 	// No supernatural gunowners allowed here
 	if (!pOwner->IsAlive())
 	{
-		Assert(false);
+#if DEBUG
+		static int deathTicks[MAX_PLAYERS_ARRAY_SAFE]{};
+		if (!*deathTicks)
+			for (int i = 0; i < ARRAYSIZE(deathTicks); ++i) deathTicks[i] = -1;
+		const int ownerIdx = pOwner->entindex();
```
sunzenshen, line 1149: I suppose I'm a little unfamiliar with the expected range of entindex, but according to this wiki, there could be up to 4096 entities:

https://developer.valvesoftware.com/wiki/Entity_index

Since this is used to index into deathTicks[MAX_PLAYERS_ARRAY_SAFE], where MAX_PLAYERS_ARRAY_SAFE is =34, would it make sense to add another AssertMsg regarding the range of deathTicks?

Rainyan: Since the owner of a `CBaseCombatWeapon` must be `CBaseCombatCharacter`, and all players are guaranteed to be indexed in the `1...MAX_PLAYERS` range, as long as we can assume all our base combat characters are players, this index can't go out of range. The only way I could see this breaking is if we added combat characters who are not players in the future. But I've added an extra assert, just in case.

Issue #1418, opened 2025-10-07
Game shouldn't crash when playing a deathmatch map with bots


AdamTadeusz:
Caused by bots joining TEAM_UNASSIGNED which messes up the actions somehow.

To reproduce:

set sv_neo_gamemode_enforcement to 1
launch a map with some bots (I picked ntre_bullet_tdm specifically)

What should happen:

The game doesn't crash

What happens:

The game crashes

To fix:

Remove the IsTeamplay() check in CNEOBotManager::MaintainBotQuota() (or better yet the entirety of line 346)

Rainyan:
I can repro this with the steps, and it produces the following call stack for the exception:

```
>	server.dll!Action<CNEOBot>::IsOutOfScope() Line 1006	C++
 	server.dll!Action<CNEOBot>::InvokeUpdate(CNEOBot * me, Behavior<CNEOBot> * behavior, float interval) Line 1357	C++
 	server.dll!Behavior<CNEOBot>::Update(CNEOBot * me, float interval) Line 258	C++
 	server.dll!CNEOBotIntention::Update() Line 2528	C++
 	server.dll!INextBot::Update() Line 133	C++
 	server.dll!NextBotPlayer<CNEO_Player>::Update() Line 976	C++
 	server.dll!CNEOBot::Update() Line 817	C++
 	server.dll!NextBotPlayer<CNEO_Player>::PhysicsSimulate() Line 743	C++
 	server.dll!CNEOBot::PhysicsSimulate() Line 683	C++
 	server.dll!Physics_SimulateEntity(CBaseEntity * pEntity) Line 2019	C++
 	server.dll!Physics_RunThinkFunctions(bool simulating) Line 2075	C++
 	server.dll!CServerGameDLL::GameFrame(bool simulating) Line 1281	C++
 	[External Code]	
```

Incidentally, this call stack turns out to be identical to the one in issue comments of #1355 

Maybe these two issues are in fact symptoms of the same underlying problem? Although I'm not quite sure what that is.

Rainyan:
I could also raise the exception with call stack (same as the alternative crash at #1355):

```
 	0000000000000000()	Unknown
>	server.dll!Action<CNEOBot>::ApplyResult(CNEOBot * me, Behavior<CNEOBot> * behavior, ActionResult<CNEOBot> result) Line 1569	C++
 	server.dll!Behavior<CNEOBot>::Update(CNEOBot * me, float interval) Line 258	C++
[...]

Issue #1355, opened 2025-09-18
If the Juggernaut dies whilst targeted by a bot the game will crash


DESTROYGIRL:
To reproduce:

1. `map ntre_terminal_jgr`
2. Spawn some bots
3. Wait for round to start
4. Take control over the juggernaut
5. Die to an enemy bot, or suicide near one
6. Crash
Sometimes it doesn't happen but I find it to be more consistent with 5+ bots in the game

Rainyan:
Unable to reproduce the crash.

I assume the "Die to an enemy bot" part of the repro steps means dying after taking control of the Juggernaut, and not dying as a regular (non JGR) player?

AdamTadeusz:
I was able to replicate the issue, not sure what the problem is exactly though

<img width="2559" height="1439" alt="Image" src="https://github.com/user-attachments/assets/ade32679-67d3-4018-9859-12f9a04173f6" />

Rainyan:
There appears to be multiple memory violations that can be triggered if the JGR is killed by a bot while it is mid-air. You can trigger these by noclipping and letting the bots shoot you, or timing a jump/fall such that you die before you land.

Here's the two that I caught (not sure if this is a complete list):

```
Exception thrown: read access violation.
under was 0xFFFFFFFFFFFFFFBF.

Call stack:
>	server.dll!Action<CNEOBot>::IsOutOfScope() Line 1006	C++
 	server.dll!Action<CNEOBot>::InvokeUpdate(CNEOBot * me, Behavior<CNEOBot> * behavior, float interval) Line 1357	C++
 	server.dll!Behavior<CNEOBot>::Update(CNEOBot * me, float interval) Line 258	C++
 	server.dll!CNEOBotIntention::Update() Line 2528	C++
 	server.dll!INextBot::Update() Line 133	C++
 	server.dll!NextBotPlayer<CNEO_Player>::Update() Line 976	C++
 	server.dll!CNEOBot::Update() Line 817	C++
 	server.dll!NextBotPlayer<CNEO_Player>::PhysicsSimulate() Line 743	C++
 	server.dll!CNEOBot::PhysicsSimulate() Line 683	C++
 	server.dll!Physics_SimulateEntity(CBaseEntity * pEntity) Line 2019	C++
 	server.dll!Physics_RunThinkFunctions(bool simulating) Line 2075	C++
 	server.dll!CServerGameDLL::GameFrame(bool simulating) Line 1281	C++
 	[External Code]	
[...]

PR #1428, opened 2025-10-14
Reduce occurrence of JGR crashes on destruction

sunzenshen:
## Description
Attempt to fix the crash when bots are attacking a JGR player and that player is destroyed.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
@@ -236,6 +236,13 @@ EventDesiredResult< CNEOBot > CNEOBotMainAction::OnOtherKilled( CNEOBot *me, CBa
 	// make sure we forget about this guy
 	me->GetVisionInterface()->ForgetEntity( victim );
 
+	CNEO_Player* pNeoVictim = ToNEOPlayer(victim);
+	if (pNeoVictim && pNeoVictim->GetClass() == NEO_CLASS_JUGGERNAUT)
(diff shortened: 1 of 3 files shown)


src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
```diff
@@ -236,6 +236,13 @@ EventDesiredResult< CNEOBot > CNEOBotMainAction::OnOtherKilled( CNEOBot *me, CBa
 	// make sure we forget about this guy
 	me->GetVisionInterface()->ForgetEntity( victim );
 
+	CNEO_Player* pNeoVictim = ToNEOPlayer(victim);
+	if (pNeoVictim && pNeoVictim->GetClass() == NEO_CLASS_JUGGERNAUT)
+	{
+		me->GetVisionInterface()->ForgetAllKnownEntities();
+		me->GetIntentionInterface()->Reset();
```
sunzenshen, line 243: This isn't really a serious fix, but the reduction (but not elimination) of crashes with this patch job may lend credence to the theory that it's actually the dead, observing, spawn-restricted teammates of the JGR player that are triggering the exception crash. (Discussion: https://github.com/NeotokyoRebuild/neo/issues/1418 https://github.com/NeotokyoRebuild/neo/issues/1355)

Also the more JGR-side teammates are dead, the likelier it is to trigger a crash anyway, as it just increases the chance that the condition will pop up again.

PR #1430, opened 2025-10-15
Default hammer configurations

### LINKED ISSUE #1178: Hammer should be readily configured in the steam install
Since we don't have the source to hammer and the gameconfig.txt files require the absolute path to everything probably we need some scripts that set everything up and ready for compiling maps for rebuild upon installation

DESTROYGIRL:
## Description
Default configurations for the valve hammer editor (not hammer++) in our steam install
This also moves our FGD out of the assets repo and into the code repo. This is because our FGD needs to be in this bin folder to function, and it saves users having to copy it over. Additionally when making entity changes two seperate PRs had to be made (very annoying)
For unity lights.rad and detail.vbsp were also moved.

This removes the bin folder exclusion from the gitignore, PLEASE can the reviewers check if anything is missed by doing this and needs to be added to the ignore

[neoAssets](https://github.com/NeotokyoRebuild/neoAssets/pull/75)

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1178

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -680,5 +680,5 @@ FodyWeavers.xsd
 
 !/game/bin/
 !/game/bin/x64/
-!/game/bin/**/rebuild.fgd
-!/game/bin/[Gg]ame[Cc]onfig.txt
+!/game/bin/rebuild.fgd
+!/game/bin/**/[Gg]ame[Cc]onfig.txt


sunzenshen:
As I am not as familiar with Hammer these days, and there was a mention that a tutorial draft was pending on this review... I wonder if one way to review this PR is to draft the tutorial on a WYSIWYG page like in Google Docs, and then have the reviewers work through the tutorial.

.gitignore
```diff
@@ -692,6 +694,3 @@ FodyWeavers.xsd
 /mp
 
 /game/neo/cache/DownloadCache.db
-/game/neo/lights.rad
-/game/neo/rebuild.fgd
```
Rainyan, line 696: I think we should still ignore `/game/neo/rebuild.fgd`, so that devs don't accidentally re-commit it to this old location (since the `rebuild.fgd` now lives in the `bin` folder). This could happen for example if they have stale local files, and they forgot to `git clean` after switching branches back to master from some earlier revision.

So maybe something like:
```diff
 # Old /mp directory should no longer be used
 /mp

+# This is source-controlled now, but it exists in the "bin" dir, so we ignore this old location
+/game/neo/rebuild.fgd
```

.gitignore
```diff
@@ -312,13 +316,11 @@ mono_crash.*
 [Dd]ebugPublic/
 [Rr]elease/
 [Rr]eleases/
-x64/
 x86/
 [Ww][Ii][Nn]32/
 [Aa][Rr][Mm]/
 [Aa][Rr][Mm]64/
 bld/
-[Bb]in/
```
Rainyan, line 321: We probably shouldn't touch these since they're meant to exclude generic build products from dev tools. Instead we could add some exceptions at the bottom of the ignore file, using the `!` gitignore syntax. I'd propose something like this for example:

```diff
diff --git a/.gitignore b/.gitignore
index 71d643fd..5ae578d4 100644
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ out
 /game/neo/screenshots
 
 # Binary Assets
+/game/**/bin/**
 /game/neo/maps
 /game/neo/mapsrc
 /game/neo/materials
@@ -165,38 +166,15 @@ obj*/
 VTune\ Profiler\ Results/
 
 # Specific Source build products
+bin/**
 *.project
 
-client.pdb
-client.dll
-client.lib
-game_shader_dx9.pdb
-game_shader_dx9.dll
-game_shader_dx9.lib
-server.pdb
-server.dll
-server.lib
-
 vgui_controls.lib
 fgdlib.lib
 tier1.lib
 mathlib.lib
 raytrace.lib
 
-client.so
-client.so.dbg
-game_shader_dx9.so
-game_shader_dx9.so.dbg
-server.so
-server.so.dbg
-server_srv.so
-server_srv.so.dbg
-
-client.dylib
-client.dylib.dSYM/
-server.dylib
-server.dylib.dSYM/
-
 # files generated by running a mod
 config.cfg
 voice_ban.dt
@@ -316,11 +294,13 @@ mono_crash.*
 [Dd]ebugPublic/
 [Rr]elease/
 [Rr]eleases/
-x86/
+x64/**
+x86/**
 [Ww][Ii][Nn]32/
 [Aa][Rr][Mm]/
 [Aa][Rr][Mm]64/
 bld/
+[Bb]in/**
 [Oo]bj/
 [Ll]og/
 [Ll]ogs/
@@ -694,3 +674,8 @@ FodyWeavers.xsd
 /mp
 
 /game/neo/cache/DownloadCache.db
+
+!/game/bin/
+!/game/bin/x64/
+!/game/bin/**/rebuild.fgd
+!/game/bin/[Gg]ame[Cc]onfig.txt
```

I also removed the shader ignores from the file in that above example since they're somewhat redundant with the bin dir ignore. Same thing with the client/server library file ignores, although they weren't modified in this PR. But we might as well clean them up too.

.gitignore
```diff
@@ -691,7 +673,12 @@ FodyWeavers.xsd
...
+# This is source-controlled now, but it exists in the "bin" dir, so we ignore this old location
 /game/neo/rebuild.fgd
-/game/neo/detail.vbsp
+
+/game/neo/cache/DownloadCache.db
+
+!/game/bin/
+!/game/bin/x64/
+!/game/bin/**/rebuild.fgd
+!/game/bin/[Gg]ame[Cc]onfig.txt
```
Rainyan: Sorry, I typo'd the syntax here on lines 683-684. We'd instead want the exclusions to be:

```diff
diff --git a/.gitignore b/.gitignore
index 51bd5ebe..eb1317a8 100644
--- a/.gitignore
+++ b/.gitignore
@@ -680,5 +680,5 @@ FodyWeavers.xsd

 !/game/bin/
 !/game/bin/x64/
-!/game/bin/**/rebuild.fgd
-!/game/bin/[Gg]ame[Cc]onfig.txt
+!/game/bin/rebuild.fgd
+!/game/bin/**/[Gg]ame[Cc]onfig.txt
```

so the double-asterisk `/**/` token from line 683 should go with the GameConfig entry at line 684 (since we wanna match both the base "bin" dir and the "x64" dir GameConfigs).

And for the FGD, we only want to match the base "bin" dir entry, since the FGD is not supposed to exist in the "x64" dir at all.

Rainyan (changes requested):

.gitignore
DESTROYGIRL: I was questioning the fgd line but since I'm not especially familiar with the syntax I ignored it :P

Rainyan (approved):

sunzenshen (approved):

PR #1431, opened 2025-10-15
Silent footsteps in aim, revert walk speeds to parity

DESTROYGIRL:
## Description
Issues were raised how the player makes footstep sounds in aim despite being slower than walking. This resolves this issue whilst preserving the new silent walking change. Additionally, class walk speeds are reverted to parity and rather than having a standard threshold for footsteps it depends on the movement speed of each class, giving it the same behaviour without moving away from parity movement

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #1097


DESTROYGIRL:
We aren't really silencing the player just because they are walking, we've just changed the limit. If you go over that limit, by wiggle running for example you will make sounds
Also what you described is basically what was implemented, currently all players have the same walk speed of 90 but the threshold for steps is raised slightly to account for slight increases in speed but not too high
But the adverse affect of this is that it changes movement and I'm unsure if people were really aware of that side effect as there were also complaints, hence why this PR basically does the same thing that was intended without messing with movement

src/game/shared/baseplayer_shared.cpp
```diff
@@ -691,27 +691,26 @@ void CBasePlayer::UpdateStepSound( surfacedata_t *psurface, const Vector &vecOri
...
+	constexpr float SILENT_THRESHOLD_GRACE = 0.7f;
 #endif //NEO
 	// play the sound
 	// 65% volume if ducking
 	if ( GetFlags() & FL_DUCKING )
 	{
 		fvol *= 0.65;
 #ifdef NEO
-		if (neoPlayer->IsWalking() && speed <= NEO_CROUCH_WALK_SPEED * SILENT_THRESHOLD_GRACE)
+		if ((neoPlayer->IsInAim() || neoPlayer->IsWalking()) && speed <= (neoPlayer->GetCrouchSpeed() * SILENT_THRESHOLD_GRACE))
```
Agiel, line 705: Can this not just be a speed check? What if the player is using an analog input to walk slowly? And I just want to point out that changing the speeds at which players are silent is a much bigger change than changing the walking speed.

DESTROYGIRL: > What if the player is using an analog input to walk slowly?

Can probably add an else if that uses the original values

AdamTadeusz: I'm messing with this pr locally to add what you suggested Agiel but then we're basically back at square one, unless its in an else if like DESTROGIRL suggested, but then you have a disconnect where you can be silent at a speed thats above the base silent walk speed if you're holding the aim or walk button, but if you reach that same speed without those two then you make noise when walking at that same speed, which is what my original silent footsteps pr was trying to fix

Agiel:
Approved to get this over with. We can keep discussing if we want parity or not later.

PR #1432, opened 2025-10-17
Parity grenade damage and radius

AdamTadeusz:
## Description

Like Agiel said, grenade damage and damage radius was not parity, a value of 270 for both matches og values better

To test:

map nt_dimensions_test 
neo_bot_quota 0 (in nt;re)
cl_neo_hud_health_as_percentage 1 (in nt;re)
pick assault
setang 89 0 0
setpos 1600 -1600 187
setpos 1400 -1600 187
setpos 1600 -1600 187
throw a grenade below yourself, open the console, tap the up arrow twice to select the second last run command and hit enter

the player's health should be 24% in both


Masterkatze:
"Hit it Joe!"

PR #1442, opened 2025-10-25
Fixes for ghost beacons

### LINKED ISSUE #1426: Spectators can see ghost markers before the ghost carrier does
### Build Info

v19.1

### Description

title

### To Reproduce

1. Spectate a person when they are picking up the ghost

### Expected behavior

Ghost markers should pop for the spectator after the ghost boot up sequence has finished

### Actual behavior

The spectator can see the ghost markers instantly

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

### LINKED ISSUE #1441: Ghost beacon max distance is incorrect
### Build Info

b3e313f4479a49c9769739aaebced8ed5fa6b8c1

### Description

In OGNT, the maximum distance at which a ghost beacon is visible is 1800 units. This is slightly more than 45 meters. We are using this incorrect 45 meters value both in the cvar `cl_neo_ghost_view_distance`, and as a hardcoded magic value in a few places around the code base.

### To Reproduce

* In OG NT, to `map nt_tarmac_ctg;`

* Place a bot with `bot_mimic 1;` to these coordinates, shoot it once to settle it, and then use `bot_mimic 0;` so it stays put: `setpos -1930.968750 730.968750 -230.968750;`

* Use `setpos` to go to this following location with `noclip` enabled. Note that you have to offset the Z value by your model viewheight, for recon it's by -57 units (`-230.968750 - 57 == -287.96875` for the recon Z height): `setpos -131.000000 730.968750 -287.96875;`

* Then, cross the beacon max distance limit by modifying the `setpos` X value from `-131` to `-130.99` and note how you can now hear the beacon sound effect, since you're in range: `setpos -130.99 730.968750 -287.96875;`

* Then finally, calculating the `setpos` X difference to the bot's coords (Y and Z axes are equal) to verify, we get a difference of about `-1799.96875` which approaches the `-1800` limit where the beacon are cut off.

* 1800 units equals 45.72 meters. Whereas 45 meters would be 1771.65 units, so it's off by 0.72 meters from the parity value.

* For NT;RE, you could do the same test to compare, if you wanted to.

### Expected behavior

Ghost beacons distance limit is at 1800 Hammer units.

### Actual behavior

Ghost beacons distance limit is at around 1771.6535 Hammer units.

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

11

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

Rainyan:
## Description
Fix a problem with spectators seeing the spectated ghoster's beacons instantly, instead of after the delay specified by the cvar `neo_ghost_delay_secs`. For the default case, we can derive the ghost equip time from its `m_flNextPrimaryAttack`, but for the `neo_ctg_ghost_beacons_when_inactive` mode I also had to add a new time variable (`m_flPickupTime`) for tracking when the ghost was initially picked up.

Also fixes an inaccuracy for the ghost beacons maximum visible distance.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1426 
- fixes #1441

### DIFF
--- a/src/game/shared/neo/weapons/weapon_ghost.h
+++ b/src/game/shared/neo/weapons/weapon_ghost.h
@@ -51,8 +51,8 @@ public:
 		return deltaTime + timeOffset >= sv_neo_ghost_delay_secs.GetFloat();
 	}
 
-	// If there are enemies within this ghost's range, returns true and passes the distance to the nearest enemy by reference.
-	// If there are no enemies within range, the distance will not be written into.
+	// If the enemy is within this ghost's range, returns true and passes its distance by reference.
+	// If the enemy is not in range, the distance will not be written into.
 	[[nodiscard]] bool BeaconRange(CBaseEntity* enemy, float& outDistance) const;
 
 	virtual void ItemPreFrame(void) OVERRIDE;


src/game/shared/neo/neo_player_shared.h
```diff
@@ -22,6 +22,10 @@
 #include "weapon_bits.h"
 #include "neo_enums.h"
 
+// This default value is not a typo. The OGNT ghost beacon distance is 1800 Hammer units/inches, which equals a little over 45 meters.
+// Since we can represent this value exactly with floating point, it's not really a problem to store it as meters here.
+static ConVar cl_neo_ghost_view_distance("cl_neo_ghost_view_distance", "45.72", FCVAR_REPLICATED, "How far can the ghost user see players in meters.");
```
Agiel: I realize you've just moved this, but since we're touching all of this anyway we should probably fix the naming. This should be sv_ right?
Same goes for the neo_ctg_ghost_beacons_when_inactive cvar I guess.

Agiel (changes requested):
Good work!

src/game/shared/neo/neo_player_shared.h
Rainyan: Good catch! Also cleaned up some of the cl\_... cvars touched.

src/game/client/neo/ui/neo_hud_ghost_beacons.cpp
```diff
@@ -24,16 +24,16 @@
...
-ConVar neo_ghost_beacon_scale_with_distance("neo_ghost_beacon_scale_with_distance", "0", FCVAR_ARCHIVE,
+ConVar cl_neo_ghost_beacon_scale_with_distance("cl_neo_ghost_beacon_scale_with_distance", "0", FCVAR_ARCHIVE,
 	"Toggles the scaling of ghost beacons with distance.", true, 0, true, 1);
 
-ConVar neo_ghost_beacon_alpha("neo_ghost_beacon_alpha", "150", FCVAR_ARCHIVE,
+ConVar cl_neo_ghost_beacon_alpha("cl_neo_ghost_beacon_alpha", "150", FCVAR_ARCHIVE,
 	"Alpha channel transparency of HUD ghost beacons.", true, 0, true, 255);
 
-ConVar neo_ghost_delay_secs("neo_ghost_delay_secs", "3.3", FCVAR_CHEAT | FCVAR_REPLICATED,
+ConVar cl_neo_ghost_delay_secs("cl_neo_ghost_delay_secs", "3.3", FCVAR_CHEAT | FCVAR_REPLICATED,
```
Agiel: sv_ ?

Rainyan: Hmm, it's marked replicated, but appears to only be declared in that client.dll UI file. So I guess the FCVAR_REPLICATED bit flag is incorrect?

Agiel: Hm, true. Move it to player_shared? Something like this shouldn't be client side imo, even if it's cheat protected.

Rainyan: Also I'm getting a bunch of console spam with:
```
For FCVAR_REPLICATED, ConVar must be defined in client and game .dlls (sv_neo_ghost_view_distance)
For FCVAR_REPLICATED, ConVar must be defined in client and game .dlls (sv_neo_ghost_view_distance)
```
so I guess I have to see what I did wrong with that cvar refactor to `neo_player_shared.h`

Rainyan: > Hm, true. Move it to player_shared? Something like this shouldn't be client side imo, even if it's cheat protected.

Yeah I guess we should ideally not network the beacon data until the client is legitimately allowed to display it, plus/minus some leeway to account for network delay. But we probably need to still have such a clientside timer for the correct moment to display the UI element, because we can't rely on network timing alone for it.

It should maybe be a replicated server cvar, which is used for `ShouldTransmit` server-side, and which the client UI can then receive to figure out when to display the beacon UI. I'll see if I can sneak it into this PR a bit later today.

Agiel (approved):

Masterkatze (approved):

PR #1464, opened 2025-11-06
Reflect skybox in most water materials

### LINKED ISSUE #1452: [FEATURE] 3D Skybox in water reflections
### Description

Give mappers the option to have the 3D skybox render in water reflections. This would vastly enhance the visuals for maps with large bodies of water, like rogue

### Proposed Solution

Add an option to the sky_camera entity that makes the 3D skybox render in water reflections, which is turned off by default, as most maps don't require this

### Alternatives Considered

Don't, because it might hurt performance for lower end systems

### Additional Context

- This was implemented in CSGO as a material parameter
- [I've achieved this before by changing the reflect texture in the water material, but the view was very broken.](https://cdn.discordapp.com/attachments/1252260488772059166/1347607230630989864/2025-03-07_16-28-38.mp4?ex=69058cc9&is=69043b49&hm=cdd478653e71e8fc2b91b22669c81b49257651c3d5415af1dc0133461465a557&)
- There are various comments about the 3d skybox and reflections in viewrender.cpp

DESTROYGIRL:
## Description
Makes skybox geometry render in water reflections when "Reflect All" is enabled in options.
In the issue I made for this I said it ought to be an option to be enabled per map for performance reasons, but I think this is unnessecary as it's only avaliable on the highest setting, and I did not notice any meaningful impact on performance when testing with vprof (on rogue)

The implementation can be viewed best on rogue, but also shrine and isolation

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1452

### DIFF
--- a/src/game/server/SkyCamera.cpp
+++ b/src/game/server/SkyCamera.cpp
@@ -112,15 +112,16 @@ void CSkyCamera::Spawn( void )
 	m_skyboxData.area = engine->GetArea( m_skyboxData.origin );
 	
 #ifdef NEO
-	if (m_skyboxData.reflectMode.Get() == REFLECT_SKYBOX_WATERCLIPZ)
+	if ( m_skyboxData.reflectMode.Get() == REFLECT_SKYBOX_WATERCLIPZ )
 	{
-		if (m_strWaterLevelDesignator != NULL_STRING)
+		auto pDesignator = gEntList.FindEntityByName( nullptr, m_strWaterLevelDesignator );
+		if ( m_strWaterLevelDesignator != NULL_STRING && pDesignator )
 		{
-			m_skyboxData.waterLevel = gEntList.FindEntityByName(nullptr, m_strWaterLevelDesignator)->GetAbsOrigin().z;
+			m_skyboxData.waterLevel = pDesignator->GetAbsOrigin().z;
 		}
 		else
 		{
-			Warning("sky_camera: Skybox water level entity not specified!\n");
+			Warning( "sky_camera: Skybox water level entity not found/specified!\n" );
 		}
 	}
 #endif


AdamTadeusz:
There's a lot of Assertion failures in the console

image

Also I notice when moving my mouse back and forth that the reflection in the water lags behind a bit, like its attached to a spring, this is on rogue specifically, but this still looks great in general good job!

AdamTadeusz:
wrt the springiness, it seems to be tied to mouse movement specifically, skybox reflections remain perfect when simply moving back and forth laterally without any rotation

DESTROYGIRL:
@AdamTadeusz Hopefully these issues should be fixed now, not putting it up for review yet though as I want to test some things

src/game/server/SkyCamera.cpp
```diff
@@ -107,6 +111,20 @@ void CSkyCamera::Spawn( void )
 	m_skyboxData.origin = GetLocalOrigin();
 	m_skyboxData.area = engine->GetArea( m_skyboxData.origin );
 	
+#ifdef NEO
+	if (m_skyboxData.reflectMode.Get() == REFLECT_SKYBOX_WATERCLIPZ)
+	{
+		if (m_strWaterLevelDesignator != NULL_STRING)
+		{
+			m_skyboxData.waterLevel = gEntList.FindEntityByName(nullptr, m_strWaterLevelDesignator)->GetAbsOrigin().z;
```
sunzenshen: Wonder if there's a situation where FindEntityByName returns a NULL pointer, in case something goes wrong or if there is no water level at all in the map.

Maybe we could be proactive and check to see if the entity is present?

```cpp
CBaseEntity *pWaterLevelEntity = gEntList.FindEntityByName(nullptr, m_strWaterLevelDesignator);
if (pWaterLevelEntity)
{
	m_skyboxData.waterLevel = pWaterLevelEntity->GetAbsOrigin().z;
}
else
{
	Warning("sky_camera: Water level designator entity '%s' not found! Please ensure the entity exists and is correctly named.\n", m_strWaterLevelDesignator.ToCStr());
}
```

sunzenshen (approved):

Masterkatze (approved):

PR #1469, opened 2025-11-09
NeoUI - Refactor root NeoUI usages, sound within NeoUI

### LINKED ISSUE #1468: NeoUI - Refactor root around main menu buttons, add rollover sound support within NeoUI
The main menu buttons are done in some loop which is very awkward to use with the NeoUI, it'll be better to instead use the NeoUI directly as intended. Also add in rollover sound support within NeoUI, but only enabled for the main menu at the moment.

nullsystem:
## Description
* Refactor neo_root to not loop around buttons, instead just directly place and use the buttons like the rest of the menus
* Put sound rollover into within the NeoUI API and remove manual handling within neo_root. Off by default, just add the new flag to per section and can customize to point to different paths from external to the NeoUI API.

## Toolchain

## Linked Issues

* fixes #1468

### DIFF
--- a/src/game/client/neo/ui/neo_root.cpp
+++ b/src/game/client/neo/ui/neo_root.cpp
@@ -364,8 +364,8 @@ CNeoRoot::CNeoRoot(VPANEL parent)
 
 	for (int i = 0; i < MMBTN__TOTAL; ++i)
 	{
-		const char *label = BTNS_LOCALIZE[i];
-		wchar_t *localizedWszStr = g_pVGuiLocalize->Find(label);
+		const char *pszLocalizeKey = BTNS_LOCALIZE[i];
+		wchar_t *localizedWszStr = g_pVGuiLocalize->Find(pszLocalizeKey);
 		Assert(localizedWszStr);
 		if (localizedWszStr)
 		{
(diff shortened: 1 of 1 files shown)


Masterkatze (approved):

src/game/client/neo/ui/neo_root.cpp
```diff
@@ -364,18 +362,19 @@ CNeoRoot::CNeoRoot(VPANEL parent)
...
 	SetScheme(neoscheme);
 
-	for (int i = 0; i < BTNS_TOTAL; ++i)
+	for (int i = 0; i < MMBTN__TOTAL; ++i)
 	{
-		const char *label = BTNS_INFO[i].label;
-		if (wchar_t *localizedWszStr = g_pVGuiLocalize->Find(label))
+		const char *label = BTNS_LOCALIZE[i];
+		wchar_t *localizedWszStr = g_pVGuiLocalize->Find(label);
+		Assert(localizedWszStr);
```
sunzenshen, line 369: nit: This Assert could be considered redundant since there's immediately a branch based on localizedWszStr one line later.  My interpretation as a reviewer unfamiliar with this code is that the expected case is that localizedWszStr=true, which seems to imply that the else case later is an unexpected case. If that is true, arguably some sort of error logging would be warranted, though that might add some i/o cost if that else case happens more often than expected.

sunzenshen: Overall the Release behavior of this branch looks okay to me.  Feel free to resolve this conversation and merge if this Assert still fits your expected behavior on Debug builds.

nullsystem: Added `Warning` ERROR message: https://github.com/NeotokyoRebuild/neo/pull/1469/commits/eff5a92e63f59b4c654c200e3d5e801382eb8e69

PR #1473, opened 2025-11-13
Configurable IFF Markers and Player Xray during gameplay

AdamTadeusz:
## Description

Makes the IFF markers customizable from the options menu, using the same serialisation concept that is used for the crosshair.

ALSO implements the player xray proposed feature.
[Discord initial discussion](https://discord.com/channels/124171198245502976/1439306600966782986)
[Google Doc](https://docs.google.com/document/d/10ay3eY6C3WoidSAAULwRoXpWa6IioYZKjV1huyrsR6g/edit?tab=t.0)

---> [neoassets PR](https://github.com/NeotokyoRebuild/neoAssets/pull/82) <---

### DIFF
--- a/src/game/client/neo/ui/neo_ui.cpp
+++ b/src/game/client/neo/ui/neo_ui.cpp
@@ -1401,6 +1401,7 @@ void Tabs(const wchar_t **wszLabelsList, const int iLabelsSize, int *iIndex, con
 		vgui::surface()->GetFullscreenViewport(oldX, oldY, oldW, oldH);
 		vgui::surface()->SetFullscreenViewport(c->rWidgetArea.x0, c->rWidgetArea.y0, c->irWidgetWide, c->irWidgetTall);
 		vgui::surface()->PushFullscreenViewport();
+
 		const auto *pFontI = &c->fonts[c->eFont];
 		for (int i = 0, iXPosTab = 0; i < iLabelsSize; ++i, iXPosTab += iTabWide)
 		{
@@ -1436,6 +1437,7 @@ void Tabs(const wchar_t **wszLabelsList, const int iLabelsSize, int *iIndex, con
 			vgui::surface()->DrawSetColor(COLOR_WHITE);
 			vgui::surface()->DrawFilledRect(c->irWidgetWide, 0, c->irWidgetWide-2, c->irWidgetTall);
 		}
+
 		vgui::surface()->PopFullscreenViewport();
 		vgui::surface()->SetFullscreenViewport(oldX, oldY, oldW, oldH);
 


AdamTadeusz:
sorry for another push I forgot we were disabling vision modes when spectating players in eye and using xray, that's redundant now

src/game/client/neo/ui/neo_root_settings.cpp
```diff
@@ -1314,3 +1357,62 @@ void NeoSettings_Crosshair(NeoSettings *ns)
...
+	NeoUI::RingBoxBool(L"Show name", &pMarker->bShowName);
+	NeoUI::RingBoxBool(L"Prepend clantag", &pMarker->bPrependClantagToName);
+	NeoUI::SliderInt(L"Max name length (including clantag)", &pMarker->iMaxNameLength, 0, MAX_MARKER_STRLEN, 1);
+
+#ifdef GLOWS_ENABLE
+	NeoUI::HeadingLabel(L"Player Xray");
+	NeoUI::RingBoxBool(L"Enable Xray",  &pHud->bEnableXray);
+	NeoUI::Slider(L"Outline Width", &pHud->flOutlineWidth, 0, 2, 2, 0.25f);
+	NeoUI::Slider(L"Center Opacity", &pHud->flCenterOpacity, 0, 1, 2, 0.1f);
+	NeoUI::Slider(L"Texture Opacity (Cloak highlight)", &pHud->flTexturedOpacity, 0, 1, 2, 0.1f);
```
sunzenshen, line 1448: After I copy over the `neo\materials\dev\halo_textured_add_to_screen.vmt` from neoAssets into the `git:neo/game/neo/materials/dev` folder, I'm not seeing the diagonal pattern with my setup:

Image

(At least the bright white cloak goes away when opacity is set to 0.  Same for the team color fill-in for behind walls.)

Image

I think this is just me not getting my setup right, so maybe someone else could sanity check.

sunzenshen (changes requested):
The effect looks really good in game.  There are some merge conflicts from the last settings related PR merge, so those need to be resolved.

Not 100% sure what's going on with my lack of diagonal lines texture, but another reviewer can sanity check me.

AdamTadeusz:
oopse poopsie @sunzenshen I added the shader to neoassets too, I thought I saw the compiled shader in the changed files list and thought it would copy over when building, adding the compiled shader to neoassets is how new shader authoring works

sunzenshen (approved):
While I think some aspects of the UX of the configuration are a bit unclear (specifically Xray vs non-Xray config tabs and which ones go into effect when Player Xray is enabled), I think the most expedient way to get feedback on the user experience is to show this to playtesters as an initial version, so I think this is worth trying.

src/game/client/neo/ui/neo_root_settings.cpp
sunzenshen: This was fixed with @AdamTadeusz's [update to neoAssets](https://github.com/NeotokyoRebuild/neoAssets/pull/82/commits/a6b6e4b4aad2eac8bd706ce2ef6b6c5b5ee74721) with the compiled shader:

Screenshot 2025-11-20 233225

Screenshot 2025-11-20 233104

src/game/client/c_basecombatcharacter.cpp
```diff
@@ -165,7 +165,7 @@ void C_BaseCombatCharacter::UpdateGlowEffect( void )
 		float r, g, b;
 		GetGlowEffectColor( &r, &g, &b );
 #ifdef NEO
-		m_pGlowEffect = new CGlowObject( this, Vector( r, g, b ), 1.0, true, true );
+		m_pGlowEffect = new CGlowObject( this, Vector( r, g, b ), 1.0, true, false );
 #else
 		m_pGlowEffect = new CGlowObject( this, Vector( r, g, b ), 1.0, true );
 #endif
```
Rainyan: Nit: since false is the default value, we could remove this NEO ifdef block entirely and use the SDK default from line 170.

src/game/client/c_world.cpp
```diff
@@ -15,6 +15,9 @@
 #include "eventlist.h"
 // NVNT haptic include for notification of world precache
 #include "haptics/haptic_utils.h"
+#if defined NEO && GLOWS_ENABLE
```
Rainyan: Maybe should be `&& defined GLOWS_ENABLE` instead for consistency (since it's being ifdef'd elsewhere instead of checking for value).

src/game/client/glow_outline_effect.cpp
```diff
@@ -12,43 +12,31 @@
...
+ConVar glow_outline_effect_enable("glow_outline_effect_enable", "0", FCVAR_ARCHIVE, "Enable entity outline glow effects.", 
+	[](IConVar* var [[maybe_unused]], const char* pOldValue [[maybe_unused]], float flOldValue [[maybe_unused]])
 	{
-		if (auto player = UTIL_PlayerByIndex(i))
-		{
-			float r, g, b;
-			NEORules()->GetTeamGlowColor(player->GetTeamNumber(), r, g, b);
-			player->SetGlowEffectColor(r, g, b);
-			player->SetClientSideGlowEnabled(!flOldValue);
+		bool wantGlow = glow_outline_effect_enable.GetBool();
```
Rainyan: Nit: unused variable.

src/game/client/neo/ui/neo_hud_friendly_marker.cpp
```diff
@@ -12,24 +12,51 @@
...
+	ConVarRef conVar( pConVar );
+	const char* newValue = conVar.GetString();
+
+	NeoIFFMarkerOption conVarChanged = NEOIFFMARKER_OPTION_SQUAD;
+	if (pConVar->GetName() == "cl_neo_friendly_marker") { conVarChanged = NEOIFFMARKER_OPTION_FRIENDLY; }
+	else if (pConVar->GetName() == "cl_neo_spectator_marker") { conVarChanged = NEOIFFMARKER_OPTION_SPECTATOR; }
+#ifdef GLOWS_ENABLE
+	else if (pConVar->GetName() == "cl_neo_friendly_xray_marker") { conVarChanged = NEOIFFMARKER_OPTION_FRIENDLY_XRAY; }
+	else if (pConVar->GetName() == "cl_neo_squad_xray_marker") { conVarChanged = NEOIFFMARKER_OPTION_SQUAD_XRAY; }
+	else if (pConVar->GetName() == "cl_neo_spectator_xray_marker") { conVarChanged = NEOIFFMARKER_OPTION_SPECTATOR_XRAY; }
```
Rainyan: Shouldn't rely on address-based comparisons for string literals, because AFAIK this depends on implementation:

> [lex.string]
[Note 4: String literal objects are potentially non-unique ([intro.object]).
Whether successive evaluations of a string-literal yield the same or a different object is unspecified.
— end note]

MSVC also emits a warning for this:
> warning C4130: '==': logical operation on address of string constant

Maybe use `FStrEq` instead (it tries the cheap ptr compare path first, for implementations that do support this)

src/game/client/neo/ui/neo_hud_friendly_marker.h
```diff
@@ -1,15 +1,70 @@
...
+#define NEO_SQUAD_MARKER_DEFAULT "0;1;0;0;0.50;1;0;1;1;32;"
+#define NEO_FRIENDLY_MARKER_DEFAULT "0;1;0;1;0.50;1;0;0;0;32;"
+#define NEO_SPECTATOR_MARKER_DEFAULT "0;0;0;1;0.50;1;1;1;1;32;"
+#ifdef GLOWS_ENABLE
+#define NEO_SQUAD_XRAY_MARKER_DEFAULT "4;0;0;0;0.50;1;0;1;0;32;"
+#define NEO_FRIENDLY_XRAY_MARKER_DEFAULT "4;0;0;0;0.50;1;0;0;0;32;"
+#define NEO_SPECTATOR_XRAY_MARKER_DEFAULT "4;0;0;0;0.50;1;0;1;1;32;"
+#endif // GLOWS_ENABLE
+constexpr int NEO_IFFMARKER_SEQMAX = 32;
+constexpr int MAX_MARKER_STRLEN = 48 + NEO_MAX_CLANTAG_LENGTH + 1;
```
Rainyan: Nit: could we rename this `..._SIZE` or something like that, because "strlen" typically means the length without the null terminator, so this could be confusing and lead to off-by-one memory bugs if someone incorrectly assumes it means `MAX_MARKER_STRLEN+1` is the null terminator.

E.g. `sizeof("foo") == 3+1`, vs `strlen("foo") == 3`

src/game/client/neo/ui/neo_hud_friendly_marker.cpp
```diff
@@ -150,133 +191,260 @@ void CNEOHud_FriendlyMarker::DrawPlayerForTeam(C_Team *team, const C_NEO_Player
...
+	int textSize = 0;
+
+	auto DisplayText = [this, &textSize, x](const char *textASCII, int y, int maxLength) {
+			wchar_t textUTF[MAX_MARKER_STRLEN];
+			COMPILE_TIME_ASSERT(sizeof(textUTF) == sizeof(wchar_t) * MAX_MARKER_STRLEN);
+			g_pVGuiLocalize->ConvertANSIToUnicode(textASCII, textUTF, min(sizeof(textUTF), sizeof(wchar_t) * maxLength));
+			int textWidth, textHeight;
+			vgui::surface()->GetTextSize(m_hFont, textUTF, textWidth, textHeight);
+			vgui::surface()->DrawSetTextPos(x - (textWidth / 2), y - textHeight);
+			vgui::surface()->DrawPrintText(textUTF, min(V_wcslen(textUTF), maxLength));
```
Rainyan: Nit: we can get the text length from the `g_pVGuiLocalize->ConvertANSIToUnicode` return value (minus 1 to omit the null terminator). So we can skip the `min(V_wcslen(...` calculation. Assuming the API comment for the function is accurate.

src/game/client/neo/ui/neo_hud_friendly_marker.cpp
```diff
@@ -150,133 +191,260 @@ void CNEOHud_FriendlyMarker::DrawPlayerForTeam(C_Team *team, const C_NEO_Player
...
+	NEOIFFMARKERVARTYPE_INT,   // NEOIFFMARKER_SEGMENT_I_MAXNAMELENGTH
+};
+
+bool ImportMarker(FriendlyMarkerInfo *crh, const char *pszSequence)
+{
+	int iPrevSegment = 0;
+	int iSegmentIdx = 0;
+	NeoIFFMarkerVariant vars[NEOIFFMARKER_SEGMENT__TOTAL] = {};
+
+	const int iPszSequenceSize = V_strlen(pszSequence);
```
Rainyan: Nit: would prefer `...Len` instead of `...Size` for the var name to avoid the "sizeof" vs "strlen" confusion.

Rainyan (changes requested):
Looks great!

Left some feedback, and additionally:

* At https://github.com/AdamTadeusz/neo/blob/a122f1963acbe902217d64d99fd3d5bdb8e932e6/src/game/client/glow_outline_effect.cpp#L330-L331 the variables `nSrcWidth` and `nSrcHeight` are unused for the NEO ifdef path. Perhaps guard these with a `#ifndef NEO` to fix the compiler warning?

* The added main menu tab kinda cramps the headers (see the spacing between "Mouse/Controller" and "Audio" tab headers):
Screenshot 2025-11-25 003425

Even for high-ish resolutions, it's a snug fit:

Screenshot 2025-11-25 003823

Maybe a simple fix for now would be to rename "Mouse/Controller" to something shorter like "Aiming"?

Rainyan (approved):
LGTM

One nit I have about the UI: if the user doesn't know about the existence of the horizontal scrolling, it might be easy to not realise there's more tabs available. Perhaps it'd be a little more obvious if we had an arrow symbol to indicate more items?

arr

But this is completely fine to merge by me. Also tested with gamepad, and the shoulder buttons worked well to move between tabs with the horizontal scrolling.

src/game/client/neo/ui/neo_root_settings.h
```diff
@@ -297,6 +312,22 @@ struct NeoSettings
...
 		CONVARREF_DEF(cl_neo_crosshair_hip_fire);
+
+		// Friendly Markers
+		CONVARREF_DEFNOGLOBALPTR(cl_neo_squad_marker);
+		CONVARREF_DEFNOGLOBALPTR(cl_neo_friendly_marker);
+		CONVARREF_DEFNOGLOBALPTR(cl_neo_spectator_marker);
+#ifdef GLOWS_ENABLE
+		CONVARREF_DEFNOGLOBALPTR(cl_neo_squad_xray_marker);
+		CONVARREF_DEFNOGLOBALPTR(cl_neo_friendly_xray_marker);
+		CONVARREF_DEFNOGLOBALPTR(cl_neo_spectator_xray_marker);
```
Rainyan, line 323: We should maybe update the comment at https://github.com/AdamTadeusz/neo/blob/fcc9075b44763b3b8623cd32356e5b4b525b55e1/src/game/client/neo/ui/neo_root_settings.h#L19 regarding `CONVARREF_DEFNOGLOBALPTR`, since it's no longer used just for volume settings.

Rainyan (dismissed):

Rainyan:
Also, should we default-enable this feature? It seems like a huge usability win, and it would be nice to get as many people as possible to playtest it.

sunzenshen:
Setting the max value (64) of "Initial offset" in the IFF customization settings causes the player crosshair to be moved to the top center-left of the screen:

https://github.com/user-attachments/assets/61636605-2801-4838-9ff2-96670550b80f

Restarting the game resets the crosshair to be in the center again, but as far as I can tell dying or round resets or changing the setting back to a different value does not reset the crosshair position.

sunzenshen (dismissed):
Open to the idea of merging this in for testing with the odd extreme offset bug if we open a follow up issue.  We can just tell testers that the "Initial offset" setting is jank and get their feedback early.

sunzenshen (approved):
The horizontal scroll issue explains why the "Initial offset" setting was working earlier.

I checked the new build and it LGTM:

Image

PR #1474, opened 2025-11-13
Don't allow players to enable xray when alive and mp_forcecamera is set to 0

### LINKED ISSUE #1470: players shouldn't be able to see enemies through walls
the command glow_outline_effect_enable can be used after respawning to highlight friends and enemies

Image

AdamTadeusz:
## Linked Issues

- fixes #1470


DESTROYGIRL (approved):
Works, I noticed I was still able to enable it when in the ragdoll death cam, but I guess it wouldn't matter much at that point

PR #1478, opened 2025-11-16
Add commands for adjusting bots' class

Rainyan:
## Description
Add the following (mainly) debug commands:

* Add `bot_class` cvar for adjusting which class bots should spawn as the next time. This var name and behaviour mimics the `bot_class` cvar of OGNT, with the exception of default value `-1` (don't force class but instead use a random one), instead of `0` (always spawn recon bots).

* Add `bot_changeclass` command for switching all bots' classes right now instead of at next spawn (if allowed by gamerules/other cvars, it's left for the `RequestSetClass` logic to decide).

Reasoning for adding: `bot_class` is parity, and `bot_changeclass` is "SDK parity" prior to nextbots, although I can't remember if it was ever working for the NT classes. Also handy for people who want to test class-specific things with the bots but don't have the development tools set up to edit code.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/neo_bot.cpp
+++ b/src/game/server/neo/bot/neo_bot.cpp
@@ -176,7 +176,6 @@ CON_COMMAND_F(neo_bot_add, "Add a bot.", FCVAR_GAMEDLL)
 				pBot->SetAttribute(CNEOBot::QUOTA_MANANGED);
 			}
 
-			pBot->RequestClassOnProfile();
 			engine->SetFakeClientConVarValue(pBot->edict(), "name", pBot->GetPlayerName() );
 			pBot->RequestSetSkin(RandomInt(0, 2));
 			pBot->HandleCommand_JoinTeam(iTeam);
@@ -2546,11 +2545,6 @@ NeoClass CNEOBot::ChooseRandomClass() const
 
 	AssertMsg(false, "fell through the logic");
 	return NEO_CLASS_RECON;
-}
-
-void CNEOBot::RequestClassOnProfile()
-{
-	RequestSetClass(ChooseRandomClass());
 }
 
 CNEOBotIntention::CNEOBotIntention(CNEOBot *bot)
--- a/src/game/server/neo/bot/neo_bot.h
+++ b/src/game/server/neo/bot/neo_bot.h
@@ -429,7 +429,6 @@ public:
 	float m_flLastShouldAimTime = 0.0f;
 
 	CNEOBotProfile m_profile = {};
-	void RequestClassOnProfile();
 	NeoClass ChooseRandomClass() const;
 	int m_iIntendTeam = 0;
 	int m_iProfileIdx = -1;
--- a/src/game/server/neo/bot/neo_bot_manager.cpp
+++ b/src/game/server/neo/bot/neo_bot_manager.cpp
@@ -402,7 +402,6 @@ void CNEOBotManager::MaintainBotQuota()
 		if ( pBot )
 		{
 			pBot->SetAttribute( CNEOBot::QUOTA_MANANGED );
-			pBot->RequestClassOnProfile();
 			engine->SetFakeClientConVarValue( pBot->edict(), "name", pBot->GetPlayerName() );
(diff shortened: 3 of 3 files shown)


src/game/server/neo/neo_player.cpp
```diff
@@ -3805,3 +3830,19 @@ void CNEO_Player::SpectatorTakeoverPlayerRevert(bool bHardReset)
...
+		return;
+	}
+	const int botClass = V_atoi(args.Arg(1));
+	for (int i = 1; i <= gpGlobals->maxClients; ++i)
+	{
+		auto* player = assert_cast<CNEO_Player*>(UTIL_PlayerByIndex(i));
+		if (player && player->IsBot() && player->GetTeamNumber() >= FIRST_GAME_TEAM)
+			player->RequestSetClass(botClass);
+	}
+}
```
sunzenshen: If we happen to suggest any changes, maybe add a newline here even if this missing newline probably fine with modern toolchains.

src/game/server/neo/neo_player.cpp
```diff
@@ -150,6 +150,10 @@ ConVar sv_neo_change_threshold_interval("sv_neo_change_threshold_interval", "0.2
 ConVar sv_neo_dm_max_class_dur("sv_neo_dm_max_class_dur", "10", FCVAR_REPLICATED, "The time in seconds when the player can change class on respawn during deathmatch.", true, 0.0f, true, 60.0f);
 ConVar sv_neo_warmup_godmode("sv_neo_warmup_godmode", "0", FCVAR_REPLICATED, "If enabled, everyone is invincible on idle and warmup.", true, 0.0f, true, 1.0f);
 
+ConVar bot_class("bot_class", "-1", 0, "Force all bots to spawn with the specified class number, or -1 to disable.", true, -1, true, NEO_CLASS_ENUM_COUNT-1);
+void BotChangeClassFn(const CCommand& args);
+ConCommand bot_changeclass("bot_changeclass", BotChangeClassFn, "Force all bots to switch to the specified class number.");
```
sunzenshen, line 155: Do we want to provide range limits like with bot_class (-1 thru NEO_CLASS_ENUM_COUNT-1), or do we want to keep this as a power user command where they can specify any value?

Rainyan: The `ConCommand` bounds are checked inside the callback, I don't think there exists a constructor for clamping the inputs. Also it takes a variadic number of inputs inside the `CCommand` parameter.

sunzenshen: Then again, I tried setting some out of bound values, and those didn't cause any noticeable issues, so perhaps this is fine.

That said, I tried to change bot_class and bot_changeclass back to -1 and it doesn't restore the heterogeneous class compositions for the teams.  Not sure if that's an edge case we want to solve for what are debugging commands.

sunzenshen: > The `ConCommand` bounds are checked inside the callback, I don't think there exists a constructor for clamping the inputs. Also it takes a variadic number of inputs inside the `CCommand` parameter.

Ah I see, that explains why things were okay when I tried setting out of bound values, so looks good from cursory testing.

sunzenshen (approved):
Some optional nits but LGTM:

Image

Image

src/game/server/neo/neo_player.cpp
Rainyan: I don't think this is UB though, if I'm reading this right, from the C++20 spec \[lex.phases\] 5.2.2:
> A source file that is not empty and that (after splicing) does not end in a new-line character shall be processed as if an additional new-line character were appended to the file[.](https://eel.is/c++draft/lex.phases#1.2.sentence-5)

This looks to be the same wording as with C++11 spec, so I'd assume all reasonable compilers will conform to it by now(?)

sunzenshen: Yeah, at this point it's probably fine with any modern toolchain. 👍

AdamTadeusz (approved):

PR #1480, opened 2025-11-16
Fix m_rfAttackers... off by 1

### LINKED ISSUE #1479: `m_rfAttackers...` off by 1
Those are 1-indexed and could happen to get used in `MAX_PLAYER` index, just `+ 1` the array size so it can handle that index.

nullsystem:
## Description
* The array is used as 1-indexed and so `MAX_PLAYER` numbered index could be used. So instead increase the array size by 1 so it can handle that.
* 0 index never used, but better that than converting to 0-index when using the array everytime

## Toolchain

- Linux GCC Distro Native - Arch GCC 15

## Linked Issues

* fixes #1479

### DIFF
--- a/src/game/client/neo/c_neo_player.cpp
+++ b/src/game/client/neo/c_neo_player.cpp
@@ -109,9 +109,9 @@ IMPLEMENT_CLIENTCLASS_DT(C_NEO_Player, DT_NEO_Player, CNEO_Player)
 END_RECV_TABLE()
 
 BEGIN_PREDICTION_DATA(C_NEO_Player)
-	DEFINE_PRED_ARRAY(m_rfAttackersScores, FIELD_INTEGER, (MAX_PLAYERS + 1), FTYPEDESC_INSENDTABLE),
-	DEFINE_PRED_ARRAY(m_rfAttackersAccumlator, FIELD_FLOAT, (MAX_PLAYERS + 1), FTYPEDESC_INSENDTABLE),
-	DEFINE_PRED_ARRAY(m_rfAttackersHits, FIELD_INTEGER, (MAX_PLAYERS + 1), FTYPEDESC_INSENDTABLE),
+	DEFINE_PRED_ARRAY(m_rfAttackersScores, FIELD_INTEGER, MAX_PLAYERS_ARRAY_SAFE, FTYPEDESC_INSENDTABLE),
+	DEFINE_PRED_ARRAY(m_rfAttackersAccumlator, FIELD_FLOAT, MAX_PLAYERS_ARRAY_SAFE, FTYPEDESC_INSENDTABLE),
+	DEFINE_PRED_ARRAY(m_rfAttackersHits, FIELD_INTEGER, MAX_PLAYERS_ARRAY_SAFE, FTYPEDESC_INSENDTABLE),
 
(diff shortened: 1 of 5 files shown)


Masterkatze (dismissed):

src/game/client/neo/c_neo_player.cpp
```diff
@@ -109,9 +109,9 @@ IMPLEMENT_CLIENTCLASS_DT(C_NEO_Player, DT_NEO_Player, CNEO_Player)
 END_RECV_TABLE()
 
 BEGIN_PREDICTION_DATA(C_NEO_Player)
-	DEFINE_PRED_ARRAY(m_rfAttackersScores, FIELD_INTEGER, MAX_PLAYERS, FTYPEDESC_INSENDTABLE),
-	DEFINE_PRED_ARRAY(m_rfAttackersAccumlator, FIELD_FLOAT, MAX_PLAYERS, FTYPEDESC_INSENDTABLE),
-	DEFINE_PRED_ARRAY(m_rfAttackersHits, FIELD_INTEGER, MAX_PLAYERS, FTYPEDESC_INSENDTABLE),
+	DEFINE_PRED_ARRAY(m_rfAttackersScores, FIELD_INTEGER, (MAX_PLAYERS + 1), FTYPEDESC_INSENDTABLE),
+	DEFINE_PRED_ARRAY(m_rfAttackersAccumlator, FIELD_FLOAT, (MAX_PLAYERS + 1), FTYPEDESC_INSENDTABLE),
```
Rainyan: Nit: we have a `MAX_PLAYERS_ARRAY_SAFE` definition, which might be preferred over `MAX_PLAYERS + 1` because it makes it more obvious this is really correct and the `+ 1` is not a mistake.

src/game/client/neo/ui/neo_hud_killer_damage_info.cpp
```diff
@@ -135,7 +135,7 @@ void CNEOHud_KillerDamageInfo::resetHUDState()
 		auto *neoAttacker = dynamic_cast<C_NEO_Player *>(UTIL_PlayerByIndex(pIdx));
 		if (neoAttacker)
 		{
-			for (int i = 0; i < MAX_PLAYERS; ++i)
+			for (int i = 0; i <= MAX_PLAYERS; ++i)
```
Rainyan: Nit: I wonder if it would be more readable to do something like:
```cpp
for (int i = 0; i < _ARRAYSIZE(neoAttacker->m_rfAttackersScores); ++i)
```
so it's obvious we're looping over all the array indices, and it's known to never overflow at compile time. Since the from i=0 up to <= limit can look like a off-by-one overflow mistake, even though it's correct here. Well, maybe personal preference.

Rainyan (dismissed):

src/game/client/neo/ui/neo_hud_killer_damage_info.cpp
nullsystem: I think I'll change it to `< MAX_PLAYERS_ARRAY_SAFE` mainly cause it's over multiple arrays of same `MAX_PLAYERS_ARRAY_SAFE` size.

Rainyan (approved):

PR #1487, opened 2025-11-17
Update key bindings on NT;RE update

### LINKED ISSUE #1332: Add binding on update
There been new bindings (EX: music, killer/damage HUD) but those doesn't get picked up on existing installs, requiring it to be manually set in by the player. I'm guessing some convar or file on what the user's config version currently is and just add in the default binding of the new binding if the binding is not filled in already.

nullsystem:
## Description
* Keeps two hidden client convars: `cl_neo_cfg_version_major` and `cl_neo_cfg_version_minor` to keep track of which version the config was set for. Default value of "0" so it can recognize config previous to `cl_neo_cfg_version_...`. Major/minor versions are checked and set on client startup and will only set the new keybinds if the bind wasn't set, and the default key wasn't set.
* The config version follows NT;RE `major.minor` version and extracts the version from the git tag with cmake, which puts it into `neo_version_number.h` header
* Reviewers/testers can set `cl_neo_cfg_version_...` and unset relevant keybinds to emulate upgrading from older config to current
* Currently sets mp3, kdinfo, ping, and +aim -> toggle_aim if config version < 22.0/unset
* Updated github workflow to fully fetch tags and history as it's required now

## Toolchain

## Linked Issues

* fixes #1332


src/game/client/cdll_client_int.cpp
```diff
@@ -1342,6 +1366,37 @@ void CHLClient::PostInit()
 		g_pCVar->FindVar("neo_clantag")->InstallChangeCallback(NeoConVarStrLimitChangeCallback<NEO_MAX_CLANTAG_LENGTH>);
 		g_pCVar->FindVar("cl_neo_crosshair")->InstallChangeCallback(NeoConVarStrLimitChangeCallback<NEO_XHAIR_SEQMAX>);
 		g_pCVar->FindVar("sv_use_steam_networking")->SetValue(false);
+
+		// ConVarRef so that it properly sets the ConVar value
+		ConVarRef cvr_cl_neo_cfg_version_major("cl_neo_cfg_version_major");
+		ConVarRef cvr_cl_neo_cfg_version_minor("cl_neo_cfg_version_minor");
+		const int iCfgVerMajor = cvr_cl_neo_cfg_version_major.GetInt();
+		const int iCfgVerMinor = cvr_cl_neo_cfg_version_minor.GetInt();
+		if (iCfgVerMajor < NEO_VERSION_MAJOR || iCfgVerMinor < NEO_VERSION_MINOR)
```
sunzenshen: Is this considered a valid major difference?

22.1 -> 20.2 

(iCfgVerMajor=22 < NEO_VERSION_MAJOR=20)  false
||
(iCfgVerMinor=1 < NEO_VERSION_MINOR=2) true

false || true = true

which would then be considered an "upgrade".

I think perhaps you meant to check the minor version only when the major versions are *equal*?

(I'm assuming that we're not worried about handling downgrade scenarios, since that seems unlikely with a rolling Steam release that should only increment versions.)

nullsystem: Well guess might as well check major equal just in-case.

sunzenshen (approved):

sunzenshen (approved):

PR #1492, opened 2025-11-21
Update weapon cycle times in the script files

### LINKED ISSUE #1485: Firerates for some weapons are off
Image

Cycle times should be changed to the values in the "Actual cycle time" column in the weapon scripts(?)

bauxiteDYS:
## Description

Some weapon cycle times were off in the script files leading to incorrect weapon rate of fire, this should fix or make them close to the original game as it is currently played

Note: Actual cycle time of SRS seems to be unknown, weapon has interesting behaviour from what I can tell and the cycle time seems to vary from 1.0 s when quick switching (perhaps lower) to 1.333 s without switching, or 1.4 when firing more than 2 shots.

## Toolchain
The usual tools and stuff I guess

## Linked Issues

- fixes #1485


game/neo/scripts/weapon_m41.txt
```diff
@@ -14,7 +14,7 @@ WeaponData
 	"bucket_position"	"0"
 	"Damage"		"29"
 	"Penetration"		"32.0"
-	"CycleTime"		"0.1"		// time between shots
+	"CycleTime"		"0.166666"		// time between shots
```
Rainyan, line 17: I notice this was marked with "Actual Cycle Time" value of `0.1060606061` in the table, in #1485 Which one is correct?

game/neo/scripts/weapon_m41s.txt
```diff
@@ -14,7 +14,7 @@ WeaponData
 	"bucket_position"	"0"
 	"Damage"		"26"
 	"Penetration"		"20.0"
-	"CycleTime"		"0.1"		// time between shots
+	"CycleTime"		"0.166666"		// time between shots
```
Rainyan, line 17: Same question as with https://github.com/NeotokyoRebuild/neo/pull/1492/files#r2553608649

game/neo/scripts/weapon_srs.txt
```diff
@@ -14,7 +14,7 @@ WeaponData
 	"bucket_position"	"0"
 	"Damage"		"100"
 	"Penetration"		"65.0"
-	"CycleTime"		"1.0"
+	"CycleTime"		"1.4"
```
Rainyan, line 17: This seems to also differ from the values table in the issue?

game/neo/scripts/weapon_m41.txt
bauxiteDYS: The table uses the OG NT script value, OG NT uses a hardcoded value for the weapon that uses this script file, as found by Agiel's decompilation that value is 0.166, not 0.1.

game/neo/scripts/weapon_m41s.txt
bauxiteDYS: The table uses the OG NT script value, OG NT uses a hardcoded value for the weapon that uses this script file, as found by Agiel's decompilation that value is 0.166, not 0.1.

game/neo/scripts/weapon_srs.txt
bauxiteDYS: This value is derived from experiments and is similar to the OG NT Plugin `srs limiter` which caps the minimum ROF at `1.38 s`, however we don't know what OG NT is exactly doing with this weapon's ROF, Agiel seemed to think it actually uses a hardcoded value of `1.25 s` - but it doesn't seem to be the case from in-game measurements. The value in the table however is from the OG script, which again seems to be ignored by the game.

Rainyan: Ah ok, that makes sense since the SM timers have tickrate-dependent accuracy.

PR #1494, opened 2025-11-22
 Change how lean toggling works to allow for reset

### LINKED ISSUE #1493: Lean Toggle should turn off when sprinting/jumping just like OGNT
Pressing lean left toggle and running will lean you left after running. There is no indicator to show this so you will get a surprise lean if you run around for a while or keep jumping.

VantoNortim:
## Description
Lean is only bound to one key now and hold/toggle can be changed in the menu
If it's in toggle mode jumping/sprinting will reset it, in hold mode holding the key will make the player lean in after the action just like before

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1493

### DIFF
--- a/src/game/client/cdll_client_int.cpp
+++ b/src/game/client/cdll_client_int.cpp
@@ -1359,6 +1359,22 @@ static void SetupBindIfNotSet(const char *pszBindName, const ButtonCode_t bc)
 	}
 }
 
+static void UnbindIfExists(const char *pszBindName)
+{
+	const ButtonCode_t btnCode = gameuifuncs->GetButtonCodeForBind(pszBindName);
+	if (btnCode)
+	{
+		const char* bindBtnName = g_pInputSystem->ButtonCodeToString(btnCode);
+		if (bindBtnName && bindBtnName[0])
+		{
+			char szCmd[128];
+
+			V_sprintf_safe(szCmd, "unbind \"%s\"\n", bindBtnName);
+			engine->ClientCmd_Unrestricted(szCmd);
+		}
+	}
+}
+
 #endif // NEO
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


game/neo/scripts/kb_act.lst
```diff
@@ -7,8 +7,6 @@
 "+moveright"			"#Valve_Move_Right"
 "+leanl"	            "Lean left"
 "+leanr"	            "Lean right"
-"toggle_leanl"	        "Lean left (toggle)"
-"toggle_leanr"	        "Lean right (toggle)"
```
sunzenshen, line 11: Since these end up unbinding the lean keys if a user upgrades, maybe the key binding update system needs to take this into consideration.

See: https://github.com/NeotokyoRebuild/neo/pull/1487

VantoNortim: Yes, I can add a bind for those once it's in 👍

src/game/server/neo/neo_player.cpp
```diff
@@ -705,16 +705,21 @@ void CNEO_Player::CheckLeanButtons()
 	{
 		return;
 	}
-	
-	m_bInLean = NEO_LEAN_NONE;
-	if ((m_nButtons & IN_LEAN_LEFT) && !(m_nButtons & IN_LEAN_RIGHT || IsSprinting()))
```
sunzenshen, line 710: You might need to replicate the in air or sprinting check in place of what was removed here, and just before the leanleft/leanright assignments below.  Otherwise, the server vs client prediction might diverge.

That is, this logic might need to be closer to C_NEO_Player::CheckLeanButtons() above

Maybe something long the lines of:

```cpp
    if (IsSprinting() || IsAirborne())
    {
        m_bInLean = NEO_LEAN_NONE;
        return;
    }

    const bool leanLeft  = (m_nButtons & IN_LEAN_LEFT)  != 0;
    const bool leanRight = (m_nButtons & IN_LEAN_RIGHT) != 0;
```

src/game/client/neo/c_neo_player.cpp
```diff
@@ -534,22 +534,39 @@ void C_NEO_Player::CheckVisionButtons()
...
-	if ((m_nButtons & IN_LEAN_LEFT) && !(m_nButtons & IN_LEAN_RIGHT || IsSprinting()))
+	if (IsSprinting() || IsAirborne()) {
+		m_bInLean = NEO_LEAN_NONE;
+		
+		if (!cl_neo_lean_hold.GetBool()) {
+			IN_LeanReset();
+		}
+		
+		return;
+	}
```
sunzenshen, line 554: The above logic may need to be replicated in neo_player.cpp below.

game/neo/scripts/kb_act.lst
sunzenshen: I think it is already merged, you may need to rebase to be able to use that functionality.

src/game/server/neo/neo_player.cpp
VantoNortim: You cannot lean in those cases but I might as well add it back.

PR #1495, opened 2025-11-23
Lean toggle reset on sprint and jump plus hold priority over toggle except when toggling twice

### LINKED ISSUE #1493: Lean Toggle should turn off when sprinting/jumping just like OGNT

AdamTadeusz:
## Description

solves the same problem as #1493 but keeps separate keybinds for hold and toggle lean. Also allows a hold bind to lift a toggle bind for the same behavior and adds that to the lean and walk binds

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #1493

### DIFF
--- a/src/game/client/in_main.cpp
+++ b/src/game/client/in_main.cpp
@@ -393,7 +393,11 @@ void KeyDown( kbutton_t *b, const char *c )
 		b->down[1] = k;
 	else
 	{
+#ifdef NEO
+		if ( c && c[0] )
+#else
 		if ( c[0] )
+#endif // NEO
 		{
 			DevMsg( 1,"Three keys down for a button '%c' '%c' '%c'!\n", b->down[0], b->down[1], c[0]);
 		}
@@ -437,7 +441,7 @@ void KeyDownHoldReplaceToggle( kbutton_t *b, const char *c )
 		b->down[1] = k;
 	else
 	{
-		if ( c[0] )
+		if ( c && c[0] )
 		{
 			DevMsg( 1,"Three keys down for a button '%c' '%c' '%c'!\n", b->down[0], b->down[1], c[0]);
 		}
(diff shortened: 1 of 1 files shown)


sunzenshen (approved):
For what it's worth, this works for how I usually use lean, but interested to see what someone with more experience with the movement mechanics would think.

VantoNortim (dismissed):
Working as intended. However some sides notes, if you can't actually jump/run it will untoggle your lean. On the other hand falling does not even though it unleans you.

`if (IsSprinting() || IsAirborne()) {
	IN_LeanToggleReset();
	return;
}`

Could have these reset in c_neo_player but this is fine as well.

src/game/client/in_main.cpp
```diff
@@ -498,7 +574,11 @@ void IN_Attack2Down( const CCommand &args ) { KeyDown(&in_attack2, args[1] );}
 void IN_Attack2Up( const CCommand &args ) {KeyUp(&in_attack2, args[1] );}
 void IN_UseDown ( const CCommand &args ) {KeyDown(&in_use, args[1] );}
 void IN_UseUp ( const CCommand &args ) {KeyUp(&in_use, args[1] );}
+#ifdef NEO
+void IN_JumpDown(const CCommand& args) { KeyDown(&in_jump, args[1]); }
+#else
 void IN_JumpDown ( const CCommand &args ) {KeyDown(&in_jump, args[1] );}
+#endif // NEO
```
Rainyan: Nit: the NEO path and the default path are identical, so we could remove this ifdef guard

src/game/client/in_main.cpp
```diff
@@ -405,6 +405,51 @@ void KeyDown( kbutton_t *b, const char *c )
...
+		b->down[1] = k;
+		return; // state hasn't changed
+	}
+	else if (!b->down[0])
+		b->down[0] = k;
+	else if (!b->down[1])
+		b->down[1] = k;
+	else
+	{
+		if ( c[0] )
```
Rainyan: For line 440: Since we're checking for nullptr at line 417, should we also do the same check here, to avoid a nullptr dereference? That being said, the same thing occurs at line 396 in Valve code. But maybe it'd be safe to add the nullptr check for both?

Rainyan (approved):

sunzenshen (dismissed):

Rainyan (approved):

PR #1506, opened 2025-11-24
ghost EHandle

AdamTadeusz:
## Description

There is a rare issue that's come up a few times where the ghost marker gets placed on objects which are not the ghost. This PR aims to hopefully solve this issue by creating a networked [chandle](https://developer.valvesoftware.com/wiki/CHandle) for the ghost object, and using that instead client side to get the ghost position. This PR also makes the position of the ghost marker when carried by players behave more like in the original

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #

### DIFF
--- a/src/game/client/neo/ui/neo_hud_ghost_marker.cpp
+++ b/src/game/client/neo/ui/neo_hud_ghost_marker.cpp
@@ -171,28 +171,8 @@ void CNEOHud_GhostMarker::DrawNeoHudElement()
 				ghostColor = (iGhostingTeam == TEAM_JINRAI) ? COLOR_JINRAI : COLOR_NSF;
 			}
 		}
-
-		// Use PVS over networked-given position if possible as it'll give a smoother visual
-		// NEO NOTE (Rain): this assignment was segfaulting on Linux, so I've split the logic
-		// and added extra guards, so we can catch it better if it still happens.
-		Vector ghostPos = NEORules()->GetGhostPos();
-
-		if (const int ghosterPlayerIdx = NEORules()->GetGhosterPlayer();
-			ghosterPlayerIdx > 0)
-		{
-			if (auto ghosterPlayer = static_cast<CNEO_Player*>(UTIL_PlayerByIndex(ghosterPlayerIdx)))
-			{
-				if (auto pWeapon = static_cast<C_NEOBaseCombatWeapon*>(ghosterPlayer->GetActiveWeapon());
-					pWeapon->GetNeoWepBits() & NEO_WEP_GHOST &&	ghosterPlayer->IsVisible())
-				{
-					constexpr const int GHOST_MARKER_STANDING_PLAYER_OFFSET = 32;
-					constexpr const int GHOST_MARKER_CROUCHING_PLAYER_OFFSET = 24;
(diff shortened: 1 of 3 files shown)


src/game/client/neo/ui/neo_hud_ghost_marker.cpp
```diff
@@ -205,32 +175,19 @@ void CNEOHud_GhostMarker::DrawNeoHudElement()
 		// Use PVS over networked-given position if possible as it'll give a smoother visual
 		// NEO NOTE (Rain): this assignment was segfaulting on Linux, so I've split the logic
 		// and added extra guards, so we can catch it better if it still happens.
```
Rainyan: This comment can probably be removed, since we're removing the code it's referring to

src/game/client/neo/ui/neo_hud_ghost_marker.cpp
```diff
@@ -205,32 +175,19 @@ void CNEOHud_GhostMarker::DrawNeoHudElement()
...
+		Vector ghostPos = NEORules()->GetGhostPos();
 
 		if (const int ghosterPlayerIdx = NEORules()->GetGhosterPlayer();
 			ghosterPlayerIdx > 0)
 		{
 			if (auto ghosterPlayer = static_cast<CNEO_Player*>(UTIL_PlayerByIndex(ghosterPlayerIdx)))
 			{
-				if (ghosterPlayer->IsVisible())
+				if (auto pWeapon = static_cast<C_NEOBaseCombatWeapon*>(ghosterPlayer->GetActiveWeapon());
+					pWeapon->GetNeoWepBits() & NEO_WEP_GHOST &&	ghosterPlayer->IsVisible())
```
Rainyan: Since this is a HUD draw function that runs fairly often, a perf nit: ideally we should check `IsVisible` first, because it's non-virtual, and won't have the dynamic dispatch overhead of needing to indirect through the vtable.

Analysing the release compile, at least MSVC sadly isn't smart enough to reorder this automatically despite all of the function calls being `const` qualified. Ghidra decompile:
```c
plVar10 = (longlong *)(**(code **)(*plVar8 + 0x720))(plVar8);
uVar11 = (**(code **)(*plVar10 + 0xba8))(plVar10);
if (((uVar11 & 8) == 0) || (cVar1 = thunk_FUN_180454130(plVar8), cVar1 == '\0'))
```
where the last `thunk_FUN` is the `IsVisible` check.

Rainyan (approved):

src/game/shared/neo/neo_gamerules.h
```diff
@@ -323,7 +324,8 @@ class CNEORules : public CHL2MPRules, public CGameEventListener
 	int GetGhosterTeam() const { return m_iGhosterTeam; }
 	int GetGhosterPlayer() const { return m_iGhosterPlayer; }
 	bool GhostExists() const { return m_bGhostExists; }
-	Vector GetGhostPos() const { return m_vecGhostMarkerPos; }
+	Vector GetGhostPos() const;
```
Rainyan, line 327: Nit: this could return a `const Vector&` instead of `Vector` since we're not returning a local value. Same with `GetJuggernautMarkerPos` actually.

AdamTadeusz: ill open a new PR to add these to expedite the approval process

PR #1512, opened 2025-11-26
getGhostPos const Vector&

AdamTadeusz:
## Description

https://github.com/NeotokyoRebuild/neo/pull/1506#discussion_r2560991606

### DIFF
--- a/src/game/shared/neo/neo_gamerules.cpp
+++ b/src/game/shared/neo/neo_gamerules.cpp
@@ -4322,7 +4322,7 @@ const Vector& CNEORules::GetGhostPos() const
 #endif // GAME_DLL
 }
 
-const Vector& CNEORules::GetGhostMarkerPos() const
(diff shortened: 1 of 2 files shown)


src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -4322,7 +4322,7 @@ Vector CNEORules::GetGhostPos() const
 #endif // GAME_DLL
 }
 
-Vector CNEORules::GetGhostMarkerPos() const
+const Vector& CNEORules::GetGhostMarkerPos() const
```
Rainyan: Sadly we can't return by reference from this method, because it returns a local at line 4339, so this reference would dangle. But it should be safe for the main `GetGhostPos`.

Rainyan (changes requested):
Could we also sneak in the same change for `Vector GetJuggernautMarkerPos() const` in the `neo_gamerules.h` header, at line 332

PR #1526, opened 2025-11-27
Strip surrounding whitespace from player names

### LINKED ISSUE #1520: IFF breaks when a player has a name made up of space characters
Image

Rainyan:
## Description
Strip surrounding whitespace from player name variables `name` (Steam name), `neo_name`, and `neo_clantag`.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1520

### DIFF
--- a/src/game/shared/neo/neo_gamerules.cpp
+++ b/src/game/shared/neo/neo_gamerules.cpp
@@ -3556,7 +3556,9 @@ void CNEORules::ClientSettingsChanged(CBasePlayer *pPlayer)
 	const char *pszOldNeoClantag = pNEOPlayer->GetNeoClantag();
 	if (V_strcmp(pszOldNeoClantag, pszNeoClantag) != 0)
 	{
-		V_strncpy(pNEOPlayer->m_szNeoClantag.GetForModify(), pszNeoClantag, NEO_MAX_CLANTAG_LENGTH);
+		V_strncpy(pNEOPlayer->m_szNeoClantag.GetForModify(),
+			(FStrEq(pszNeoClantag, "#empty") ? "" : pszNeoClantag),
+			NEO_MAX_CLANTAG_LENGTH);
 		m_bThinkCheckClantags = true;
 	}
 
--- a/src/game/server/neo/neo_player.cpp
+++ b/src/game/server/neo/neo_player.cpp
@@ -1827,8 +1827,16 @@ bool CNEO_Player::SetNeoPlayerName(const char *newNeoName)
 	// NEO NOTE (nullsystem): Generally it's never NULL but just incase
 	if (newNeoName)
 	{
-		V_memcpy(m_szNeoName.GetForModify(), newNeoName, sizeof(m_szNeoName)-1);
-		m_bNeoNameHasSet = (m_szNeoName.Get()[0] != 0);
+		if (FStrEq(newNeoName, "#empty"))
+		{
+			m_szNeoName.GetForModify()[0] = '\0';
+			m_bNeoNameHasSet = false;
+		}
+		else
+		{
+			V_memcpy(m_szNeoName.GetForModify(), newNeoName, sizeof(m_szNeoName)-1);
+			m_bNeoNameHasSet = (m_szNeoName.Get()[0] != 0);
+		}
 		return m_bNeoNameHasSet;
 	}
 	return false;


sunzenshen (approved):

src/game/client/neo/ui/neo_ui.h
```diff
@@ -491,8 +491,19 @@ enum TextEditFlag_
...
+/*2W*/ void TextEdit(const wchar_t *wszLeftLabel, wchar_t *wszText, const int iMaxWszTextSize,
+					 const TextEditFlags flags = TEXTEDITFLAG_NONE,
+					 const NextFn&& next=nullptr)
+{
+	BeginMultiWidgetHighlighter(2);
+	Label(wszLeftLabel);
+	TextEdit(wszText, iMaxWszTextSize, flags);
+	EndMultiWidgetHighlighter();
+	if constexpr (std::is_invocable_v<NextFn>) next();
+}
```
nullsystem: I strongly would not want templating in NeoUI, I don't see why there's a need for a next function + templating when just calling `VarTrimmer` after calling `TextEdit` suffice.

AdamTadeusz (dismissed):

src/game/client/neo/ui/neo_root_settings.cpp
```diff
@@ -1000,6 +1001,40 @@ static const wchar_t *EQUIP_UTILITY_PRIORITY_LABELS[NeoSettings::EquipUtilityPri
...
+		const bool hasDoubleBlankInPos = (
+			(i + 1 < zeroIdx) &&
+			std::iswblank(input[i]) &&
+			std::iswblank(input[i + 1]));
+
+		if (hasBadCharInPos || hasDoubleBlankInPos)
+		{
+			std::wmemmove(&input[i], &input[i + 1], zeroIdx-i);
+			NeoUI::CurrentContext()->iTextSelCur = NeoUI::CurrentContext()->iTextSelCur = i+hasDoubleBlankInPos;
+			NeoUI::CurrentContext()->iTextSelCur = NeoUI::CurrentContext()->iTextSelStart = i+hasDoubleBlankInPos;
```
sunzenshen, line 1031: Since contents are shifted left by one, should `i` and  `zeroIdx` be shifted left to compensate?

Otherwise it's possible that we are not checking some characters because of the increment in the 
`for (int i = 0; i < zeroIdx; ++i)` above?

e.g.
```cpp
    std::wmemmove(&input[i], &input[i + 1], zeroIdx - i);
    --zeroIdx;
    --i; 
    NeoUI::CurrentContext()->iTextSelCur = NeoUI::CurrentContext()->iTextSelCur = i+hasDoubleBlankInPos;
    NeoUI::CurrentContext()->iTextSelCur = NeoUI::CurrentContext()->iTextSelStart = i+hasDoubleBlankInPos;
```

src/game/client/neo/ui/neo_root_settings.cpp
```diff
@@ -1000,6 +1001,40 @@ static const wchar_t *EQUIP_UTILITY_PRIORITY_LABELS[NeoSettings::EquipUtilityPri
...
+		}
+	}
+	// Block leading spaces
+	if (std::iswblank(input[0]))
+	{
+		std::wmemmove(input, &input[1], maxlen - 1);
+		NeoUI::CurrentContext()->iTextSelCur = NeoUI::CurrentContext()->iTextSelCur = 0;
```
sunzenshen, line 1041: Is this double assignment of `NeoUI::CurrentContext()->iTextSelCur =` intentional for intent of what is going on in two consecutive assignments?  Otherwise it seems like the first assignment is redundant unless there's some operator overload of `->` I am not aware of.

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -3365,10 +3345,20 @@ void CNEORules::ClientSettingsChanged(CBasePlayer *pPlayer)
 		pNEOPlayer->Weapon_SetZoom(pNEOPlayer->m_bInAim);
 	}
 
-	const char *pszSteamName = engine->GetClientConVarValue(pPlayer->entindex(), "name");
-
 	const bool clientAllowsNeoName = (0 == StrToInt(engine->GetClientConVarValue(engine->IndexOfEdict(pNEOPlayer->edict()), "cl_onlysteamnick")));
-	const char *pszNeoName = engine->GetClientConVarValue(pNEOPlayer->entindex(), "neo_name");
+
+	char szSteamName[MAX_PLACE_NAME_LENGTH] = "";
```
sunzenshen: Just to sanity check but is there any chance this was intended as 

`MAX_PLAYER_NAME_LENGTH`
instead of 
`MAX_PLACE_NAME_LENGTH` as above?

They both seem to exist but wondering if there's a significance to using `_PLACE_` above instead of `_PLAYER_`.

src/game/client/neo/ui/neo_root_settings.cpp
Rainyan: Yes, this seems to indeed be a typo of some kind, not sure what happened here. Addressed in https://github.com/NeotokyoRebuild/neo/commit/2b16761deeffe66494eeeac61d6ca572b66a3639

src/game/shared/neo/neo_gamerules.cpp
Rainyan: Seems to be a typo, addressed in https://github.com/NeotokyoRebuild/neo/commit/7716e9cde5e2d0d569de0101ad40f80970702bae

src/game/client/neo/ui/neo_root_settings.cpp
Rainyan: Good catch, fixed in https://github.com/NeotokyoRebuild/neo/commit/fa79e0a390cd771567ee1324d803b5f82a82cdc6

src/game/client/cdll_client_int.h
```diff
@@ -126,7 +126,36 @@ extern AchievementsAndStatsInterface* g_pAchievementsAndStatsInterface;
...
+	if (bStaticCallbackChangedCVar)
+	{
+		return;
+	}
+
+	char mutStr[STR_LIMIT_SIZE];
+	ConVarRef cvarRef(cvar);
+	V_strcpy_safe(mutStr, cvarRef.GetString());
+
+	if (V_strlen(cvarRef.GetString()) >= STR_LIMIT_SIZE)
```
sunzenshen: Optional nit: 
Just to sanity check, but is it intended that Q_UnicodeRepair will be called only when the string is over STR_LIMIT_SIZE?

I did the see the SDK comment:

// Repair invalid Unicode strings by dropping truncated characters and fixing improperly-double-encoded UTF-16 sequences.
// Unlike conversion functions which replace with '?' by default, a repair operation assumes that you know that something
// is wrong with the string (eg, mid-sequence truncation) and you just want to do the best possible job of cleaning it up.
// You can pass a REPLACE or FAIL policy if you would prefer to replace characters with '?' or clear the entire string.
// Returns nonzero on success, or 0 if the policy is FAIL and an invalid sequence was found.

And I suppose this should only be called when something is known to be wrong, but is the string size the only error case we know of in this context?

src/game/client/cdll_client_int.h
```diff
@@ -126,7 +126,36 @@ extern AchievementsAndStatsInterface* g_pAchievementsAndStatsInterface;
...
 
-
+#ifdef NEO
+template <int STR_LIMIT_SIZE>
+	requires (STR_LIMIT_SIZE > 0)
+// De-mangle bad Unicode and trim leading and trailing whitespace.
+static void NeoConVarFixPrintable(IConVar *cvar, const char *, float)
+{
+	// prevent reentrancy
+	static bool bStaticCallbackChangedCVar = false;
```
sunzenshen, line 136: Optional nit because to my understanding, only 1 translation unit from cdll_client_int.cpp is using  NeoConVarFixPrintable, but if cdll_client_int.h is #included in other contexts, wouldn't this static get duplicated?

Would an inline have accomplished the same intentions?

```cpp
template <int STR_LIMIT_SIZE>
    requires (STR_LIMIT_SIZE > 0)
inline void NeoConVarFixPrintable(IConVar *cvar, const char *, float)
{ ... }
```

src/game/client/neo/ui/neo_root_settings.cpp
```diff
@@ -1000,6 +1007,42 @@ static const wchar_t *EQUIP_UTILITY_PRIORITY_LABELS[NeoSettings::EquipUtilityPri
...
+			V_IsMeanSpaceW(input[i]));
+
+		const bool hasDoubleBlankInPos = (
+			(i + 1 < zeroIdx) &&
+			std::iswblank(input[i]) &&
+			std::iswblank(input[i + 1]));
+
+		if (hasBadCharInPos || hasDoubleBlankInPos)
+		{
+			std::wmemmove(&input[i], &input[i + 1], zeroIdx-i);
```
sunzenshen: optional nit:

supposedly Q_memmove is the equivalent Valve utility, e.g.

```cpp
Q_memmove(&input[i], &input[i + 1], (zeroIdx - i) * sizeof(wchar_t));
```

src/game/client/cdll_client_int.h
Rainyan: Oh whoops, you're right. For some reason I assumed the template with the same `STR_LIMIT_SIZE` would resolve to the same definition but that's obviously not the case here.

src/game/client/cdll_client_int.h
Rainyan: I can't remember why that length check is there, but yeah we should probably just always call that function.

PR #1530, opened 2025-12-03
Fix all warnings

Rainyan:
### Motivation

Currently, we are emitting about 7700+ compiler warnings total from a full debug+release build, across all of the CI jobs:

```sh
# Grepping the CI runner build logs case-insensitively for "warn".
# Not all of these results are warnings, but most of them are.
$ grep -ri warn . | wc -l
7769
```

Even if we assume "only" about a fourth of these warnings for the average build (Win/Linux + Release/Debug), we can still ballpark estimate there being thousands of irrelevant warnings being shown. This much noise in the build output renders the warnings not very useful, and it's unlikely the programmer will notice any new warnings introduced by their code changes from amidst them all.

This PR aims to eliminate all of those warnings to make compiler warnings useful again.

Doing the above test again on this PR's CI run, we see all the remaining grep hits are not from compiler errors:
```sh
# Now we have 0 compiler warnings!
$ grep -ri warn .
./0_Windows Native Libraries Debug.txt:2026-01-30T13:12:58.8697615Z -- Treat compile warnings as errors: ON
./0_Windows Native Libraries Debug.txt:2026-01-30T13:24:11.1965320Z -- Treat compile warnings as errors: ON
./1_Windows Native Resources.txt:2026-01-30T13:13:01.6324293Z -- Treat compile warnings as errors: ON
./1_Windows Native Resources.txt:2026-01-30T13:13:26.0546744Z -- Installing: D:/a/neo/neo/src/build/windows-release/install/neo-20260130-6375e04-resources/neo/sound/rogue/ht_nadewarn_lp.wav
./2_Linux Native Libraries Release.txt:2026-01-30T13:13:36.8785758Z -- Treat compile warnings as errors: ON
./2_Linux Native Libraries Release.txt:2026-01-30T13:27:32.1921560Z -- Treat compile warnings as errors: ON
./3_Windows Native Libraries Release.txt:2026-01-30T13:13:08.1192301Z -- Treat compile warnings as errors: ON
./3_Windows Native Libraries Release.txt:2026-01-30T13:24:24.4757435Z -- Treat compile warnings as errors: ON
./4_Linux Native Libraries Debug.txt:2026-01-30T13:13:51.2221108Z -- Treat compile warnings as errors: ON
./4_Linux Native Libraries Debug.txt:2026-01-30T13:25:33.1112196Z -- Treat compile warnings as errors: ON
```

### Summary of the changes

* Fix all compiler warnings for MSVC 2022 and the compilers in Steam Runtime SDK versions 3 and 4 (most importantly the ones used by the CI)
* Turn warnings into errors
* Update docs to reflect the changes

This is a refactor in spirit, but there have technically been some changes where a bug was discovered and fixed, and/or undefined behaviour was tamed as part of the process. In practical terms though, there should be no functional changes introduced.

## Toolchain

The following compilers are documented as "supported" in `CONTRIBUTING.md` by this PR, and they've been confirmed to not emit any warnings in release nor debug presets:

- [x] Windows MSVC v143 - VS 2022 C++ x64/x86 build tools (latest stable update as of 2026/01/30)
- [x] Windows MSVC 2022 version used by the `windows-2025` CI [runner image](https://github.com/actions/runner-images/blob/main/images/windows/Windows2025-Readme.md) (Microsoft.VisualStudio.Component.VC.Tools.x86.x64)
- [x] Linux GCC 10 steamrt3 'sniper'
- [x] Linux GCC 14 steamrt3 'sniper'
- [ ] ~~Linux Clang 11 steamrt3 'sniper'~~ wontfix: compile issues with c++20 stdlib includes
- [x] Linux GCC 14 steamrt4
- [x] Linux Clang 19 steamrt4

Of the list above, the compilers currently used by the GitHub CI builds are:
* The `windows-2025` runner image version of MSVC for Windows
* The GCC 10 steamrt3 inside the 'sniper' container for Linux

While we don't use steamrt4 for CI yet, support for it was included to make the migration smoother in the future, either to GCC 14 or Clang 19 as provided by the image.

The Windows runner image [is not currently pinnable](https://github.com/actions/runner-images/issues/2894#issuecomment-3431933836), so sporadic breakage is possible. However, we are now using the `windows-2025` label instead of `windows-latest` to provide some more stability for whenever `windows-latest` is bumped to some newer MSVC version by Microsoft.

The Linux steamrt image could be pinned; we're currently using `sniper/sdk:latest` but as it seems rather stable, pinning is probably not necessary for now. Some safeguards have been implemented by this PR [as documented](https://github.com/Rainyan/rebuild/blob/6fb9eddcd6700a6fce6f09cefd767b050f5ee7bf/CONTRIBUTING.md?plain=1#L141-L144), such as disabling warnings-as-errors for tagged release builds, to minimize breakage from unexpected build environment changes.

## Other compilers

Additionally, the following compilers were verified to not emit any warnings, although not going to "officially" maintain support for them (nor any other compilers outside the ones listed in "Toolchain"), to avoid an unreasonable maintenance burden:

- [x] Linux GCC 11.4.0-1ubuntu1\~22.04.2
- [x] Linux GCC 12.3.0-1ubuntu1\~22.04.2
- [X] Linux GCC 13.3.0-6ubuntu2\~24.04
- [X] Linux GCC 14.3.0 (compiled from source)
- [X] Linux GCC 15.2.0 (compiled from source)
- [X] Linux GCC 16-20251221 Snapshot (the latest GCC dev release as of 2025/12/28)
- [x] Linux Clang 14.0.0-1ubuntu1.1
- [x] Linux Clang 21.1.0 (the latest Clang dev release as of 2025/12/28)

These were tested on Pop!_OS 22.04 LTS.

## Out of scope

I decided to drop support for the problematic Clang 11 compiler in the steamrt3 container, because it's over 5 years old and unlikely to get patches or backports anymore, and because we do not currently use it for the CI anyway. If you want to use Clang, the steamrt4 option is the recommended one (alternatively see the "Other compilers" section of this message, although beware support for them to compile our code base not being actively maintained.

Likewise I decided to leave MSVC 2026 outside this PR, since the PR is large enough already, and because we are currently still targeting MSVC 2022 for Windows. This could be something to do in the future though.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -253,7 +253,7 @@ if(OS_WINDOWS)
             /MT
         )
 
-        if(NOT CMAKE_SIZEOF_VOID_P EQUAL 8)
+        if(NOT CPU_64BIT)
             # In x64 compilers, /Oy and /Oy- are not available.
             add_compile_options(/Oy-)
         endif()
--- a/src/tier1/CMakeLists.txt
+++ b/src/tier1/CMakeLists.txt
@@ -101,7 +101,7 @@ if(OS_WINDOWS)
     )
 
     # Currently only need exception handling for 32-bit CPU detection code
-    if(CMAKE_SIZEOF_VOID_P EQUAL 4)
+    if(CPU_32BIT)
         set_source_files_properties(processor_detect.cpp PROPERTIES COMPILE_FLAGS /EHsc)
     endif()
 endif()


src/CMakeLists.txt
```diff
@@ -115,6 +115,12 @@ elseif(CMAKE_BUILD_TYPE STREQUAL "Release" OR CMAKE_BUILD_TYPE STREQUAL "RelWith
     )
 endif()
 
+add_compile_definitions(
+	# Whether to validate narrowing conversions done with narrow_cast<T>(...)
+	# If not, will simply alias narrow_cast to static_cast
+	$<$<CONFIG:Debug>:PARANOID_NARROWING>
```
Masterkatze: It's better to put this define under `if(CMAKE_BUILD_TYPE STREQUAL "Debug")` rather than using generator expression so IDE can know this and highlight the corresponding `#ifdef` branch.

src/CMakeLists.txt
```diff
@@ -235,17 +248,16 @@ if(OS_WINDOWS)
...
-            /Ot # FavorSizeOrSpeed - Favor Fast Code
-            /GF # EnableStringPooling
-            /Gy # EnableFunctionLevelLinking
+            /O2 # Optimization - Maximize Speed
             #/Zo
-            /Oy-
             /MT
         )
 
+        if(NOT CMAKE_SIZEOF_VOID_P EQUAL 8)
```
Masterkatze: We have `CPU_64BIT` variable from `utils.cmake`

src/tier1/CMakeLists.txt
```diff
@@ -99,7 +100,10 @@ if(OS_WINDOWS)
         processor_detect.cpp
     )
 
-    set_source_files_properties(processor_detect.cpp PROPERTIES COMPILE_FLAGS /EHsc)
+    # Currently only need exception handling for 32-bit CPU detection code
+    if(CMAKE_SIZEOF_VOID_P EQUAL 4)
```
Masterkatze: We have `CPU_64BIT` variable from `utils.cmake`

Masterkatze (approved):
Apart of a couple CMake nitpicks - LGTM

PR #1532, opened 2025-12-04
JGR scores with kills instead of points, respawn cooldown after JGR death

DESTROYGIRL:
## Description
- To score points in JGR, the juggernaut must kill enemies. By default 30 kills to win
- Instead of spawning the instant the JGR dies, the dead team will spawn after an 8 second cooldown
- JGR starts locked for 20 seconds after round start
- XP of all players is reset every round
- Enabled use of comp settings in JGR
- Default round timelimit reduced to match CTG

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/shared/neo/neo_gamerules.cpp
+++ b/src/game/shared/neo/neo_gamerules.cpp
@@ -4347,13 +4347,8 @@ bool CNEORules::JuggernautItemExists() const
 		auto jgr = assert_cast<CNEO_Juggernaut*>(m_hJuggernaut.Get());
 		return jgr;
 	}
-	else
-	{
-		return m_bJuggernautItemExists;
-	}
-#else
-	return m_bJuggernautItemExists;
 #endif
+	return m_bJuggernautItemExists;
 }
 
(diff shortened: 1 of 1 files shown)


kassibuss:
20 kills might be too little for the win condition

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -2095,9 +2072,11 @@ void CNEORules::JuggernautActivated(CNEO_Player *pPlayer)
 {
 	if (GetGameType() == NEO_GAME_TYPE_JGR)
 	{
+		m_hJuggernaut = pPlayer;
```
sunzenshen, line 2075: This is me just trying to spitball nit ideas, but I wonder if there's any value to debug Assert whether JuggernautItemExists() like the similar check in CNEORules::IsJuggernautLocked().

DESTROYGIRL: Probably a good idea, I added that

AdamTadeusz (approved):

PR #1535, opened 2025-12-08
Spread out bot teammate path planning

sunzenshen:
## Description
Introduces a NavArea reservation system for bots, with the goal of reducing bots clumping together.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- Related: https://github.com/NeotokyoRebuild/neoAssets/pull/85

### DIFF
--- a/src/game/server/neo/bot/neo_bot_path_cost.cpp
+++ b/src/game/server/neo/bot/neo_bot_path_cost.cpp
@@ -109,7 +109,9 @@ float CNEOBotPathCost::operator()(CNavArea* baseArea, CNavArea* fromArea, const 
 
 		// ------------------------------------------------------------------------------------------------
 		// New path reservation related cost adjustments
-		if (neo_bot_path_friendly_reservation_enable.GetBool() && !m_bIgnoreReservations)
+		if ( neo_bot_path_friendly_reservation_enable.GetBool()
+			&& !m_bIgnoreReservations
+			&& (m_routeType != FASTEST_ROUTE) )
 		{
 			cost += CNEOBotPathReservations()->GetPredictedFriendlyPathCount(area->GetID(), m_me->GetTeamNumber()) * neo_bot_path_reservation_penalty.GetFloat();
 		}


src/game/server/neo/bot/neo_bot_path_cost.cpp
```diff
@@ -109,7 +109,9 @@ float CNEOBotPathCost::operator()(CNavArea* baseArea, CNavArea* fromArea, const
 
 		// ------------------------------------------------------------------------------------------------
 		// New path reservation related cost adjustments
-		if (neo_bot_path_friendly_reservation_enable.GetBool() && !m_bIgnoreReservations)
+		if ( neo_bot_path_friendly_reservation_enable.GetBool()
+			&& !m_bIgnoreReservations
+			&& (m_routeType != FASTEST_ROUTE) )
```
sunzenshen, line 114: After reviewing the different route types, decided that it would be appropriate to ignore friendly path reservations when determining the fastest path, as that is used for melee combat and simpler item collection routines.

To recap the different options and my understanding of them:

- DEFAULT_ROUTE, used in attacking another player. Adds a cost randomization fudge to encourage some path randomization
- FASTEST_ROUTE, no special handling originally
- SAFEST_ROUTE, we actually disabled some of this code since it was relying on some SDK routines we didn't support.  I decided it was out of scope to consider special behavior for this PR, since this option doesn't do much prior to this PR.
- RETREAT_ROUTE, also does nothing special before this PR. Maybe in a future PR we could consider where threats are looking, but that's out of scope for this PR.

src/game/server/neo/bot/behavior/nav_entities/neo_bot_nav_ent_destroy_entity.cpp
```diff
@@ -133,8 +134,7 @@ ActionResult< CNEOBot >	CNEOBotNavEntDestroyEntity::Update( CNEOBot *me, float i
 	{
 		m_repathTimer.Start( RandomFloat( 1.0f, 2.0f ) );
 
-		CNEOBotPathCost cost( me, FASTEST_ROUTE );
-		m_path.Compute( me, target->GetAbsOrigin(), cost );
+		CNEOBotPathCompute( me, m_path, target->GetAbsOrigin(), FASTEST_ROUTE );
```
sunzenshen, line 137: The reason why I changed all of the Compute/UpdateChase calls for our version of the NextBots is that no matter what a bot is doing, I want other friendlies to have the option to consider the path that bot has reserved.  If you want a bot to ignore other friendly paths, use the FASTEST_ROUTE option like this example.

src/game/server/neo/bot/neo_bot.h
```diff
@@ -800,121 +806,6 @@ inline const CNEOBot* ToNEOBot(const CBaseEntity* pEntity)
 	return static_cast<const CNEOBot*>(pEntity);
 }
 
-//--------------------------------------------------------------------------------------------------------------
```
sunzenshen, line 803: This entire declaration has been moved to neo_bot_path_cost.  I originally was tinkering with this cost, but ended up settling with a smaller tweak.  I did try to preserve the existing structure in the new location, and marked where I made the additional cost change.

src/game/server/neo/bot/neo_bot_path_cost.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+//-------------------------------------------------------------------------------------------------
+CNEOBotPathCost::CNEOBotPathCost(CNEOBot* me, RouteType routeType)
+{
+	m_me = me;
+	m_routeType = routeType;
+	m_stepHeight = me->GetLocomotionInterface()->GetStepHeight();
+	m_maxJumpHeight = me->GetLocomotionInterface()->GetMaxJumpHeight();
+	m_maxDropHeight = me->GetLocomotionInterface()->GetDeathDropHeight();
+	m_bIgnoreReservations = !neo_bot_path_friendly_reservation_enable.GetBool();
```
sunzenshen, line 23: Most of the constructor is the same, but this line is new, for disabling the reservation feature completely.

src/game/server/neo/bot/neo_bot_path_cost.cpp
```diff
@@ -0,0 +1,128 @@
...
+				const float combatDangerCost = 4.0f;
+				dist *= combatDangerCost * area->GetCombatIntensity();
+			}
+#endif
+		}
+
+
+		float cost = (dist * preference);
+
+		// ------------------------------------------------------------------------------------------------
```
sunzenshen, line 110: I tried my best to keep everything the same by comparing with the original declaration.  The new behavior is marked with this comment block, but I'm open to removing these comments if the reviewers would prefer.

PR #1534, opened 2025-12-05
Bots retreat under fire or while reloading

sunzenshen:
## Description
Bots retreat when fired upon by enemies or when they need to reload.

## Toolchain
- Windows MSVC VS2022

PR #1538, opened 2025-12-10
Bots slow down if far ahead of teammates

sunzenshen:
## Description
Leverages the walk intention interface to slow down the travel speed of bots to allow their teammates to catch up when approaching the CTG objective.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #1534 
- related #1535 

(Requires the above PRs to be accepted before, as currently bots in a team will clump together which presents a friendly fire and splash damage risk.)

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
@@ -232,6 +232,18 @@ QueryResultType CNEOBotSeekAndDestroy::ShouldHurry( const INextBot *me ) const
 // Should I slow down to let my teammates catch up?
 QueryResultType	CNEOBotSeekAndDestroy::ShouldWalk( const CNEOBot *me, const QueryResultType qShouldAimQuery ) const
 {
+	// ShouldAim query shorts cuts ShouldWalk to ANSWER_YES as aiming and running blocks each other
+	if (qShouldAimQuery == ANSWER_YES)
+	{
+		return ANSWER_YES;
+	}
+
+	const CKnownEntity* threat = me->GetVisionInterface()->GetPrimaryKnownThreat();
+	if (threat != NULL)
+	{
+		// I need to decide how to deal with a threat
+		return ANSWER_UNDEFINED;
+	}
 
 	if (NEORules()->GetGameType() == NEO_GAME_TYPE_CTG)
 	{
@@ -252,6 +264,13 @@ QueryResultType	CNEOBotSeekAndDestroy::ShouldWalk( const CNEOBot *me, const Quer
 				}
 
 				iAliveTeammates++;
+
+				const CNEO_Player *pNeoPlayer = ToNEOPlayer(pPlayer);
+				if (pNeoPlayer && pNeoPlayer->IsCarryingGhost())
+				{
+					// Maybe I can catch up to my team's ghost carrier
+					return ANSWER_UNDEFINED;
+				}
 
 				// Check if this teammate is closer to the ghost
 				float flTeammateSqDistToGhost = (pPlayer->GetAbsOrigin() - vGhostPos).LengthSqr();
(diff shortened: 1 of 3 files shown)


sunzenshen:
Requires the bot spreadout and retreat behaviors to be merged in before this PR is viable, because  as is this will cause bots to clump together into an amorphous blob:

https://github.com/user-attachments/assets/5d12a66a-4908-4294-9597-85fd10789b9a

nullsystem (dismissed):

src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
```diff
@@ -227,6 +227,82 @@ QueryResultType CNEOBotSeekAndDestroy::ShouldHurry( const INextBot *me ) const
...
+			return ANSWER_YES;
+		}
+	}
+
+	return ANSWER_UNDEFINED;
+}
+
+
+//---------------------------------------------------------------------------------------------
+QueryResultType CNEOBotSeekAndDestroy::ShouldAim(const CNEOBot* me, const bool bWepHasClip) const
```
sunzenshen, line 300: Needed to allow CNEOBotSeekAndDestroy::ShouldWalk to instantiate the abstract class.

I decided to revert my original experiment with aiming to slow down since walking already was enough to let teammates catch up with the recent related bot updates.

sunzenshen:
With the other bot updates for path spread-out and retreating merged, the pacing looks more natural in my opinion:

https://github.com/user-attachments/assets/1ebb7ed7-de7f-4851-9feb-ff94ac7c2e01

src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
```diff
@@ -227,6 +227,82 @@ QueryResultType CNEOBotSeekAndDestroy::ShouldHurry( const INextBot *me ) const
...
+					// Maybe I can catch up to my team's ghost carrier
+					return ANSWER_UNDEFINED;
+				}
+
+				// Check if this teammate is closer to the ghost
+				float flTeammateSqDistToGhost = (pPlayer->GetAbsOrigin() - vGhostPos).LengthSqr();
+				if (flTeammateSqDistToGhost < flMySqDistToGhost)
+				{
+					// Maybe I can catch up to my teammate up ahead
+					return ANSWER_UNDEFINED;
```
AdamTadeusz, line 280: So if I'm reading this correctly then only one bot per team can ever be walking at a time because they are close to the objective. Perhaps an early return here only if any team mate is closer to the objective than "me" by some tweakable value, that way some number of bots closest to the ghost can all walk at once.

If you wanted to add something like that you would probably also need to count how many alive bots are within that tolerance where a bot is considered to be the same distance from the ghost as "me", so in the case where all the bots on a team are close together they don't all walk.

AdamTadeusz (approved):

src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
sunzenshen: How many bots should be slowing down is also a good design question. In experimenting with CTG tactics in another branch, I'm wondering if slowing down any of the bots is a generally optimal strategy. For example, if the ghost was spawned closer to one team, it could make sense to just try to rush, but on the other team the same approach would lead to getting picked off one by one.  I think it might be a good idea to let this PR wait until we decide on a default bot strategy for CTG, so that we have some behavior to compare with before and after this tweak.

sunzenshen: I guess one potential heuristic we could try when the other PRs slot into place is whether we should have teams that are closer to the objective should rush, but the other team that is calculated to be farther should engage in this logic of slowing down, so they can attack together in force.

Also, while this patch currently only has 1 bot slow down for the second up front bot on the team, perhaps we could calculate a ratio of how lopsided team counts are, so teams that are evenly stacked will have fewer bots slowing down than teams that are greatly outnumbered by the other team.

PR #1540, opened 2025-12-11
Bots avoid bumping into friends

sunzenshen:
## Description
Bots when they are not going anywhere in particular will check if they are touching a teammate and then move away.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
@@ -135,43 +135,43 @@ void CNEOBotTacticalMonitor::MonitorArmedStickyBombs( CNEOBot *me )
 
 #endif //NEO
 //-----------------------------------------------------------------------------------------
-void CNEOBotTacticalMonitor::AvoidBumpingEnemies( CNEOBot *me )
+void CNEOBotTacticalMonitor::AvoidBumpingFriends( CNEOBot *me )
 {
-	if ( me->GetDifficulty() < CNEOBot::HARD )
-		return;
+	const float avoidRange = 32.0f;
 
-	const float avoidRange = 200.0f;
+	CUtlVector< CNEO_Player * > friendVector;
+	CollectPlayers( &friendVector, me->GetTeamNumber(), COLLECT_ONLY_LIVING_PLAYERS );
 
-	CUtlVector< CNEO_Player * > enemyVector;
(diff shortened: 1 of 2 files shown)


src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
```diff
@@ -127,43 +127,43 @@ void CNEOBotTacticalMonitor::MonitorArmedStickyBombs( CNEOBot *me )
...
 #endif //NEO
 //-----------------------------------------------------------------------------------------
-void CNEOBotTacticalMonitor::AvoidBumpingEnemies( CNEOBot *me )
+void CNEOBotTacticalMonitor::AvoidBumpingFriends( CNEOBot *me )
 {
-	if ( me->GetDifficulty() < CNEOBot::HARD )
-		return;
-
-	const float avoidRange = 200.0f;
+	const float avoidRange = 75.0f;
```
AdamTadeusz: perhaps this value should be lowered while we look at this erratic behavior around the objective? A standard humanoid hull is 32 hu wide, that seems like a good value, though I don't know exactly what end result you are looking for.

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.h
```diff
@@ -31,5 +31,5 @@ class CNEOBotTacticalMonitor : public Action< CNEOBot >
 	void MonitorArmedStickyBombs(CNEOBot* me);
 #endif
 
-	void AvoidBumpingEnemies(CNEOBot* me);
```
AdamTadeusz, line 36: are we certain we don't want the bots to avoid bumping into enemies?

sunzenshen: I think the likelihood that bots are going to be in close proximity with their enemies is relatively low.  In the case where they did get into a situation where they are bumping into their enemies, I don't want to interrupt the bot's combat routine to mess with their movement keys like this function does. My speculation is that this function was more for a Left 4 Dead situation where an enemy like a Witch was passive, and you don't want your friendly AI to accidentally walk over them.

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
sunzenshen: 32 hammer units seems to work okay with normal 5v5 player numbers:

https://github.com/user-attachments/assets/6b077b7c-f236-4b86-821e-9a88442635fd

AdamTadeusz: Parental_Advisory_label svg

PR #1542, opened 2025-12-11
Recon bots super jump when under fire

sunzenshen:
## Description
Recon bots super jump when under fire, after checking whether they have some room to do so.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
@@ -211,8 +211,6 @@ void CNEOBotTacticalMonitor::ReconConsiderSuperJump( CNEOBot *me )
 	me->GetLocomotionInterface()->Jump();
 	me->PressJumpButton();
 	me->SuperJump();
-
-	pNeoMe->SuitPower_Drain(SUPER_JMP_COST);
 }
 
 


src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
```diff
@@ -170,6 +171,113 @@ void CNEOBotTacticalMonitor::AvoidBumpingEnemies( CNEOBot *me )
...
+	Vector end = start + Vector( 0, 0, ceilingCheckHeight );
+	UTIL_TraceLine( start, end, MASK_PLAYERSOLID, pNeoMe, COLLISION_GROUP_PLAYER_MOVEMENT, &tr );
+
+	if ( tr.fraction < 1.0f )
+	{
+		return;
+	}
+
+	// Trajectory checks
+	const float trajectoryCheckDistance = 256.0f;
```
sunzenshen: Honestly was kind of looking for feedback on how far trajectory checks should go in Hammer units.

Probably should have different distances for the ceiling, diagonal, and feet checks.  Perhaps we should even have another traceline at eye/body level to determine if there's a wall in the way?

sunzenshen:
https://github.com/user-attachments/assets/3a75f813-07d1-4fc7-9f23-5fb8e960de10

https://github.com/user-attachments/assets/b67d9c13-7cea-4ef3-affd-d599051243ae

https://github.com/user-attachments/assets/de19d3e9-5d13-486b-9f69-98665cc0bf75

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
AdamTadeusz: I think a super jump can be very short and still effective at avoiding damage, 256 seems like too big a distance to me

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
```diff
@@ -170,6 +171,113 @@ void CNEOBotTacticalMonitor::AvoidBumpingEnemies( CNEOBot *me )
...
+
+	const CKnownEntity *pKnownThreat = me->GetVisionInterface()->GetPrimaryKnownThreat();
+	if ( !pKnownThreat || !pKnownThreat->GetEntity() )
+	{
+		return;
+	}
+
+	CBaseEntity *pThreatEntity = pKnownThreat->GetEntity();
+
+	if ( !pNeoMe->IsThreatFiringAtMe( pThreatEntity ) )
```
AdamTadeusz: I see a lot of instances where recons are taking damage but their primary threat isn't shooting at them so they just eat it. Would checking GetLastDamageTime here too be worthwhile? Presumably whether we are in cover or not also shouldn't matter if the bot is currently taking damage

AdamTadeusz: Also I guess whether we have a known threat or not probably shouldn't matter if the bot is taking damage regardless

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
```diff
@@ -170,6 +171,113 @@ void CNEOBotTacticalMonitor::AvoidBumpingEnemies( CNEOBot *me )
...
+	}
+
+	// Don't want to waste aux jumping behind cover
+	if ( !me->IsLineOfFireClear( pThreatEntity, CNEOBot::LINE_OF_FIRE_FLAGS_DEFAULT ) )
+	{
+		return;
+	}
+
+	// Check that bot isn't only moving sideways which wastes aux power
+	if ( ( pNeoMe->m_nButtons & IN_SPEED ) == 0 || ( pNeoMe->m_nButtons & ( IN_FORWARD | IN_BACK ) ) == 0 )
```
AdamTadeusz: Do we really care whether the bot is holding down their sprint key? As long as the bot is moving in a viable direction, if the bot has a good reason to super jump then they probably should. Of course IsAllowedToSuperJump will return false if the bot isn't sprinting, so if the bot passed all the other checks the bot would have to be made to sprint right before the jump

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
```diff
@@ -170,6 +171,113 @@ void CNEOBotTacticalMonitor::AvoidBumpingEnemies( CNEOBot *me )
...
+	{
+		return;
+	}
+
+	// Ceiling check
+	const float ceilingCheckHeight = me->GetLocomotionInterface()->GetMaxJumpHeight();
+	trace_t tr;
+	Vector start = pNeoMe->GetAbsOrigin();
```
AdamTadeusz: Is this starting position too low maybe for sideways checks? The slightest bit of uneven terrain will cause the traceline to hit the ground, maybe lift start by a few hu? (and if so then lower end by the same amount for the ceiling check)

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
AdamTadeusz: Also if the bot isn't moving in a viable direction but they have a good reason to super jump to evade, they should probably just pick a random direction and do it anyway

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
sunzenshen: I was actually kicking myself seeing bots superjump into obstacles, but I didn't think of the bullet avoidance aspect.  I guess it's okay if they are able to traverse at least a human body's width, and given that is about 32 hammer units, maybe 100 hu is good enough and is a round number.  I'll give that a try.

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
sunzenshen: I believe we can cheat a little and initiate SuperJump() even if the criteria of IsAllowedToSuperJump() is not fully met.

For picking a random direction, I think we can afford to let the bot decide which direction they want to go in the next tick, or if they want to decide to fight by only strafing I think that's actually fine by me.

sunzenshen: I think making the jump too random might mess with prediction, but currently I am testing on a local server so I can't notice anything with prediction yet.

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
sunzenshen: After some thought and experimentation, maybe these checks are not really needed. Especially given the discussion about how it's okay for bots to "waste" a superjump as long as it helps them dodge any bullets.

In combination with the damage trigger and needing to be already moving forwards and backwards, I think the occurrence of super jumps has been reasonable, given that the edge cases are so infrequent.  Plus bot lifetimes are so short that if they don't use it suboptimally now, when will they use it?

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
sunzenshen: As mentioned below, I decided to scrap these traceline checks for now because we would prefer the bot try to dodge and waste their aux, than to have them not use the aux and then get hit.

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
sunzenshen: For now, decided to just rely on GetLastDamageTime.  There's a chance that bots don't really need to worry about a threat if that person has horrible aim, so only jumping when getting hit might be reasonable enough.  Plus it's hard to tell sometimes when you've hit as an attacker, so it's not always obvious that this is the heuristic we are using now as long as it's not pointed out.

AdamTadeusz (approved):

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
```diff
@@ -211,8 +211,6 @@ void CNEOBotTacticalMonitor::ReconConsiderSuperJump( CNEOBot *me )
 	me->GetLocomotionInterface()->Jump();
 	me->PressJumpButton();
 	me->SuperJump();
-
-	pNeoMe->SuitPower_Drain(SUPER_JMP_COST);
```
sunzenshen, line 215: In theory, sometimes the forced sprint will lead to a debit in SUPER_JMP_COST.  But before I was thinking that maybe it doesn't.

In the end, I decided it was fine theoretically if Recon bots get a free superjump, as in reality with their current routines, it seems extremely rare that they can super jump 3 times in a row (they tend to want to fight after a super jump, so most people would not notice if they got a free superjump).  Plus sometimes they waste their superjumps, so being more generous with their  aux accounting helps them compensate for their lack of knowledge of human bunny hop techniques.

src/game/server/neo/neo_player.cpp
```diff
@@ -1443,10 +1443,14 @@ void CNEO_Player::SuperJump(void)
...
+bool CNEO_Player::IsAllowedToSuperJump(bool isBot)
 {
-	if (!IsSprinting())
-		return false;
+	// NEOJANK: Bots are exempt from certain checks due to their their erratic input control
+	if (!isBot)
+	{
+		if (!IsSprinting())
+			return false;
+	}
```
DESTROYGIRL, line 1453: Instead of adding a parameter I would just use `IsBot()` inside the function, probably cleaner but it's whatever

sunzenshen: I meant to address this earlier and then I forgot.  I'll make a PR to simplify this.

sunzenshen: Created #1587 to address this.

PR #1544, opened 2025-12-11
Basic Achievements

DESTROYGIRL:
## Description
Introduces some basic achievements laying the groundwork for future ones

For testing- paste your compiled DLLs and map into the steam version of the game and beat the tactical trial with less then 50 seconds on the clock. Good luck!

neoAssets >>>>> https://github.com/NeotokyoRebuild/neoAssets/pull/87

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/client/neo/c_neo_player.cpp
+++ b/src/game/client/neo/c_neo_player.cpp
@@ -47,6 +47,7 @@
 #include <materialsystem/itexture.h>
 #include "rendertexture.h"
 #include "ivieweffects.h"
+#include "iachievementmgr.h"
 #include "c_neo_killer_damage_infos.h"
 #include <vgui/ILocalize.h>
 #include <tier3.h>
@@ -140,6 +141,20 @@ static void __MsgFunc_IdleRespawnShowMenu(bf_read &)
 	}
 }
 USER_MESSAGE_REGISTER(IdleRespawnShowMenu);
+
+static void __MsgFunc_AchievementMark(bf_read &msg)
+{
+	IAchievementMgr *pAchievementMgr = engine->GetAchievementMgr();
+	if (!pAchievementMgr)
+	{
+		return;
+	}
+	char szString[2048];
+	msg.ReadString(szString, sizeof(szString));
+
+	pAchievementMgr->OnMapEvent(szString);
+}
+USER_MESSAGE_REGISTER(AchievementMark);
 
 ConVar cl_drawhud_quickinfo("cl_drawhud_quickinfo", "0", 0,
 	"Whether to display HL2 style ammo/health info near crosshair.",
--- a/src/game/shared/neo/achievements_neo.cpp
+++ b/src/game/shared/neo/achievements_neo.cpp
@@ -6,11 +6,17 @@ CAchievementMgr g_AchievementMgrNEO;	// global achievement mgr for NEO
 
 class CAchievementNEO_TutorialComplete : public CBaseAchievement
 {
-public:
 	void Init()
 	{
-		SetFlags( ACH_SAVE_GLOBAL );
-		SetGoal( 1 );
+		static const char *szComponents[] =
+		{
+			"NEO_TUTORIAL_RECON", "NEO_TUTORIAL_ASSAULT", "NEO_TUTORIAL_SUPPORT",
+		};
+		m_pszComponentNames = szComponents;
+		m_iNumComponents = ARRAYSIZE( szComponents );
+		SetComponentPrefix( "NEO_TUTORIAL" );
+		SetFlags( ACH_SAVE_GLOBAL | ACH_HAS_COMPONENTS | ACH_LISTEN_COMPONENT_EVENTS );
+		SetGoal( m_iNumComponents );
 		SetMapNameFilter( "ntre_class_tut" );
 	}
 };
@@ -18,7 +24,6 @@ DECLARE_NEO_ACHIEVEMENT( CAchievementNEO_TutorialComplete, ACHIEVEMENT_NEO_TUTOR
 
 class CAchievementNEO_Trial50Seconds : public CBaseAchievement
 {
-public:
 	void Init()
 	{
 		SetFlags( ACH_SAVE_GLOBAL );
@@ -30,7 +35,6 @@ DECLARE_NEO_ACHIEVEMENT( CAchievementNEO_Trial50Seconds, ACHIEVEMENT_NEO_TRIAL_5
 
 class CAchievementNEO_Trial40Seconds : public CBaseAchievement
 {
-public:
 	void Init()
 	{
 		SetFlags( ACH_SAVE_GLOBAL );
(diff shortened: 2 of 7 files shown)


src/game/client/CMakeLists.txt
```diff
@@ -1621,6 +1621,8 @@ target_sources_grouped(
     TARGET client
     NAME "Source Files\\Shared"
     FILES
+    ${CMAKE_SOURCE_DIR}/game/shared/neo/achievements_neo.cpp
+    ${CMAKE_SOURCE_DIR}/game/shared/neo/achievements_neo.h
```
Rainyan, line 1625: We are failing this assertion for the client dll:

`Assert( NULL == pAchievementMgr );`

>	client.dll!CAchievementMgr::Init() Line 301	C++
 	client.dll!IGameSystem::InitAllSystems() Line 220	C++
 	client.dll!CHLClient::Init(void *(*)(const char *, int *) appSystemFactory, void *(*)(const char *, int *) physicsFactory, CGlobalVarsBase * pGlobals) Line 1129	C++
 	[External Code]	

This appears to be because the server dll has already constructed the `g_AchievementMgrNEO` achievements singleton, and when we call `engine->GetAchievementMgr` from client, it will return that listen-server instance of `g_AchievementMgrNEO` which already exists.

I assume we would only want to declare this `g_AchievementMgrNEO` singleton client-sided or server-sided, but not both to avoid this issue of double achievement managers existing.

src/game/server/neo/neo_message.cpp
```diff
@@ -120,5 +122,18 @@ void CNEO_Message::InputStopTimer(inputdata_t& inputData)
...
+			if ((gpGlobals->curtime - m_flTimerStart) < 50.0f)
+			{
+				g_AchievementMgrNEO.AwardAchievement(ACHIEVEMENT_NEO_TRIAL_50_SECONDS);
+
+				if ((gpGlobals->curtime - m_flTimerStart) < 40.0f)
+				{
+					g_AchievementMgrNEO.AwardAchievement(ACHIEVEMENT_NEO_TRIAL_40_SECONDS);
+				}
+			}
+		}
```
Rainyan: It looks like we're awarding the achievement from server code, which I assume works for the time trial because it's local-hosted. But would this `g_AchievementMgrNEO.AwardAchievement` stuff used here work in a multiplayer environment, where the server is remote and different from the client? I don't have a key for the NTRE SRCDS so I can't verify this, but maybe it's worth checking?

Since I don't see `AwardAchievement` specifying which client is being awarded with the achievement, I think this might not work remotely, but I'm unsure. Should we be perhaps catching some kind of achievement related event client-side, and award from there, or something like that?

Rainyan (commented):
I can verify I can unlock the achievements with this code, but left some notes.

src/game/server/neo/neo_message.cpp
DESTROYGIRL: Yes, I misunderstood what was being done in TF, will change achievements to be clientside. I thought it to be a much more complex system but you can just do `pPlayer->AwardAchievement()` and as long as the achievement is declared clientside it should work

src/game/client/neo/c_neo_player.cpp
```diff
@@ -141,6 +142,20 @@ static void __MsgFunc_IdleRespawnShowMenu(bf_read &)
...
 
+static void __MsgFunc_AchievementMark(bf_read &msg)
+{
+	IAchievementMgr *pAchievementMgr = engine->GetAchievementMgr();
+	if (!pAchievementMgr)
+	{
+		return;
+	}
+	char szString[2048];
+	msg.ReadString(szString, sizeof(szString));
```
Rainyan, line 153: Nit: In the future, we might want to refactor this to have some means of getting the longest achievement name length instead of allocating with a magic value, since the 2 KiB buffer seems a bit overkill. Not that it has much of a difference in practice.

src/game/shared/neo/achievements_neo.cpp
```diff
@@ -0,0 +1,46 @@
+#include "cbase.h"
+#ifdef CLIENT_DLL
```
Rainyan, line 2: Nit: We can revisit this later, but if we don't need this implementation server side, then it should probably go from shared->client folder, and be dropped from the server CMakeLists, rather than ifdef-guarding the entire logic of the cpp file.

src/game/shared/neo/achievements_neo.cpp
```diff
@@ -0,0 +1,46 @@
...
+{
+	void Init()
+	{
+		SetFlags( ACH_SAVE_GLOBAL );
+		SetGoal( 1 );
+		SetMapNameFilter( "ntre_shooting_tut" );
+	}
+};
+DECLARE_NEO_ACHIEVEMENT( CAchievementNEO_Trial40Seconds, ACHIEVEMENT_NEO_TRIAL_40_SECONDS, "NEO_TRIAL_40_SECONDS" );
+#endif
```
Rainyan, line 46: Nit: Since the class implementation of both `CAchievementNEO_Trial50Seconds` and `CAchievementNEO_Trial40Seconds` is identical, we could probably refactor this as a single class called `CAchievementNEO_Trial` or similar, and then do:
```cpp
DECLARE_NEO_ACHIEVEMENT( CAchievementNEO_Trial, ACHIEVEMENT_NEO_TRIAL_50_SECONDS, "NEO_TRIAL_50_SECONDS" );
DECLARE_NEO_ACHIEVEMENT( CAchievementNEO_Trial, ACHIEVEMENT_NEO_TRIAL_40_SECONDS, "NEO_TRIAL_40_SECONDS" );
```
to avoid the redundancy of declaring two classes with identical behaviour.

Better yet, since in `src/game/shared/neo/achievements_neo.h`, we declare:
```cpp
class CAchievementNEOBasic : public CBaseAchievement
{
public:
	void Init()
	{
		SetFlags( ACH_SAVE_GLOBAL );
		SetGoal( 1 );
	}
};
```
we could inherit from `CAchievementNEOBasic` instead of `CBaseAchievement`, so we can get the `SetFlags` and `SetGoal` from the base implementation in the override:
```cpp
class CAchievementNEO_Trial : public CAchievementNEOBasic
{
public:
	virtual void Init() override
	{
		CAchievementNEOBasic::Init();
		SetMapNameFilter( "ntre_shooting_tut" );
	}
};
```

DESTROYGIRL: It was my assumption that all achievement classes needed to be unique, as that's what all other games have done, and is what the macros (the basic ones implemented in this pr and ones defined in baseachievement) do as well. I only set them up with their own classes to use `SetMapNameFilter`, otherwise would have used `DECLARE_BASIC_NEO_ACHIEVEMENT`. But further investigation can be done to see how it actually works

src/game/shared/neo/achievements_neo.cpp
DESTROYGIRL: Following the structure in TF/HL/SDK of all achievement related systems being shared, maybe in case we wanted some to be exclusively granted server side somehow

src/game/client/neo/c_neo_player.cpp
DESTROYGIRL: Ur right, there is probably a limit defined somewhere. Will change it in the next pr

PR #1541, opened 2025-12-11
Fix: Bot didn't have plan when repathing

sunzenshen:
## Description
Sometimes a bot doesn't have a path plan when the repathing code triggers around a friendly which leads to an access error.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #1535

PR #1545, opened 2025-12-12
Bots evade from grenades

sunzenshen:
## Description
In their tactical monitor, bots check for grenade entities around them and then attempt to evade any grenades found using retreating logic.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #1534 
^ (reference for new behavior class but tweaked to avoid a grenade entity)


src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -1916,6 +1916,12 @@ void CNEOBot::RepathIfFriendlyBlockingLineOfFire()
 		return;
 	}
 
+	const PathFollower* pPath = GetCurrentPath();
+	if (pPath == NULL)
```
sunzenshen: From: https://github.com/NeotokyoRebuild/neo/pull/1541/changes#diff-3b7f13a60108836218d7b05121bf08885f6ebabc4e3628d60213db76c6ffde67R1923

PR #1546, opened 2025-12-12
background map panning

AdamTadeusz:
## Description

The camera moves slightly with the cursor when loaded into a background map. Also fixed the background map fov not being 75

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #
- related #

### DIFF
--- a/src/game/client/view.cpp
+++ b/src/game/client/view.cpp
@@ -739,7 +739,9 @@ float CViewRender::GetZFar()
 }
 
 
+#ifdef NEO
 ConVar cl_neo_background_pan("cl_neo_background_pan", "1", FCVAR_ARCHIVE, "Pan the camera with the cursor in the main menu background maps");
+#endif // NEO
 //-----------------------------------------------------------------------------
 // Sets up the view parameters
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


src/game/client/view.cpp
```diff
@@ -736,6 +739,7 @@ float CViewRender::GetZFar()
 }
 
 
+ConVar cl_neo_background_pan("cl_neo_background_pan", "1", FCVAR_ARCHIVE, "Pan the camera with the cursor in the main menu background maps");
```
DESTROYGIRL, line 743: This needs an ifdef neo around it

```suggestion
#ifdef NEO
ConVar cl_neo_background_pan("cl_neo_background_pan", "1", FCVAR_ARCHIVE, "Pan the camera with the cursor in the main menu background maps");
#endif
```

src/game/client/view.cpp
```diff
@@ -785,6 +789,44 @@ void CViewRender::SetUpViews()
...
+				vgui::input()->GetCursorPos(x, y);
+				float flX = ((float)x / viewEye.width) - 0.5;
+				float flY = ((float)y / viewEye.height) - 0.5;
+
+				// S curve with a minimum value of -0.5 and maximum value of 0.5
+				constexpr float CURVE_STEEPNESS = 5;
+				flX = (1 / (1 + pow(2, -CURVE_STEEPNESS * flX))) - 0.5;
+				flY = (1 / (1 + pow(2, -CURVE_STEEPNESS * flY))) - 0.5;
+
+				constexpr int CAMERA_MOVEMENT_MULTIPIER = 5;
```
DESTROYGIRL: The effect is very good but I think it's best applied subtly and 5 is a bit much. Consider changing it to 2 or 3, but it's down to u

PR #1548, opened 2025-12-13
Fix eye angles going out of netprop range

### LINKED ISSUE #1375: m_angEyeAngles can go out of range on players which will spam the console
SourceTV can do this as well, which will flood the server console forever.

Example: (on ntre_isolation_ctg)
`DataTable warning: player: Out-of-range value (358.000000/90.000000) in SendPropFloat 'm_angEyeAngles[0]', clamping.`

DESTROYGIRL:
## Description
Clamp the pitch at which the player initially spawns

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1375

### DIFF
--- a/src/game/shared/gamerules.cpp
+++ b/src/game/shared/gamerules.cpp
@@ -210,10 +210,22 @@ CBaseEntity *CGameRules::GetPlayerSpawnSpot( CBasePlayer *pPlayer )
 	pPlayer->SetLocalOrigin( pSpawnSpot->GetAbsOrigin() + Vector(0,0,1) );
 #endif // NEO
 	pPlayer->SetAbsVelocity( vec3_origin );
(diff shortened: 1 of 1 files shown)


Masterkatze (approved):

DESTROYGIRL:
You can also just go on rise and get the error a few times just by loading into the map

DESTROYGIRL:
It'd be better to clamp there I guess, but since the angle is 358 and the limits -90 to 90 I wonder if this an editor issue where mappers might tilt the entity angle to one that is in range but in the wrong direction, because 358 would be the same as -2 here right?

AdamTadeusz:
image
The clamp does place the camera at a rather unfortunate angle of 89

AdamTadeusz (approved):

PR #1557, opened 2025-12-17
Include map editor configs and FGD in resources zip

### LINKED ISSUE #1528: game/bin should be included in the resources zip
`bin` needs to be included in the resources zip alongside `neo` for the steam build so Hammer configurations and the FGD can be distributed.

DESTROYGIRL:
## Description
I am guessing

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1528

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -557,7 +557,7 @@ endif()
 
 if(NEO_INSTALL_RESOURCES)
     install(
-        DIRECTORY ${CMAKE_SOURCE_DIR}/../game/neo
+        DIRECTORY ${CMAKE_SOURCE_DIR}/../game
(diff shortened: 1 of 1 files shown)


DESTROYGIRL:
It works 🎊

Masterkatze (approved):

sunzenshen (approved):
I downloaded the artifacts from the GitHub Actions of this PR, and then pointed my Hammer.exe level editor to the FGD file.

Image

Then I opened up one of the existing tutorial maps and double clicked on an entity to see that it was recognized as an NTRE asset.

Image

(I don't know what I'm doing so I wasn't able to run the map, but I think this is enough evidence that the right files are being packaged?)

PR #1558, opened 2025-12-18
Crouch-jump constants recalc for bot ledge NavAreas

sunzenshen:
## Description
Aims to fix bots trying to climb ledges that they don't have enough crouch-jump height to pull off, by recalculating the underlying constants used by the bot locomotion interface to determine climbing eligibility for adjacent NavAreas.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #1535 (This PR made the issue more obvious because earlier all bots would follow easy paths, but the newer dispersal logic encouraged bots to take paths that were of impossible jump height differences for their class choice.)

### DIFF
--- a/src/game/server/neo/bot/neo_bot_locomotion.h
+++ b/src/game/server/neo/bot/neo_bot_locomotion.h
@@ -47,12 +47,12 @@ inline float CNEOBotLocomotion::GetMaxJumpHeight( void ) const
 			return NEO_RECON_CROUCH_JUMP_HEIGHT;
 		case NEO_CLASS_JUGGERNAUT:
 			return NEO_JUGGERNAUT_CROUCH_JUMP_HEIGHT;
-		case NEO_CLASS_ASSAULT:
-			[[fallthrough]];
 		case NEO_CLASS_SUPPORT:
-			[[fallthrough]];
+			return NEO_SUPPORT_CROUCH_JUMP_HEIGHT;
 		case NEO_CLASS_VIP:
-			return NEO_CROUCH_JUMP_HEIGHT;
+			return NEO_ASSAULT_CROUCH_JUMP_HEIGHT; // vip same as assault
+		case NEO_CLASS_ASSAULT:
+			return NEO_ASSAULT_CROUCH_JUMP_HEIGHT;
 		default:
 			Assert(false);
 			return 0.f;
--- a/src/game/shared/neo/neo_player_shared.h
+++ b/src/game/shared/neo/neo_player_shared.h
@@ -152,20 +152,24 @@ COMPILE_TIME_ASSERT(NEO_ASSAULT_CROUCH_SPEED == NEO_VIP_CROUCH_SPEED);
 // Support: 70 - 59 = 11
 // VIP: 65 - 48 = 17
 //
-// Juggernaught: 88 - 75 = 13
+// Juggernaut: 88 - 75 = 13
 //   (Base Hull Max = 70) + (NEO_JUGGERNAUT_MAXHULL_OFFSET.z = 18) = 88
 //   (Base Hull Duck Max = 59) + (NEO_JUGGERNAUT_DUCK_MAXHULL_OFFSET.z = 16) = 75
 //
-// Proposed bot crouch jump heights:
+// Then apply a uniform spare height budget gap of 7 units to all classes.
+// Formula: (Class Jump Height) + (Lift) - (Spare Gap)
+//
+// Precalculated bot crouch jump heights for use in bot locomotion/navigation checks:
 // ---
-// Recon: 54 + 11 = 65
-// Assault/VIP = 36 + 11 = 47
-// Support = 36 + 11 = 47
-// Juggernaut = 50.4 + 11 = 61.4 (rounded for simplicity)
+// Recon: 54 + 18 - 7 = 65
+// Assault/VIP: 36 + 17 - 7 = 46
+// Support: 36 + 11 - 7 = 40
+// Juggernaut: 50.4 + 13 - 7 = 56.4 (rounded down to 56)
 
 #define NEO_RECON_CROUCH_JUMP_HEIGHT 65.f
-#define NEO_CROUCH_JUMP_HEIGHT 47.f
-#define NEO_JUGGERNAUT_CROUCH_JUMP_HEIGHT 62.f
+#define NEO_ASSAULT_CROUCH_JUMP_HEIGHT 46.f
+#define NEO_SUPPORT_CROUCH_JUMP_HEIGHT 40.f
+#define NEO_JUGGERNAUT_CROUCH_JUMP_HEIGHT 56.f
 
 // END OF NEO MOVEMENT DEFINITIONS
 //////////////////////////////////////////////////////


sunzenshen:
As a demonstration of the before/after A/B test for this fix, I used a cherry-pickable version of PR #1345 to demonstrate the effect these height constants have on pathing:

Before (on branch #1345 without the fixes from this #1558 PR):

https://github.com/user-attachments/assets/8914e9b2-78e3-44bb-b2eb-45f4f505a0d7

After (by cherry picking #1345 onto a temporary copy of this branch #1558):

https://github.com/user-attachments/assets/84f0c700-cec8-4857-9f21-bab9b3462af6

(All that said, does it feel like in the second video that having an exact jump+lift threshold is a bit tough for Support bots to navigate?  But would that limit their ability to traverse areas that human players can with exact timing?)

src/game/shared/neo/neo_player_shared.h
```diff
@@ -120,8 +120,52 @@ COMPILE_TIME_ASSERT(NEO_RECON_CROUCH_SPEED > NEO_ASSAULT_CROUCH_SPEED);
...
+//
+// Juggernaught: 88 - 75 = 13
+//   (Base Hull Max = 70) + (NEO_JUGGERNAUT_MAXHULL_OFFSET.z = 18) = 88
+//   (Base Hull Duck Max = 59) + (NEO_JUGGERNAUT_DUCK_MAXHULL_OFFSET.z = 16) = 75
+//
+// Proposed bot crouch jump heights:
+// ---
+// Recon: 54 + 11 = 65
+// Assault/VIP = 36 + 11 = 47
+// Support = 36 + 11 = 47
```
sunzenshen: To start the conversation, in my testing on different maps, it seems like Support struggles the most when trying to jump onto elevation changes that are within this threshold, while Assault and Recon are generally more adaptable given their more conservative thresholds.  Maybe we should consider some sort of standard lift gap and make that consistent between all the classes?  For example, what about a standard gap of 7?

e.g.

Recon: 54 + 18 - 7 = 65
Assault/VIP: 36 + 17 - 7 = 46
Support: 36 + 11 - 7 =  40
Juggernaut: 50.4 + 13 - 7 =  56.4 (or round down to 56)

sunzenshen: It's really important for behavioral smoothness for Support bots to have a different navigation height clearance so I went ahead with applying the consistent 7 unit gap to all classes, such that Recon has the same height budget as before.

sunzenshen:
Clicked the rebase branch in the GitHub web UI, so this rebased build should have the bot commands feature.

One can test the jump heights attempts in maps by using the following commands:

```
sv_neo_bot_cmdr_enable 1;
neo_ctg_round_timelimit 99;
```

Then one can press use on the paused teammate bots to get them to follow you, and then use the ping button to have bots walk between different areas that might be suspicious.

src/game/server/neo/bot/neo_bot_locomotion.h
```diff
@@ -38,19 +38,21 @@ class CNEOBotLocomotion : public PlayerLocomotion
 
 inline float CNEOBotLocomotion::GetMaxJumpHeight( void ) const
 {
+	// NEO JANK: Assumes [MD]'s g_bMovementOptimizations = true, where we assume sv_gravity is 800 for navigation.
+	// Changing that setting can potentially break bot navigation.
```
Rainyan, line 42: Would be good to debug-assert sv_gravity == 800 here if we're hardcoding that assumption, so that this blows up more prominently when that is not the case.

src/game/shared/neo/neo_player_shared.h
```diff
@@ -120,8 +120,56 @@ COMPILE_TIME_ASSERT(NEO_RECON_CROUCH_SPEED > NEO_ASSAULT_CROUCH_SPEED);
...
+// Recon: 54 + 18 - 7 = 65
+// Assault/VIP: 36 + 17 - 7 = 46
+// Support: 36 + 11 - 7 = 40
+// Juggernaut: 50.4 + 13 - 7 = 56.4 (rounded down to 56)
+
 #define NEO_RECON_CROUCH_JUMP_HEIGHT 65.f
-#define NEO_CROUCH_JUMP_HEIGHT 56.f
+#define NEO_ASSAULT_CROUCH_JUMP_HEIGHT 46.f
+#define NEO_SUPPORT_CROUCH_JUMP_HEIGHT 40.f
+#define NEO_JUGGERNAUT_CROUCH_JUMP_HEIGHT 56.f
```
Rainyan, line 172: Are these values intended to reflect the actual maximum crouch-jump heights for the classes? The recon (65 units) and assault (46 units) values are accurate, however the support value seems incorrect. Supports have the same max jump height as the assault, AFAIK.

You can test this in the `devbox` map with `sv_neo_can_change_classes_anytime 1`, using the ledges that go up in 1 unit increments (there's a visual marker per every 16 units).

As for the JGR, I have no idea, couldn't figure out how to spawn as one. I guess I can try to recompile the devbox in JGR mode later...

Also should be noted, the 65 and 46 values are for the crouch-first-then-jump (or possibly crouch & jump on same exact tick? i'm unsure) maximums; for a more typical user input (jump-then-crouch), the max heights are about 1 unit less. But I guess the bots should know how to use the optimum one, because some jumps in some maps do indeed use the crouch-first value as a form of a "trick jump", or just by coincidence, and we probably do want the bots to be able to perform these jumps that normal players also could.

Rainyan: Oh, I see you've touched upon this in https://github.com/NeotokyoRebuild/neo/pull/1558#discussion_r2649367093 already.

Rainyan: Personally not a huge fan of fuzzing the variables with some magic offset, especially with a descriptive name like `<CLASSNAME>_CROUCH_JUMP_HEIGHT` - would prefer we separated any such magic as its own value separate from what we know is the real max height clearance. But for the sake of getting these nav improvements in the game, I guess we can merge this and leave the refactoring for another time.

One way we could decouple the real jump heights could perhaps be to return `<CLASSNAME>_CROUCH_JUMP_HEIGHT + /* some offset as required */` from `CNEOBotLocomotion::GetMaxJumpHeight`, so it was more obvious what's going on when looking at the code, rather than directly modifying the defines to differ from the empirical values.

sunzenshen: I know what you mean with questioning where the extra fuzz value even came from, but the rationale for keeping it was that the original Recon crouch jump height also has a questionable origin as I couldn't think of any other reason why it was the original value.  I'd like to start off with a consistent set of values that are conservatively almost guaranteed to be appropriate for bots in most reasonable circumstances, and I'm open to following up later with another PR that investigates constants that are more accurate to the true max jump behavior.

src/game/server/neo/bot/neo_bot_locomotion.h
sunzenshen: New PR opened to address this in #1626.

src/game/shared/neo/neo_player_shared.h
sunzenshen: New PR opened to address this in #1626.

PR #1443, opened 2025-10-25
Don't reset until after the post round intermission.

Agiel:
## Description
Set a flag when a winner has been determined, and check that when it's time to start the new round.

## Toolchain

- Windows MSVC VS2022

## Linked Issues
- fixes #1437
- fixes #1439

PR #1569, opened 2025-12-23
Fix games not ending if match is won by hitting score limit

### LINKED ISSUE #1563: Games never end if won by hitting score limit
### Build Info

v22.0

### Description

At the end of a game, when a team wins by score, the map does not change. Probably caused by #1443 

### To Reproduce

1. Win a map by score
2. Be stuck in limbo

### Expected behavior

Map should change

### Actual behavior

_No response_

### Operating System

- [ ] Windows
- [ ] Linux

### Version/Distro

Wind blows, love dies

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

DESTROYGIRL:
## Description
Dont set `m_flNeoNextRoundStartTime = FLT_MAX` or else the match end logic is never run

I am sleepy. Maybe I am wrong

Testers: Ensure this does not break scenarios like `neo_round_limit` = `0` where `neo_score_limit` has a value and vice versa... and maybe check a different gamemode like JGR just to be safe
Also make sure that `m_flNeoNextRoundStartTime` was not set to float max for a proper reason, I assume it was originally there to prevent new rounds starting during the map change?

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1563 
- related #1443

### DIFF
--- a/src/game/client/neo/c_neo_player.cpp
+++ b/src/game/client/neo/c_neo_player.cpp
@@ -557,7 +557,7 @@ void C_NEO_Player::CheckVisionButtons()
 
 void C_NEO_Player::CheckLeanButtons()
 {
-	if (!IsAlive())
+	if (!IsAlive() || GetFlags() & FL_FROZEN)
 	{
 		return;
 	}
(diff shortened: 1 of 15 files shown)


sunzenshen (approved):
LGTM:

https://github.com/user-attachments/assets/1347b9a7-9af6-4c4d-bc99-ddca88635da9

PR #1570, opened 2025-12-23
Improve release assets compression

Rainyan:
### Summary

* Change the GitHub release assets' archive format from ZIP to 7-Zip
* Change the archive compression algorithm from DEFLATE to LZMA

Optimize release assets for file size, because the compress/decompress performance difference between these settings is negligible for our data.

At time of this commit, the total (release + debug) assets size before this patch is about 946 MB. After this patch, the total size is about 638 MB. So this change results in about 33% size (and download time) reduction for the release assets.

### Users

Because the Windows file explorer GUI natively supports 7-Zip decompression since Windows 11 22H2 (released late 2022), this change should not affect the regular user experience for people with an up-to-date Windows machine. And since 7z is a fairly well-known format, I doubt it should cause confusion even for those whose OS doesn't have native extraction capability.

### Steam release script

This is a breaking change for the Steam release script, because both the archive format (ZIP -> 7-Zip) and the file extension (`<filename>.zip` -> `<filename>.7z`) have changed. As such, this PR should not be merged before confirming compatibility with the Steam release script.

The actual file name patterns aside from the file extension remain unchanged.

### The asset format

Alternatively, we could've changed the ZIP algorithm used to LZMA, as support for it was [added to the ZIP spec](https://pkwaredownloads.blob.core.windows.net/pkware-general/Documentation/APPNOTE-6.3.0.TXT) already in 2006, but as of this PR date, Windows file explorer still does not natively handle LZMA-compressed ZIP archives extraction. Therefore, to avoid users confusing LZMA-compressed ZIPs as being corrupted, I chose to instead change the entire archive format to 7-Zip (which is fully compatible for native .7z extraction with modern Windows 11).

### Compressing intermediate build artifacts

I also experimented with changing the upload-artifact action's [compression config](https://github.com/actions/upload-artifact?tab=readme-ov-file#altering-compressions-level-speed-v-size) between 1 (best speed) and 9 (best compression) from the default value of 6, but this made no meaningful difference in the CI job speeds (+/- a couple seconds, which could've been random load variance). So I decided to not change that from the default.

## Toolchain
N/A

### DIFF
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -337,7 +337,7 @@ jobs:
       uses: actions/download-artifact@v4
 
     - name: Packing artifacts
-      run: 'parallel 7z a -tzip -mx=9 "{}.zip" "./{}/*" ::: neo-*'
+      run: 'parallel 7z a -t7z -m0=lzma -mx=9 "{}.7z" "./{}/*" ::: neo-*'
 
     - name: Create & upload latest build
       env:
@@ -347,7 +347,7 @@ jobs:
         # Delete old "latest" (delete step always returns true so that even if there's no release, it still uploads)
         gh release delete latest --cleanup-tag --yes --repo "$REPO_INFO" || true
         # Create new "latest"
-        gh release create latest neo-*.zip --prerelease --title "Latest Master Build" --repo "$REPO_INFO"
+        gh release create latest neo-*.7z --prerelease --title "Latest Master Build" --repo "$REPO_INFO"
 
   upload-release:
     name: Upload Release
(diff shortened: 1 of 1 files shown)


Rainyan:
There's an example output of the 7z assets for preview here: https://github.com/Rainyan/rebuild/releases/tag/latest

And in contrast, the current latest dev release assets of the main branch: https://github.com/NeotokyoRebuild/neo/releases/tag/latest

blaberry:
LGTM, lmk when this is merged and I'll update the build script accordingly.

DESTROYGIRL (approved):
FYI, the release linked isn't valid anymore but I'm going to take your word for it

sunzenshen (approved):

PR #1573, opened 2025-12-24
Reduce bot ammo waste from spraying

sunzenshen:
## Description
Reduces bot ammo waste from spraying, by reducing the minimum time spraying based on difficulty, while also interrupting sprays when the line of fire checks fail when firing at an enemy.

Also incidentally fixes bots being able to shoot as the Juggernaut, as the player weapon relied on a novel weapon cooling mechanic that led to the refusal to fire for bots.

## Toolchain
- Windows MSVC VS2022


src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
```diff
@@ -18,7 +18,6 @@ ConVar neo_bot_path_lookahead_range( "neo_bot_path_lookahead_range", "300" );
 ConVar neo_bot_sniper_aim_error( "neo_bot_sniper_aim_error", "0.01", FCVAR_CHEAT );
 ConVar neo_bot_sniper_aim_steady_rate( "neo_bot_sniper_aim_steady_rate", "10", FCVAR_CHEAT );
 ConVar neo_bot_debug_sniper( "neo_bot_debug_sniper", "0", FCVAR_CHEAT );
-ConVar neo_bot_fire_weapon_min_time( "neo_bot_fire_weapon_min_time", "1", FCVAR_CHEAT );
```
sunzenshen, line 21: Before, bots would hold down the fire button for an entire second after trying to fire at an enemy with an automatic weapon.  They would continue to do so even when the target ducked behind cover or happened to cross line of fire with a friendly, during which the bot laying down the spray would waste ammo into a wall/obstruction or a friendly without pause.

Masterkatze (approved):

Issue #1510, opened 2025-11-25
Grenades should not block each other


Rainyan:
### Build Info

v21.1-alpha

### Description

If multiple players throw utility from the same position at the same time, the thrown utility gets stuck in air.

### To Reproduce

* Set commands:
```
sv_neo_can_change_classes_anytime 1;
bot_mimic 1;
bot_class 2;
map devbox;
```
* Spawn in the level and join a player team
* Herd the bots to overlap the same position (should already occur for map `devbox` initially)
* Use `bot_mimic` to synchronize a nade throw
  * Tested with `weapon_grenade` and `weapon_smokegrenade`
  * Yet to be tested: test with `weapon_remotedet`

### Expected behavior

Nades do not interact.

### Actual behavior

Nades get stuck to each other.

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

11

### Additional context and Screenshots

https://github.com/user-attachments/assets/96bf61e0-e487-48b2-acbd-e3d05acb5335

PR #1575, opened 2025-12-24
Fix projectiles colliding with dropped weapons

### LINKED ISSUE #1567: Projectiles collide with dropped weapons
Tested on oilstain

DESTROYGIRL:
## Description
In #915 and #891 projectiles were set to use `COLLISION_GROUP_WEAPON` which is probably incorrect as they collide with dropped weapons which is not parity. Changing them to `COLLISION_GROUP_PROJECTILE` does not appear to bring back the colliding with players issue nor have I found any side effects of this change...

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1567


Rainyan:
Fixes #1510

PR #1578, opened 2025-12-29
Make unmanned JGR solid, with soft-collide post-death transitions

Rainyan:
## Description

This commit makes the JGR prop solid when it is uncontrolled and idle in the world. Once a player occupies the JGR, "it" (really the player entity) starts (or rather, continues) abiding by the usual player collision rules.

Once the JGR dies, the unmanned prop will transition back to being solid through a brief period of soft-collision. The soft-collision period is used to avoid nearby players getting stuck inside the collider during the collision mode transition.

The soft collisions are implemented as a child point_push object, which will push away any players inside the collider until the JGR think routine has confirmed its vicinity is clear for becoming solid again.

The easiest way to test this is to use `bot_mimic 1` to have a bot take over the JGR. Then, shoot the JGR until it explodes at very close range to see the soft-collisions in action.

## Motivations

The JGR prop being so huge felt to me a little awkward to rush quickly towards for pressing `+use` on it, because you could accidentally phase through it when rushing. It also feels a bit strange trying to shoot at someone taking over the JGR, since they can stand inside the prop. I think it would make sense for the non-controlled JGR objective to have solid collisions for a better game feel for the above reasons.

@DESTROYGIRL Maybe you could take a look, since you are kind of the "code owner" of this particular game mode? Or if you think it should go through a design doc then so be it, but I was hoping this would be uncontroversial enough to pass as a regular PR.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/shared/neo/neo_juggernaut.cpp
+++ b/src/game/shared/neo/neo_juggernaut.cpp
@@ -222,13 +222,14 @@ void CNEO_Juggernaut::MakePushThink()
 {
 	if (IsMarkedForDeletion()) return;
 
+	constexpr const char* pushEnt = "point_push";
+
 	if (m_hPush.Get())
 	{
-		AssertMsg1(false, "push already exists; called %s multiply?", __FUNCTION__);
+		AssertMsg2(false, "%s already exists; called %s multiply?", pushEnt, __FUNCTION__);
 		return;
 	}
 
-	constexpr const char* pushEnt = "point_push";
 	if (!(m_hPush = CreateEntityByName(pushEnt)))
 	{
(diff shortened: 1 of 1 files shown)


sunzenshen (approved):

Rainyan:
> Very cool! Though in testing I found that the soft collisions could quite easily be kept on as the point_push doesn't seem to push a player far enough to enable the hard collision even without any resistance from them. Maybe worth tweaking the push radius / magnitude?

Yeah, I noticed the same with bots standing nearby, but figured it would be safer to err on the side of caution, to minimize the chances of someone ever getting stuck. We could probably fine-tune the value based on future playtests for sure.

PR #1579, opened 2025-12-29
JGR mode behavior dispatcher for bots

sunzenshen:
## Description
Encapsulates bot behaviors related to the JGR game mode such that modifications to the behavior of Juggernaut mode bots can be isolated away from general seek and destroy behavior.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/CMakeLists.txt
+++ b/src/game/server/CMakeLists.txt
@@ -1440,10 +1440,10 @@ target_sources_grouped(
     neo/bot/behavior/neo_bot_dead.h
     neo/bot/behavior/neo_bot_jgr_capture.cpp
     neo/bot/behavior/neo_bot_jgr_capture.h
-    neo/bot/behavior/neo_bot_jgr_escort.cpp
-    neo/bot/behavior/neo_bot_jgr_escort.h
     neo/bot/behavior/neo_bot_jgr_enemy.cpp
     neo/bot/behavior/neo_bot_jgr_enemy.h
+    neo/bot/behavior/neo_bot_jgr_escort.cpp
+    neo/bot/behavior/neo_bot_jgr_escort.h
     neo/bot/behavior/neo_bot_jgr_juggernaut.cpp
     neo/bot/behavior/neo_bot_jgr_juggernaut.h
     neo/bot/behavior/neo_bot_jgr_seek.cpp
--- a/src/game/server/neo/bot/behavior/neo_bot_jgr_capture.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_jgr_capture.cpp
@@ -6,7 +6,7 @@
 #include "neo_juggernaut.h"
 
 //---------------------------------------------------------------------------------------------
-CNEOBotJgrCapture::CNEOBotJgrCapture( CBaseEntity *pObjective )
+CNEOBotJgrCapture::CNEOBotJgrCapture( CNEO_Juggernaut *pObjective )
 {
 	m_hObjective = pObjective;
 }
@@ -81,11 +81,22 @@ ActionResult<CNEOBot> CNEOBotJgrCapture::Update( CNEOBot *me, float interval )
 		return Done( "Became the Juggernaut" );
 	}
 
+	const int iJuggernautPlayer = NEORules()->GetJuggernautPlayer();
+	if ( iJuggernautPlayer > 0 )
+	{
+		CNEO_Player *pJuggernautPlayer = ToNEOPlayer( UTIL_PlayerByIndex( iJuggernautPlayer ) );
+		if ( pJuggernautPlayer && pJuggernautPlayer != me )
+		{
+			return Done( "Juggernaut captured by someone else" );
+		}
+	}
+
 	if ( me->GetAbsOrigin().DistToSqr( m_hObjective->GetAbsOrigin() ) < CNEO_Juggernaut::GetUseDistanceSquared() )
 	{
-		CNEO_Juggernaut *pJuggernaut = static_cast<CNEO_Juggernaut*>(m_hObjective.Get());
+		CNEO_Juggernaut *pJuggernaut = m_hObjective.Get();
 		if ( NEORules()->IsJuggernautLocked() || (pJuggernaut && pJuggernaut->m_bLocked) )
 		{
+			Assert( NEORules()->IsJuggernautLocked() == (pJuggernaut && pJuggernaut->m_bLocked) );
 			me->ReleaseUseButton();
 			return SuspendFor( new CNEOBotRetreatToCover( 2.0f ), "Juggernaut is locked, taking cover to wait for it to unlock" );
 		}
@@ -97,8 +108,8 @@ ActionResult<CNEOBot> CNEOBotJgrCapture::Update( CNEOBot *me, float interval )
 		me->ReleaseLeftButton();
 		me->ReleaseRightButton();
 
-		const Vector vecToTarget = m_hObjective->WorldSpaceCenter() - me->EyePosition();
-		Vector vecToTargetDir = vecToTarget;
+		const Vector vecObjectiveCenter = m_hObjective->WorldSpaceCenter();
+		Vector vecToTargetDir = vecObjectiveCenter - me->EyePosition();
 		vecToTargetDir.NormalizeInPlace();
 
 		// Ensure we are facing the target before attempting to use
@@ -107,7 +118,7 @@ ActionResult<CNEOBot> CNEOBotJgrCapture::Update( CNEOBot *me, float interval )
 		const float flDot = vecEyeDirection.Dot( vecToTargetDir );
 		const bool bIsFacing = flDot > 0.9f;
 
-		me->GetBodyInterface()->AimHeadTowards( m_hObjective->WorldSpaceCenter(), IBody::CRITICAL, 0.1f, NULL, "Looking at Juggernaut objective to use" );
+		me->GetBodyInterface()->AimHeadTowards( vecObjectiveCenter, IBody::CRITICAL, 0.1f, NULL, "Looking at Juggernaut objective to use" );
 
 		if ( m_useAttemptTimer.HasStarted() )
 		{
@@ -116,7 +127,7 @@ ActionResult<CNEOBot> CNEOBotJgrCapture::Update( CNEOBot *me, float interval )
 				return Done( "Use timer elapsed, failed to capture" );
 			}
 		}
-		else if ( me->GetBodyInterface()->IsHeadAimingOnTarget() && bIsFacing )
+		else if ( bIsFacing && me->GetBodyInterface()->IsHeadAimingOnTarget() )
 		{
 			m_useAttemptTimer.Start( CNEO_Juggernaut::GetUseDuration() + 1.0f );
 			me->PressUseButton( CNEO_Juggernaut::GetUseDuration() + 1.0f );
--- a/src/game/server/neo/bot/behavior/neo_bot_jgr_enemy.h
+++ b/src/game/server/neo/bot/behavior/neo_bot_jgr_enemy.h
@@ -10,15 +10,15 @@ class CNEOBotJgrEnemy : public Action< CNEOBot >
 public:
 	CNEOBotJgrEnemy( void );
 
-	virtual ActionResult< CNEOBot >	OnStart( CNEOBot *me, Action< CNEOBot > *priorAction );
-	virtual ActionResult< CNEOBot >	Update( CNEOBot *me, float interval );
-	virtual ActionResult< CNEOBot >	OnResume( CNEOBot *me, Action< CNEOBot > *interruptingAction );
+	virtual ActionResult< CNEOBot >	OnStart( CNEOBot *me, Action< CNEOBot > *priorAction ) override;
+	virtual ActionResult< CNEOBot >	Update( CNEOBot *me, float interval ) override;
+	virtual ActionResult< CNEOBot >	OnResume( CNEOBot *me, Action< CNEOBot > *interruptingAction ) override;
 
-	virtual EventDesiredResult< CNEOBot > OnStuck( CNEOBot *me );
-	virtual EventDesiredResult< CNEOBot > OnMoveToSuccess( CNEOBot *me, const Path *path );
-	virtual EventDesiredResult< CNEOBot > OnMoveToFailure( CNEOBot *me, const Path *path, MoveToFailureType reason );
+	virtual EventDesiredResult< CNEOBot > OnStuck( CNEOBot *me ) override;
+	virtual EventDesiredResult< CNEOBot > OnMoveToSuccess( CNEOBot *me, const Path *path ) override;
+	virtual EventDesiredResult< CNEOBot > OnMoveToFailure( CNEOBot *me, const Path *path, MoveToFailureType reason ) override;
 
-	virtual const char *GetName( void ) const	{ return "JgrEnemy"; }
+	virtual const char *GetName( void ) const override	{ return "JgrEnemy"; }
 
 private:
 	ChasePath m_chasePath;
--- a/src/game/server/neo/bot/behavior/neo_bot_jgr_escort.h
+++ b/src/game/server/neo/bot/behavior/neo_bot_jgr_escort.h
@@ -10,15 +10,15 @@ class CNEOBotJgrEscort : public Action< CNEOBot >
 public:
 	CNEOBotJgrEscort( void );
 
-	virtual ActionResult< CNEOBot >	OnStart( CNEOBot *me, Action< CNEOBot > *priorAction );
-	virtual ActionResult< CNEOBot >	Update( CNEOBot *me, float interval );
-	virtual ActionResult< CNEOBot >	OnResume( CNEOBot *me, Action< CNEOBot > *interruptingAction );
+	virtual ActionResult< CNEOBot >	OnStart( CNEOBot *me, Action< CNEOBot > *priorAction ) override;
+	virtual ActionResult< CNEOBot >	Update( CNEOBot *me, float interval ) override;
+	virtual ActionResult< CNEOBot >	OnResume( CNEOBot *me, Action< CNEOBot > *interruptingAction ) override;
 
-	virtual EventDesiredResult< CNEOBot > OnStuck( CNEOBot *me );
-	virtual EventDesiredResult< CNEOBot > OnMoveToSuccess( CNEOBot *me, const Path *path );
-	virtual EventDesiredResult< CNEOBot > OnMoveToFailure( CNEOBot *me, const Path *path, MoveToFailureType reason );
+	virtual EventDesiredResult< CNEOBot > OnStuck( CNEOBot *me ) override;
+	virtual EventDesiredResult< CNEOBot > OnMoveToSuccess( CNEOBot *me, const Path *path ) override;
+	virtual EventDesiredResult< CNEOBot > OnMoveToFailure( CNEOBot *me, const Path *path, MoveToFailureType reason ) override;
 
-	virtual const char *GetName( void ) const	{ return "jgrEscort"; }
+	virtual const char *GetName( void ) const override	{ return "jgrEscort"; }
 
 private:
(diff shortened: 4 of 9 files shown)


DESTROYGIRL:
Got an assert here after a round had ended. Probably not a fault of changes in this PR but ???
Could be that `m_bJuggernautItemExists` is not set correctly after a round ends

```
	server.dll!assert_cast<CNEO_Juggernaut *,CBaseEntity>(CBaseEntity * pSource) Line 576	C++
 	server.dll!CNEORules::JuggernautItemExists() Line 4354	C++
 	server.dll!CNEOBotJgrSeek::RecomputeSeekPath(CNEOBot * me) Line 87	C++
 	server.dll!CNEOBotSeekAndDestroy::OnResume(CNEOBot * me, Action<CNEOBot> * interruptingAction) Line 201	C++
 	server.dll!Action<CNEOBot>::InvokeOnResume(CNEOBot * me, Behavior<CNEOBot> * behavior, Action<CNEOBot> * interruptingAction) Line 1529	C++
 	server.dll!Action<CNEOBot>::ApplyResult(CNEOBot * me, Behavior<CNEOBot> * behavior, ActionResult<CNEOBot> result) Line 1697	C++
 	server.dll!Action<CNEOBot>::InvokeUpdate(CNEOBot * me, Behavior<CNEOBot> * behavior, float interval) Line 1379	C++
 	server.dll!Action<CNEOBot>::InvokeUpdate(CNEOBot * me, Behavior<CNEOBot> * behavior, float interval) Line 1379	C++
 	server.dll!Action<CNEOBot>::InvokeUpdate(CNEOBot * me, Behavior<CNEOBot> * behavior, float interval) Line 1379	C++
 	server.dll!Behavior<CNEOBot>::Update(CNEOBot * me, float interval) Line 258	C++
 	server.dll!CNEOBotIntention::Update() Line 2664	C++
 	server.dll!INextBot::Update() Line 133	C++
 	server.dll!NextBotPlayer<CNEO_Player>::Update() Line 976	C++
 	server.dll!CNEOBot::Update() Line 837	C++
 	server.dll!NextBotPlayer<CNEO_Player>::PhysicsSimulate() Line 743	C++
 	server.dll!CNEOBot::PhysicsSimulate() Line 693	C++
 	server.dll!Physics_SimulateEntity(CBaseEntity * pEntity) Line 2019	C++
 	server.dll!Physics_RunThinkFunctions(bool simulating) Line 2075	C++
 	server.dll!CServerGameDLL::GameFrame(bool simulating) Line 1281	C++

File: src/game/server/neo/bot/behavior/neo_bot_jgr_capture.cpp [hunk 1]
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	if ( me->GetClass() == NEO_CLASS_JUGGERNAUT )
+	{
+		return Done( "Became the Juggernaut" );
+	}
+
+	if ( me->GetAbsOrigin().DistToSqr( m_hObjective->GetAbsOrigin() ) < CNEO_Juggernaut::GetUseDistanceSquared() )
+	{
+		CNEO_Juggernaut *pJuggernaut = static_cast<CNEO_Juggernaut*>(m_hObjective.Get());
```
Rainyan: While we are guarded in the "happy path" by `neo_bot_jgr_seek.cpp:62`:
```cpp
if ( FStrEq( classname, "neo_juggernaut" ) )
```
before the `CNEOBotJgrCapture` call, I think for future-proofing against bugs, it would be nice to refactor `CNEOBotJgrCapture::m_hObjective` to be of type `CHandle<CNEO_Juggernaut>` instead of `CHandle<CBaseEntity>`.

This would move the pointer cast responsibility to the caller (last change of the diff below), so they'd have to actively cast to `CNEO_Juggernaut*` to make the call, which should make it less likely that someone accidentally passes in an unrelated `CBaseEntity*`, which might not trigger a compiler error but would be ill-formed if we downcast it to `CNEOBotJgrCapture*` from some unrelated ptr.

Something like this perhaps?

```diff
diff --git a/src/game/server/neo/bot/behavior/neo_bot_jgr_capture.cpp b/src/game/server/neo/bot/behavior/neo_bot_jgr_capture.cpp
index 14ff9ddd..0e9d0fd9 100644
--- a/src/game/server/neo/bot/behavior/neo_bot_jgr_capture.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_jgr_capture.cpp
@@ -6,7 +6,7 @@
 #include "neo_juggernaut.h"
 
 //---------------------------------------------------------------------------------------------
-CNEOBotJgrCapture::CNEOBotJgrCapture( CBaseEntity *pObjective )
+CNEOBotJgrCapture::CNEOBotJgrCapture( CNEO_Juggernaut *pObjective )
 {
 	m_hObjective = pObjective;
 }
@@ -83,7 +83,7 @@ ActionResult<CNEOBot> CNEOBotJgrCapture::Update( CNEOBot *me, float interval )
 
 	if ( me->GetAbsOrigin().DistToSqr( m_hObjective->GetAbsOrigin() ) < CNEO_Juggernaut::GetUseDistanceSquared() )
 	{
-		CNEO_Juggernaut *pJuggernaut = static_cast<CNEO_Juggernaut*>(m_hObjective.Get());
+		CNEO_Juggernaut *pJuggernaut = m_hObjective.Get();
 		if ( NEORules()->IsJuggernautLocked() || (pJuggernaut && pJuggernaut->m_bLocked) )
 		{
 			me->ReleaseUseButton();
diff --git a/src/game/server/neo/bot/behavior/neo_bot_jgr_capture.h b/src/game/server/neo/bot/behavior/neo_bot_jgr_capture.h
index d1fe7fe8..984697a6 100644
--- a/src/game/server/neo/bot/behavior/neo_bot_jgr_capture.h
+++ b/src/game/server/neo/bot/behavior/neo_bot_jgr_capture.h
@@ -8,7 +8,7 @@
 class CNEOBotJgrCapture : public Action<CNEOBot>
 {
 public:
-	CNEOBotJgrCapture( CBaseEntity *pObjective );
+	CNEOBotJgrCapture( CNEO_Juggernaut *pObjective );
 	virtual ~CNEOBotJgrCapture() { }
 
 	virtual const char *GetName() const override { return "jgrCapture"; }
@@ -20,7 +20,7 @@ public:
 	virtual ActionResult<CNEOBot> OnResume( CNEOBot *me, Action<CNEOBot> *interruptingAction ) override;
 
 private:
-	CHandle<CBaseEntity> m_hObjective;
+	CHandle<CNEO_Juggernaut> m_hObjective;
 	CountdownTimer m_useAttemptTimer;
 	PathFollower m_path;
 	CountdownTimer m_repathTimer;
diff --git a/src/game/server/neo/bot/behavior/neo_bot_jgr_seek.cpp b/src/game/server/neo/bot/behavior/neo_bot_jgr_seek.cpp
index 05d21403..d2abd7e9 100644
--- a/src/game/server/neo/bot/behavior/neo_bot_jgr_seek.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_jgr_seek.cpp
@@ -61,7 +61,7 @@ ActionResult< CNEOBot > CNEOBotJgrSeek::Update( CNEOBot *me, float interval )
 
 					if ( FStrEq( classname, "neo_juggernaut" ) )
 					{
-						return SuspendFor( new CNEOBotJgrCapture( m_hTargetEntity ), "Capturing Juggernaut" );
+						return SuspendFor( new CNEOBotJgrCapture( static_cast<CNEO_Juggernaut*>(m_hTargetEntity.Get()) ), "Capturing Juggernaut" );
 					}
 				}
 			}
```

src/game/server/neo/bot/behavior/neo_bot_jgr_capture.cpp
```diff
@@ -0,0 +1,140 @@
...
+		if ( NEORules()->IsJuggernautLocked() || (pJuggernaut && pJuggernaut->m_bLocked) )
```
Rainyan, line 97: Would we want to assert for a case where the gamerules and the `pJuggernaut` are in disagreement about this "locked" condition? Or can these get out of sync normally?

src/game/server/neo/bot/behavior/neo_bot_jgr_capture.cpp
```diff
@@ -0,0 +1,140 @@
...
+		// Stop moving while using
+		m_path.Invalidate();
+		me->ReleaseForwardButton();
+		me->ReleaseBackwardButton();
+		me->ReleaseLeftButton();
+		me->ReleaseRightButton();
+
+		const Vector vecToTarget = m_hObjective->WorldSpaceCenter() - me->EyePosition();
+		Vector vecToTargetDir = vecToTarget;
+		vecToTargetDir.NormalizeInPlace();
```
Rainyan: Nit: this could probably simplify to
```cpp
Vector vecToTargetDir = m_hObjective->WorldSpaceCenter() - me->EyePosition();
vecToTargetDir.NormalizeInPlace();
```
without needing the intermediate Vector be constructed.

src/game/server/neo/bot/behavior/neo_bot_jgr_capture.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+		// Ensure we are facing the target before attempting to use
+		Vector vecEyeDirection;
+		me->EyeVectors( &vecEyeDirection );
+		const float flDot = vecEyeDirection.Dot( vecToTargetDir );
+		const bool bIsFacing = flDot > 0.9f;
+
+		me->GetBodyInterface()->AimHeadTowards( m_hObjective->WorldSpaceCenter(), IBody::CRITICAL, 0.1f, NULL, "Looking at Juggernaut objective to use" );
```
Rainyan: Perf nit: could reuse cached `m_hObjective->WorldSpaceCenter()` from line 100 to avoid needing to recompute it here - at least MSVC doesn't know how to automatically optimize these into one call for release builds. Since there's temp Vector allocation and all sorts of bells and whistles going on in the `WorldSpaceCenter` implementation that we can avoid running twice.

src/game/server/neo/bot/behavior/neo_bot_jgr_capture.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+		if ( m_useAttemptTimer.HasStarted() )
+		{
+			if ( m_useAttemptTimer.IsElapsed() )
+			{
+				return Done( "Use timer elapsed, failed to capture" );
+			}
+		}
+		else if ( me->GetBodyInterface()->IsHeadAimingOnTarget() && bIsFacing )
```
Rainyan: Perf nit: comparing the local `bIsFacing` is faster than the virtual call (2 CPU instructions for `if bIsFacing` vs 9 for the method for MSVC release build), so preferably test the least expensive `bIsFacing` first for a logical AND. This is a micro-optimization I wouldn't usually point out, but might as well, assuming we're doing the `WorldSpaceCenter` cache for the above lines anyway.

src/game/server/CMakeLists.txt
```diff
@@ -1438,6 +1438,16 @@ target_sources_grouped(
     neo/bot/behavior/neo_bot_behavior.h
     neo/bot/behavior/neo_bot_dead.cpp
     neo/bot/behavior/neo_bot_dead.h
+    neo/bot/behavior/neo_bot_jgr_capture.cpp
+    neo/bot/behavior/neo_bot_jgr_capture.h
+    neo/bot/behavior/neo_bot_jgr_escort.cpp
+    neo/bot/behavior/neo_bot_jgr_escort.h
+    neo/bot/behavior/neo_bot_jgr_enemy.cpp
```
Rainyan: Dumb OCD nit: these are not alphabetically sorted :D

src/game/server/neo/bot/behavior/neo_bot_jgr_juggernaut.h
```diff
@@ -0,0 +1,30 @@
...
+public:
+	CNEOBotJgrJuggernaut( float duration = -1.0f ) : CNEOBotSeekAndDestroy( duration ), m_bStuckCrouchPhase( false ) {}
+
+	virtual ActionResult< CNEOBot >	OnStart( CNEOBot *me, Action< CNEOBot > *priorAction );
+	virtual ActionResult< CNEOBot >	Update( CNEOBot *me, float interval );
+	virtual const char *GetName( void ) const { return "jgrJuggernaut"; }
+
+	virtual EventDesiredResult< CNEOBot > OnStuck( CNEOBot *me );
+
+	virtual QueryResultType	ShouldWalk( const CNEOBot *me, const QueryResultType qShouldAimQuery ) const;
```
Rainyan: I assume the intention here is to declare an override of the virtual base class method `ShouldWalk`? If so, this doesn't actually work, because the base `CNEOBotBehavior::ShouldWalk` is declared `final`, i.e. it cannot be overridden. So this declaration would be a kind of overload (rather than override) that may or may not shadow the base method, depending on which kind of a pointer we are invoking the method from.

You can verify this by adding the `override` specifier to this declaration, and seeing the compiler emit an error:

> 'CNEOBotJgrJuggernaut::ShouldWalk': method with override specifier 'override' did not override any base class methods [MSVC latest, not sure what GCC/Clang emits here]

We should probably remove the `final` specifier from the base declaration to allow for this override. And ideally mark these virtuals explicitly with the `override` specifier, to make it easier to catch these kinds of things at compile-time.

src/game/server/neo/bot/behavior/neo_bot_jgr_seek.cpp
```diff
@@ -0,0 +1,112 @@
...
+		return Done( "Game mode is no longer JGR" );
+	}
+
+	if ( me->GetClass() != NEO_CLASS_JUGGERNAUT )
+	{
+		int iJuggernautPlayer = NEORules()->GetJuggernautPlayer();
+		if (iJuggernautPlayer > 0)
+		{
+			CNEO_Player* pJuggernaut = ToNEOPlayer(UTIL_PlayerByIndex(iJuggernautPlayer));
+			if (pJuggernaut && pJuggernaut != me && pJuggernaut->IsAlive())
```
Rainyan, line 29: Perhaps Assert for the case where `me->GetClass() != NEO_CLASS_JUGGERNAUT` at line 21, but `pJuggernaut` retrieved from `NEORules()->GetJuggernautPlayer()` ended up `== me`, since that game logic disagreement should probably never happen. So that we don't hide the error if it does ever occur.

src/game/server/neo/bot/behavior/neo_bot_jgr_seek.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+	if ( me->GetClass() == NEO_CLASS_JUGGERNAUT )
+	{
+		return SuspendFor( new CNEOBotJgrJuggernaut, "I am the Juggernaut now" );
+	}
+
+	// Juggernaut objective capture logic
+	if (m_bGoingToTargetEntity && m_hTargetEntity)
+	{
+		const float useRangeSq = 100.0f * 100.0f;
```
Rainyan: Should we use the `CNEO_Juggernaut::GetUseDistanceSquared` here, instead of a magic number? Also, that function return value one doesn't seem to agree with this value; I'm not sure which one is correct though.

There is also the SDK native `PLAYER_USE_RADIUS` value that could be used, although that one's also slightly different from both this magic value and the `CNEO_Juggernaut::GetUseDistanceSquared` value.

Not quite sure which is the best here; on one hand I could see an argument made for following the SDK-native `+use` distance to match what a player would expect from the classic Half-Life use distance, as that might be intuitive for long-time players to eyeball. But the exact value is somewhat offtopic to this PR perhaps. But regardless, it'd be nice to unify these checks to all use the same singular authoritative value somewhere to avoid ambiguity - probably the `CNEO_Juggernaut::GetUseDistanceSquared`.

src/game/server/neo/bot/behavior/neo_bot_jgr_seek.cpp
```diff
@@ -0,0 +1,112 @@
...
+		if ( me->GetAbsOrigin().DistToSqr( m_hTargetEntity->GetAbsOrigin() ) < useRangeSq ) 
+		{
+			if ( !m_hTargetEntity->IsPlayer() )
```
Rainyan: On the nitpick territory, but there might be a refactor opportunity here to change `CNEOBotSeekAndDestroy::m_hTargetEntity` type from `CHandle<CBaseEntity>` to something like `CHandle<INEOBotSeekAndDestroyEntity>` (and have the relevant entities implement that interface), so that we could explicitly check `INEOBotSeekAndDestroyEntity::IsJuggernaut` here, rather than the inverse case ("Is not player, therefore it is Juggernaut"). Because the moment we add some third SeekAndDestroy target entity for some future game mode, this logic would become incorrect. So it'd be nice to account for that in advance.

src/game/server/neo/bot/behavior/neo_bot_jgr_escort.h
```diff
@@ -0,0 +1,27 @@
...
+
+	virtual EventDesiredResult< CNEOBot > OnStuck( CNEOBot *me );
+	virtual EventDesiredResult< CNEOBot > OnMoveToSuccess( CNEOBot *me, const Path *path );
+	virtual EventDesiredResult< CNEOBot > OnMoveToFailure( CNEOBot *me, const Path *path, MoveToFailureType reason );
+
+	virtual const char *GetName( void ) const	{ return "jgrEscort"; }
+
+private:
+	ChasePath m_chasePath;
+};
```
Rainyan, line 25: Preferably specify virtual `override`s explicitly.

src/game/server/neo/bot/behavior/neo_bot_jgr_enemy.h
```diff
@@ -0,0 +1,27 @@
...
+
+	virtual EventDesiredResult< CNEOBot > OnStuck( CNEOBot *me );
+	virtual EventDesiredResult< CNEOBot > OnMoveToSuccess( CNEOBot *me, const Path *path );
+	virtual EventDesiredResult< CNEOBot > OnMoveToFailure( CNEOBot *me, const Path *path, MoveToFailureType reason );
+
+	virtual const char *GetName( void ) const	{ return "JgrEnemy"; }
+
+private:
+	ChasePath m_chasePath;
+};
```
Rainyan, line 25: Preferably specify virtual `override`s explicitly.

Rainyan (changes requested):
Looks good for the most part, but I'd like to see the [`ShouldWalk` declaration's clash with the base class `final` specifier](https://github.com/NeotokyoRebuild/neo/pull/1579#discussion_r2659150889) fixed.

For the other nits/comments, feel free to address/ignore based on personal preference, I don't necessarily require changes for those if you don't see them necessary.

src/game/server/neo/bot/behavior/neo_bot_jgr_capture.cpp
sunzenshen: So, it turns out that this proposed Assert occasionally fires but not too frequently, so I'm going to keep the assert in for @DESTROYGIRL to look at later.

src/game/server/neo/bot/behavior/neo_bot_jgr_juggernaut.h
sunzenshen: It turns out that I've done something similar for #1538, but trying this method of ShouldWalk ended up not actually satisfactorily stopping sprint even when the answer was ANSWER_YES here.

It turned out to be easier to cancel sprint just before calling PressFireButton in neo_bot_behavior, so I'm just going to do that and let the primary maintainers of JGR mode figure out what is going on with ShouldWalk in the future.

src/game/server/neo/bot/behavior/neo_bot_jgr_seek.cpp
sunzenshen: Personally feel that the use press from the bots registers better when they are closer than CNEO_Juggernaut::GetUseDistanceSquared, but 0.8 of that seems reliable enough, so I'd rather let the JGR maintainers figure out how they want to tune this after this PR merges..

Rainyan (approved):

PR #1581, opened 2025-12-29
Fix neo_restart_this leaking edicts

### LINKED ISSUE #1580: neo_restart_this should not leak edicts
### Build Info

f50a0d7c

### Description

When using `neo_restart_this` to restart the map, edicts are not correctly cleaned up and/or reused. This causes edict starvation and ultimately a game crash when repeated often enough for a sufficiently complex map.

### To Reproduce

* Load `map ntre_terminal_jgr`
* Have bots enabled to fill both teams with players
* Join a team to start the match
* Use `neo_restart_this 1` multiple times in a row

### Expected behavior

Game does not crash.

### Actual behavior

Game crashes with a message:
```
---------------------------
Engine Error
---------------------------
ED_Alloc: no free edicts
---------------------------
OK   
---------------------------
```

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

11

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

MSVC 19.44.35222.0

_No response_

Rainyan:
## Description
For some reason, calling RestartGame directly from the neo_restart_this change callback does not clean up edicts correctly, which leads to an edict starvation and a subsequent game crash when using `neo_restart_this 1` repeatedly in a sufficiently complex map.

Because we don't have support for the SDK native `mp_restartgame_immediate`, and we don't want `neo_restart_this` to mimic the SDK native `mp_restartgame` center of screen printout, this commit works around those limitations by calling `mp_restartgame` with a magic value that we intercept from `CHL2MPRules::CheckRestartGame`, and then monkey with the values to achieve parity-ish map reset behaviour while sidestepping the edict leak problem.

This is a complete hack and should be revisited, but since the command is essentially broken currently, this is better than nothing for the time being.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1580

### DIFF
--- a/src/game/shared/hl2mp/hl2mp_gamerules.cpp
+++ b/src/game/shared/hl2mp/hl2mp_gamerules.cpp
@@ -1241,6 +1241,11 @@ void CHL2MPRules::CheckRestartGame( void )
 
 	if ( iRestartDelay > 0 )
 	{
+#ifdef NEO
+		if (iRestartDelay != MAGIC_NEO_RESTART_THIS)
+		{
+#endif
+
 		if ( iRestartDelay > 60 )
 			iRestartDelay = 60;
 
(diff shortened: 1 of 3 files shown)


Masterkatze (approved):
LGTM - yes
Better than nothing - also yes

PR #1583, opened 2025-12-30
Bots scavenge for primary weapons

sunzenshen:
## Description
When bots do not have a primary weapon, usually from trying capture the ghost, search in a radius around the bot for any available primary weapons to scavenge.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_seek_weapon.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_seek_weapon.cpp
@@ -8,15 +8,15 @@
 CBaseEntity *FindNearestPrimaryWeapon( const Vector &searchOrigin, bool bShortCircuit )
 {
 	constexpr float flSearchRadius = 1000.0f;
-	CBaseEntity *pClosestWeapon = NULL;
+	CBaseEntity *pClosestWeapon = nullptr;
 	float flClosestDistSq = FLT_MAX;
 
 	// Iterate through all available weapons, looking for the nearest primary 
-	CBaseEntity *pEntity = NULL;
-	while ( ( pEntity = gEntList.FindEntityByClassnameWithin( pEntity, "weapon_*", searchOrigin, flSearchRadius ) ) != NULL )
+	CBaseEntity *pEntity = nullptr;
+	while ( ( pEntity = gEntList.FindEntityByClassnameWithin( pEntity, "weapon_*", searchOrigin, flSearchRadius ) ) )
 	{
 		CBaseCombatWeapon *pWeapon = pEntity->MyCombatWeaponPointer();
-		if ( pWeapon && pWeapon->GetOwner() == NULL && pWeapon->HasAnyAmmo() && pWeapon->GetSlot() == 0 )
+		if ( pWeapon && !pWeapon->GetOwner() && pWeapon->HasAnyAmmo() && pWeapon->GetSlot() == 0 )
 		{
 			// Prioritize getting any primary class weapon that is NOT the ghost
 			if ( FStrEq( pEntity->GetClassname(), "weapon_ghost" ) )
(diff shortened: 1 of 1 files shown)


sunzenshen:
To test, `bot_mimic 1` and have the bots drop their primary weapon.  After leading them away from their weapons, and setting `bot_mimic 0` the bots should route to the nearest weapons to pick them up.

https://github.com/user-attachments/assets/7c55224f-61a0-4095-8468-4d3bc70f9cbf

Example from another branch with CTG related improvements where a bot will drop the ghost if an enemy has caught up with them and then try to arm themselves with a better weapon:

https://github.com/user-attachments/assets/1d1ff172-24f7-44e6-bfeb-9e0f824cb0aa

sunzenshen:
To record some discussions from Discord:

- Idea: Potentially have a cooldown timer on weapons dropped by non-bot friendlies, so that bots don't steal you weapon when you want to drop your weapon for the ghost, but then change your mind.
   - A: Currently bots rarely drop their primary weapon, and even with the WIP CTG branch bots rarely drop their weapon unless to pick up the ghost.  But that scenario could definitely come into play when we introduce other behaviors related to dropping weapons.  That said, I'm shooting for making a minimum viable PR in this scope, with the idea that we can revisit the consequences of other behaviors we introduce in the future.  Another reason for limiting the scope of this PR is that I want to keep the calculations lean for the MVP, and timers would add more state that reviewers will need to consider for this PR.

- Idea: bots swapping guns if there is a better option on the ground
   - A: Also interested in exploring something like this in the future, but this is an example of an idea that would interact with scenarios like the one above where there are more interactions to consider.  I'd like to keep this PR simple so that we can review things like the shape of the new files like neo_bot_seek_weapon, so that we don't need to consider that in the scope of a future additional PR.

- Idea: Trading weapons at the start or even during freeze time.
    - A: This also seems like a good idea for a new behavior that only comes into play when the game is currently in freezetime.

- Idea: Juggling weapons, such as moving the ghoster's discarded weapon so that they can rearm themselves easily.
   - A: I think we could address this as a separate behavior in a future PR.  In my WIP CTG PR, I have ghoster team escorts assigning themselves roles based on their relative distance from the ghost carrier, and it might make sense to introduce a juggler role as a dedicated behavior class.  That way we can consider the consequences of introducing that behavior in a controlled scope and known working but simpler weapon pickup commit.
   
- Idea: Have bots drop their gun for an unarmed human player if there is no active danger
   - A: I think this could work as an augmentation to #1345 once that is merged.  My thinking is that as-is you're rarely in a controlled situation where a bot happens to be waiting around and where it's clear that you would like their gun (read: the bots are running all over the place and it would be hard to catch up).  But with the bot command feature, my thinking was that you could have a behavior for following bots where if their commander is looking at them and the commander has no primary, the bot could toss their weapon at the commander.  This could also work for swapping a ghost with a bot, as you could drop your weapon near your follower, and when they drop the ghost they would pick up your nearby discarded weapon.

AdamTadeusz (dismissed):

src/game/server/neo/bot/behavior/neo_bot_seek_weapon.cpp
```diff
@@ -0,0 +1,148 @@
...
+	constexpr float flSearchRadius = 1000.0f;
+	CBaseEntity *pClosestWeapon = NULL;
+	float flClosestDistSq = FLT_MAX;
+
+	// Iterate through all available weapons, looking for the nearest primary 
+	CBaseEntity *pEntity = NULL;
+	while ( ( pEntity = gEntList.FindEntityByClassnameWithin( pEntity, "weapon_*", searchOrigin, flSearchRadius ) ) != NULL )
+	{
+		CBaseCombatWeapon *pWeapon = pEntity->MyCombatWeaponPointer();
+		if ( pWeapon && pWeapon->GetOwner() == NULL && pWeapon->HasAnyAmmo() && pWeapon->GetSlot() == 0 )
```
DESTROYGIRL: Correct me if I'm wrong but shouldn't every instance of `NULL` be `nullptr` in this file? (since `NULL` is just a macro for `0`?)
And in that case can't you just do `!pWeapon->GetOwner()`

sunzenshen: That's a good point.  In general for C++ using nullptr is the Effective Modern C++ way to do things, but I've been caught between trying to follow the conventions of the code base, but this is a new file so maybe it's better to implement things in a cleaner way since I'm not bound by the original Source Engine SDK coding conventions in this case.

DESTROYGIRL (approved):

PR #1586, opened 2026-01-01
Warn user about unsupported FOV value

### LINKED ISSUE #1201: Aggressive skybox culling at higher fov
see https://github.com/ValveSoftware/Source-1-Games/issues/2657

Looks like this isn't something we can fix ourselves

Example of issue in rebuild:

Image

Rainyan:
## Description
We are restricted to supporting a max FOV of 90 by a limitation in the engine library renderer, for which we do not have access. As such, this patch resolves #1201 by warning the user about their unsupported FOV.

We probably don't want to hard-limit the upper bound of `neo_fov` to 90, because it could lead to unhappy ultrawide/multi-monitor users, so this is the best(?) compromise for now.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- closes #1201

### DIFF
--- a/src/game/client/neo/ui/neo_root.cpp
+++ b/src/game/client/neo/ui/neo_root.cpp
@@ -2303,8 +2303,8 @@ void CNeoRoot::MainLoopPopup(const MainLoopParam param)
 				{
 					wchar_t warning[77+1];
 					V_swprintf_safe(warning,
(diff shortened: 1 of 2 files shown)


Masterkatze (approved):

sunzenshen (approved):
LGTM on Windows:

Image

Image

PR #1600, opened 2026-01-12
Bot CTG gameplay behaviors

sunzenshen:
## Description
Bot behaviors for playing CTG mode.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/CMakeLists.txt
+++ b/src/game/server/CMakeLists.txt
@@ -1441,8 +1441,6 @@ target_sources_grouped(
     neo/bot/behavior/neo_bot_behavior.h
     neo/bot/behavior/neo_bot_command_follow.cpp
     neo/bot/behavior/neo_bot_command_follow.h
-    neo/bot/behavior/neo_bot_dead.cpp
-    neo/bot/behavior/neo_bot_dead.h
     neo/bot/behavior/neo_bot_ctg_capture.cpp
     neo/bot/behavior/neo_bot_ctg_capture.h
     neo/bot/behavior/neo_bot_ctg_carrier.cpp
@@ -1455,6 +1453,8 @@ target_sources_grouped(
     neo/bot/behavior/neo_bot_ctg_lone_wolf.h
     neo/bot/behavior/neo_bot_ctg_seek.cpp
     neo/bot/behavior/neo_bot_ctg_seek.h
+    neo/bot/behavior/neo_bot_dead.cpp
+    neo/bot/behavior/neo_bot_dead.h
     neo/bot/behavior/neo_bot_jgr_capture.cpp
     neo/bot/behavior/neo_bot_jgr_capture.h
     neo/bot/behavior/neo_bot_jgr_enemy.cpp
--- a/src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
@@ -125,9 +125,9 @@ void CNEOBotGhostEquipmentHandler::EquipBestWeaponForGhoster( CNEOBot *me )
 {
 	const CKnownEntity *threat = me->GetVisionInterface()->GetPrimaryKnownThreat();
 
-	CNEOBaseCombatWeapon *pActive = dynamic_cast<CNEOBaseCombatWeapon*>( me->GetActiveWeapon() );
-	CNEOBaseCombatWeapon *pGhost = dynamic_cast<CNEOBaseCombatWeapon*>( me->Weapon_GetSlot( 0 ) );
-	CNEOBaseCombatWeapon *pSecondary = dynamic_cast<CNEOBaseCombatWeapon*>( me->Weapon_GetSlot( 1 ) );
+	CNEOBaseCombatWeapon *pActive =    assert_cast<CNEOBaseCombatWeapon*>( me->GetActiveWeapon() );
+	CNEOBaseCombatWeapon *pGhost =     assert_cast<CNEOBaseCombatWeapon*>( me->Weapon_GetSlot( 0 ) );
+	CNEOBaseCombatWeapon *pSecondary = assert_cast<CNEOBaseCombatWeapon*>( me->Weapon_GetSlot( 1 ) );
 
 	// Sanity check: if we don't have the ghost, we shouldn't be in this behavior, but handle gracefully
 	if ( !pGhost ) 
@@ -177,12 +177,12 @@ void CNEOBotGhostEquipmentHandler::UpdateGhostCarrierCallout( CNEOBot *me, const
 {
 	CNEO_Player *pBestCallout = nullptr;
 	float flBestCalloutMoved = -1.0f;
-	float flBestCalloutDist = FLT_MAX;
+	float flBestCalloutDistSq = FLT_MAX;
 	bool bBestCalloutIsNew = false;
 	
 	bool bConsideringOnlyLoSEnemies = false;
 	
-	const Vector vecMyPos = me->GetAbsOrigin();
+	const Vector& vecMyPos = me->GetAbsOrigin();
 
 	for ( int i = 0; i < enemies.Count(); ++i )
 	{
@@ -205,7 +205,7 @@ void CNEOBotGhostEquipmentHandler::UpdateGhostCarrierCallout( CNEOBot *me, const
 			continue;
 		}
 
-		float flDistToMe = vecMyPos.DistTo( pPlayer->GetAbsOrigin() );
+		float flDistToMeSq = vecMyPos.DistToSqr( pPlayer->GetAbsOrigin() );
 
 		// Also take into account how much the enemy has moved since we last reported them
 		Vector vecLast = m_enemyLastPos[ idx ];
@@ -226,16 +226,16 @@ void CNEOBotGhostEquipmentHandler::UpdateGhostCarrierCallout( CNEOBot *me, const
 				pBestCallout = pPlayer;
 				bBestCalloutIsNew = bIsNew;
 				flBestCalloutMoved = flMoved;
-				flBestCalloutDist = flDistToMe;
+				flBestCalloutDistSq = flDistToMeSq;
 			}
 			else
 			{
-				if ( flDistToMe < flBestCalloutDist )
+				if ( flDistToMeSq < flBestCalloutDistSq )
 				{
 					pBestCallout = pPlayer;
 					bBestCalloutIsNew = bIsNew;
 					flBestCalloutMoved = flMoved;
-					flBestCalloutDist = flDistToMe;
+					flBestCalloutDistSq = flDistToMeSq;
 				}
 			}
 			continue;
@@ -252,7 +252,7 @@ void CNEOBotGhostEquipmentHandler::UpdateGhostCarrierCallout( CNEOBot *me, const
 			pBestCallout = pPlayer;
 			bBestCalloutIsNew = bIsNew;
 			flBestCalloutMoved = flMoved;
-			flBestCalloutDist = flDistToMe;
+			flBestCalloutDistSq = flDistToMeSq;
 			continue;
 		}
 
@@ -261,7 +261,7 @@ void CNEOBotGhostEquipmentHandler::UpdateGhostCarrierCallout( CNEOBot *me, const
 		{
 			pBestCallout = pPlayer;
 			bBestCalloutIsNew = true;
-			flBestCalloutDist = flDistToMe;
+			flBestCalloutDistSq = flDistToMeSq;
 			flBestCalloutMoved = 0;
 			continue;
 		}
@@ -276,10 +276,10 @@ void CNEOBotGhostEquipmentHandler::UpdateGhostCarrierCallout( CNEOBot *me, const
 		if ( bIsNew )
 		{
 			// Both New: Tie breaker is distance to me (closest first)
-			if ( flDistToMe < flBestCalloutDist )
+			if ( flDistToMeSq < flBestCalloutDistSq )
 			{
 				pBestCallout = pPlayer;
-				flBestCalloutDist = flDistToMe;
+				flBestCalloutDistSq = flDistToMeSq;
 			}
 		}
 		else
@@ -291,11 +291,11 @@ void CNEOBotGhostEquipmentHandler::UpdateGhostCarrierCallout( CNEOBot *me, const
 			if ( FloatMakePositive( diff ) < 100.0f )
 			{
 				// Roughly same movement, tie breaker is closest distance
-				if ( flDistToMe < flBestCalloutDist )
+				if ( flDistToMeSq < flBestCalloutDistSq )
 				{
 					pBestCallout = pPlayer;
 					flBestCalloutMoved = flMoved;
-					flBestCalloutDist = flDistToMe;
+					flBestCalloutDistSq = flDistToMeSq;
 				}
 			}
 			else if ( diff > 0.0f )
@@ -303,7 +303,7 @@ void CNEOBotGhostEquipmentHandler::UpdateGhostCarrierCallout( CNEOBot *me, const
 				// Distinctly moved more
 				pBestCallout = pPlayer;
 				flBestCalloutMoved = flMoved;
-				flBestCalloutDist = flDistToMe;
+				flBestCalloutDistSq = flDistToMeSq;
 			}
 		}
 	}
@@ -583,8 +583,8 @@ void CNEOBotCtgCarrier::UpdateFollowPath( CNEOBot *me, const CUtlVector<CNEO_Pla
 
 	if ( pTargetTeammate )
 	{
-		float flDistToTeammate = me->GetAbsOrigin().DistTo( pTargetTeammate->GetAbsOrigin() );
-		if ( flDistToTeammate < 100.0f )
+		float flDistToTeammateSq = me->GetAbsOrigin().DistToSqr( pTargetTeammate->GetAbsOrigin() );
+		if ( flDistToTeammateSq < ( 100.0f * 100.0f ) )
 		{
 			m_chasePath.Invalidate();
 			return;
--- a/src/game/server/neo/bot/behavior/neo_bot_ctg_escort.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ctg_escort.cpp
@@ -92,7 +92,8 @@ ActionResult< CNEOBot >	CNEOBotCtgEscort::Update( CNEOBot *me, float interval )
 			// Look away from the carrier to cover their blind spots
 			Vector vecFromCarrier = me->GetAbsOrigin() - pGhostCarrier->GetAbsOrigin();
 			vecFromCarrier.z = 0.0f; // Bias towards horizontal scanning
-			if ( VectorNormalize( vecFromCarrier ) > 0.1f )
+			// CNEOBotTacticalMonitor::AvoidBumpingFriends handles <32hu distance
+			if ( VectorNormalize( vecFromCarrier ) > 32.0f )
 			{
 				// Look at a point far away in the opposite direction of the carrier
 				Vector vecLookTarget = me->EyePosition() + ( vecFromCarrier * 500.0f );
@@ -176,7 +177,7 @@ ActionResult< CNEOBot >	CNEOBotCtgEscort::Update( CNEOBot *me, float interval )
 
 		m_chasePath.Invalidate();
 
-		Vector vecMoveTarget = m_vecGoalPos;
+		Vector& vecMoveTarget = m_vecGoalPos;
 
 
 		if ( m_role == ROLE_SCREEN && pBotGhostCarrier )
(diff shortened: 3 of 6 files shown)


src/game/server/CMakeLists.txt
```diff
@@ -1460,6 +1472,8 @@ target_sources_grouped(
     neo/bot/behavior/neo_bot_scenario_monitor.h
     neo/bot/behavior/neo_bot_seek_and_destroy.cpp
     neo/bot/behavior/neo_bot_seek_and_destroy.h
+    neo/bot/behavior/neo_bot_seek_weapon.cpp
+    neo/bot/behavior/neo_bot_seek_weapon.h
```
sunzenshen, line 1483: TODO: Rebase when #1583 gets merged in.

src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
```diff
@@ -0,0 +1,501 @@
...
+		}
+	}
+
+	// We dropped the ghost to hunt a threat.
+	if ( m_bPursuingDropThreat )
+	{
+		// First, ensure we have a weapon.
+		if ( !me->Weapon_GetSlot( 0 ) )
+		{
+			return SuspendFor( new CNEOBotSeekWeapon(), "Scavenging for weapon to hunt threat" );
```
sunzenshen, line 65: Dependency on #1583 comes from behaviors related to dropping weapons/ghost and needing to scavenge for weapons.

src/game/server/neo/bot/neo_bot_path_cost.cpp
```diff
@@ -114,6 +114,15 @@ float CNEOBotPathCost::operator()(CNavArea* baseArea, CNavArea* fromArea, const
...
 		{
 			cost += CNEOBotPathReservations()->GetPredictedFriendlyPathCount(area->GetID(), m_me->GetTeamNumber()) * neo_bot_path_reservation_penalty.GetFloat();
+
+			if (m_routeType == SAFEST_ROUTE)
+			{
+				// NEO Jank Cheat: Incorporate enemy bot paths so that we don't run directly into their line of fire
+				// Intended for use by ghost carrier team, to emulate a team that knows where enemies are likely to ambush
+				// Compensates for bots' lack of meta knowledge by making them prefer routes not reserved by enemies
+				// Adheres to cheat against bots but not against humans philosophy by not considering human players' positions
+				cost += CNEOBotPathReservations()->GetPredictedFriendlyPathCount(area->GetID(), GetEnemyTeam(m_me->GetTeamNumber())) * neo_bot_path_reservation_penalty.GetFloat() * 2;
```
sunzenshen, line 125: This is why I changed the SAFEST_ROUTE parameters for other path calculations not related to the ghost.  There wasn't any existing handling with this case beyond some stubs related to combat areas, so this was the cleanest way I could think of for choosing paths as a ghoster team.

src/game/server/neo/bot/behavior/neo_bot_get_health.cpp
```diff
@@ -175,7 +175,7 @@ ActionResult< CNEOBot >	CNEOBotGetHealth::OnStart( CNEOBot *me, Action< CNEOBot
 	m_healthKit = s_possibleHealth;
 	m_isGoalCharger = m_healthKit->ClassMatches( "*charger*" );
 
-	if (!CNEOBotPathCompute(me, m_path, m_healthKit->WorldSpaceCenter(), SAFEST_ROUTE))
+	if (!CNEOBotPathCompute(me, m_path, m_healthKit->WorldSpaceCenter(), DEFAULT_ROUTE))
```
sunzenshen, line 178: See neo_bot_path_cost.cpp for why these route type changes.

sunzenshen:
Rebased on top of the bot command branch, so now the following debug commands are available:

```
sv_neo_bot_cmdr_enable 1;
neo_ctg_round_timelimit 99;
```

This PR doesn't really need bot commands for testing, but might be worth thinking of test scenarios where the CTG behaviors and bot commands may interact.

I guess one possible interaction to look into after a first pass of feedback is that a bot holding the ghost will not report enemy positions to teammates, as they will be in the follow behavior.  Maybe I should move the enemy reporting code into neo_bot_tactical_monitor and make it a general behavior to report ghoster targets (if the bot has the ghost equipped), instead of making it limited to the ghost carrier behavior.

AdamTadeusz:
One problem I've noticed with the bots "seeing" the ghoster is that they will immediately lock onto them and pop their cloak as if they are about to engage in combat, even if the ghoster is on the other side of the map, which causes the non ghoster team to not really check their angles at all. I realise that a more sophisticated solution will require a lot more work so this isnt a deal breaker for me, but I think its something we should consider in the future.

One way to cheaply check whether the ghoster is "around the corner" and worth pursuing with the kind of determination bots show at the moment from across the map could be to check whether the ghoster is in the bot's PVS given how the potentially visible set tends to extend past corners for some distance, though I don't know how this check works with the ghoster always being networked. And again I don't think this is something that has to be fixed in this PR necessarily

src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
```diff
@@ -0,0 +1,653 @@
...
+				}
+				pBot->GetVisionInterface()->AddKnownEntity( pFocus ); // keep after ForgetEntity
+			}
+		}
+
+		me->GetBodyInterface()->AimHeadTowards( pFocus->WorldSpaceCenter(), IBody::IMPORTANT, 1.0f, nullptr, "Ghost carrier focus look" );
+	}
+}
+
+void CNEOBotGhostEquipmentHandler::EquipBestWeaponForGhoster( CNEOBot *me )
```
AdamTadeusz, line 124: I don't know if we will keep this feature, but perhaps the bots can consider the value of sv_neo_ctg_ghost_beacons_when_inactive when deciding what weapon to equip

AdamTadeusz (approved):

src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
sunzenshen: I considered this idea and even started implementing it, but then I started thinking that honestly the bot callouts for their bot teammates are accurate enough that I'd rather not give bots another unfair advantage over human players. I think it's fine to keep the bots following the base rules for CTG even when sv_neo_ctg_ghost_beacons_when_inactive is true, as I think it's more fun when you can disrupt a bot's callouts by aggro'ing them, etc.

sv_neo_ctg_ghost_beacons_when_inactive  is pretty much a cheat for human players who are too shy to pick up the ghost, and bots may not need that extra incentive for now.  Bots are also much more fluent than humans at swapping to their sidearm and engaging targets. I think it's fine if the bots have a consistent time penalty of having to switch to their sidearm from the ghost.

(We can reconsider this in a new PR, but I'd like to keep this discussion out of scope for this PR.)

src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
```diff
@@ -0,0 +1,653 @@
...
+{
+	const CKnownEntity *threat = me->GetVisionInterface()->GetPrimaryKnownThreat();
+
+	CNEOBaseCombatWeapon *pActive = dynamic_cast<CNEOBaseCombatWeapon*>( me->GetActiveWeapon() );
+	CNEOBaseCombatWeapon *pGhost = dynamic_cast<CNEOBaseCombatWeapon*>( me->Weapon_GetSlot( 0 ) );
+	CNEOBaseCombatWeapon *pSecondary = dynamic_cast<CNEOBaseCombatWeapon*>( me->Weapon_GetSlot( 1 ) );
```
Rainyan: This seems to be running at a 0.1 seconds update interval which isn't that often, so it doesn't have too bad of a performance impact, but these could probably be swapped for `assert_cast` to make the cast faster for release builds.

src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
```diff
@@ -0,0 +1,653 @@
...
+void CNEOBotGhostEquipmentHandler::UpdateGhostCarrierCallout( CNEOBot *me, const CUtlVector<CNEO_Player*> &enemies )
+{
+	CNEO_Player *pBestCallout = nullptr;
+	float flBestCalloutMoved = -1.0f;
+	float flBestCalloutDist = FLT_MAX;
+	bool bBestCalloutIsNew = false;
+	
+	bool bConsideringOnlyLoSEnemies = false;
+	
+	const Vector vecMyPos = me->GetAbsOrigin();
```
Rainyan: `vecMyPos` could be made `const Vector&` to avoid the copy

src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
```diff
@@ -0,0 +1,653 @@
...
+	{
+		Vector bestPos = CNEO_Player::VECTOR_INVALID_WAYPOINT;
+		float flNearestCapDistSq = FLT_MAX;
+		for( int i=0; i<NEORules()->m_pGhostCaps.Count(); ++i )
+		{
+			CNEOGhostCapturePoint *pCapPoint = dynamic_cast<CNEOGhostCapturePoint*>( UTIL_EntityByIndex( NEORules()->m_pGhostCaps[i] ) );
+			if (!pCapPoint)
+			{
+				continue;
+			}
```
Rainyan, line 402: It's a bit unfortunate that we have to dynamic_cast here, might be nicer if `m_pGhostCaps` stored `CUtlVector<CHandle<CNEOGhostCapturePoint>>` to avoid it, and the risk of the cached entindex going stale. Maybe out of scope for this PR, but we should probably refactor it at  some point.

src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
```diff
@@ -0,0 +1,653 @@
...
+		{
+			// regroup with nearest friendlies if alone for too long
+			pTargetTeammate = pTeammateNearestToMe;
+		}
+	}
+
+	if ( pTargetTeammate )
+	{
+		float flDistToTeammate = me->GetAbsOrigin().DistTo( pTargetTeammate->GetAbsOrigin() );
+		if ( flDistToTeammate < 100.0f )
```
Rainyan: Nit: could optimize to squared check < 100*100 to avoid the sqrt

src/game/server/neo/bot/behavior/neo_bot_ctg_escort.cpp
```diff
@@ -0,0 +1,349 @@
...
+	{
+		m_lostSightOfCarrierTimer.Invalidate();
+
+		if ( !m_bHasGoal )
+		{
+			// Asymmetric defense: No goal cap zone, so defend the carrier.
+			// Look away from the carrier to cover their blind spots
+			Vector vecFromCarrier = me->GetAbsOrigin() - pGhostCarrier->GetAbsOrigin();
+			vecFromCarrier.z = 0.0f; // Bias towards horizontal scanning
+			if ( VectorNormalize( vecFromCarrier ) > 0.1f )
```
Rainyan: What does this > 0.1 comparison do? If it's a look direction thing, should we be using a dot product? Or am I thinking about this wrong

src/game/server/neo/bot/behavior/neo_bot_ctg_escort.cpp
```diff
@@ -0,0 +1,349 @@
...
+				// Hold position
+				m_chasePath.Invalidate();
+				m_path.Invalidate();
+			}
+			return Continue();
+		}
+
+		m_chasePath.Invalidate();
+
+		Vector vecMoveTarget = m_vecGoalPos;
```
Rainyan: This could probably be `Vector&` to avoid the copy.

src/game/server/neo/bot/behavior/neo_bot_ctg_escort.cpp
```diff
@@ -0,0 +1,349 @@
...
+	const int iMyTeam = me->GetTeamNumber();
+
+	// Collect candidates and process roles in one pass
+	float flBestDistToGoal = FLT_MAX;
+	float flBestDistToCarrier = FLT_MAX;
+	float flSecondBestDistToCarrier = FLT_MAX;
+	CNEO_Player* pBestToCarrier = nullptr;
+	CNEO_Player* pSecondBestToCarrier = nullptr;
+
+	Vector vecCarrierOrigin = pGhostCarrier->GetAbsOrigin();
```
Rainyan: Could be `const Vector&` to avoid the vector copy

src/game/server/neo/bot/behavior/neo_bot_ctg_escort.cpp
```diff
@@ -0,0 +1,349 @@
...
+
+	for ( int i = 1; i <= gpGlobals->maxClients; i++ )
+	{
+		CNEO_Player *pPlayer = ToNEOPlayer( UTIL_PlayerByIndex( i ) );
+		if ( !pPlayer || !pPlayer->IsAlive() || pPlayer->GetTeamNumber() != iMyTeam || pPlayer == pGhostCarrier )
+		{
+			continue;
+		}
+
+		Vector vecPlayerOrigin = pPlayer->GetAbsOrigin();
```
Rainyan: Preferably `const Vector&`

src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
```diff
@@ -0,0 +1,501 @@
...
+			// Aggressively reload due to lack of backup
+			me->PressReloadButton();
+		}
+	}
+
+	// We dropped the ghost to hunt a threat.
+	if ( m_bPursuingDropThreat )
+	{
+		// First, ensure we have a weapon.
+		if ( !me->Weapon_GetSlot( 0 ) )
```
Rainyan, line 63: Might've been an odd coincidence, but while observing the bots in this PR, I spotted one occurrence of the lone ghoster dropping the ghost, switching to pistol, and then running straight ahead, which resulted in them picking up the ghost again in their back. Then they tried to fight the enemy team with the pistol whilst still having the ghost in their back passively broadcasting their exact location to the enemy.

This is actually a pretty common mistake for newer players, so maybe we'd actually want the easier bots to be susceptible. But maybe there's some way to detect the inactive ghost in the inventory and drop it if it's not wanted by the bot (and also avoid the ghost pickup if the bot wants a gun and not a ghost). Hackier solutions could involve being able to toss the ghost to the opposite direction of where the bot wants to go, or being able to refuse the ghost pickup at will, although personally I'd prefer the bots have as little non-human special powers as possible.

But anyway, these kinds of things are probably out of scope for this PR and could be enhanced at some later time, so I don't have any actual request for changes here.

src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
```diff
@@ -0,0 +1,501 @@
...
+
+	// Occasionally reconsider which cap zone is our goal
+	if ( !m_capPointUpdateTimer.HasStarted() || m_capPointUpdateTimer.IsElapsed() )
+	{
+		m_closestCapturePoint = CNEO_Player::VECTOR_INVALID_WAYPOINT;
+		float flNearestCapDist = FLT_MAX;
+		
+		if ( NEORules()->m_pGhostCaps.Count() > 0 )
+		{
+			Vector vecStart = me->IsCarryingGhost() ? me->GetAbsOrigin() : m_hGhost->GetAbsOrigin();
```
Rainyan: Would prefer `const Vector& vecStart = ...` to avoid making a copy

src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
```diff
@@ -0,0 +1,501 @@
...
+				}
+			}
+		}
+		m_capPointUpdateTimer.Start( RandomFloat( 0.5f, 1.0f ) );
+	}
+
+	float flDistGhostToGoal = FLT_MAX;
+	if ( m_closestCapturePoint != CNEO_Player::VECTOR_INVALID_WAYPOINT )
+	{
+		Vector vecStart = me->IsCarryingGhost() ? me->GetAbsOrigin() : m_hGhost->GetAbsOrigin();
```
Rainyan: Preferably `const Vector& vecStart`

src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
```diff
@@ -0,0 +1,501 @@
...
+	
+	for ( int i = 1; i <= gpGlobals->maxClients; i++ )
+	{
+		CNEO_Player *pPlayer = ToNEOPlayer( UTIL_PlayerByIndex( i ) );
+		if ( pPlayer && pPlayer->IsAlive() && pPlayer->GetTeamNumber() != me->GetTeamNumber() )
+		{
+			iEnemyTeamCount++;
+			if ( m_closestCapturePoint != CNEO_Player::VECTOR_INVALID_WAYPOINT )
+			{
+				float d = pPlayer->GetAbsOrigin().DistTo( m_closestCapturePoint );
```
Rainyan: Nit: could probably refactor these comparisons to squared distance? Or if it gets too messy then no big deal.

src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.h
```diff
@@ -0,0 +1,44 @@
...
+	CountdownTimer m_lookAroundTimer;
+	CountdownTimer m_stalemateTimer;
+
+	CountdownTimer m_capPointUpdateTimer;
+	Vector m_closestCapturePoint;
+
+	CUtlVector< CNavArea * > m_visibleAreas;
+};
+
+#endif // NEO_BOT_CTG_LAST_STAND_H
```
Rainyan: Mismatching preprocessor guard comment for the closing `NEO_BOT_CTG_LONE_WOLF_H` at the end of this line 44.

AFAIK we also have full support for `#pragma once` on all platforms/compilers that we target btw, which would be terser than the `ifndef ... define ... endif` pattern.

src/game/server/neo/bot/behavior/neo_bot_ctg_seek.cpp
```diff
@@ -0,0 +1,144 @@
...
+			{
+				if ( me->IsLineOfSightClear( m_hTargetEntity, CBaseCombatCharacter::IGNORE_ACTORS ) )
+				{
+					const char *classname = m_hTargetEntity->GetClassname();
+					if ( FStrEq( classname, "weapon_ghost" ) )
+					{
+						CBaseCombatWeapon* pWeapon = m_hTargetEntity->MyCombatWeaponPointer();
+						if ( pWeapon && !pWeapon->GetOwner() )
+						{
+							CWeaponGhost *pGhost = dynamic_cast<CWeaponGhost*>( m_hTargetEntity.Get() );
```
Rainyan: Nit: `assert_cast` should be safe here for better perf than `dynamic_cast`, and it should be safe to do since we do the classname check above.

src/game/server/CMakeLists.txt
```diff
@@ -1443,6 +1443,18 @@ target_sources_grouped(
     neo/bot/behavior/neo_bot_command_follow.h
     neo/bot/behavior/neo_bot_dead.cpp
     neo/bot/behavior/neo_bot_dead.h
+    neo/bot/behavior/neo_bot_ctg_capture.cpp
```
Rainyan: Nit: the `...neo_bot_dead...` would alphabetically go below these newly added behaviors

src/game/server/neo/bot/behavior/neo_bot_ctg_escort.cpp
sunzenshen: The original intent was to make sure the vector length isn't zero or negative but maybe this is a redundant check (on reflection, those values seem unlikely for a length of a vector, even when a bot is on top of a ghost carrier).  Maybe this should be larger than 32 hu in order to skip this forced look if a bot is directly inside of a ghost carrier, but I was kind of relying on the teammate bumping avoidance behavior to sort out any rough spots. (Taking a look though.)

src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
sunzenshen: I agree that ideally we would find a solution for the accidental ghost pickup, but honestly this edge case was why I introduced m_bPursuingDropThreat above.  The scenario I had in mind was a ghoster trying to drop the ghost and fumbling the drop, while in full or imminent view of an enemy.  In that case, continuing to try to drop the ghost, especially in a claustrophobic environment seems to me a bigger risk than pursuing the enemy while having their location be known.

Plus diagnosing why a bot might pick up the wrong weapon might be easier to code review in a PR that doesn't have the other changes.

src/game/server/neo/bot/behavior/neo_bot_ctg_escort.cpp
Rainyan: Ah I see, the function returns the magnitude before the normalization.

Rainyan (approved):

PR #1601, opened 2026-01-12
Dynamic bot jump path penalty based on class max

sunzenshen:
## Description
Penalizes bot path NavArea candidates that involve jumping, scaled to the max jump height for the class of that bot.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/neo_bot_path_cost.cpp
+++ b/src/game/server/neo/bot/neo_bot_path_cost.cpp
@@ -12,7 +12,7 @@ ConVar neo_bot_path_friendly_reservation_enable("neo_bot_path_friendly_reservati
 ConVar neo_bot_path_around_friendly_cooldown("neo_bot_path_around_friendly_cooldown", "2.0", FCVAR_CHEAT,
 	"How often to check for friendly path dispersion", false, 0, false, 60);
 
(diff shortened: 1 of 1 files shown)


src/game/server/neo/bot/neo_bot_path_cost.cpp
```diff
@@ -12,6 +12,9 @@ ConVar neo_bot_path_friendly_reservation_enable("neo_bot_path_friendly_reservati
 ConVar neo_bot_path_around_friendly_cooldown("neo_bot_path_around_friendly_cooldown", "2.0", FCVAR_CHEAT,
 	"How often to check for friendly path dispersion", false, 0, false, 60);
 
+ConVar neo_bot_path_jump_penalty_multiplier("neo_bot_path_jump_penalty_multiplier", "100.0", FCVAR_CHEAT,
+	"Maximum penalty multiplier for jump height changes in pathfinding", false, 0.01f, false, 1000.0f);
```
sunzenshen, line 16: This PR will be much easier to test with #1345, but with a 100 multiplier value, bots will recognize the easiest jump point near the fence area in dawn, and will avoid trying to jump onto the trailer in threadplate (unless the ghost spawns on it, where then the bots will path with jumps onto the trailer with no problem)

AdamTadeusz (approved):

PR #1603, opened 2026-01-12
Slightly opaque smoke particles in thermals and draw glows when drawing smoke fog overlay, don't reset glows enabled setting when mp_forcecamera is toggled, draw smoke particles normally in spectate but not the fog overlay

### LINKED ISSUE #1595: Smoke Grenade particles should be visible slightly when in thermals or when spectating and using xray

### LINKED ISSUE #1596: Glowing objects should still render their glow effect when the observer is in range of a smoke grenade

AdamTadeusz:
## Description

So to clarify the smoke particles bit, the behavior now should be that when you are looking through the eyes of a player you should see the smoke particles and smoke fog overlay the same way that player sees them. In other spectating cases the particles should have the same opacity as when viewed normally without thermals, but the fog overlay does not draw for spectators regardless of how close they are to the smoke grenade

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #1595
- fixes #1596

### DIFF
--- a/src/game/client/smoke_fog_overlay.cpp
+++ b/src/game/client/smoke_fog_overlay.cpp
@@ -169,9 +169,9 @@ void UpdateThermalOverride()
 		}
 	}
 #ifdef GLOWS_ENABLE
-	else if (localPlayer->IsObserver() && glow_outline_effect_enable.GetBool() && (localPlayer->GetTeamNumber() == TEAM_SPECTATOR || mp_forcecamera.GetInt() == OBS_ALLOW_ALL))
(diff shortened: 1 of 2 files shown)


Rainyan (approved):
LGTM, for reference:

### OGNT

devbox0001

devbox0002

### NT;RE before this patch

(the git hash in the screenshots is stale from not doing a full rebuild...)

devbox0001

devbox0002

### NT;RE with this patch applied

devbox0001

devbox0002

PR #1610, opened 2026-01-14
Don't glow dm players

AdamTadeusz:
## Description

This also fixes a problem where the glows don't update when the game mode changes. a RecvProxy on the m_nGameTypeSelected variable doesn't work, event messages probably could work but they arrive before the neorules data (including the selected game mode) arrives and anyway the developer wiki says to not run entity logic there and to use PostDataUpdate instead, but the comment in the source code describing PostdataUpdate suggests OnDataChanged should be used instead, except the neorules object doesn't extend C_BaseEntity, probably why the game rules proxy exists.

We could probably detect the start of a new round in a similar fashion to how we detect the game mode change here to do stuff that should be done to every player on spawn client side, although a user message could work too not sure which one would be the best way to do it


sunzenshen (approved):
LGTM:

https://github.com/user-attachments/assets/127072bf-6154-4c49-9d8c-fef4b8cd8d17

(Also checked that glows still happen in CTG mode at least.)

PR #1611, opened 2026-01-15
New spectate commands

### LINKED ISSUE #1647: [FEATURE] Spectator binds for observing matches
### Description

The [plugin](https://github.com/Rainyan/sourcemod-nt-spec-quicktarget?tab=readme-ov-file#console-commands) we've used in OGNT probably doesn't work in NT;RE. It'd be nice to have the functionality implemented natively.

### Proposed Solution

The most important binds are the number keys to jump to specific players. 1-5 for Jinrai and 6-0 for NSF. This could probably be implemented by building on #1611.

Additionally the binds to jump to the last event of various kinds would be nice to haves.

### Alternatives Considered

_No response_

### Additional Context

_No response_

AdamTadeusz:
## Description

This PR now adds commands for spectating a player based on their position in the new top hud, as well as spectating a player closest to the center of the screen and spectating the fastest player.

The following commands require the new squad hud to work reliably/at all

spec_player_by_hud_position works similar to sm_spec_slot, except theres no guarantee that the first 5 numbers will correspond to jinrai entirely, or that the next 5 numbers will correspond to nsf entirely. In a 3v3 scenario, spec_player_by_hud_position 0-2 will return jinrai players, and spec_player_by_hud_position 3-5 will return nsf players. In a 10v10 scenario spec_player_by_hud_position 0-9 will return jinrai players, and 10-19 will return nsf players. This makes the same binds work for any tournament format, though in games with more than 10 players you are likely to run out of binds.

select_next/previous_alive_player_in_hud change which player in the top hud is selected, without changing the observer mode to that player. Everytime the selection changes that player will flash with an orange colour for one second. The command spectate_player_selected_in_hud can then be used to change the spectate target to the currently selected one.

spec_next/previous_entity_in_hud works similar to select, except the spectate target is changed immediately. The next and previous targets are determined by the current target, not where the selection is.

The following commands will work regardless of the new squad hud being enabled

spec_player_under_mouse will change the spectate target to whoever's center of mass is closes to the center of the spectator's screen

spec_fastest_player will change the spectate target to whoever is currently moving the fastest.

I have included an example config to try out the new commands. I recommend backing up your current config/controls beforehand

(Edit)

I have also added 
    sm_spec_last_hurt — Target on the last player who was damaged.
    sm_spec_last_shooter — Target on the last player who fired their weapon.
    sm_spec_last_event — Target on the latest event of any kind from the list below:
    sm_spec_last_attacker — Target on the last player who did damage.
    sm_spec_last_killer — Target on the last player who got a kill.
    sm_spec_last_ghoster — Target on the last ghost carrier.

In the form of
    spec_last_hurt
    spec_last_shooter
    spec_last_event
    spec_last_attacker
    spec_last_killer
    spec_last_ghoster

(Edit 2)

I have also added a command to automatically spectate the killer of the current spectate target

-fixes #1647

### DIFF
--- a/src/game/client/neo/ui/neo_hud_round_state.cpp
+++ b/src/game/client/neo/ui/neo_hud_round_state.cpp
@@ -554,17 +554,8 @@ void CNEOHud_RoundState::DrawNeoHudElement()
 	if (!g_PR)
 		return;
 
-	if (cl_neo_squad_hud_original.GetBool())
-	{
-		DrawPlayerList();
-		return;
-	}
-
-	// Draw players on top
-	int leftCount = 0;
-	int rightCount = 0;
-	if (NEORules()->IsTeamplay())
-	{
+	if (NEORules()->IsTeamplay() && (!cl_neo_squad_hud_original.GetBool() || localPlayerSpecOrNoTeam))
+	{ // Sort player list even if not drawing new hud so spectators can use commands
 		for (int i = 0; i < MAX_PLAYERS; i++)
 		{ // First pass update player values, and count them while we're doing it
 			constexpr const int INDEX_SHIFT = 6;
@@ -619,7 +610,19 @@ void CNEOHud_RoundState::DrawNeoHudElement()
 		}
 
 		m_nPlayerList.Sort([](const playerIndexAndTheirValue *first, const playerIndexAndTheirValue *second)->int{return second->playerValue - first->playerValue;});
+	}
 
+	if (cl_neo_squad_hud_original.GetBool())
+	{
+		DrawPlayerList();
+		return;
+	}
+
+	// Draw players on top
+	int leftCount = 0;
+	int rightCount = 0;
+	if (NEORules()->IsTeamplay())
+	{
 		// Fade background to make names easier to see
 		surface()->DrawSetColor(COLOR_DARK);
 		surface()->DrawFilledRectFade((m_iLeftOffset - 2) - (m_iLeftPlayersTotal * m_ilogoSize) - (m_iLeftPlayersTotal * 2), 0,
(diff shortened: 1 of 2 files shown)


AdamTadeusz:
Now I'm wondering if it would be better to workout the player closest to the center client side and have a command sent to the server to spectate a specific player instead.

src/game/client/clientmode_shared.cpp
```diff
@@ -823,6 +823,47 @@ int ClientModeShared::HandleSpectatorKeyInput( int down, ButtonCode_t keynum, co
...
+			C_NEO_Player* pPlayer = ToNEOPlayer(UTIL_PlayerByIndex(i));
+			if (currentTarget != pPlayer && pNeoPlayer->IsValidObserverTarget(pPlayer) && pPlayer->IsAlive())
+			{
+				Vector vecForward;
+				AngleVectors( pNeoPlayer->EyeAngles(), &vecForward );
+
+				Vector vecToTarget = pPlayer->WorldSpaceCenter() - pNeoPlayer->EyePosition();
+				vecToTarget.NormalizeInPlace();
+				float dotProduct = DotProduct(vecForward, vecToTarget);
+				if (dotProduct > targetDotProduct && dotProduct > 0)
```
Agiel: `dotProduct > 0` seems a bit generous here. Maybe do 0.5 for a 60º cone?

src/game/client/neo/ui/neo_hud_round_state.cpp
Agiel: Is there a way we can do the sorting before this block? That way the spec binds would work with the classic squad hud as well (would be on the player to also disable sort_players_by_class_alive_and_star for the order to make sense).

AdamTadeusz: most definitely yes

Agiel:
This should close #1647

sunzenshen (approved):

Issue #1330, opened 2025-08-30
Dedicated servers running sv_neo_comp 1 will flood console after some time


DESTROYGIRL:
20250824_bebf05c (v18.0)

Example:
`Server has been empty for 2059 seconds on this map, cycling map as per mp_mapcycle_empty_timeout_seconds`
`DMSpawn file not found: maps/dm_locs/ntre_redlight_ctg.loc`
And it will keep spamming this message until a player joins again.

`mp_mapcycle_empty_timeout_seconds` is `1800` by default so to avoid this servers running comp would have to set it to 0, but this functionality should prolly still be restored

DESTROYGIRL:
When I say spam I ought to have specified it was a constant stream of messages many times each second- like something was failing- making the console unreadable

PR #1620, opened 2026-01-20
Add default mapcycle

Rainyan:
## Description

Add a default mapcycle file.

The name of the mapcycle file is determined by the cvar `mapcyclefile`, according to the logic of `::DetermineMapCycleFilename`. By default, it is `mapcycle.txt` which is used here.

Without a mapcycle file, the next map action will simply reload the current map, which is typically not what the user/server operator wants.

This map list is adapted from the OG NT default mapcycle, with the following changes:

* The following maps use the JGR variant: `sentinel`, `terminal`

* The following CTG map is added: `rogue`

* All maps are of the `ntre_...` filename variant, to match the NT;RE naming scheme, compared to `nt_...` of OG NT map names

## Toolchain
N/A

## Linked Issues
- fixes #
- related #1330


Rainyan:
> @blaberry May correct me if I am wrong but a default mapcycle file is automatically generated as part of the steam build with all the maps included (not featuring backgrounds, tutorials / devbox)

Ah, I didn't realise that. Perhaps this PR is redundant then.

PR #1622, opened 2026-01-22
Store achievement progress in steam cloud

### LINKED ISSUE #1562: Tutorial completion achievement progress is not saved between sessions
Progress appears to never be saved to steam or anywhere, so restarting the game or steam makes it reset.

DESTROYGIRL:
## Description
This fixes achievement progress being removed on game restart and should allow for achievement progress to be saved across systems. This also improves communication with the steam stats API for component type achievements

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1562

### DIFF
--- a/src/game/shared/achievementmgr.cpp
+++ b/src/game/shared/achievementmgr.cpp
@@ -161,7 +161,7 @@ static void WriteAchievementGlobalState( KeyValues *pKV, bool bPersistToSteamClo
 #ifdef NEO
 								else
 								{
-									Warning("CAchievementMgr: Failed to write to steam cloud! Expected size %d got size %d\n", filesize, nRead);
+									Warning("%s: Failed to write to steam cloud! Expected size %d got size %d\n", __FUNCTION__, filesize, nRead);
 								}
 #endif
                             }
--- a/src/game/shared/achievementmgr.h
+++ b/src/game/shared/achievementmgr.h
@@ -34,11 +34,7 @@ public:
         SteamCloudPersist_On,
     };
 
-#ifdef NEO
-	CAchievementMgr( SteamCloudPersisting ePersistToSteamCloud = SteamCloudPersist_On );
-#else
 	CAchievementMgr( SteamCloudPersisting ePersistToSteamCloud = SteamCloudPersist_Off );
-#endif
 
     //=============================================================================
     // HPE_END
--- a/src/game/shared/baseachievement.cpp
+++ b/src/game/shared/baseachievement.cpp
@@ -241,7 +241,11 @@ void CBaseAchievement::IncrementCount( int iOptIncrement )
 			bool bRet = steamapicontext->SteamUserStats()->SetStat( pszProgressName, m_iCount );
 			if ( !bRet )
 			{
+#ifdef NEO
+				DevWarning( "ISteamUserStats::SetStat failed to set progress value in Steam for achievement %s\n", pszProgressName );
+#else
 				DevMsg( "ISteamUserStats::GetStat failed to set progress value in Steam for achievement %s\n", pszProgressName );
+#endif
 			}
 
 			m_pAchievementMgr->SetDirty( true );
@@ -531,7 +535,7 @@ void CBaseAchievement::SetComponentBits( uint64 iComponentBits )
 		bool bRet = steamapicontext->SteamUserStats()->SetStat( pszProgressName, m_iCount );
 		if ( !bRet )
 		{
-			DevMsg( "ISteamUserStats::GetStat failed to set progress value in Steam for achievement %s\n", pszProgressName );
+			DevWarning( "ISteamUserStats::SetStat failed to set progress value in Steam for achievement %s\n", pszProgressName );
 		}
 
 		m_pAchievementMgr->SetDirty( true );
--- a/src/game/shared/neo/achievements_neo.cpp
+++ b/src/game/shared/neo/achievements_neo.cpp
@@ -2,7 +2,7 @@
 #ifdef CLIENT_DLL
 #include "achievements_neo.h"
 
-CAchievementMgr g_AchievementMgrNEO;	// global achievement mgr for NEO
+CAchievementMgr g_AchievementMgrNEO( CAchievementMgr::SteamCloudPersist_On );	// global achievement mgr for NEO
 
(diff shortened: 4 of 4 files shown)


src/game/shared/baseachievement.cpp
```diff
@@ -519,6 +519,25 @@ void CBaseAchievement::SetComponentBits( uint64 iComponentBits )
...
+	// if this achievement's progress should be stored in Steam, set the steam stat for it
+	if ( StoreProgressInSteam() && steamapicontext->SteamUserStats() )
+	{
+		// Set the Steam stat with the same name as the achievement.  Only cached locally until we upload it.
+		char pszProgressName[1024];
+		Q_snprintf( pszProgressName, 1024, "%s_STAT", GetStat() );
+		bool bRet = steamapicontext->SteamUserStats()->SetStat( pszProgressName, m_iCount );
+		if ( !bRet )
+		{
+			DevMsg( "ISteamUserStats::GetStat failed to set progress value in Steam for achievement %s\n", pszProgressName );
```
Rainyan: Possibly a typo in the `DevMsg` method name (The print message says `GetStat` but the `bRet` error originated from the `SetStat` call) 

Also maybe we want `DevWarning` instead of `DevMsg`, since this is something going wrong.

src/game/shared/achievementmgr.cpp
```diff
@@ -154,6 +158,12 @@ static void WriteAchievementGlobalState( KeyValues *pKV, bool bPersistToSteamClo
                                     // Write out the data to steam cloud
                                     pRemoteStorage->FileWrite(szFilename, pData, filesize);
                                 }
+#ifdef NEO
+								else
+								{
+									Warning("CAchievementMgr: Failed to write to steam cloud! Expected size %d got size %d\n", filesize, nRead);
```
Rainyan: Nit: Technically not an error in `CAchievementMgr`, since we're inside the free-standing static function `WriteAchievementGlobalState`. Perhaps instead format as `Warning("%s: ...", __FUNCTION__`, where `__FUNCTION__` will be automatically substituted by the current function name by the compiler

src/game/shared/achievementmgr.cpp
```diff
@@ -141,7 +141,11 @@ static void WriteAchievementGlobalState( KeyValues *pKV, bool bPersistToSteamClo
                         if (pData)
                         {
                             // Read in the data from the file system GameState.txt file
+#ifdef NEO
+							FileHandle_t	handle = filesystem->Open(szFilename, "rb");
```
Rainyan, line 145: Why do we need to open the text file in binary mode? Encoding related?

src/game/shared/achievementmgr.h
```diff
@@ -34,7 +34,11 @@ class CAchievementMgr : public CAutoGameSystemPerFrame, public CGameEventListene
         SteamCloudPersist_On,
     };
 
+#ifdef NEO
+	CAchievementMgr( SteamCloudPersisting ePersistToSteamCloud = SteamCloudPersist_On );
+#else
 	CAchievementMgr( SteamCloudPersisting ePersistToSteamCloud = SteamCloudPersist_Off );
+#endif
```
Rainyan: Opinionated, but perhaps instead of modifying the base SDK constructor defaults, it might be cleaner to instantiate our NEO-specific instance with the desired value:

```diff
diff --git a/src/game/shared/achievementmgr.h b/src/game/shared/achievementmgr.h
index 7a990fbf..c33261e2 100644
--- a/src/game/shared/achievementmgr.h
+++ b/src/game/shared/achievementmgr.h
@@ -34,11 +34,7 @@ public:
         SteamCloudPersist_On,
     };

-#ifdef NEO
-       CAchievementMgr( SteamCloudPersisting ePersistToSteamCloud = SteamCloudPersist_On );
-#else
        CAchievementMgr( SteamCloudPersisting ePersistToSteamCloud = SteamCloudPersist_Off );
-#endif

     //=============================================================================
     // HPE_END
diff --git a/src/game/shared/neo/achievements_neo.cpp b/src/game/shared/neo/achievements_neo.cpp
index caef7bf8..b6a005ff 100644
--- a/src/game/shared/neo/achievements_neo.cpp
+++ b/src/game/shared/neo/achievements_neo.cpp
@@ -2,7 +2,7 @@
 #ifdef CLIENT_DLL
 #include "achievements_neo.h"

-CAchievementMgr g_AchievementMgrNEO;   // global achievement mgr for NEO
+CAchievementMgr g_AchievementMgrNEO(CAchievementMgr::SteamCloudPersist_On);    // global achievement mgr for NEO

 class CAchievementNEO_TutorialComplete : public CBaseAchievement
 {
 ```
 
 But if you prefer this way, I'm ok with that too, so I don't necessarily require a change here.

Comment (DESTROYGIRL) [reply src/game/shared/achievementmgr.h]: I tried this initially and got compiler errors but works fine now so... guess I screwed something up -_-

File: src/game/shared/achievementmgr.cpp (hunk 3 above)
Comment (DESTROYGIRL) [reply src/game/shared/achievementmgr.cpp:145]: `filesystem->Size()` never returned the same size as `filesystem->Read()` hence the warning added below here. Changing this though they always matched up. Not sure if I'm doing it wrong or just the steam cloud persitance section here has broken at some point in time because it's not used in TF2

Comment (Rainyan) [reply src/game/shared/achievementmgr.cpp:145]: It seems that `filesystem->Read` in text mode returns the number of characters read, and this doesn't include the `\r` carriage return that Windows uses for newlines, so `nRead` is less by the number of newlines in the file compared to the actual file size. So I guess reading as binary is indeed the best choice here, to avoid the `\n` vs `\r\n` size confusion.

**Review by Rainyan — APPROVED** [review]
Tested the achievements with debug dlls in the NT;RE appid location, with:
```
achievement_debug 1;
achievement_reset_all;
```
and quitting the game to check the gamestate.txt contents, and then with:
```
achievement_debug 1;
achievement_unlock_all;
```
and quitting to store the gamestate.txt.

Seems that the gamestate.txt updates fine for both Windows & Linux.

Also verified that earning (or debug-resetting) an achievement on NT;RE running in Linux Steam client will cloud sync that achievement status update with a Windows Steam client, and vice versa.

PR #1627, opened 2026-01-23
Only reset XP down to sergeant inbetween JGR rounds

DESTROYGIRL:
## Description
Rather than resetting all XP only rank players down to sergeant if they had higher than 10 xp, otherwise players keep the xp they had

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/shared/neo/neo_gamerules.cpp
+++ b/src/game/shared/neo/neo_gamerules.cpp
@@ -3094,9 +3094,9 @@ void CNEORules::ResetJGR()
 	for (int i = 1; i <= gpGlobals->maxClients; ++i)
 	{
 		auto pPlayer = static_cast<CNEO_Player *>(UTIL_PlayerByIndex(i));
-		if (pPlayer)
+		if (pPlayer && pPlayer->m_iXP.Get() > 10)
 		{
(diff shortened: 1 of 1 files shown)


src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -3094,9 +3094,9 @@ void CNEORules::ResetJGR()
 	for (int i = 1; i <= gpGlobals->maxClients; ++i)
 	{
 		auto pPlayer = static_cast<CNEO_Player *>(UTIL_PlayerByIndex(i));
-		if (pPlayer)
+		if (pPlayer && pPlayer->m_iXP.Get() > 10)
 		{
-			pPlayer->m_iXP.GetForModify() = 0;
+			pPlayer->m_iXP.GetForModify() = 10;
```
sunzenshen, line 3099: Could maybe use the constant XP_SERGEANT from `src/game/shared/neo/neo_weapon_loadout.h`

DESTROYGIRL: Those should probably be moved into neo_enums or something because I did not see them and they don't seem to be used inside of gamerules for xp stuff currently

sunzenshen (approved):

PR #1629, opened 2026-01-25
Bots can trade/donate their primary weapon with their commander

sunzenshen:
## Description
- When a commander of bots has no primary and looks at one of their subordinates, that subordinate will throw their weapon at the commander as as donation.
- Can also be used to trade weapons if a commander throws their weapon at the subordinate, as the subordinate will collect the nearest weapon after throwing their weapon.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
@@ -73,7 +73,7 @@ ActionResult< CNEOBot >	CNEOBotCommandFollow::Update(CNEOBot *me, float interval
 ActionResult< CNEOBot > CNEOBotCommandFollow::CheckCommanderWeaponRequest(CNEOBot *me)
 {
 	// Can I even drop my primary? (e.g. Juggernaut's BALC)
-	CNEOBaseCombatWeapon *pMyPrimary = dynamic_cast<CNEOBaseCombatWeapon*>(me->Weapon_GetSlot(0));
+	CNEOBaseCombatWeapon *pMyPrimary = assert_cast<CNEOBaseCombatWeapon*>(me->Weapon_GetSlot(0));
 	if (!pMyPrimary || !pMyPrimary->CanDrop())
 	{
 		return Continue();
@@ -104,7 +104,7 @@ ActionResult< CNEOBot > CNEOBotCommandFollow::CheckCommanderWeaponRequest(CNEOBo
 	Vector vecCmdrFacing;
 	pCommander->EyeVectors(&vecCmdrFacing);
 
-	if (vecCmdrFacing.Dot(vecToBot) <= 0.99f)
+	if (vecCmdrFacing.Dot(vecToBot) <= 0.95f)
 	{
 		m_commanderLookingAtMeTimer.Invalidate();
 		m_bWasCommanderLookingAtMe = false;
@@ -121,7 +121,8 @@ ActionResult< CNEOBot > CNEOBotCommandFollow::CheckCommanderWeaponRequest(CNEOBo
 
 	if (m_commanderLookingAtMeTimer.GetElapsedTime() > 0.2f)
 	{
-		me->GetBodyInterface()->AimHeadTowards(pCommander->EyePosition(), IBody::CRITICAL, 0.2f, NULL, "Noticed commander looking at me without a primary");
+		// The throw behavior triggers drop when the bot is looking at the commander, so practively aim at feet to avoid overthrowing
+		me->GetBodyInterface()->AimHeadTowards(pCommander->GetAbsOrigin(), IBody::IMPORTANT, 0.2f, NULL, "Noticed commander looking at me without a primary");
 	}
 
 	if (m_commanderLookingAtMeTimer.GetElapsedTime() > 1.0f)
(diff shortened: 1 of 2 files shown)


src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
```diff
@@ -279,19 +276,6 @@ ActionResult< CNEOBot >	CNEOBotTacticalMonitor::Update( CNEOBot *me, float inter
 		return result;
 	}
 
-	if (sv_neo_bot_cmdr_enable.GetBool())
-	{
-		if (me->m_hLeadingPlayer.Get() || me->m_hCommandingPlayer.Get())
-		{
-			return SuspendFor(new CNEOBotCommandFollow, "Following commander");
```
sunzenshen, line 286: It turns out that launching into the bot commander following behavior was blocking other tactical monitor behaviors, so I moved this behavior launch point into neo_bot_scenario_monitor.

src/game/server/neo/bot/behavior/neo_bot_scenario_monitor.cpp
```diff
@@ -64,10 +66,26 @@ ActionResult< CNEOBot >	CNEOBotScenarioMonitor::OnStart( CNEOBot *me, Action< CN
...
+
 
 //-----------------------------------------------------------------------------------------
 ActionResult< CNEOBot >	CNEOBotScenarioMonitor::Update( CNEOBot *me, float interval )
 {
+	if (sv_neo_bot_cmdr_enable.GetBool())
+	{
+		if (me->m_hLeadingPlayer.Get() || me->m_hCommandingPlayer.Get())
+		{
+			return SuspendFor(new CNEOBotCommandFollow, "Following commander");
```
sunzenshen, line 80: Moved from neo_bot_tactical_monitor, to unblock other tactical monitor behaviors.

src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
```diff
@@ -53,6 +56,81 @@ ActionResult< CNEOBot >	CNEOBotCommandFollow::Update(CNEOBot *me, float interval
...
+		// Check if commander has dropped primary
+		if (!pCommander->Weapon_GetSlot(0))
+		{
+			// Check if looking at me
+			Vector vecToBot = me->EyePosition() - pCommander->EyePosition();
+			vecToBot.NormalizeInPlace();
+			Vector vecCmdrFacing;
+			pCommander->EyeVectors(&vecCmdrFacing);
+
+			if (vecCmdrFacing.Dot(vecToBot) > 0.99f)
```
AdamTadeusz: I struggle to aim at the right spot on a bot to get them to drop their weapon outside of the pre-round freeze, when a bot stands very close to me

AdamTadeusz (dismissed):

src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
sunzenshen: I get what you mean, but when I tried adding some bot movement away from the player, the bot path is often erratic enough that the bot walks away from the player's cursor, even if the movement vector is theoretically in line of the 2 players.  That's why I associated this trading feature with bot commanding, as you really need to be able to control a bot's standing position before you try to initiate a trade.

AdamTadeusz (dismissed):

src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
```diff
@@ -53,11 +56,98 @@ ActionResult< CNEOBot >	CNEOBotCommandFollow::Update(CNEOBot *me, float interval
...
 
 	return Continue();
 }
 
+
+//---------------------------------------------------------------------------------------------
+ActionResult< CNEOBot > CNEOBotCommandFollow::CheckCommanderWeaponRequest(CNEOBot *me)
+{
+	// Can I even drop my primary? (e.g. Juggernaut's BALC)
+	CNEOBaseCombatWeapon *pMyPrimary = dynamic_cast<CNEOBaseCombatWeapon*>(me->Weapon_GetSlot(0));
```
Rainyan: We could use `assert_cast` here for performance

src/game/server/neo/bot/behavior/neo_bot_throw_weapon_at_player.cpp
```diff
@@ -0,0 +1,63 @@
...
+	else
+	{
+		return Done( "No primary weapon to throw" );
+	}
+
+	if ( me->GetBodyInterface()->IsHeadAimingOnTarget() )
+	{
+		me->PressDropButton();
+		return Done("Weapon dropped");
+	}
```
Rainyan, line 55: The bots seem to quite often yeet the gun over the receiver's head, especially at close range (spawn freezetime etc). Maybe ideally we would toss slightly lower than towards the receiver's eyes to account for the arc of the throw?

Rainyan (dismissed):

src/game/server/neo/bot/behavior/neo_bot_throw_weapon_at_player.cpp
sunzenshen: After some testing, the weapons travel surprisingly far even when bots aim at the player's feet, so I settled for that aim target now.

AdamTadeusz (approved):

PR #1630, opened 2026-01-27
Outline Alpha

AdamTadeusz:
## Description

Adds a variable for controlling the opacity of the xray outline

Depends on https://github.com/NeotokyoRebuild/neoAssets/pull/97

## Toolchain

- Windows MSVC VS2022

### DIFF
--- a/src/game/client/glow_outline_effect.cpp
+++ b/src/game/client/glow_outline_effect.cpp
@@ -32,7 +32,7 @@ ConVar glow_outline_effect_enable("glow_outline_effect_enable", "1", FCVAR_ARCHI
 	}
 );
 ConVar glow_outline_effect_width( "glow_outline_effect_width", "1.f", FCVAR_ARCHIVE, "Width of glow outline effect.", true, 0.f, false, 0.f);
-ConVar glow_outline_effect_alpha( "glow_outline_effect_alpha", "1.f", FCVAR_ARCHIVE, "Alpha of glow outline effect.", true, 0.f, true, 1.f);
+ConVar glow_outline_effect_alpha( "glow_outline_effect_alpha", "0.5f", FCVAR_ARCHIVE, "Alpha of glow outline effect.", true, 0.f, true, 1.f);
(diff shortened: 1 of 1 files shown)


sunzenshen (approved):
LGTM:

Outline Opacity values:
(2.00 width for all screenshots)

1.00

Image


0.80

Image


0.60

Image

0.40

Image

0.20

Image

0.10

Image

src/game/client/glow_outline_effect.cpp
```diff
@@ -32,6 +32,7 @@ ConVar glow_outline_effect_enable("glow_outline_effect_enable", "1", FCVAR_ARCHI
 	}
 );
 ConVar glow_outline_effect_width( "glow_outline_effect_width", "1.f", FCVAR_ARCHIVE, "Width of glow outline effect.", true, 0.f, false, 0.f);
+ConVar glow_outline_effect_alpha( "glow_outline_effect_alpha", "0.5f", FCVAR_ARCHIVE, "Alpha of glow outline effect.", true, 0.f, true, 1.f);
```
Rainyan, line 35: Nit: trailing `f` is technically not detected, but instead the parsing happens to break the loop of whatever token iterator it's using at an unidentified character. So `"0.5🙂"` would equally work by happenstance. But I imagine ideally we should not rely on this implementation detail and rather not append the `f` postfix for parsed values.

(I believe the "e" scientific notation does work, assuming whatever the SDK implementation does follows the C(++) spec, but that's tangential). Hexadecimal repr, no idea. Might work?

Rainyan (approved):

src/game/client/glow_outline_effect.cpp
AdamTadeusz: we need to go over the convar flags for a lot of the convars, so thats something we can take a look at in that pr as wel

AdamTadeusz: new issue #1634

PR #1633, opened 2026-01-29
Bots can climb ladders

### LINKED ISSUE #1588: Bots are not able to climb ladders
### Build Info

20269103_fdb1747

### Description

On the current master branch, bots fail to reliably scale ladders. The issue may have multiple contributing causes, such as pathfinding path costs, locomotion logic conflicts, and behavior system overrides.

1. Path Cost Calculation Penalties -  In CNEOBotPathCost::operator(), the check for height change (deltaZ) may prematurely reject valid ladder paths. If the vertical distance exceeds m_maxJumpHeight, the path is often discarded as "too high to reach", ignoring the presence of a ladder pointer which should validate the traversal.  \neo\src\game\server\NextBot\simple_bot.h was used as an implementation example for CNEOBotPathCost, and it appears to have this same lack of checking for ladders.  That said, when an exception is added for ladders and elevators, we encounter downstream problems that make the situation worse (where before the bots would just ignore the path, now they will get actively stuck when the try the following behaviors...).

2. Locomotion Logic Conflict (Crouch-Jump) -  When a bot mounts a ladder, IsOnGround() returns false. CNEOBotLocomotion::Update might interpret this as being "in the air" and automatically presses the Crouch button to facilitate a crouch-jump. In Source Engine, holding crouch while on a ladder might be causing the player entity to detach, preventing ascent.

3. Stuck Monitor False Positives - It kind of looks like that the StuckMonitor logic in CNEOBotMainAction frequently misinterprets this state as being stuck. Consequently, CNEOBotMainAction::OnStuck triggers a recovery maneuver involving Jumping and Strafing (Left/Right). Executing this maneuver while on a ladder may force the bot to jump off or detach immediately.  This hasn't been confirmed as a cause but the behavior looks really similar to standard OnStuck behavior.

4. Aim Focus Distractions - While climbing, the bot's behavior systems UpdateLookingAroundForEnemies might remain active, interrupting the bot when climbing a ladder and effectively disarming the bot mid-climb.

5. ??? - There may be other factors preventing the bots from climbing ladders.

### To Reproduce

- Load a map with vertical ladders, such as ntre_dawn_ctg.
- Spawn a friendly bot (e.g., neo_bot_add 1).
- Order the bot (via command system implemented in #1345) to navigate to a destination that requires climbing a ladder (e.g., from the foot of the ladder to a platform above).
- Observe the bot's behavior when it reaches the ladder.

### Expected behavior

The bot should attach to the ladder and climb steadily to the NavArea at the top, maintaining a view angle roughly aligned with the ladder's ascent direction.

### Actual behavior

The bots spin around, strafe, and jump at the bottom of the ladder.

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

Windows 11

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

sunzenshen:
## Description
Bots now attempt to cross NavAreas that involve ladders.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1588

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
@@ -23,8 +23,7 @@ ActionResult<CNEOBot> CNEOBotLadderClimb::OnStart( CNEOBot *me, Action<CNEOBot> 
 	me->StopLookingAroundForEnemies();
 
 	// Timeout based on ladder length
-	const float averageClimbSpeed = 200.0f;
-	float estimatedClimbTime = m_ladder->m_length / averageClimbSpeed + 1.0f;
+	float estimatedClimbTime = m_ladder->m_length / MAX_CLIMB_SPEED + 1.0f;
 	m_timeoutTimer.Start( estimatedClimbTime );
 
 	m_bHasBeenOnLadder = false;
@@ -69,7 +68,7 @@ ActionResult<CNEOBot> CNEOBotLadderClimb::Update( CNEOBot *me, float interval )
 	}
 
 	// Get current position and target
-	Vector myPos = mover->GetFeet();
+	const Vector& myPos = mover->GetFeet();
 	float currentZ = myPos.z;
 	float targetZ = m_bGoingUp ? m_ladder->m_top.z : m_ladder->m_bottom.z;
 
--- a/src/game/client/neo/ui/neo_root.cpp
+++ b/src/game/client/neo/ui/neo_root.cpp
@@ -1776,6 +1776,7 @@ static constexpr const wchar_t *CREDITSPEOPLELABEL_NAMES[] = {
 	L"Viktor \"Slick Vick\" Svensson",
 	L"Ryan \"Stenchy\" Anderson",
 	L"[title]", // Special Thanks
+	L"Dragoteryx",
 	L"Kasietti",
 	L"Kerim \"Nbc66\" Camdzic",
 	L"Tony \"omega\" Sergi"
--- a/src/game/server/neo/bot/behavior/neo_bot_ladder_approach.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ladder_approach.cpp
@@ -58,10 +58,10 @@ ActionResult<CNEOBot> CNEOBotLadderApproach::Update( CNEOBot *me, float interval
 	ILocomotion *mover = me->GetLocomotionInterface();
 	IBody *body = me->GetBodyInterface();
 
-	Vector myPos = mover->GetFeet();
+	const Vector& myPos = mover->GetFeet();
 
 	// Are we climbing up or down the ladder?
-	Vector targetPos = m_bGoingUp ? m_ladder->m_bottom : m_ladder->m_top;
+	const Vector& targetPos = m_bGoingUp ? m_ladder->m_bottom : m_ladder->m_top;
 
 	// Calculate 2D vector from bot to ladder mount point
 	Vector2D to = ( targetPos - myPos ).AsVector2D();


src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
```diff
@@ -0,0 +1,126 @@
...
+	if ( !m_ladder )
+	{
+		return Done( "No ladder specified" );
+	}
+
+	// Ignore enemies while climbing
+	me->StopLookingAroundForEnemies();
+
+	// Timeout based on ladder length
+	float estimatedClimbTime = m_ladder->m_length / 100.0f + 2.0f;
```
AdamTadeusz: surprised with this value, for a vertical ladder the climb speed is 200hu regardless of class, though this value does vary with the view angle

https://github.com/user-attachments/assets/7e14e520-dbd4-4427-a8f5-40eb7e9832ed

sunzenshen: Admittedly I was leaning towards the longer side because I was worried about bots not having enough time budget.  I don't really have a good intuition of how long ladders are on average in Hammer units.

sunzenshen:
Note: Rebased on top of the fix all warnings PR.

src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
```diff
@@ -0,0 +1,127 @@
...
+	else if ( m_bHasBeenOnLadder )
+	{
+		// We were on the ladder but got knocked off - return to reevaluate situation
+		// Since ladder approach should use ChangeTo climbing behavior, this Done will return to the state BEFORE ladder approach
+		return Done( "Knocked off ladder - reevaluating situation" );
+	}
+
+	// Get current position and target
+	Vector myPos = mover->GetFeet();
+	float currentZ = me->GetAbsOrigin().z;
```
AdamTadeusz: Are there any instances where me->GetAbsOrigin().z != myPos.z?

src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
```diff
@@ -0,0 +1,127 @@
...
+	if ( !m_ladder )
+	{
+		return Done( "No ladder specified" );
+	}
+
+	// Ignore enemies while climbing
+	me->StopLookingAroundForEnemies();
+
+	// Timeout based on ladder length
+	const float averageClimbSpeed = 200.0f;
```
Rainyan: Maybe use `MAX_CLIMB_SPEED` rather than hard code

README.md
```diff
@@ -339,4 +339,5 @@ shaders\fxc\sdk_screenspaceeffect_vs20.vcs
     * Updated as of the TF2 SDK update: [Commit 0759e2e8e179d5352d81d0d4aaded72c1704b7a9](https://github.com/ValveSoftware/source-sdk-2013/commit/0759e2e8e179d5352d81d0d4aaded72c1704b7a9)
 * [Nbc66/source-sdk-2013-ce](https://github.com/Nbc66/source-sdk-2013-ce) - Community Edition for additional fixes prior to the TF2 SDK update
 * [tonysergi/source-sdk-2013](https://github.com/tonysergi/source-sdk-2013) - tonysergi's commits that were missing from the original SDK
+* [Dragoteryx/drgbase](https://github.com/Dragoteryx/drgbase) - Reference for NextBot ladder climbing behavior
```
Rainyan, line 342: Should we also add a line at https://github.com/NeotokyoRebuild/neo/blob/fb61f3c9b06469fa3971a29fb409b4988c494949/src/game/client/neo/ui/neo_root.cpp#L1778-L1782 ?

sunzenshen: I reviewed the NextBot ladder climbing code that I used for reference, and it looks like one person was responsible for all of the code, so that makes it easy to credit.

The credits now look like this on my branch:

image

PR #1638, opened 2026-01-30
Server commands for setting team score

### LINKED ISSUE #1637: [FEATURE] Add server commands for manually setting the scores of Jinrai and NSF
### Description

Would fix having to completely replay the map if teams are playing a tourney game

### Proposed Solution

Add in admin commands for adjusting the amount of points both NSF and Jinrai have won

### Alternatives Considered

OGNT servers use a plugin to address this, but the plugin is not compatible with NT;RE

### Additional Context

_No response_

DESTROYGIRL:
## Description
badda bing badda boom

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1637

### DIFF
--- a/src/game/shared/neo/neo_gamerules.cpp
+++ b/src/game/shared/neo/neo_gamerules.cpp
@@ -508,6 +508,34 @@ static inline void FireLegacyEvent_NeoRoundEnd()
 }
 
 #ifdef GAME_DLL
+CON_COMMAND( sv_neo_score_set_jinrai, "Set point count for team Jinrai" )
+{
+	if ( 2 != args.ArgC() )
+	{
+		Msg( "Usage: %s <score>\n", __FUNCTION__ );
+		return;
+	}
+	
+	CTeam *jinrai = GetGlobalTeam( TEAM_JINRAI );
+	Assert( jinrai );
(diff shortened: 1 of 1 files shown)


src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -508,6 +508,34 @@ static inline void FireLegacyEvent_NeoRoundEnd()
...
+{
+	if ( 2 != args.ArgC() )
+	{
+		Msg( "Usage: %s <score>\n", __FUNCTION__ );
+		return;
+	}
+	
+	CTeam *jinrai = GetGlobalTeam( TEAM_JINRAI );
+	Assert( jinrai );
+	jinrai->SetScore( atoi( args[1] ) );
```
sunzenshen, line 521: Maybe could use V_atoi from `\neo\src\public\tier1\strtools.h`

Rainyan (approved):

src/game/shared/neo/neo_gamerules.cpp
sunzenshen: As discussed offline, maybe we could consider a different PR that maps atoi to whatever implementation we want.

PR #1639, opened 2026-01-31
Add cvars for controlling XP rewards.

### LINKED ISSUE #1462: XP rewards should be able to be changed (only for comp servers?)
For the few previous tournaments we have used [Agiel's plugin](https://github.com/Agiel/nt-wincond) to change various XP related rewards, such as not getting survivor bonus when winning a round whilst staying alive or getting a set amount of XP for capturing the ghost instead of a whole rank up. It would be nice to able to do it natively. 

(If we arent changing any XP mechanics for the whole game at the moment, then itd maybe be simpler if you could only use the related commands when a server has sv_neo_comp enabled?)

Agiel:
## Description
Add cvars from WinCond plugin for controlling XP rewards.

*sv_neo_cap_reward* - How much XP to reward for capturing the ghost. 0 = Rank up. (Default: 0)
*sv_neo_cap_reward_dead* - Whether dead players should receive the ghost capture reward. (Default: 0)
*sv_neo_survivor_bonus* - Whether surviving players on the winning team should receive extra XP. (Default: 1)
*sv_neo_ghost_carrier_bonus* - Whether the ghost carrier on the winning team should receive extra XP. (Default: 1)

Some liberal refactoring was made to reduce code duplication. The bot takeover logic was a bit confusing but I think I got it right.

Tested with and without possessing a bot:

|                                                    | sv_neo_cap_reward_dead 0 | sv_neo_cap_reward_dead 1 |
|------------------------------------------|---------------------------------------------------|--------------------------------------------------|
| sv_neo_cap_reward 0 | ✅ | ✅ |
|sv_neo_cap_reward 3  | ✅ | ✅ |

## Toolchain
- CachyOS, gcc v15.2.1

## Linked Issues
- fixes #1462

### DIFF
--- a/src/game/shared/neo/neo_gamerules.cpp
+++ b/src/game/shared/neo/neo_gamerules.cpp
@@ -1619,13 +1619,13 @@ void CNEORules::AwardRankUp(CNEO_Player *pClient)
 	{
 		if (pClient->m_iXP.Get() < ranks[i])
 		{
-            pClient->AddPoints(ranks[i] - pClient->m_iXP, false, true);
+			pClient->AddPoints(ranks[i] - pClient->m_iXP, false, true);
 			return;
 		}
 	}
 
 	// If we're beyond max rank, just award +1 point.
-    pClient->AddPoints(1, false, true);
+	pClient->AddPoints(1, false, true);
 }
 
 // Return remaining time in seconds. Zero means there is no time limit.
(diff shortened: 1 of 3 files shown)


src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -131,6 +131,10 @@ static_assert(DEF_TEAMPLAYERTHRES <= ((MAX_PLAYERS - 1) / 2));
 ConVar sv_neo_readyup_teamplayersthres("sv_neo_readyup_teamplayersthres", V_STRINGIFY(DEF_TEAMPLAYERTHRES), FCVAR_REPLICATED, "The exact total players per team to be in and ready up to start a game.", true, 0.0f, true, (MAX_PLAYERS - 1) / 2);
 ConVar sv_neo_readyup_skipwarmup("sv_neo_readyup_skipwarmup", "1", FCVAR_REPLICATED, "Skip the warmup round when already using ready up.", true, 0.0f, true, 1.0f);
 ConVar sv_neo_readyup_autointermission("sv_neo_readyup_autointermission", "0", FCVAR_REPLICATED, "If disabled, skips the automatic intermission at the end of the match.", true, 0.0f, true, 1.0f);
+ConVar sv_neo_cap_reward("sv_neo_cap_reward", "0", FCVAR_REPLICATED, "How much XP to reward for capturing the ghost or escaping as VIP. 0 = Rank up.", true, 0.0f, false, 0.0f);
```
sunzenshen, line 134: My understanding was that the last argument is the maximum possible value for the float/numerical input of the ConVar, so setting a range of 0.0-0.0 prevents the server operator from being able to set a value different from 0.  It seems like we should specify a reasonable maximum score reward.

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -1660,13 +1619,13 @@ void CNEORules::AwardRankUp(CNEO_Player *pClient)
 	{
 		if (pClient->m_iXP.Get() < ranks[i])
 		{
-			pClient->AddPoints(ranks[i] - pClient->m_iXP, false);
+            pClient->AddPoints(ranks[i] - pClient->m_iXP, false, true);
```
sunzenshen: If a human player captures the ghost while they are taking over the bot, would setting true for bIgnorePlayerTakeover award the rank up to the human player?  My understanding from earlier discussions was that the bot that was taken over should get the rank up, so that human players cannot exploit the bot takeovers to accelerate their rank up progress.

sunzenshen: Oh, doh, I saw the note below for the actual CTG scoring, and I didn't read above that this is awarding points for deathmatch, not a team mode.

sunzenshen: Have you been able to take over another player while in DM or TDM?  In my experience either of those modes forces a spawn, so there's no opportunity to take over a bot.  Maybe we could leave this function alone, since it's related to DM mode?

src/game/shared/neo/neo_gamerules.cpp
Agiel: As far as I understand, the second to last argument tells whether the max value should apply or not. Here I've set it to false.

src/game/shared/neo/neo_gamerules.cpp
Agiel: Did you misplace this comment? AwardRankUp is only used in CTG and VIP.

sunzenshen: I think I clicked the show more buttons too much and ended up seeing a function declaration related to DM. 😅

Issue #548, opened 2024-08-06
Firing guns should not draw tracers from the players own viewmodel


blaberry:
### Build Info

20240629_96f57be

### Description

Currently, if you fire a weapon there are client side tracers drawn simulating bullets flying out of your gun

### To Reproduce

1. Shoot gun
2. See tracer

### Expected behavior

No tracers for first person shots

### Actual behavior

There are tracers

### Operating System

- [X] Windows
- [X] Linux

### Version/Distro

Win 10

AdamTadeusz:
Should the tracers be visible coming out of your own gun in thirdperson? And what about when using cl_first_person_uses_world_model?

blaberry:
I think limiting it do when drawing the viewmodels is fine. That's parity behaviour afaik. Should probably be tested though.

PR #1640, opened 2026-02-01
Weapon Tracers

DESTROYGIRL:
## Description
Weapon tracers for MX, PZ and BALC3

Includes fixed `r_drawtracers_firstperson` cvar for clients

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/client/fx_tracer.cpp
+++ b/src/game/client/fx_tracer.cpp
@@ -100,15 +100,20 @@ void TracerCallback( const CEffectData &data )
 	if ( !r_drawtracers.GetBool() )
 		return;
 
+#ifdef NEO
+	C_BasePlayer *pPlayer = EffectPlayerSource( data );
+	const bool bPlayerFirstperson = ( pPlayer && !pPlayer->ShouldDrawThisPlayer() );
+#endif
+
 	if ( !r_drawtracers_firstperson.GetBool() )
 	{
 #ifdef NEO
-		C_BasePlayer *pPlayer = EffectPlayerSource( data );
+		if ( bPlayerFirstperson )
 #else
 		C_BasePlayer *pPlayer = dynamic_cast<C_BasePlayer*>( data.GetEntity() );
-#endif
 
 		if ( pPlayer && !pPlayer->ShouldDrawThisPlayer() )
+#endif
 			return;
 	}
 
@@ -116,9 +121,14 @@ void TracerCallback( const CEffectData &data )
 	Vector vecStart = GetTracerOrigin( data );
 	float flVelocity = data.m_flScale;
 	bool bWhiz = (data.m_fFlags & TRACER_FLAG_WHIZ);
+
+#ifdef NEO
+	if ( bPlayerFirstperson )
+#else
 	int iEntIndex = data.entindex();
 
 	if ( iEntIndex && iEntIndex == player->index )
+#endif
 	{
 		Vector	foo = data.m_vStart;
 		QAngle	vangles;


src/game/client/fx_tracer.cpp
```diff
@@ -18,6 +18,22 @@ ConVar r_drawtracers_firstperson( "r_drawtracers_firstperson", "1", FCVAR_ARCHIV
 
 #define	TRACER_SPEED			5000 
 
+#ifdef NEO
+// null if not sourced from a player or their weapon
+C_BasePlayer *EffectPlayerSource( const CEffectData &data )
+{
+	C_BaseCombatWeapon *pWep = dynamic_cast<C_BaseCombatWeapon*>( data.GetEntity() );
```
Rainyan: Maybe we could do something like:

```cpp
pEnt = data.GetEntity();
pWep = ( pEnt && pEnt->IsBaseCombatWeapon() ) ? assert_cast<C_BaseCombatWeapon*>( pEnt ) : nullptr;
```
or similar, to avoid the more expensive dynamic cast?

Rainyan:
Related: #548

src/game/client/fx_tracer.cpp
```diff
@@ -18,6 +18,23 @@ ConVar r_drawtracers_firstperson( "r_drawtracers_firstperson", "1", FCVAR_ARCHIV
...
+{
+	C_BaseEntity *pEnt = data.GetEntity();
+	C_BaseCombatWeapon *pWep = ( pEnt && pEnt->IsBaseCombatWeapon() ) ? assert_cast<C_BaseCombatWeapon*>( pEnt ) : nullptr;
+	if ( pWep )
+	{
+		return ToBasePlayer( pWep->GetOwner() );
+	}
+	else
+	{
+		return ToBasePlayer( data.GetEntity() );
```
Rainyan: Nit: could reuse `pEnt` here to avoid a double `data.GetEntity()` ehandle lookup

Rainyan (commented):
Visually comparing the MX and PZ tracers to OG NT, I'm thinking maybe OG is using the `FX_PlayerTracer` (rather than `FX_Tracer`) for its effects; for example try to do `if (true)` to unconditionally enter this block and compare visually: https://github.com/NeotokyoRebuild/neo/pull/1640/files#diff-b365f62ed579ea4d1f8f9ad32cb204cce6b77e9bd4915d1b5064bae86ed45103L100-L114

But maybe this also depends on graphics settings/resolution. I could see us wanting the more prominent tracers currently in this PR for the BALC since it's a big gun, but maybe worth experimenting with this other tracer function for the parity visuals?

DESTROYGIRL:
Thanks. Were you able to check if it was the same effect used for other players, not just the local player? People preferred `FX_PlayerTracer` in first person anyway so I was going to change it for that, but wasn't sure about third person

Rainyan:
Apologies about the delay. Here are some rough visual comparisons of what I'm seeing:

**OG NT**:

[ognt.webm](https://github.com/user-attachments/assets/6071286e-aee9-47c0-93fe-217f13f92ae4)

**This PR**:

[pr.webm](https://github.com/user-attachments/assets/4e83ede5-d391-4aec-9c92-15aaf7d258c0)

We can probably agree the visuals are quite distinctly different here for the shooter.

Here's also the alternative path of:

```diff
diff --git a/src/game/client/fx_tracer.cpp b/src/game/client/fx_tracer.cpp
index 893d8d9e..254774dd 100644
--- a/src/game/client/fx_tracer.cpp
+++ b/src/game/client/fx_tracer.cpp
@@ -118,7 +118,7 @@ void TracerCallback( const CEffectData &data )
        bool bWhiz = (data.m_fFlags & TRACER_FLAG_WHIZ);
        int iEntIndex = data.entindex();

-       if ( iEntIndex && iEntIndex == player->index )
+       if ( true || iEntIndex && iEntIndex == player->index )
        {
                Vector  foo = data.m_vStart;
                QAngle  vangles;
```
applied on top of this PR:

[pr-mods.webm](https://github.com/user-attachments/assets/f0c17d46-4336-485b-9026-d12849f8aa00)

It's not quite parity with OG NT, but I'd say it's closer.

I haven't gotten around to comparing 3rd person because I am not a useful person, but maybe I'll try and force myself to pretend I am one again in the near future.

PR #1651, opened 2026-02-07
Bots throw grenades from cover

sunzenshen:
## Description
Bots decide whether they can throw a grenade at a threat after retreating to cover.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.cpp
@@ -26,6 +26,12 @@ Action< CNEOBot > *CNEOBotGrenadeDispatch::ChooseGrenadeThrowBehavior( CNEOBot *
 		return nullptr;
 	}
 
+	// Prefer shooting if we have a clear line of fire
+	if ( me->IsLineOfFireClear( threat->GetEntity(), CNEOBot::LINE_OF_FIRE_FLAGS_DEFAULT ) )
+	{
+		return nullptr;
+	}
+
 	CNEO_Player *pNEOPlayer = ToNEOPlayer( me->GetEntity() );
 	if ( !pNEOPlayer )
 	{
@@ -75,31 +81,34 @@ Action< CNEOBot > *CNEOBotGrenadeDispatch::ChooseGrenadeThrowBehavior( CNEOBot *
 	// Should I toss a smoke grenade?
 	if ( pSmokeGrenade )
 	{
-		if ( pNEOPlayer->GetClass() == NEO_CLASS_SUPPORT )
-		{
-			// I can see through smoke
-			return new CNEOBotGrenadeThrowSmoke( pSmokeGrenade, threat );
-		}
-
-		bool bEnemySupportInField = false;
 		for ( int i = 1; i <= gpGlobals->maxClients; i++ )
 		{
 			CNEO_Player *pPlayer = ToNEOPlayer( UTIL_PlayerByIndex( i ) );
-			if ( pPlayer && pPlayer->IsAlive() && pPlayer->GetClass() == NEO_CLASS_SUPPORT )
+			if ( !pPlayer || !pPlayer->IsAlive() || pPlayer == pNEOPlayer )
 			{
-			    if ( !pPlayer->InSameTeam(pNEOPlayer) )
+				continue;
+			}
+
+			if ( pPlayer->InSameTeam( pNEOPlayer ) ) 
+			{
+				if ( !pPlayer->IsBot() && pPlayer->GetClass() != NEO_CLASS_SUPPORT )
 				{
-					bEnemySupportInField = true; // Enemy support could see through smoke
-					break;
+					// Avoid blocking the vision of a friendly human
+					// (Bots benefit from concealment without the disorientation)
+					return nullptr;
+				}
+			}
+			else
+			{
+				if ( pPlayer->GetClass() == NEO_CLASS_SUPPORT )
+				{
+					// Avoid giving an enemy with thermal vision a free smoke screen
+					return nullptr;
 				}
 			}
 		}
 
-		// Enemy team does not have Support players in the field
-		if (!bEnemySupportInField)
-		{
-			return new CNEOBotGrenadeThrowSmoke( pSmokeGrenade, threat );
-		}
+		return new CNEOBotGrenadeThrowSmoke( pSmokeGrenade, threat );
 	}
 
 	// Should I toss a frag grenade? 
--- a/src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.cpp
@@ -32,19 +32,13 @@ Action< CNEOBot > *CNEOBotGrenadeDispatch::ChooseGrenadeThrowBehavior( const CNE
 		return nullptr;
 	}
 
-	CNEO_Player *pNEOPlayer = ToNEOPlayer( me->GetEntity() );
-	if ( !pNEOPlayer )
-	{
-		return nullptr;
-	}
-
 	CWeaponGrenade *pFragGrenade = nullptr;
 	CWeaponSmokeGrenade *pSmokeGrenade = nullptr;
 
-	int iWeaponCount = pNEOPlayer->WeaponCount();
+	int iWeaponCount = me->WeaponCount();
 	for ( int i=0; i<iWeaponCount; ++i )
 	{
-		CBaseCombatWeapon *pWep = pNEOPlayer->GetWeapon( i );
+		CBaseCombatWeapon *pWep = me->GetWeapon( i );
 		if ( !pWep )
 		{
 			continue;
@@ -87,12 +81,12 @@ Action< CNEOBot > *CNEOBotGrenadeDispatch::ChooseGrenadeThrowBehavior( const CNE
 		for ( int i = 1; i <= gpGlobals->maxClients; i++ )
 		{
 			CNEO_Player *pPlayer = ToNEOPlayer( UTIL_PlayerByIndex( i ) );
-			if ( !pPlayer || !pPlayer->IsAlive() || pPlayer == pNEOPlayer )
+			if ( !pPlayer || !pPlayer->IsAlive() || pPlayer == me )
 			{
 				continue;
 			}
 
-			if ( pPlayer->InSameTeam( pNEOPlayer ) ) 
+			if ( pPlayer->InSameTeam( me ) ) 
 			{
 				if ( !pPlayer->IsBot() && pPlayer->GetClass() != NEO_CLASS_SUPPORT )
 				{
--- a/src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.cpp
@@ -14,7 +14,7 @@ ConVar sv_neo_bot_grenade_use_smoke("sv_neo_bot_grenade_use_smoke", "1", FCVAR_N
 ConVar sv_neo_bot_grenade_throw_cooldown("sv_neo_bot_grenade_throw_cooldown", "10", FCVAR_NONE, "Cooldown in seconds between grenade throws for bots");
 
 //---------------------------------------------------------------------------------------------
-Action< CNEOBot > *CNEOBotGrenadeDispatch::ChooseGrenadeThrowBehavior( CNEOBot *me, const CKnownEntity *threat )
+Action< CNEOBot > *CNEOBotGrenadeDispatch::ChooseGrenadeThrowBehavior( const CNEOBot *me, const CKnownEntity *threat )
 {
 	if (!sv_neo_bot_grenade_use_frag.GetBool() && !sv_neo_bot_grenade_use_smoke.GetBool())
 	{
@@ -51,24 +51,22 @@ Action< CNEOBot > *CNEOBotGrenadeDispatch::ChooseGrenadeThrowBehavior( CNEOBot *
 		}
 
 		CNEOBaseCombatWeapon *pNeoWep = static_cast< CNEOBaseCombatWeapon * >( pWep );
-		if ( pNeoWep ) 
+
+		auto bits = pNeoWep->GetNeoWepBits();
+		if ( sv_neo_bot_grenade_use_frag.GetBool() && (bits & NEO_WEP_FRAG_GRENADE) )
 		{
-			auto bits = pNeoWep->GetNeoWepBits();
-			if ( sv_neo_bot_grenade_use_frag.GetBool() && (bits & NEO_WEP_FRAG_GRENADE) )
+			pFragGrenade = static_cast< CWeaponGrenade * >( pNeoWep );
+			if ( pSmokeGrenade )
 			{
-				pFragGrenade = static_cast< CWeaponGrenade * >( pNeoWep );
-				if ( pSmokeGrenade )
-				{
-					break; // found both
-				}
+				break; // found both
 			}
-			else if ( sv_neo_bot_grenade_use_smoke.GetBool() && (bits & NEO_WEP_SMOKE_GRENADE) )
+		}
+		else if ( sv_neo_bot_grenade_use_smoke.GetBool() && (bits & NEO_WEP_SMOKE_GRENADE) )
+		{
+			pSmokeGrenade = static_cast< CWeaponSmokeGrenade * >( pNeoWep );
+			if ( pFragGrenade )
 			{
-				pSmokeGrenade = static_cast< CWeaponSmokeGrenade * >( pNeoWep );
-				if ( pFragGrenade )
-				{
-					break; // found both
-				}
+				break; // found both
 			}
 		}
 	}
--- a/src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.cpp
@@ -19,6 +19,11 @@ ConVar sv_neo_bot_grenade_use_smoke("sv_neo_bot_grenade_use_smoke", "1", FCVAR_N
 //---------------------------------------------------------------------------------------------
 Action< CNEOBot > *CNEOBotGrenadeDispatch::ChooseGrenadeThrowBehavior( CNEOBot *me, const CKnownEntity *threat )
 {
+	if (!sv_neo_bot_grenade_use_frag.GetBool() && !sv_neo_bot_grenade_use_smoke.GetBool())
+	{
+		return nullptr;
+	}
+
 	if ( !threat || !threat->GetEntity() || !threat->GetEntity()->IsPlayer() || !threat->GetEntity()->IsAlive() )
 	{
 		return nullptr;
@@ -26,11 +31,6 @@ Action< CNEOBot > *CNEOBotGrenadeDispatch::ChooseGrenadeThrowBehavior( CNEOBot *
 
 	CNEO_Player *pNEOPlayer = ToNEOPlayer( me->GetEntity() );
 	if ( !pNEOPlayer )
-	{
-		return nullptr;
-	}
-
-	if (!sv_neo_bot_grenade_use_frag.GetBool() && !sv_neo_bot_grenade_use_smoke.GetBool())
 	{
 		return nullptr;
 	}
@@ -69,11 +69,36 @@ Action< CNEOBot > *CNEOBotGrenadeDispatch::ChooseGrenadeThrowBehavior( CNEOBot *
 
 	Vector vecThreatPos = threat->GetLastKnownPosition();
 
+	// Should I toss a smoke grenade?
+	if ( pSmokeGrenade )
+	{
+		if ( pNEOPlayer->GetClass() == NEO_CLASS_SUPPORT )
+		{
+			// I can see through smoke
+			return new CNEOBotGrenadeThrowSmoke( pSmokeGrenade, threat );
+		}
+
+		for ( int i = 1; i <= gpGlobals->maxClients; i++ )
+		{
+			CNEO_Player *pPlayer = ToNEOPlayer( UTIL_PlayerByIndex( i ) );
+			if ( pPlayer && pPlayer->IsAlive() && pPlayer->GetClass() == NEO_CLASS_SUPPORT )
+			{
+			    if ( !pPlayer->InSameTeam(pNEOPlayer) )
+				{
+					return nullptr; // Enemy support could see through smoke
+				}
+			}
+		}
+
+		// Enemy team does not have Support players in the field
+		return new CNEOBotGrenadeThrowSmoke( pSmokeGrenade, threat );
+	}
+
 	// Should I toss a frag grenade? 
 	if ( pFragGrenade )
 	{
 		float flDistToThreatSqr = me->GetAbsOrigin().DistToSqr( vecThreatPos );
-		float flSafeRadius = sv_neo_grenade_blast_radius.GetFloat();
+		float flSafeRadius = CNEOBotGrenadeThrowFrag::GetFragSafetyRadius();
 		float flSafeRadiusSqr = flSafeRadius * flSafeRadius;
 
 		if ( flDistToThreatSqr < flSafeRadiusSqr )
@@ -110,17 +135,6 @@ Action< CNEOBot > *CNEOBotGrenadeDispatch::ChooseGrenadeThrowBehavior( CNEOBot *
 		}
 
 		return new CNEOBotGrenadeThrowFrag( pFragGrenade, threat );
-	}
-
-	// Should I toss a smoke grenade?
-	if ( pSmokeGrenade )
-	{
-		CNEO_Player *pThreatPlayer = ToNEOPlayer( threat->GetEntity() );
-		// Only smoke if threat is NOT support (supports have thermal vision)
-		if ( pThreatPlayer && pThreatPlayer->GetClass() != NEO_CLASS_SUPPORT )
-		{
-			return new CNEOBotGrenadeThrowSmoke( pSmokeGrenade, threat );
-		}
 	}
 
 	return nullptr;
--- a/src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.cpp
@@ -1,0 +1,119 @@
+#include "cbase.h"
+#include "neo_gamerules.h"
+#include "neo_player.h"
+#include "bot/neo_bot.h"
+#include "bot/behavior/neo_bot_grenade_dispatch.h"
+#include "bot/behavior/neo_bot_grenade_throw_frag.h"
+#include "bot/behavior/neo_bot_grenade_throw_smoke.h"
+#include "weapon_neobasecombatweapon.h"
+#include "weapon_grenade.h"
+#include "weapon_smokegrenade.h"
+
+extern ConVar sv_neo_grenade_blast_radius;
+extern ConVar sv_neo_grenade_fuse_timer;
+extern ConVar sv_neo_grenade_throw_intensity;
+
+//---------------------------------------------------------------------------------------------
+Action< CNEOBot > *CNEOBotGrenadeDispatch::ChooseGrenadeThrowBehavior( CNEOBot *me, const CKnownEntity *threat )
+{
+	if ( !threat || !threat->GetEntity() || !threat->GetEntity()->IsPlayer() || !threat->GetEntity()->IsAlive() )
+	{
+		return nullptr;
+	}
+
+	CNEO_Player *pNEOPlayer = ToNEOPlayer( me->GetEntity() );
+	if ( !pNEOPlayer )
+	{
+		return nullptr;
+	}
+
+	CWeaponGrenade *pFragGrenade = nullptr;
+	CWeaponSmokeGrenade *pSmokeGrenade = nullptr;
+
+	for ( int i=0; i<pNEOPlayer->WeaponCount(); ++i )
+	{
+		CBaseCombatWeapon *pWep = pNEOPlayer->GetWeapon( i );
+		if ( !pWep )
+		{
+			continue;
+		}
+
+		if ( ( pFragGrenade = dynamic_cast< CWeaponGrenade * >( pWep ) ) )
+		{
+			if ( pSmokeGrenade )
+			{
+				break; // found both
+			}
+		}
+		else if ( ( pSmokeGrenade = dynamic_cast< CWeaponSmokeGrenade * >( pWep ) ) )
+		{
+			if ( pFragGrenade )
+			{
+				break; // found both
+			}
+		}
+	}
+
+	if ( !pFragGrenade && !pSmokeGrenade )
+	{
+		return nullptr;
+	}
+
+	Vector vecThreatPos = threat->GetLastKnownPosition();
+
+	// Should I toss a frag grenade? 
+	if ( pFragGrenade )
+	{
+		float flDistToThreatSqr = me->GetAbsOrigin().DistToSqr( vecThreatPos );
+		float flSafeRadius = sv_neo_grenade_blast_radius.GetFloat();
+		float flSafeRadiusSqr = flSafeRadius * flSafeRadius;
+
+		if ( flDistToThreatSqr < flSafeRadiusSqr )
+		{
+			return nullptr;
+		}
+
+		float flMaxThrowDist = sv_neo_grenade_throw_intensity.GetFloat() * sv_neo_grenade_fuse_timer.GetFloat();
+		if ( flDistToThreatSqr > ( flMaxThrowDist * flMaxThrowDist ) )
+		{
+			return nullptr;
+		}
+
+		if ( NEORules()->IsTeamplay() )
+		{
+			bool bTeammateTooClose = false;
+			for ( int i = 1; i <= gpGlobals->maxClients; i++ )
+			{
+				CNEO_Player *pPlayer = ToNEOPlayer( UTIL_PlayerByIndex( i ) );
+				if ( pPlayer && pPlayer->IsAlive() && pPlayer != pNEOPlayer && pPlayer->InSameTeam(me) )
+				{
+					if ( pPlayer->GetAbsOrigin().DistToSqr( vecThreatPos ) < flSafeRadiusSqr )
+					{
+						bTeammateTooClose = true;
+						break;
+					}
+				}
+			}
+
+			if ( bTeammateTooClose )
+			{
+				return nullptr;
+			}
+		}
+
+		return new CNEOBotGrenadeThrowFrag( pFragGrenade, threat );
+	}
+
+	// Should I toss a smoke grenade?
+	if ( pSmokeGrenade )
+	{
+		CNEO_Player *pThreatPlayer = ToNEOPlayer( threat->GetEntity() );
+		// Only smoke if threat is NOT support (supports have thermal vision)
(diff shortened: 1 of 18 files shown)


sunzenshen:
Demo with sv_neo_grenade_debug_behavior enabled:
- https://streamable.com/la2yok
- https://streamable.com/e3o1y1
- https://streamable.com/2t4ofz

src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.cpp
```diff
@@ -0,0 +1,117 @@
...
+	if ( pSmokeGrenade )
+	{
+		if ( pNEOPlayer->GetClass() == NEO_CLASS_SUPPORT )
+		{
+			// I can see through smoke
+			return new CNEOBotGrenadeThrowSmoke( pSmokeGrenade, threat );
+		}
+
+		bool bEnemySupportInField = false;
+		for ( int i = 1; i <= gpGlobals->maxClients; i++ )
```
sunzenshen, line 81: To start the conversation, I wonder if we should check if there are any human non-support players on the bot's team before throwing smoke. Bots don't have too much trouble navigating smoked areas, but I imagine humans could struggle to navigate on maps with fewer routes like bullet.

Rainyan: One fuzzy measure could be calculating a numeric value for how "useful" the smoke would be, based on number of nearby supports, and whether they are friendly/enemy or bot/human, and try to make it such that smokes that may hinder human friendlies are to be avoided, even if they were marginally useful for some bots. And likewise avoid smokes that give advantage to human enemies based on some metric like that.

sunzenshen: After watching a friend try the game the other evening, I decided that for the scope of this review, I'd like to make the use of smokes somewhat conservative, avoiding disoriented humans and enemies that could exploit the smoke screen. (As one may guess, the very first thing they did was walk into a smoke cloud and then immediately became lost for the duration. 🥲 )

Then another PR could explore a more sophisticated handling of smokes.  For example, maybe we could calculate a path to the smoke target, estimate the movement time needed, and figure out if a player could realistically reach and view the smoke based on their class movement.

Rainyan: We've definitely entered the bikeshed, but: it might be interesting to try a kind of [VR Chaperone](https://help.steampowered.com/en/faqs/view/30FC-2296-D4CD-58DA) style effect inside smokes, to represent the in-smoke fumbling player having tactile feedback of physically hitting/brushing against the wall (it's basically solving for the same thing IRL):

chaperone

I imagine this might be able to reuse the friendly HUD X-ray shader stuff, but I'm not too familiar with that code.

<br>

> Then another PR could explore a more sophisticated handling of smokes. For example, maybe we could calculate a path the the smoke target, estimate the movement time needed, and figure out if a player could realistically reach and view the smoke based on their class movement.

Definitely, I think we should implement some kind of code to calculate the full flight path of a grenade ahead of time, and that could be used for stuff like these bot throws, but also for training modes to practice nade throws etc like in CS (the smaller PiP screens show where that nade is going to land):

cs2_grenprev

src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.cpp
```diff
@@ -0,0 +1,126 @@
...
+	for ( int i=0; i<iWeaponCount; ++i )
+	{
+		CBaseCombatWeapon *pWep = pNEOPlayer->GetWeapon( i );
+		if ( !pWep )
+		{
+			continue;
+		}
+
+		CNEOBaseCombatWeapon *pNeoWep = static_cast< CNEOBaseCombatWeapon * >( pWep );
+		if ( pNeoWep )
```
Rainyan: `static_cast` of a non-null pointer can never return nullptr so we can skip this if-check

src/game/server/neo/bot/behavior/neo_bot_grenade_throw_frag.cpp
```diff
@@ -0,0 +1,153 @@
...
+	{
+		return true;
+	}
+
+	const CNEO_Player *pMePlayer = ToNEOPlayer( me->GetEntity() );
+
+	for ( int i = 1; i <= gpGlobals->maxClients; i++ )
+	{
+		CNEO_Player *pPlayer = ToNEOPlayer( UTIL_PlayerByIndex( i ) );
+		if ( pPlayer && pPlayer->IsAlive() && pPlayer->InSameTeam( pMePlayer ) )
```
Rainyan: Nit: we could check `if (me == pPlayer) continue;` here to avoid doing the safe distance check for `me` again, since we already checked it at the beginning of this function.

src/game/server/neo/bot/behavior/neo_bot_grenade_throw_frag.cpp
```diff
@@ -0,0 +1,153 @@
...
+	{
+		return false; // Too far to throw
+	}
+
+	if ( !NEORules()->IsTeamplay() )
+	{
+		return true;
+	}
+
+	const CNEO_Player *pMePlayer = ToNEOPlayer( me->GetEntity() );
```
Rainyan: Nit: since we don't need to call any `CNEO_Player` methods here, this `pMePlayer` is redundant. I.e. you can simply replace all uses of `pMePlayer` with `me` (because they both are pointers to the same memory address, just represented as different pointer types).

Image

src/game/server/neo/bot/behavior/neo_bot_grenade_throw_frag.cpp
```diff
@@ -0,0 +1,153 @@
...
+	{
+		CNEO_Player *pPlayer = ToNEOPlayer( UTIL_PlayerByIndex( i ) );
+		if ( pPlayer && pPlayer->IsAlive() && pPlayer->InSameTeam( pMePlayer ) )
+		{
+			if ( pPlayer->GetAbsOrigin().DistToSqr( vecTarget ) < flSafeRadiusSqr )
+			{
+				trace_t tr;
+				UTIL_TraceLine( vecTarget, pPlayer->WorldSpaceCenter(), MASK_SHOT_HULL, nullptr, COLLISION_GROUP_NONE, &tr );
+
+				if ( tr.fraction == 1.0f || tr.m_pEnt == pPlayer )
```
Rainyan: Why do we test fraction == 1 here, doesn't that mean that the trace did not hit? Or could this trace fail in some way that might disagree with the grenade blast's damage trace?

src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.h
```diff
@@ -0,0 +1,20 @@
...
+
+class CNEOBot;
+class CKnownEntity;
+
+extern ConVar sv_neo_bot_grenade_throw_cooldown;
+
+class CNEOBotGrenadeDispatch
+{
+public:
+	static Action< CNEOBot > *ChooseGrenadeThrowBehavior( CNEOBot *me, const CKnownEntity *threat );
```
Rainyan: Nit: since `me` isn't mutated, we could mark it `const CNEOBot *` here (and in the `IsFragSafe` prototype also), to at least theoretically allow the compiler to optimize these better.

src/game/server/neo/bot/behavior/neo_bot_grenade_throw.cpp
```diff
@@ -0,0 +1,264 @@
...
+		m_hThreatGrenadeTarget = nullptr;
+		m_vecThreatLastKnownPos = vec3_invalid;
+	}
+}
+
+//---------------------------------------------------------------------------------------------
+// Used to anticipate the emergence point from cover ahead of a path
+// Is calculated in reverse of path to find the farthest point from bot
+// (assuming "familiar" position is closer to the bot than the "obscured" position)
+Vector CNEOBotGrenadeThrow::FindEmergencePointAlongPath( CNEOBot *me, const Vector &familiarPos, const Vector &obscuredPos )
```
Rainyan: Nit: `me` could be `const`

Rainyan: There's also quite a few `Vector` copies being made by this function whereas we could probably use `const Vector&` lifetime-wise, for example:

```diff
diff --git a/src/game/server/neo/bot/behavior/neo_bot_grenade_throw.cpp b/src/game/server/neo/bot/behavior/neo_bot_grenade_throw.cpp
index 377d682b..b3427544 100644
--- a/src/game/server/neo/bot/behavior/neo_bot_grenade_throw.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_grenade_throw.cpp
@@ -33,7 +33,7 @@ CNEOBotGrenadeThrow::CNEOBotGrenadeThrow( CNEOBaseCombatWeapon *pWeapon, const C
 // Used to anticipate the emergence point from cover ahead of a path
 // Is calculated in reverse of path to find the farthest point from bot
 // (assuming "familiar" position is closer to the bot than the "obscured" position)
-Vector CNEOBotGrenadeThrow::FindEmergencePointAlongPath( CNEOBot *me, const Vector &familiarPos, const Vector &obscuredPos )
+const Vector& CNEOBotGrenadeThrow::FindEmergencePointAlongPath( CNEOBot *me, const Vector &familiarPos, const Vector &obscuredPos )
 {
 	CNavArea *familiarArea = TheNavMesh->GetNavArea( familiarPos );
 	CNavArea *obscuredArea = TheNavMesh->GetNavArea( obscuredPos );
@@ -41,7 +41,7 @@ Vector CNEOBotGrenadeThrow::FindEmergencePointAlongPath( CNEOBot *me, const Vect
 	if ( familiarArea && obscuredArea )
 	{
 		ShortestPathCost cost;
-		Vector vecGoal = obscuredPos;
+		const Vector& vecGoal = obscuredPos;
 		if ( NavAreaBuildPath( familiarArea, obscuredArea, &vecGoal, cost ) )
 		{
 			// search backwards from obscured position to find the first point visible to me
@@ -61,7 +61,7 @@ Vector CNEOBotGrenadeThrow::FindEmergencePointAlongPath( CNEOBot *me, const Vect
 					}
 				}
 
-				Vector vecTest = area->GetCenter();
+				const Vector& vecTest = area->GetCenter();
 
 				if ( me->IsLineOfFireClear( vecTest, CNEOBot::LINE_OF_FIRE_FLAGS_SHOTGUN ) )
 				{
diff --git a/src/game/server/neo/bot/behavior/neo_bot_grenade_throw.h b/src/game/server/neo/bot/behavior/neo_bot_grenade_throw.h
index 0ba625c8..66da6b71 100644
--- a/src/game/server/neo/bot/behavior/neo_bot_grenade_throw.h
+++ b/src/game/server/neo/bot/behavior/neo_bot_grenade_throw.h
@@ -36,7 +36,7 @@ protected:
 		THROW_TARGET_WAIT = 1,
 	};
 
-	static Vector FindEmergencePointAlongPath( CNEOBot *me, const Vector &familiarPos, const Vector &obscuredPos );
+	static const Vector& FindEmergencePointAlongPath( CNEOBot *me, const Vector &familiarPos, const Vector &obscuredPos );
 	
 	virtual ThrowTargetResult UpdateGrenadeTargeting( CNEOBot *me, CNEOBaseCombatWeapon *pWeapon ) = 0;
 
diff --git a/src/game/server/neo/bot/behavior/neo_bot_grenade_throw_frag.cpp b/src/game/server/neo/bot/behavior/neo_bot_grenade_throw_frag.cpp
index bbc574c6..48380ac0 100644
--- a/src/game/server/neo/bot/behavior/neo_bot_grenade_throw_frag.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_grenade_throw_frag.cpp
@@ -66,7 +66,7 @@ CNEOBotGrenadeThrow::ThrowTargetResult CNEOBotGrenadeThrowFrag::UpdateGrenadeTar
 		// and throw at the furthest visible point along that path
 		if ( me->IsLineOfSightClear( m_vecThreatLastKnownPos ) && m_scanTimer.IsElapsed() )
 		{
-			Vector vecThrowTarget = FindEmergencePointAlongPath( me, m_vecThreatLastKnownPos, m_hThreatGrenadeTarget->GetAbsOrigin() );
+			const Vector& vecThrowTarget = FindEmergencePointAlongPath( me, m_vecThreatLastKnownPos, m_hThreatGrenadeTarget->GetAbsOrigin() );
 
 			if ( vecThrowTarget != vec3_invalid )
 			{
diff --git a/src/game/server/neo/bot/behavior/neo_bot_grenade_throw_smoke.cpp b/src/game/server/neo/bot/behavior/neo_bot_grenade_throw_smoke.cpp
index c09922c4..5252291b 100644
--- a/src/game/server/neo/bot/behavior/neo_bot_grenade_throw_smoke.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_grenade_throw_smoke.cpp
@@ -35,7 +35,7 @@ CNEOBotGrenadeThrow::ThrowTargetResult CNEOBotGrenadeThrowSmoke::UpdateGrenadeTa
 	
 	if (m_vecTarget == vec3_invalid)
 	{
-		Vector vecThrowTarget = FindEmergencePointAlongPath(me, me->GetAbsOrigin(), m_vecThreatLastKnownPos);
+		const Vector& vecThrowTarget = FindEmergencePointAlongPath(me, me->GetAbsOrigin(), m_vecThreatLastKnownPos);
 
 		if (vecThrowTarget != vec3_invalid)
 		{
```

src/game/server/neo/bot/behavior/neo_bot_grenade_throw.cpp
```diff
@@ -0,0 +1,264 @@
...
+	if ( m_giveUpTimer.IsElapsed() )
+	{
+		return Done( "Grenade throw timed out" );
+	}
+
+	if ( sv_neo_grenade_debug_behavior.GetBool() && m_hThreatGrenadeTarget.Get() )
+	{
+		// DEBUG Colors: Red = Frag, Gray = Smoke, Purple = Unknown
+		Vector vecStart = me->GetEntity()->WorldSpaceCenter();
+		Vector vecVictim = m_hThreatGrenadeTarget->GetAbsOrigin();
```
Rainyan: These could be `const Vector&`

src/game/server/neo/bot/behavior/neo_bot_grenade_throw_frag.cpp
```diff
@@ -0,0 +1,153 @@
...
+		}
+		
+		if ( m_vecTarget == vec3_invalid )
+		{
+			// continue investigating last known position
+			m_PathFollower.Update(me);
+			return THROW_TARGET_WAIT;
+		}
+	}
+	Assert( m_vecTarget != vec3_invalid );
```
Rainyan, line 91: I hit this assertion while testing. Call stack when hitting the assert:

```
>	server.dll!CNEOBotGrenadeThrowFrag::UpdateGrenadeTargeting(CNEOBot * me, CNEOBaseCombatWeapon * pWeapon) Line 88	C++
 	server.dll!CNEOBotGrenadeThrow::Update(CNEOBot * me, float interval) Line 187	C++
 	server.dll!Action<CNEOBot>::InvokeUpdate(CNEOBot * me, Behavior<CNEOBot> * behavior, float interval) Line 1387	C++
 	server.dll!Action<CNEOBot>::InvokeUpdate(CNEOBot * me, Behavior<CNEOBot> * behavior, float interval) Line 1379	C++
 	server.dll!Action<CNEOBot>::InvokeUpdate(CNEOBot * me, Behavior<CNEOBot> * behavior, float interval) Line 1379	C++
 	server.dll!Action<CNEOBot>::InvokeUpdate(CNEOBot * me, Behavior<CNEOBot> * behavior, float interval) Line 1379	C++
 	server.dll!Behavior<CNEOBot>::Update(CNEOBot * me, float interval) Line 258	C++
 	server.dll!CNEOBotIntention::Update() Line 2704	C++
 	server.dll!INextBot::Update() Line 133	C++
 	server.dll!NextBotPlayer<CNEO_Player>::Update() Line 976	C++
 	server.dll!CNEOBot::Update() Line 836	C++
 	server.dll!NextBotPlayer<CNEO_Player>::PhysicsSimulate() Line 743	C++
 	server.dll!CNEOBot::PhysicsSimulate() Line 692	C++
 	server.dll!Physics_SimulateEntity(CBaseEntity * pEntity) Line 2019	C++
 	server.dll!Physics_RunThinkFunctions(bool simulating) Line 2075	C++
 	server.dll!CServerGameDLL::GameFrame(bool simulating) Line 1281	C++
 	[External Code]	
```

What probably happened is we entered the first `if` but not the second, and therefore also skipped the `else if` (terrible visualisation below):
```cpp
 |
 |
 |
 |  if (pPrimaryThreat && pPrimaryThreat->GetEntity() && me->IsLineOfFireClear(pPrimaryThreat->GetEntity(), CNEOBot::LINE_OF_FIRE_FLAGS_DEFAULT))
 |  {
  ------
        |
  ------|
  |
  |        if ( m_scanTimer.IsElapsed() )
  |        {
  |            // ASSIGN TO m_vecTarget
  |        }
  |   }
  |   else if (m_vecTarget == vec3_invalid)
  |   {
  |        // ASSIGN TO m_vecTarget OR RETURN
  |   }
  |
   ----
      |
      v

      Assert(m_vecTarget != vec3_invalid);
```
so we never assigned a non-invalid value, and neither did we early-return. Perhaps changing the `else if` to a plain `if` would help with catching the `vec3_invalid` case? Not sure what the best fix is, though.

src/game/server/neo/bot/behavior/neo_bot_grenade_throw.cpp
```diff
@@ -0,0 +1,264 @@
...
+		if ( vecForward.Dot( vecToTarget ) >= 0.95f )
+		{
+			bAimOnTarget = true;
+		}
+	}
+	
+	// DEBUG: Targeting decisions
+	if ( sv_neo_grenade_debug_behavior.GetBool() )
+	{
+		Vector vecStart = me->GetEntity()->WorldSpaceCenter();
```
Rainyan: could be `const Vector& vecStart`

src/game/server/neo/bot/behavior/neo_bot_grenade_throw.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+	// NEO JANK: Force the throwing phase if aimed.
+	// We ignore bWeaponReady/m_flNextPrimaryAttack because we are manually controlling the throw
+	// and want to bypass the weapon's internal scheduled throw to ensure it happens NOW.
+	if ( bAimOnTarget )
+	{
+		// Play first person throw animation
+		pWep->SendWeaponAnim( ACT_VM_THROW );
+
+		CNEO_Player* pPlayer = ToNEOPlayer(me->GetEntity());
```
Rainyan: Nit: You could replace uses of `pPlayer` with `me` and avoid this `ToNEOPlayer` line, since `CNEOBot` already is a child class of `CNEO_Player` and can be used interchangeably:

```cpp
// because:
class CNEOBot : public NextBotPlayer< CNEO_Player >;

// and:
template < typename PlayerType > // <-- PlayerType templated as CNEO_Player
class NextBotPlayer : public PlayerType;

// therefore:
static_assert(std::derived_from<CNEOBot, CNEO_Player> == true);
```

src/game/server/neo/bot/behavior/neo_bot_grenade_throw.cpp
```diff
@@ -0,0 +1,264 @@
...
+		{
+			static_cast<CWeaponGrenade*>(pWep)->ThrowGrenade( pPlayer );
+		}
+		else if ( pWep->GetNeoWepBits() & NEO_WEP_SMOKE_GRENADE )
+		{
+			static_cast<CWeaponSmokeGrenade*>(pWep)->ThrowGrenade( pPlayer );
+		}
+		else
+		{
+			DevMsg( "CNEOBotGrenadeThrow: Unknown grenade type!\n" );
```
Rainyan: Nit: would `DevWarning` be more fitting than `DevMsg` here, so we get the angry red print color to make it stand out more in the console printout?

src/game/server/neo/bot/behavior/neo_bot_grenade_throw_frag.h
```diff
@@ -0,0 +1,28 @@
...
+
+#include "neo_bot_grenade_throw.h"
+
+class CNEOBotGrenadeThrowFrag : public CNEOBotGrenadeThrow
+{
+public:
+	CNEOBotGrenadeThrowFrag( CNEOBaseCombatWeapon *pWeapon, const CKnownEntity *threat );
+	virtual ~CNEOBotGrenadeThrowFrag() override { }
+
+	virtual const char *GetName( void ) const override { return "GrenadeThrowFrag"; }
```
Rainyan: Nit: `GetName` could be specified `final` so the compiler knows it can inline this when given a compatible pointer without needing to check vtable (because there cannot possibly be any higher-priority override for a method that is `final`)

src/game/server/neo/bot/behavior/neo_bot_grenade_throw_smoke.cpp
```diff
@@ -0,0 +1,54 @@
...
+	return Continue();
+}
+
+CNEOBotGrenadeThrow::ThrowTargetResult CNEOBotGrenadeThrowSmoke::UpdateGrenadeTargeting( CNEOBot *me, CNEOBaseCombatWeapon *pWeapon )
+{
+	Assert( m_vecThreatLastKnownPos != vec3_invalid );
+	
+	if (m_vecTarget == vec3_invalid)
+	{
+		Vector vecThrowTarget = FindEmergencePointAlongPath(me, me->GetAbsOrigin(), m_vecThreatLastKnownPos);
```
Rainyan: Assuming [we refactored the function being called here](https://github.com/NeotokyoRebuild/neo/pull/1651/changes#r2829866024) to return a const reference, this could be `const Vector& vecThrowTarget` to avoid the extra copy.

src/game/server/neo/bot/behavior/neo_bot_grenade_throw_smoke.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include "neo_bot_grenade_throw.h"
+
+class CNEOBotGrenadeThrowSmoke : public CNEOBotGrenadeThrow
+{
+public:
+	CNEOBotGrenadeThrowSmoke( CNEOBaseCombatWeapon *pWeapon, const CKnownEntity *threat );
+	virtual ~CNEOBotGrenadeThrowSmoke() override { }
+
+	virtual const char *GetName( void ) const override { return "GrenadeThrowSmoke"; }
```
Rainyan: Nit: `const override final` might be preferable here for `GetName` as inline hint for the compiler

src/game/server/neo/bot/behavior/neo_bot_retreat_to_cover.cpp
```diff
@@ -53,7 +55,33 @@ class CTestAreaAgainstThreats :  public IVision::IForEachKnownEntity
...
 				if ( m_area->IsPotentiallyVisible( threatArea ) )
-					++m_exposedThreatCount;
+				{
+					// Is there smoke in this area that I can use for concealment?
+					bool bObscuredBySmoke = false;
+					CNEO_Player *pThreatPlayer = ToNEOPlayer( known.GetEntity() );
+					// Support class can see through smoke
+					if ( pThreatPlayer && (pThreatPlayer->GetClass() != NEO_CLASS_SUPPORT) )
+					{
+						ScopedSmokeLOS smokeScope( false );
```
Rainyan, line 65: What does this `smokeScope` do?

Rainyan (commented):
Looks good for the most part, there was one assertion failure we could fix and I also left some notes/nits.

src/game/server/neo/bot/behavior/neo_bot_grenade_throw_frag.cpp
sunzenshen: My understanding is that if the fraction is 1.0, the traceline was able to at least reach the coordinate that we pointed to without any obstructions in the way, which would be the player's WorldSpaceCenter in this case.  The reason I consider it a valid target in this case is that even if the traceline didn't hit the player, the traceline had no obstacles between the start of the traceline and the designated end point of the trace, which means that at least the player's recent position was exposed to the blast radius.  Part of my caution is that I'd rather not risk a grenade throw if either criteria is true, to avoid player frustration with a potential teamkill.

src/game/server/neo/bot/behavior/neo_bot_retreat_to_cover.cpp
sunzenshen: This is kind of jank, but ScopedSmokeLOS is an RAII wrapper that when given false (is not support, or rather does not have thermal vision), makes it so that smoke entities from the smoke grenade block tracelines that use the MASK_BLOCKLOS filter.  The variable this RAII line toggles is m_bNextEntitySeesThroughSmoke, which when false means to block line of sight with smoke entities.  It takes advantage of the fact that the bot code is single threaded, hence that approach.

src/game/server/neo/bot/behavior/neo_bot_grenade_throw_frag.cpp
sunzenshen: I noticed occasionally that bots stared at each other if they were in line of sight, so I think the culprit is not assigning the target when there was line of sight but the throttle timer did not elapse.  I think it might separate the cases better if we check if an initial target was set in the case of the direct line of sight case.

sunzenshen:
Added some tweaks on top of the code review feedback, where the tweaks can be seen by themselves with:
- https://github.com/NeotokyoRebuild/neo/pull/1651/changes/a2c5ce3cf3f8985471001ef732077b10bbb269f3

src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
```diff
@@ -479,24 +479,24 @@ bool CNEOBaseCombatWeapon::Deploy(void)
 			{
 				pOwner->SetMaxSpeed(pOwner->GetNormSpeed_WithWepEncumberment(this));
 			}
-		}
 
-		if (pOwner->m_nButtons & IN_ZOOM && IsAllowedToZoom(this))
```
sunzenshen, line 492: Hit an invalid pOwner exception here, so moved the block under the existing `if (pOwner)` above

src/game/server/neo/bot/behavior/neo_bot_grenade_throw_frag.cpp
sunzenshen: Oh, I see what you might be getting at here.  The traceline could be interrupted by any other MASK_SHOT_HULL entity, like an enemy, so we probably should confirm that we are actually concerned about the entity that was actually hit.

src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.cpp
```diff
@@ -0,0 +1,129 @@
...
+	if ( me->IsLineOfFireClear( threat->GetEntity(), CNEOBot::LINE_OF_FIRE_FLAGS_DEFAULT ) )
+	{
+		return nullptr;
+	}
+
+	CNEO_Player *pNEOPlayer = ToNEOPlayer( me->GetEntity() );
+	if ( !pNEOPlayer )
+	{
+		return nullptr;
+	}
```
Rainyan: Nit: there isn't really a need to cast this here (I think?), could simplify to:

```diff
--- a/src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.cpp
@@ -32,19 +32,13 @@ Action< CNEOBot > *CNEOBotGrenadeDispatch::ChooseGrenadeThrowBehavior( const CNE
                return nullptr;
        }

-       CNEO_Player *pNEOPlayer = ToNEOPlayer( me->GetEntity() );
-       if ( !pNEOPlayer )
-       {
-               return nullptr;
-       }
-
        CWeaponGrenade *pFragGrenade = nullptr;
        CWeaponSmokeGrenade *pSmokeGrenade = nullptr;

-       int iWeaponCount = pNEOPlayer->WeaponCount();
+       int iWeaponCount = me->WeaponCount();
        for ( int i=0; i<iWeaponCount; ++i )
        {
-               CBaseCombatWeapon *pWep = pNEOPlayer->GetWeapon( i );
+               CBaseCombatWeapon *pWep = me->GetWeapon( i );
                if ( !pWep )
                {
                        continue;
@@ -87,12 +81,12 @@ Action< CNEOBot > *CNEOBotGrenadeDispatch::ChooseGrenadeThrowBehavior( const CNE
                for ( int i = 1; i <= gpGlobals->maxClients; i++ )
                {
                        CNEO_Player *pPlayer = ToNEOPlayer( UTIL_PlayerByIndex( i ) );
-                       if ( !pPlayer || !pPlayer->IsAlive() || pPlayer == pNEOPlayer )
+                       if ( !pPlayer || !pPlayer->IsAlive() || pPlayer == me )
                        {
                                continue;
                        }

-                       if ( pPlayer->InSameTeam( pNEOPlayer ) )
+                       if ( pPlayer->InSameTeam( me ) )
                        {
                                if ( !pPlayer->IsBot() && pPlayer->GetClass() != NEO_CLASS_SUPPORT )
                                {
```

src/game/server/neo/bot/behavior/neo_bot_grenade_throw.cpp
```diff
@@ -0,0 +1,364 @@
...
+//---------------------------------------------------------------------------------------------
+// Used to anticipate the emergence point from cover ahead of a path
+// Is calculated in reverse of path to find the farthest point from bot
+// (assuming "familiar" position is closer to the bot than the "obscured" position)
+const Vector& CNEOBotGrenadeThrow::FindEmergencePointAlongPath( const CNEOBot *me, const Vector &familiarPos, const Vector &obscuredPos )
+{
+	CNavArea *familiarArea = TheNavMesh->GetNavArea( familiarPos );
+	CNavArea *obscuredArea = TheNavMesh->GetNavArea( obscuredPos );
+
+	if ( familiarArea && obscuredArea )
```
Rainyan: I didn't measure these, but since `GetNavArea` is marked with the `VPROF` macro, it might be expected to be somewhat expensive. So perhaps it would be good to early-return here with `if(!familiarArea) return vec3_invalid`, so we could save the second call to `GetNavArea` in such a case?

src/game/server/neo/bot/behavior/neo_bot_grenade_throw.cpp
```diff
@@ -0,0 +1,364 @@
...
+	{
+		return Done( "Targeted threat is dead" );
+	}
+
+	if ( m_giveUpTimer.IsElapsed() )
+	{
+		return Done( "Grenade throw timed out" );
+	}
+
+	CNEO_Player* pNeoMe = ToNEOPlayer(me);
```
Rainyan: Could probably simplify to:
```diff
-       CNEO_Player* pNeoMe = ToNEOPlayer(me);
-       if (pNeoMe && (gpGlobals->curtime - pNeoMe->GetLastDamageTime() < 0.5f))
+
+       if (me && (gpGlobals->curtime - me->GetLastDamageTime() < 0.5f))
        {
                return Done( "Actively getting hurt, too dangerous to throw grenade" );
        }
@@ -171,7 +171,7 @@ ActionResult< CNEOBot >     CNEOBotGrenadeThrow::Update( CNEOBot *me, float interval
        if ( sv_neo_bot_grenade_debug_behavior.GetBool() && m_hThreatGrenadeTarget.Get() )
        {
                // DEBUG Colors: Red = Frag, Gray = Smoke, Purple = Unknown
-               const Vector& vecStart = me->GetEntity()->WorldSpaceCenter();
+               const Vector& vecStart = me->WorldSpaceCenter();
                const Vector& vecVictim = m_hThreatGrenadeTarget->GetAbsOrigin();
                int r = 255, g = 0, b = 255; // Default purple just in case
                if ( pWep->GetNeoWepBits() & NEO_WEP_FRAG_GRENADE )
@@ -183,7 +183,7 @@ ActionResult< CNEOBot >     CNEOBotGrenadeThrow::Update( CNEOBot *me, float interval
                        r = 128; g = 128; b = 128; // Gray
                }
                // Draw box around the bot that is considering the throw
-               NDebugOverlay::Box( me->GetEntity()->GetAbsOrigin(), me->GetEntity()->WorldAlignMins(), me->GetEntity()->WorldAlignMaxs(), r, g, b, 30, 0.1f );
+               NDebugOverlay::Box( me->GetAbsOrigin(), me->WorldAlignMins(), me->WorldAlignMaxs(), r, g, b, 30, 0.1f );
                // Arrow from bot to threat being targeted
                NDebugOverlay::HorzArrow( vecStart, vecVictim, 2.0f, r, g, b, 255, true, 0.1f );
        }
```

Probably safe to also remove the `if (me)` check, since the base `Behavior::Update` does a check before calling `InvokeUpdate`:
```cpp
if ( me == NULL || IsEmpty() )
{
	return;
}
```

Rainyan (dismissed):

src/game/server/neo/bot/behavior/neo_bot_grenade_throw.cpp
sunzenshen: Decided that it could be a good idea to Assert(me) at the top in case this doesn't hold true anymore.

sunzenshen:
Really sorry for the additional code churn and I should have done this earlier, but I was previously hoping that the grenade retreat logic from tactical monitor would fire soon enough after a grenade was thrown, but after seeing another bot run after their grenade and die, don't think it's triggering fast enough.  (To recap, the tactical monitor check is throttled to emulate a human reaction time, but I think we need to immediately trigger a search for throwers since they are both immediately near their own grenade and they are often running in the same direction of their throw.)

So I refactored out the grenade search from tactical monitor and placed it directly in the grenade retreat behavior. I also adapted the grenade retreat behavior to allow triggering without a specific grenade threat, so that we can trigger a grenade retreat immediately after a throw.

src/game/server/neo/bot/behavior/neo_bot_retreat_from_grenade.cpp
```diff
@@ -10,8 +10,48 @@
...
+	const char *pszGrenadeClass = "neo_grenade_frag";
+	
+	int iSound = CSoundEnt::ActiveList();
+	while ( iSound != SOUNDLIST_EMPTY )
+	{
+		CSound *pSound = CSoundEnt::SoundPointerForIndex( iSound );
+		if ( !pSound )
+			break;
+
+		if ( (pSound->SoundType() & SOUND_DANGER) && pSound->ValidateOwner() )
```
Rainyan, line 33: Looking at `CSound::ValidateOwner` implementation, it seems that this check might result in skipping a dangerous nade where the owner tossed it near the bot and then disconnected or was kicked before it exploded.

There doesn't seem to be any API to access `CSound::m_bHasOwner` directly, but maybe if we declared this as its friend class, we could then instead check:

```cpp
if ( (pSound->SoundType() & SOUND_DANGER) && pSound->m_bHasOwner )
```

because the `pSound->m_hOwner.Get()` gets checked for nullptr at line 41 anyway? This way we could still do the fast discard for `!m_bHasOwner` sounds, but avoid the edge case.

Rainyan (approved):

src/game/server/neo/bot/behavior/neo_bot_retreat_from_grenade.cpp
Rainyan: Actually, it looks like this is handled for us by `CSoundEnt::Think`, so I think `ValidateOwner` is indeed the appropriate thing to check.

nullsystem (approved):

PR #1653, opened 2026-02-07
NeoUI - Refactor colors theme, general NeoUI QoL

### LINKED ISSUE #1124: Menus scale weirdly on deck
Text overlaps, logo has a gap, boarders to not always contain all of the text

Image

Image

Resolution: 1280x800
Fullscreen

### LINKED ISSUE #1155: NeoUI Theming refactoring
The current theming is fixed globally within `NeoUI`. It should instead be more flexible and applied per context, as well as greater control on coloring on each widgets and elements.

nullsystem:
## Description

Aim to make theming not fixed and able to be defined outside of NeoUI. Should be more flexible to theme and more usable as an eventual standalone library. Changed slider background bar rendering looks. Also changed theme style, now have borders for hot items.

* EXTRAS:
  * Fixed HUD tab F1-F3 text color on final tab
  * Added flags and size control to NeoUI::Tabs
  * Change IFF markers to use NeoUI::Tabs
  * Fix password and blacklist not auto-selecting text box which causes anonyance for players, the text box for those are now force-active, the other buttons/ringbox/keybinds still works fine even with textbox force-active
  * Fix main menu having fonts too small and column too small on Linux, equal-ish the sizing with Windows (just did 1.7x of Windows)
  * Fix ESC not utilize for exit in main menu
  * Fix overlapping hot widgets/check in view
  * Scrolling via mouse drag is now no longer smooth scroll so we don't need to deal with partially shown widgets
  * Fix dislodge x-axis after per-row change (server filter section)
  * Scrollbar now inside the section rather than outside
* FUTURE TODO:
  * Change dPanel setting to per every NeoUI::BeginSection parameter rather than loose dPanel settings (plan to do this but too much changes/out of scope for this PR)
  * Scrollbar position restore
  * Animated hot border? Lerp smoothing from one widget to the next and rendered at EndContext

## Toolchain

- Linux GCC Distro Native - gcc version 15.2.1 20260103 - Arch Linux

## Linked Issues

* fixes #1155
* fixes #1124


Agiel (dismissed):

DESTROYGIRL:
That aside, the additions looked good but there are some things I would tweak:
Text on selected tabs is too thin / harder to read imo. Try raising the thickness a tiny bit perhaps,  (i have a 1080p monitor)
image

Might be personal preference but I don't feel like the border around whatever the cursor is hovering over needs to be so thick. I'd try halving it, but feel free to ignore
image

Masterkatze:
A bit of bikeshedding opinion, but:

1.1 I think UI element hover gives an impression of some debug feature or a placeholder effect,
1.2 And also the column header that is currently used for sorting has overlining, which is kinda strange choice, especially when placed under the tabs - it may look like it's related more to the tab rather than column header.

Maybe we can use the top/bottom arrow on the right side of a column header to indicate that it's currently sorted (and the sorting order)? Then we can get rid of overlining and only use the underlining for the selected UI element.

2. I know it's not that related to this PR, but the `NeoUI::Divider` line color could be more white? It's hard to notice and the settings look cluttered at the first glance:


But overall, it looks fine and works without any issues.

nullsystem:
Changes: https://github.com/NeotokyoRebuild/neo/pull/1653/changes/3b1d781391cfa6ad6bf68cd953129b88262d0683

@Masterkatze 
> 1.1 I think UI element hover gives an impression of some debug feature or a placeholder effect,

I did the border hover that way mainly for visibility, before it was hard to tell which widget you're hovering over.

> 1.2 ... column header ...

I'll address this in a separate PR after this, working on doing NeoUI table headers

> 2 ... diviers...

I just made them basically white-ish now

@DESTROYGIRL 

> You need to use [!$LINUX] instead of [$!LINUX] it seems

Fixed the conditional to `!$LINUX`

> ... border ...

They're a little thinner now

> Text on selected tabs is too thin / harder to read imo

I can't set font weight on Linux, I'm leaving this out for now.

PR #1657, opened 2026-02-09
Some interp stuff

bauxiteDYS:
We can see in the SDK code `FIXME_INTERP_RATIO`, I believe this is because the `cl_interp` value is being used instead of `cl_interp_ratio` if it's larger and I don't really believe that's how it should work, probably ideally want clients to have the lowest interp they asked for between the two.

This doesn't change all that behaviour, it's a very quick fix that just changes the default `cl_interp` to match a resulting value when default ratio of 2 is used assuming `cl_updaterate 66` (clamped by server). Also we don't allow clients to further increase their interp beyond 0.1s, as it's not necessary and causes greater problems in lag comp, not to mention typos and bad configs reducing the quality of the gameplay experience.

Unfortunately defaults for cl_cmdrate and cl_updaterate live in the engine so can't be changed at the moment, but these can be enforced by server settings which most server operators already do.

Ideally the clients would get the minimum interp they have asked for or the minimum server enforced interp if there is one.

Maybe this is good enough for now?

### DIFF
--- a/src/game/client/cdll_bounded_cvars.cpp
+++ b/src/game/client/cdll_bounded_cvars.cpp
@@ -145,7 +145,7 @@ float GetClientInterpAmount()
 			AssertMsgOnce( false, "GetInterpolationAmount: can't get cl_updaterate cvar." );
 		}
 #ifdef NEO
-		return 0.033;
+		return 2.0 / 66;
 #else
 		return 0.1;
 #endif


Rainyan:
Pushing some changes with go-ahead from @bauxiteDYS 

This probably needs some testing. I got fed up with fighting the ServerBounded system for cl_interp, and it was just a weird layer of abstraction over the interp rate, so I decided to remove it entirely in favor of purely using cl_interp_rate. For the full rationale, see commit message: https://github.com/NeotokyoRebuild/neo/pull/1657/changes/43e25582ff139ebf09134d2ad9e7fa7f8100ea4f

Rainyan:
@AdamTadeusz Could you test if you're seeing any stuttering with the current PR head?

You can set the effective interp with: cl_interp_ratio / cl_updaterate as shown as "lerp" in net_graph

Issue #1644, opened 2026-02-02
SRS/ZR68L viewmodel shouldn't disappear when switching to it while zoomed in


Agiel:
### Build Info

v24.0

### Description

The viewmodel for the SRS and ZR68L disappear when switching to them while zoomed in.

### To Reproduce

1. Give yourself an SRS or ZR68L
2. Equip your pistol
3. Toggle ADS
4. Switch to the rifle. 

### Expected behavior

The rifle should be equipped in hipfire state with the viewmodel visible.

### Actual behavior

The viewmodel goes invisible until you switch to another gun and back again.

### Operating System

- [ ] Windows
- [x] Linux

### Version/Distro

CachyOS

PR #1662, opened 2026-02-10
Fix SRS quickswitching

### LINKED ISSUE #1385: SRS quickswitching should be (plugin) parity
### Build Info

v19.0-alpha

### Description

The SRS bolting should be cancellable using the quickswitch mechanic.

In OGNT, the bolting animation can be bypassed entirely by quickswitching away and back to the weapon, allowing for a faster refire time. This is probably not what we want, but instead the [plugin parity](https://github.com/Agiel/nt-srs-limiter) version, which is the typical NT experience players are used to.

In OGNT plugin-parity, the quickswitch can still be used to cancel the bolting animation, but doing so no longer makes you shoot any faster. This is desirable for reasons like quickly exiting the scope, or staying scoped, or just muscle memory and habit.

In NT;RE as of v19.0-alpha, you cannot refire until the bolting animation has been played. Also, if you quickscope, the bolting animation will forcibly pull you away from the scoped-in status against the player's will. This actually makes quickswitching *slower* than not doing so. This doesn't feel good, as the player.

Here's an example clip from the v19.0-alpha playtest from deepfrog, where after performing the quickswitch, he's forcibly pulled out from the scope for the bolting animation, and then has to manually switch to a pistol to continue fighting, which feels really awkward as the controlling player (ideally he should be allowed to remain scoped until ready to shoot again):

https://github.com/user-attachments/assets/5fe59c03-2096-4b1d-a907-a6b75fe826bf

For more video comparisons of OGNT, plugin-parity, and NT;RE, see below.

### To Reproduce

* Shoot the SRS
* Instantly after shooting, switch to a different gun
* Then quickly switch back to the SRS

### Expected behavior

* The bolting animation is cancelled.
* Refire interval is respected.

### Actual behavior

* The bolting animation is not cancelled.
* The bolting animation can interfere with fastest possible refire interval, making the quickswitch *slower* than not doing it.

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

11

More examples of the behaviour:

### OG NT
This is the original parity, quickswitching completely unrestricted. Note how I can cancel the bolting animation.

This is not perfect shooting input from me, but just something I recorded quickly. You can shoot somewhat faster than this with good input.

https://github.com/user-attachments/assets/32944833-06f6-4929-8cc4-2aad728974a1

### OG NT, plugin-parity
This is OG NT, but with [Agiel's SRS limiter plugin](https://github.com/Agiel/nt-srs-limiter). The animation can be bypassed, but minimum refire interval is respected. This is probably what we want, since it most closely mirrors the NT experience people are used to.

https://github.com/user-attachments/assets/13f6dc4d-80f7-43ce-8e10-6facc5ae0fda

### NT;RE version v19.0-alpha
This is our current behaviour. Notice how I'm forced into the bolting animation, and cannot continue shooting until the animation is done. It does not feel good as the player.

https://github.com/user-attachments/assets/ded91be0-b4dd-4acc-9de8-a578752d5bfa

Rainyan:
## Description
Fix SRS quickswitching so that we don't cancel the player's scoped state nor block their shoot inputs with the bolting animation.

Note that the quickswitching to block the bolting anim is a separate mechanic from limiting the SRS fire-rate (in plugin-parity), i.e. the player should still not be able to bypass their `m_flNextAttack` fire rate limit as determined by the weapon script's cycle time.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1385
- not addressed/out of scope: #1644

### DIFF
--- a/src/game/client/hud_crosshair.cpp
+++ b/src/game/client/hud_crosshair.cpp
@@ -29,6 +29,7 @@
 
 #ifdef NEO
 #include "weapon_neobasecombatweapon.h"
+#include "weapon_srs.h"
 #include "neo_gamerules.h"
 #endif
 
@@ -498,7 +499,18 @@ void CHudCrosshair::Paint( void )
 #endif
 		);
 
-		if (!(pWeapon->GetNeoWepBits() & NEO_WEP_SRS))
+		// NEO NOTE (Rain): If we "need" to bolt the gun, don't show the xhair inaccuracy.
+		// This is a special case for SRS handling when un-scoping right after firing a bullet.
+		// If we don't do this check, the player could see the scope inaccuracy effect for a
+		// few frames after shooting, which can hinder the scope's visibility especially at long range
+		// without offering any useful info to the player in return (i.e. the SRS post-fire
+		// inaccuracy doesn't matter because they cannot shoot again anyway until it has decayed).
+		bool weaponNeedsToBeBolted = false;
+		if (pWeapon->GetNeoWepBits() & NEO_WEP_SRS)
+		{
+			weaponNeedsToBeBolted = assert_cast<CWeaponSRS*>(pWeapon)->GetNeedsBolting();
+		}
+		if (!weaponNeedsToBeBolted)
 		{
 			if (cl_neo_crosshair_scope_inaccuracy.GetBool())
 			{
--- a/src/game/shared/neo/weapons/weapon_srs.h
+++ b/src/game/shared/neo/weapons/weapon_srs.h
@@ -44,6 +44,8 @@ public:
 	virtual float GetSpeedScale(void) const OVERRIDE { return 0.725f; }
 	bool CanBePickedUpByClass(int classId) OVERRIDE;
 
+	inline bool GetNeedsBolting() const { return m_bNeedsBolting; }
+
 protected:
 	virtual float GetFastestDryRefireTime() const OVERRIDE { return 0.2f; }
 


src/game/client/hud_crosshair.cpp
```diff
@@ -498,8 +498,7 @@ void CHudCrosshair::Paint( void )
 #endif
 		);
 
-		const bool isSRS = pWeapon->GetNeoWepBits() & NEO_WEP_SRS;
-		if (!isSRS || pWeapon->GetRoundChambered())
+		if (!(pWeapon->GetNeoWepBits() & NEO_WEP_SRS))
```
Agiel: This change has removed the "dynamic" scope completely from the SRS, whereas the original intention I guess was to temporarily disable it after firing a shot.

Rainyan: Oh, whoops. I pushed a commit to fix this.

Rainyan:
> One nit I have is that I feel like OGNT initiates the reload a bit faster when empty, but that might be out of scope (pun intended) for this issue.

I think I'll leave this outside the PR for now.

Another thing I noticed is that we go to the SRS scope instantly, whereas in OGNT there is a delay (eyeballed from 60fps video, it looks like 0.1 seconds) of the viewmodel's scope being raised in front of the camera (presumably to a pos offset defined in the weapon script?), before we actually zoom and render the scoped HUD texture.

Also the OGNT SRS scope FOV change is instant, whereas we do a lerp like with the non-sniper guns, which could change the aim feel especially for quickscoping.

PR #1668, opened 2026-02-12
Restore explosion sounds

DESTROYGIRL:
## Description
OGNT hackily overwrote the `BaseExplosionEffect.Sound` which makes all explosions (other than frags / detpack) just make a debris sound. This changes a couple things:

- Nades / dets that detonate underwater will use the correct underwater explosion sound
- All other kinds of explosions play explosion sounds as they were intended by the SDK (for example, when the juggernaut is killed)

A side effect of this is that the explosion on rogue will be VERY loud. This will be fixed seperately

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/game/neo/scripts/game_sounds_weapons.txt
+++ b/game/neo/scripts/game_sounds_weapons.txt
@@ -1153,7 +1153,7 @@
 
 "BaseGrenade.Explode"
 {
-	"channel"	"CHAN_VOICE"
+	"channel"	"CHAN_WEAPON"
 	"volume"	"0.45"
 	"soundlevel"  "SNDLVL_75dB"
 


game/neo/scripts/game_sounds_weapons.txt
```diff
@@ -1120,7 +1120,52 @@
...
+	{
+			"wave"	"^weapons/underwater_explode3.wav"
+			"wave"	"^weapons/underwater_explode4.wav"
+	}
+
+}
+
 "BaseGrenade.Explode"
+{
+	"channel"	"CHAN_VOICE"
```
Rainyan: Should this be `CHAN_WEAPON`?

DESTROYGIRL: This was copied from the SDK but yeah could be changed

Rainyan (approved):

PR #1672, opened 2026-02-13
Apply godmode when comp match is paused

### LINKED ISSUE #1645: [FEATURE] sv_neo_warmup_godmode should apply when a match is paused
### Description

Sometimes it's preferable that players take no damage (or do anything to modify their score) while the match is paused. 

### Proposed Solution

Easiest solution would be to make _sv_neo_warmup_godmode_ apply, but we could also add a separate command _sv_neo_pause_godmode_. 

### Alternatives Considered

Alternatively we could just not increase score when players kill each other, but I image it'll be more work to find all places where score is modified.

### Additional Context

_No response_

Rainyan:
## Description
Make `sv_neo_warmup_godmode 1` apply also for the comp pause mode.

## Toolchain
- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues
- fixes #1645

### DIFF
--- a/src/game/server/neo/neo_player.cpp
+++ b/src/game/server/neo/neo_player.cpp
@@ -2979,12 +2979,18 @@ int	CNEO_Player::OnTakeDamage_Alive(const CTakeDamageInfo& info)
 {
 	if (m_takedamage != DAMAGE_EVENTS_ONLY)
 	{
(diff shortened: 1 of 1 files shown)


Rainyan:
Thanks GitHub, really useful auto-generated merge message. Sigh...

PR #1679, opened 2026-02-14
Don't force crouch for bots when reloading

sunzenshen:
## Description
Don't force crouch anymore when reloading because:
- The Juggernaut was always crouching, because their weapon uses 0/low Clip1() ammo to indicate a cooler state for the heat based weapon.
- When retreating, it may be more important to be fast than than to be small, so crouching is now disabled in favor of walking/sprinting which are faster than crouch walking.

## Toolchain
- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
@@ -607,8 +607,6 @@ void CNEOBotMainAction::FireWeaponAtEnemy( CNEOBot *me )
 
 			if ( m_isWaitingForFullReload )
 			{
-				me->PressCrouchButton(0.3f);
-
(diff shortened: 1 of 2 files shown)


sunzenshen:
Regarding: "The Juggernaut was always crouching, because their weapon uses 0 primary ammo to indicate a completely cool state for the heat based weapon."

I realize the PR as is partly papering over the issue of how the juggernaut bot may be attempting to reload all the time, but the crouching (when reloading) behavior was the more obvious behavior.  For sake of keeping this PR small, I think it's fine to ignore the bots thinking the BALC is empty, as the existing special case for NEO_WEP_BALC in CNEOBotMainAction::FireWeaponAtEnemy allows the bot to continue firing the weapon even if the other code considers the weapon empty/low.

Masterkatze (approved):

PR #1681, opened 2026-02-14
Make it possible to fire the supa immediately after loading a shell

### LINKED ISSUE #1678: Supa reload animation should be cancellable
### Build Info

v24

### Description

When you press the fire button while reloading it takes much longer to fire than in OGNT.

### To Reproduce

1. Take out the Supa7
2. Fire a few rounds
3. Start reloading
4. Press and hold the fire button.

### Expected behavior

The gun should fire fast (depending on where in the cycle you are).

### Actual behavior

The animation to point the gun forward has to be played out before the round goes off.

### Operating System

- [ ] Windows
- [x] Linux

### Version/Distro

CachyOS

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

Agiel:
## Description
I didn't actually time this to make sure it's parity but it feels a lot closer. Basically, I only bump nextAttack when starting the reload and not for all the sequences after. A new member m_flNextReload was added to handle the delay between each shell because it was using nextAttack before.

I left the attack delay code commented out in case we want to revisit it later.

## Toolchain

## Linked Issues

- fixes #1678

### DIFF
--- a/src/game/shared/neo/weapons/weapon_supa7.cpp
+++ b/src/game/shared/neo/weapons/weapon_supa7.cpp
@@ -11,11 +11,13 @@ BEGIN_NETWORK_TABLE(CWeaponSupa7, DT_WeaponSupa7)
 	RecvPropBool(RECVINFO(m_bSlugLoaded)),
 	RecvPropBool(RECVINFO(m_bWeaponRaised)),
 	RecvPropBool(RECVINFO(m_bShellInChamber)),
+	RecvPropFloat(RECVINFO(m_flNextReload)),
 #else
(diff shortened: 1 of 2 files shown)


sunzenshen (approved):
LGTM:

https://github.com/user-attachments/assets/a44c0e15-3978-4ae3-82f7-ce2f036abae3

Also here is some footage of the before behavior,  checking that I understand how it feels when the animation cannot be cancelled on master branch:

https://github.com/user-attachments/assets/3d65dc22-e4b5-40f0-962a-fd9caeab35fa

PR #1687, opened 2026-02-14
Fix sv_neo_ghost_spawn_bias

### LINKED ISSUE #1652: Ghost should not spawn at the same spot over multiple rounds when sv_neo_ghost_spawn_bias is enabled
### Build Info

20260201_bc59026

### Description

When using sv_neo_ghost_spawn_bias is enabled, then the ghost can spawn at the same position multiple times in a row. This is the opposite of the intended behavior of the command (and the plugin that it is based on). 

Might be caused because of the removal of the weighted ghost spawns

### To Reproduce

1. Enable Ghost Bias
2. Observe the ghost spawns

### Expected behavior

Ghost should switch its spawn position after two rounds of spawning in the same position

### Actual behavior

Ghost can spawn at the same position over a few rounds. For example, testing the spawn bias on Oilstain, the ghost spawned at the same position for 4 rounds in a row.

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

Rainyan:
## Description
Plugin-parity: shuffle ghost spawn list at match start, and then cycle through it in sequence of 2 rounds per spawn.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1652

### DIFF
--- a/src/game/shared/neo/neo_gamerules.cpp
+++ b/src/game/shared/neo/neo_gamerules.cpp
@@ -1868,17 +1868,14 @@ void CNEORules::SpawnTheGhost(const Vector *origin)
 			// Round numbers are one-indexed
 			Assert(roundNumber() > 0);
 			bool isFirstRound = (roundNumber() == 1);
-
 			if (isFirstRound)
 			{
+				// Plugin parity: we want to shuffle the list of ghost spawns once at match beginning,
+				// and then play through them in round pairs, using the cycling logic right below this if-block.
 				m_ghostSpawns.Shuffle();
 			}
-			// Round numbers are one-indexed, so we want to start incrementing on odd rounds 3, 5, 7...
-			else if (!roundNumberIsEven())
-			{
-				m_ghostSpawns_BiasHead = (m_ghostSpawns_BiasHead + 1) % m_ghostSpawns.Count();
-			}
-			desiredSpawn = m_ghostSpawns_BiasHead;
+
+			desiredSpawn = Ceil2Int(roundNumber() / 2.f) % m_ghostSpawns.Count();
 		}
 		Assert(desiredSpawn >= 0);
 		Assert(desiredSpawn < m_ghostSpawns.Count());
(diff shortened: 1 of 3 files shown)


src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -1853,58 +1846,68 @@ void CNEORules::SpawnTheGhost(const Vector *origin)
...
-					{
-						m_pGhost->SetAbsOrigin(ghostSpawn->GetAbsOrigin());
-						m_pGhost->Drop(vec3_origin);
-					}
+				m_ghostSpawns.Shuffle();
+			}
+			// Round numbers are one-indexed, so we want to start incrementing on odd rounds 3, 5, 7...
+			else if (!roundNumberIsEven())
+			{
+				m_ghostSpawns_BiasHead = (m_ghostSpawns_BiasHead + 1) % m_ghostSpawns.Count();
```
Agiel: I realize this is similar to what we were doing before but it made me kinda nervous that it wasn't just `(roundNumber / 2) % spawns`. Considering pauses or hypothetical features to rewind to a specific round and whatnot.

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -1853,58 +1846,65 @@ void CNEORules::SpawnTheGhost(const Vector *origin)
...
-						m_pGhost->Drop(vec3_origin);
-						Warning("Failed to get ghost spawn coords; spawning ghost at map origin instead!\n");
-						Assert(false);
-					}
-					else
-					{
-						m_pGhost->SetAbsOrigin(ghostSpawn->GetAbsOrigin());
-						m_pGhost->Drop(vec3_origin);
-					}
+			desiredSpawn = Ceil2Int(roundNumber() / 2.f) % m_ghostSpawns.Count();
```
Agiel, line 1878: I think I would've done `Floor((roundNumber - 1) / 2)` so we start at 0 instead of 1. But since the spawns are randomized anyway it doesn't matter.

DESTROYGIRL (approved):

PR #1691, opened 2026-02-15
Update game name stylization

Rainyan:
## Description

Update game name string stylization from `Neotokyo; Rebuild` -> `NEOTOKYO;REBUILD`.

Also use the name define instead of hard-coding it.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/game/neo/gameinfo.txt
+++ b/game/neo/gameinfo.txt
@@ -1,6 +1,6 @@
 "GameInfo"
 {
-	Game        "Neotokyo: Rebuild"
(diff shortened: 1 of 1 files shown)


sunzenshen (approved):
Nit: Not sure if semicolons are allowed in Windows title bars, but wonder if that could also be tweaked to be the new name style as well:

(See: "Neotokyo: Rebuild" on the top left)

Image

sunzenshen (approved):

PR #1695, opened 2026-02-16
Make nav_generate more precise for navarea connections

sunzenshen:
## Description
Make nav_generate more precise for navarea connections

## Toolchain
- Windows MSVC VS2022

## Related
- https://github.com/NeotokyoRebuild/neoAssets/pull/102

### DIFF
--- a/src/game/server/nav.h
+++ b/src/game/server/nav.h
@@ -19,22 +19,17 @@
  * Below are several constants used by the navigation system.
  * @todo Move these into TheNavMesh singleton.
  */
+
 #ifdef NEO
-const float GenerationStepSize = 20.0f;			// NEO: some hallways/openings were missing connection with larger value
+const float GenerationStepSize = 20.0f;			// NEO: some hallways/openings were missing connections with larger value
 #else
 const float GenerationStepSize = 25.0f;			// (30) was 20, but bots can't fit always fit
 #endif
 
-#ifdef NEO
-const float JumpHeight = 54.0f;					// NEO: See NEO_RECON_CROUCH_JUMP_HEIGHT calculations in neo_player_shared.h
-#else
 const float JumpHeight = 41.8f;					// if delta Z is less than this, we can jump up on it
-#endif
 
 #if defined(CSTRIKE_DLL)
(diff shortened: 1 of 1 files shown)


sunzenshen:
Much to my chagrin, after testing this out on a couple maps, while this change seems to work flawlessly for simple maps like Oilstain, this change can automatically introduce NavMesh breaks in maps like Transit, where in this example spawn point, running nav_generate resulted in the stairs getting disconnected, causing bots on one team to be unable to play the map, unless the navareas here are manually connected:

Screenshot 2026-02-26 192558

As such, maybe it would be better to later discuss approaches for how to fix navmeshes for after the tournament.

Rainyan (approved):
Technically I think the hullsize width of all player classes in NT is 16 (from `GetPlayerMaxs`), so a value of 20 does leave some potential routes unconnected, like this one (it's connected in the image because I was testing the value of 16):

Image

But since the bots are a bit restless, they have a hard time to actually fit in such a tight space accurately anyway, so maybe it's better to not connect them yet. But later down the line I think it would be good to support the snuggest fit for really narrow corridors, since these could plausibly appear in custom maps. Perhaps implementing the PRECISE area flag, and auto-generating any tight spaces <20 units to have the flag would help? Can probably track down where the comparison happens by searching `GenerationStepSize`.

sunzenshen:
> Perhaps implementing the PRECISE area flag, and auto-generating any tight spaces <20 units to have the flag would help? Can probably track down where the comparison happens by searching `GenerationStepSize`.

I assume you already know but, for anyone following the discussion, if we implement using specific Attributes like NAV_MESH_PRECISE and NAV_MESH_AVOID, we will also need to implement the bot behavior routines to actually respect these NavMesh attributes. so it may be a good idea to address each Attribute we want to add in a separate PR.

For example, testing the window washer route on Rise for NAV_MESH_PRECISE since bots tend to fail the ledge traversal.  In that case we might need to turn off or tweak specific behaviors like dodge strafing, etc.

AdamTadeusz (approved):

PR #1700, opened 2026-02-17
Go into overtime if Juggernaut is being activated, allow teammates to score in JGR

DESTROYGIRL:
## Description
Title and also reverts round timelimit to 4 minutes and max score to win to 20
The scoring change is experimental.
Currently there is no way of setting per gamemount max rounds / score limit and that will be done seperately

## Toolchain
- Windows MSVC VS2022


sunzenshen (approved):
LGTM:

Image

(I noticed that bots are still a little unreliable with pressing the use button to capture the JGR, but I consider that a bot specific problem out of scope of this PR.)

PR #1701, opened 2026-02-17
Cull smoke particles inside walls

DESTROYGIRL:
## Description
Parity. Removes unnecessary smoke particles to stay under the particle limit. Also adds a warning when hitting the limit so we know what might be happening...

## Toolchain
- Windows MSVC VS2022


AdamTadeusz:
image

Warning works, players are easier to spot inside of smokes but whatever

PR #1702, opened 2026-02-17
Make sound when taking damage optional

Agiel:
## Description
Add "Taking damage sounds" option. Disabled by default.

image

## Toolchain

## Linked Issues

- fixes #

### DIFF
--- a/src/game/server/player.cpp
+++ b/src/game/server/player.cpp
@@ -1000,10 +1000,6 @@ void CBasePlayer::TraceAttack( const CTakeDamageInfo &inputInfo, const Vector &v
 	}
 }
 
-#ifdef NEO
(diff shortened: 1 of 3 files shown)


src/game/server/player.cpp
```diff
@@ -1047,7 +1051,14 @@ void CBasePlayer::DamageEffect(float flDamage, int fDamageType)
 	}
 	else if ( fDamageType & DMG_BULLET )
 	{
+#ifdef NEO
+		if (cl_neo_taking_damage_sounds.GetBool())
```
Rainyan: I think `cl_neo_taking_damage_sounds` should be a client-sided `FCVAR_USERINFO`-flagged variable, that we get via `engine->GetClientConVarValue(...)` or similar(?)

Since I assume we want this to be a per-client (hence `cl_...`) preference, rather than a server setting, and we are in server-sided code here.

PR #1706, opened 2026-02-19
NeoUI - Color widgets and Begin/EndPopup

### LINKED ISSUE #1484: NeoUI - Color widget
The crosshair's color section could be compacted from 4 rows of sliders to 1 row of color tweaking widget. Maybe something like: square color preview + text edit hex format (EX: `#FFFFFFFF`), then when clicking on the preview a popup to control RGBA + preset colors.

nullsystem:
## Description
Color widgets shows the color from given RGBA values and acts as a button to show a popup where the RGBA values can be altered. This also changed how popups are handled, they are now similar to sections and can hold the same widgets and layouts as sections but have specialized handling for popup functionalities.

Keyboard and controller can exit popup with ESC/back-related bind, and showing popup on first initialization activates on the first item of the popup, although this is an optional behavior only utilized for color widgets.

Copy, server, and blacklist menus all migrated to using the new Begin/EndPopup.

## Toolchain

## Linked Issues

* fixes #1484


sunzenshen (approved):
LGTM:

Image

Image

Image

PR #1709, opened 2026-02-20
Make firstperson tracers look less bad

DESTROYGIRL:
## Description
Tracers should actually come from the gun now and not be pixel thin

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/client/clientsideeffects_test.cpp
+++ b/src/game/client/clientsideeffects_test.cpp
@@ -155,7 +155,11 @@ void FX_PlayerTracer( Vector& start, Vector& end )
 
 	//Randomly place the tracer along this line, with a random length
 	VectorMA( start, TRACER_BASE_OFFSET + random->RandomFloat( -24.0f, 64.0f ), shotDir, dStart );
+#ifdef NEO
+	VectorMA( dStart, ( length * random->RandomFloat( 0.1f, 1.6f ) ), shotDir, dEnd );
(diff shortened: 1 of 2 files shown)


src/game/client/clientsideeffects_test.cpp
```diff
@@ -164,7 +168,11 @@ void FX_PlayerTracer( Vector& start, Vector& end )
 	//materialName = ( random->RandomInt( 0, 1 ) ) ? "effects/tracer_middle" : "effects/tracer_middle2";
 	materialName = "effects/spark";
 
+#ifdef NEO
+	t = new CFXStaticLine( "Tracer", dStart, dEnd, random->RandomFloat( 2.5f, 2.75f ), 0.01f, materialName, 0 );
+#else
 	t = new CFXStaticLine( "Tracer", dStart, dEnd, random->RandomFloat( 0.5f, 0.75f ), 0.01f, materialName, 0 );
+#endif
```
Rainyan, line 175: Maybe out of scope for PR, but BALC's tracer attachment point from `GetTracerOrigin` seems somewhat misaligned:

Image

DESTROYGIRL: I think it might be because the weapon starts in that position but jumps down the next frame of the animation. The balc uses a different tracer type so the lines mentioned shouldnt be related

PR #1713, opened 2026-02-21
Give knife to support

Agiel:
## Description
See title

## Toolchain

## Linked Issues

- fixes #
- related #

### DIFF
--- a/src/game/shared/neo/weapons/weapon_knife.cpp
+++ b/src/game/shared/neo/weapons/weapon_knife.cpp
@@ -106,7 +106,7 @@ void CWeaponKnife::PrimaryAttack()
 
 bool CWeaponKnife::CanBePickedUpByClass(int classId)
 {
-	return classId != NEO_CLASS_JUGGERNAUT;
+	return (classId != NEO_CLASS_VIP) && (classId != NEO_CLASS_JUGGERNAUT);
 }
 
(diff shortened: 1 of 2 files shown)


AdamTadeusz (dismissed):
not parity, denied

sunzenshen (approved):
Would be useful to have to simplify bot navigation around breakables.

AdamTadeusz (changes requested):
VIP doesn't have third person knife animations

src/game/shared/neo/weapons/weapon_knife.cpp
```diff
@@ -106,7 +106,7 @@ void CWeaponKnife::PrimaryAttack()
 
 bool CWeaponKnife::CanBePickedUpByClass(int classId)
 {
-	return (classId != NEO_CLASS_SUPPORT) && (classId != NEO_CLASS_JUGGERNAUT);
+	return classId != NEO_CLASS_JUGGERNAUT;
```
AdamTadeusz: replace support with vip here since no vip animations?

bauxiteDYS:
No punch? 👀 
Perhaps allow Support to be able to drop the knife if they choose. There could be crazy 300 iq meta strats where they use the faster run speed from having no weapons to clutch a round, it's definitely happened one time in the history of Neotokyo.

PR #1715, opened 2026-02-21
NeoUI - Fix first-used force active textedit not properly set cursor

nullsystem:
## Description
Password input textedit have forced active and if that textedit is used first, the cursor are not set properly and when the player enters a character the cursor goes left of the character, not right. Just add an if-statement to check if selection variables set to -1 and widget is active, set it to the size of the current text.

Test it with local server, make sure no other text edits are touched, create a local server then go to server list, enter that local server, see that password section, and start typing. The cursor should stay on the most-right side/end of the password text input.

## Toolchain

- Linux GCC Distro Native GCC 15


sunzenshen (approved):
To be honest, not sure how to test if there are no servers up for this branch version, but looks sensible enough to me.

nullsystem:
> To be honest, not sure how to test if there are no servers up for this branch version, but looks sensible enough to me.

@sunzenshen I've tested it using a local server, which is password protected by default so don't need to touch other text inputs before using password input.

sunzenshen (approved):

PR #1716, opened 2026-02-21
Fix pen handling for multiple func_breakable_surf skipping glass breaking effect

### LINKED ISSUE #1685: Bullet penetrating multiple func_breakable_surf should break them all
### Build Info

196ec4792

### Description

A bullet that travels through multiple `func_breakable_surf` panes should break them all. Currently, we only break the first penetrated pane, even though the bullet makes its way through the others.

### To Reproduce

* Go to server room in ntre_rise_ctg
* Line up a shot through the `func_breakable_surf` glass panes surrounding the server

### Expected behavior

All window panes lined up will shatter as the bullet penetrates them

### Actual behavior

Only the first penetrated pane is shattered

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

11

sm90x:
## Description

Fix bullet penetration through multiple `func_breakable_surf` panes.

Two fixes:
 - Proper penetration handling for glass on both sides (glass texture and nodraw texture)
 - Saves the function penetration handling function from stopping too soon and failing to apply shatter effects to glass

~~Plus I told the ai to not apply any damage reduction with these, but I am not sure what the behavior in the original is. (easily removable if incorrect)~~

The original fix had some crash happen, I went through the firing range and it didn't crash. I also tested the double shatter multiple times, nothing bad happened.

~~disclaimer: These changes were authored by AI. I reviewed it best I could, but it's the first time I look at Source(tm) code. The solution makes sense _I guess_.~~

AI assisted in finding cause of bug.

## Toolchain

- Linux GCC 10 Sniper 3.0

## Linked Issues

- fixes #1685

### DIFF
--- a/src/game/client/c_func_breakablesurf.cpp
+++ b/src/game/client/c_func_breakablesurf.cpp
@@ -1335,7 +1335,8 @@ IMaterial *CBreakableSurfaceProxy::GetMaterial()
 
 EXPOSE_INTERFACE( CBreakableSurfaceProxy, IMaterialProxy, "BreakableSurface" IMATERIAL_PROXY_INTERFACE_VERSION );
 
-int GetBreakableSurfaceType( C_BaseEntity *pC_BaseEntity )
+#ifdef NEO
+int GetBreakableSurfaceType( const C_BaseEntity *pC_BaseEntity )
 {
 	Assert(pC_BaseEntity);
 
@@ -1344,12 +1345,8 @@ int GetBreakableSurfaceType( C_BaseEntity *pC_BaseEntity )
 		return -1;
 	}
 
-	C_BreakableSurface *pBreakableEnt = assert_cast<C_BreakableSurface *>(pC_BaseEntity);
-
-	if (!pBreakableEnt)
-	{
-		return -1;
-	}
+	auto *pBreakableEnt = assert_cast<const C_BreakableSurface *>(pC_BaseEntity);
 
 	return pBreakableEnt->m_nSurfaceType;
 }
+#endif
--- a/src/game/shared/baseentity_shared.cpp
+++ b/src/game/shared/baseentity_shared.cpp
@@ -78,8 +78,10 @@ ConVar hl2_episodic( "hl2_episodic", "0", FCVAR_REPLICATED );
 	extern bool ExtractKeyvalue( void *pObject, typedescription_t *pFields, int iNumFields, const char *szKeyName, char *szValue, int iMaxLen );
 #endif
 
+#ifdef NEO
 #ifdef CLIENT_DLL
-	extern int GetBreakableSurfaceType(C_BaseEntity *pEnt);
+	extern int GetBreakableSurfaceType(const C_BaseEntity *pEnt);
+#endif
 #endif
 
 bool CBaseEntity::m_bAllowPrecache = false;
@@ -2304,7 +2306,7 @@ void CBaseEntity::HandleShotPenetration(const FireBulletsInfo_t& info,
 		// NEO HACK: Force a fake thickness for now non-solid breakable surfaces,
 		// this prevents stopping the function at the open air case (fraction == 1.0f).
 		// This is because there is in fact nothing solid to penetrate, but still we need to 
-		// invoke the effects for the breakable surface at DoImpactEffect.
+		// invoke the effects for the breakable surface.
 		constexpr float kBreakableSurfFakeThickness = 2.3f;
 		penetrationTrace.fraction = 1.0f - (kBreakableSurfFakeThickness / MAX_PENETRATION_DEPTH);
 	}
@@ -2331,7 +2333,8 @@ void CBaseEntity::HandleShotPenetration(const FireBulletsInfo_t& info,
 	//		 would do exactly the same anyway...
 
 	// Impact the other side (will look like an exit effect)
-	if (!bIsShatterGlassSurf) {
+	if (!bIsShatterGlassSurf)
+	{
 		DoImpactEffect(penetrationTrace, GetAmmoDef()->DamageType(info.m_iAmmoType));
 	}
 


Rainyan (changes requested):
# 🏴‍☠️ Code Review by Captain Jack Sparrow
## 1. Early detection of breakable surfaces

A clever trick, that. Ye’re catchin’ the glass before it turns ghostly and slips through yer fingers. The logic be sound, though I’d mind this bit:

```cpp
const char *cn = tr.m_pEnt->GetClassname();
bIsBreakableSurface = cn && strstr(cn, "BreakableSurface");
```

Aye, it works, but it’s a tad loose — like tyin’ a ship to the dock with a frayed rope. If the client ever renames or subclasses the thing, ye might get false positives. But if that’s how the engine names ’em, then fair winds.
## 2. Manual penetration exit synthesis

Now this be the heart of yer fix, and I must say, it’s as crafty as hidin’ rum from the Royal Navy.

```cpp
penetrationTrace.endpos = tr.endpos + vecDir * 0.5f;
penetrationTrace.fraction = 1.0f - (0.5f / MAX_PENETRATION_DEPTH);
penetrationTrace.plane.normal = -tr.plane.normal;
```

Ye’re fakin’ the exit trace just enough to get the bullet past the first pane so the next one shatters proper. Smart. Very smart. But beware:

* Hardcoding 0.5f is like buryin’ treasure without a map. Works today, maybe cursed tomorrow.

* If the pane thickness varies, ye might want a constant or a cvar for this magic number.

Still, it’s a tidy workaround for Source’s… quirks, shall we say.
## 3. Added safety checks

This part:
```cpp
if (penResistance <= 0.0f || info.m_flPenetration <= 0)
{
    return;
}
```

Aye, that’s good defensive sailin’. No sense calculatin’ penetration when ye’ve got none to give.
## 4. Damage preservation for glass
```cpp
behindMaterialInfo.m_flDamage = info.m_flDamage;
```

Aye, glass shouldn’t be stealin’ damage from the shot. This be correct and proper.
## 5. Minor notes from the captain

    Ye added allsolid to the early-out check. Sensible. Keeps ye from firin’ into cursed geometry.

    The comments be thorough — almost too thorough. But better a map with too many markings than one with none.

# ⚓ Verdict

This patch sails truer than most ships I’ve commandeered. The workaround for the glass cascade bug be clever, targeted, and unlikely to cause mutiny elsewhere in the code.

If I were to polish it further, I’d:

* Replace the 0.5f magic offset with a named constant.

* Tighten the classname check if possible.

* Maybe add a debug overlay to visualize the synthetic exit trace — helps when huntin’ future ghosts.

But as it stands?

**A fine fix, mate. Wouldn’t hesitate to fly it under the black flag.**


...Oh, sorry, you expected a human review? I hate myself too little to review AI code, but in terms of behavior, it breaks the penetration through glass:

https://github.com/user-attachments/assets/f40deabb-4cb8-40df-810d-b7ff77be66b2

src/game/client/c_func_breakablesurf.cpp
```diff
@@ -1334,3 +1334,22 @@ IMaterial *CBreakableSurfaceProxy::GetMaterial()
...
+	{
+		return -1;
+	}
+
+	C_BreakableSurface *pBreakableEnt = assert_cast<C_BreakableSurface *>(pC_BaseEntity);
+
+	if (!pBreakableEnt)
+	{
+		return -1;
+	}
```
Rainyan: This check we can remove, because the above `if (!pC_BaseEntity)` guard takes care of it.

For debug builds, the above `assert_cast` will trigger for bad conversions, and for release builds it will optimize into a `static_cast`, which cannot return a nullptr here (because of the check on line 1342).

src/game/client/c_func_breakablesurf.cpp
```diff
@@ -1334,3 +1334,22 @@ IMaterial *CBreakableSurfaceProxy::GetMaterial()
 }
 
 EXPOSE_INTERFACE( CBreakableSurfaceProxy, IMaterialProxy, "BreakableSurface" IMATERIAL_PROXY_INTERFACE_VERSION );
+
+int GetBreakableSurfaceType( C_BaseEntity *pC_BaseEntity )
```
Rainyan: We could turn the input `pC_BaseEntity` pointer into `const`:

```diff
diff --git a/src/game/client/c_func_breakablesurf.cpp b/src/game/client/c_func_breakablesurf.cpp
index 4256a4bb..23748e2a 100644
--- a/src/game/client/c_func_breakablesurf.cpp
+++ b/src/game/client/c_func_breakablesurf.cpp
@@ -1335,7 +1335,7 @@ IMaterial *CBreakableSurfaceProxy::GetMaterial()

 EXPOSE_INTERFACE( CBreakableSurfaceProxy, IMaterialProxy, "BreakableSurface" IMATERIAL_PROXY_INTERFACE_VERSION );

-int GetBreakableSurfaceType( C_BaseEntity *pC_BaseEntity )
+int GetBreakableSurfaceType( const C_BaseEntity *pC_BaseEntity )
 {
        Assert(pC_BaseEntity);

@@ -1344,7 +1344,7 @@ int GetBreakableSurfaceType( C_BaseEntity *pC_BaseEntity )
                return -1;
        }

-       C_BreakableSurface *pBreakableEnt = assert_cast<C_BreakableSurface *>(pC_BaseEntity);
+       auto *pBreakableEnt = assert_cast<const C_BreakableSurface *>(pC_BaseEntity);

        if (!pBreakableEnt)
        {
diff --git a/src/game/shared/baseentity_shared.cpp b/src/game/shared/baseentity_shared.cpp
index 55ff6517..c2dac6bf 100644
--- a/src/game/shared/baseentity_shared.cpp
+++ b/src/game/shared/baseentity_shared.cpp
@@ -79,7 +79,7 @@ ConVar hl2_episodic( "hl2_episodic", "0", FCVAR_REPLICATED );
 #endif

 #ifdef CLIENT_DLL
-       extern int GetBreakableSurfaceType(C_BaseEntity *pEnt);
+       extern int GetBreakableSurfaceType(const C_BaseEntity *pEnt);
 #endif

 bool CBaseEntity::m_bAllowPrecache = false;
```

src/game/client/c_func_breakablesurf.cpp
```diff
@@ -1334,3 +1334,22 @@ IMaterial *CBreakableSurfaceProxy::GetMaterial()
 }
 
 EXPOSE_INTERFACE( CBreakableSurfaceProxy, IMaterialProxy, "BreakableSurface" IMATERIAL_PROXY_INTERFACE_VERSION );
+
```
Rainyan, line 1337: This is kind of badly documented by the NTRE project, but we're using preprocessor guards for all the blocks of NTRE specific code when inside Valve SDK files. This is mostly done to make it more obvious where the code is SDK original and where we have written something custom without needing to actively compare with the git history (and Valve also does it with their games, you'll find blocks like ifdef PORTAL, ifdef CSTRIKE etc, so stylistically it kind of fits).

So this function should be surrounded by a ifdef guard like:

```cpp
#ifdef NEO
int GetBreakableSurfaceType( C_BaseEntity *pC_BaseEntity )
{
/* function body... */
}
#endif
```

Files that are purely original to NTRE are an exception to this rule, typically those will be named with a `neo` prefix, so something like `neo_filename.cpp`

src/game/shared/baseentity_shared.cpp
```diff
@@ -77,6 +78,10 @@ ConVar hl2_episodic( "hl2_episodic", "0", FCVAR_REPLICATED );
 	extern bool ExtractKeyvalue( void *pObject, typedescription_t *pFields, int iNumFields, const char *szKeyName, char *szValue, int iMaxLen );
 #endif
 
+#ifdef CLIENT_DLL
```
Rainyan, line 82: Should use the NEO guard here:
```cpp
#ifdef NEO
#ifdef CLIENT_DLL
	extern int GetBreakableSurfaceType(C_BaseEntity *pEnt);
#endif
#endif
```
or similar

src/game/shared/baseentity_shared.cpp
```diff
@@ -2278,7 +2331,9 @@ void CBaseEntity::HandleShotPenetration(const FireBulletsInfo_t& info,
 	//		 would do exactly the same anyway...
 
 	// Impact the other side (will look like an exit effect)
-	DoImpactEffect(penetrationTrace, GetAmmoDef()->DamageType(info.m_iAmmoType));
+	if (!bIsShatterGlassSurf) {
+		DoImpactEffect(penetrationTrace, GetAmmoDef()->DamageType(info.m_iAmmoType));
```
Rainyan: Style nit: to keep consistent with the surrounding code, would prefer if we moved the opening brace to its own line

PR #1718, opened 2026-02-22
[WIP] Fix weapons visible through wall in PVS in thermal vision in Linux + OpenGL

### LINKED ISSUE #1717: Can see enemy's weapons through walls in thermal vision (Linux + OpenGL + mesa driver only)
### Build Info

master (2026-02-22) 6caf7268a278b72d575fcbd807d60c770738d824

### Description

I guess it's related to PVS as you'll have to be close to the enemy to see it and it's quite pronounced when there's a ghoster player. It's only the weapons (includes ghost, anything it's carrying activly using it or not), player models seems fine/doesn't appear. This is a Linux + OpenGL + mesa driver combination issue, Windows/Linux + vulkan doesn't show the bug.

### To Reproduce

1. Use the mesa driver on Linux
2. Launch NT;RE with `-gl` param on Linux, double check it's OpenGL with `neo_version`
3. Load up `ntre_oilstain_ctg` local server with bots
4. Play around as support, always in thermal vision mode
5. When you get close to enemy, you could see weapons the enemy having/using through walls
6. If the bot's holding the ghost, it becomes more pronounced and can see the ghost and secondary they're holding
### Expected behavior

To not see any weapons (hot or not) through walls in thermal vision mode

### Actual behavior

_No response_

### Operating System

- [ ] Windows
- [x] Linux

### Version/Distro

Arch Linux

### Machine's CPU

_No response_

### Machine's GPU

AMD Radeon RX 9070 XT

### GPU's driver

Mesa (AMD)

### Build's compiler

GNU 15.2.1

Looking at ghoster through wall

Image

nullsystem:
## Description
The fix is more general than just OpenGL, it's just an extra check to see if the weapon is visible in player's FOV or not, then it just doesn't render it. Only applied on weapons as that seems to have the issue but not like player models.

## Toolchain

- Linux GCC Distro Native GCC 15 Arch

## Linked Issues

* fixes #1717

### DIFF
--- a/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
+++ b/src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
@@ -1249,6 +1249,47 @@ int CNEOBaseCombatWeapon::RestoreData(const char* context, int slot, int type)
 	return val;
 }
 
+class CTraceFilterFindWeapon : public CTraceFilterSimple
+{
+public:
+	CTraceFilterFindWeapon( const IHandleEntity *passentity = NULL, int collisionGroup = COLLISION_GROUP_NONE )
+		: CTraceFilterSimple( passentity, collisionGroup )
+	{
+	}
+
+	IHandleEntity *targetEntityCharacter = nullptr;
+	IHandleEntity *targetEntityWeapon = nullptr;
+
+	virtual bool ShouldHitEntity( IHandleEntity *pHandleEntity, int contentsMask )
+	{
+		if ( CTraceFilterSimple::ShouldHitEntity( pHandleEntity, contentsMask ) )
+		{
+			CBaseEntity *pEntity = EntityFromEntityHandle( pHandleEntity );
+			if ( !pEntity )
+				return false;
+
+			if (pEntity->MyCombatCharacterPointer())
+			{
+				return (pEntity->MyCombatCharacterPointer() == targetEntityCharacter);
+			}
+
+			if (pEntity->MyCombatWeaponPointer())
+			{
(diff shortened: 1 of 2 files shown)


AdamTadeusz:
is it possible that baseclass:drawmodel does some kind of check like this which is why the guns aren't visible without vision modes? Or is this that exact check pulled out of one of the base classes?

Rainyan:
@nullsystem I'm still seeing the same problem with commit ebc6e9e3 manifest quite easily, for example just moving the camera around in spawn will cause some teammate's guns to blink in and out of being rendered. Notice the guns on the backs of the bots sometimes disappearing:

https://github.com/user-attachments/assets/171a762b-fb1b-4b0d-af4d-5419f31ff7f3

This was recorded on Windows, but I have the same problem on Linux with OpenGL (I don't think the renderer matters since it's the `IsAbleToSee` logic driving it.

Do you think this would work:

```cpp
if (!pTargetPlayer->IsLineOfSightClear(this, C_BaseCombatCharacter::IGNORE_ACTORS))
{
	return 0;
}
```

I can't reproduce the OpenGL bug on my system (maybe drivers/GPU related also?) so you'd have to verify if that still works as a fix for the original bug, but with the above line, I'm not seeing any of the false positive causing the models to visibly blink in and out anymore.

Rainyan:
Nevermind, I'm still seeing guns disappear, likely because a held gun assumes the abs origin position of the holder.

welp

So I guess we would instead have to do a similar trace, but to the player model's weapon attachment position, while ignoring actors, or something like that...

nullsystem:
Trace checking didn't work out, so abandoning this PR

PR #1725, opened 2026-02-23
Fix bot CTG navigation stuttering during breakaway run

sunzenshen:
## Description
The movements of the ghost carrier and their escorts would stutter if they were closer to the capture point than any enemy.  The root cause was identified as a thrashing invalidation of a valid path, resulted in stuttering between repath timer cycles.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #1600

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
@@ -507,18 +507,14 @@ void CNEOBotCtgCarrier::UpdateFollowPath( CNEOBot *me, const CUtlVector<CNEO_Pla
 			if ( !bAnyEnemyCloser )
 			{
 				m_chasePath.Invalidate();
-				m_path.Invalidate();
 
-				if ( !m_repathTimer.HasStarted() || m_repathTimer.IsElapsed() )
+				if ( !m_path.IsValid() || !m_repathTimer.HasStarted() || m_repathTimer.IsElapsed() )
 				{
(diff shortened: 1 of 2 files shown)


src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
```diff
@@ -507,18 +507,14 @@ void CNEOBotCtgCarrier::UpdateFollowPath( CNEOBot *me, const CUtlVector<CNEO_Pla
 			if ( !bAnyEnemyCloser )
 			{
 				m_chasePath.Invalidate();
-				m_path.Invalidate();
```
sunzenshen, line 510: The stuttering was caused by this Invalidate, as this makes the path only valid in the brief moment that m_repathTimer IsElapsed.  Otherwise the bots are waiting around with their path invalidated.

src/game/server/neo/bot/behavior/neo_bot_ctg_escort.cpp
```diff
@@ -201,7 +201,6 @@ ActionResult< CNEOBot >	CNEOBotCtgEscort::Update( CNEOBot *me, float interval )
 		if ( m_repathTimer.IsElapsed() )
 		{
 			m_chasePath.Invalidate();
-			m_path.Invalidate();
```
sunzenshen, line 204: My understanding at this point is that we only need to Invalidate() a path if we want the bots to stop using that path, which we don't for m_path, but we do want to stop following m_chasePath.

nullsystem (approved):

Rainyan (approved):

PR #1740, opened 2026-02-26
Bots hold cover angles towards incoming threats

sunzenshen:
## Description
Bots have a very short memory of recent unseen threats (3 seconds), so in order to encourage bots to hold threat angles for a longer period of time, we use the directional movement of the potential threat as a general heuristic of whether the threat is still pursuing the bot.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/NextBot/NextBotVisionInterface.cpp
+++ b/src/game/server/NextBot/NextBotVisionInterface.cpp
@@ -32,6 +32,14 @@
 ConVar nb_blind( "nb_blind", "0", FCVAR_CHEAT, "Disable vision" );
 ConVar nb_debug_known_entities( "nb_debug_known_entities", "0", FCVAR_CHEAT, "Show the 'known entities' for the bot that is the current spectator target" );
 
+#ifdef NEO
+ConVar neo_bot_remember_incoming_threat( "neo_bot_remember_incoming_threat", "1", FCVAR_CHEAT,
+	"If true, bots will refresh their memory timers when an unseen enemy is moving towards them.", true, 0, true, 1 );
+
+ConVar neo_bot_remember_incoming_threat_debug( "neo_bot_remember_incoming_threat_debug", "0", FCVAR_CHEAT,
+	"Show debug arrows for neo_bot_remember_incoming_threat. Yellow = Last known position. Red = Actual position.", true, 0, true, 1 );
+#endif
+
 
 //------------------------------------------------------------------------------------------
 IVision::IVision( INextBot *bot ) : INextBotComponent( bot )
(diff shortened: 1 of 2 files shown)


sunzenshen:
When neo_bot_remember_incoming_threat_debug is enabled, yellow indicates where the bot remembers was the last known location of the threat, and red indicates the difference between the last known position and the actual position.

image

The usual 3 second timer is delayed if the threat is still moving towards the bot, assuming the last known position is not visible.

Note that this does not affect what threat the bot prioritizes, it only extends the memory that a threat used to be at previous location, so if the bot does not have any other threats, it will use CNEOBot::UpdateLookingAroundForEnemies to infer what is the closest visible area that the threat could emerge from, given a last known position.  Since this tweak only affects the lifetime of a known threat, it does not directly effect which threat is chosen from GetPrimaryKnownThreat, so bots can still engage other targets while remembering the last known position of older threats.

src/game/server/NextBot/NextBotVisionInterface.cpp
```diff
@@ -464,6 +472,31 @@ void IVision::UpdateKnownEntities( void )
...
+#ifdef NEO
+					// NEO Jank: To address bots forgetting enemies that are rushing them
+					// Intended to prevent bots from looking away from known incoming threats
+					else if ( neo_bot_remember_incoming_threat.GetBool()
+						&& known.GetEntity()
+						&& known.GetEntity()->IsAlive()
+						&& GetBot()->GetEntity()->IsAlive()
+						&& GetBot()->IsEnemy( known.GetEntity() ) )
+					{
+						Vector vecToBot = GetBot()->GetEntity()->WorldSpaceCenter() - known.GetEntity()->WorldSpaceCenter();
```
Rainyan, line 484: Perf nit: `WorldSpaceCenter` is more expensive to calculate than `GetAbsOrigin`. Since we only care about the relative direction vector rather than exact position, perhaps favor the latter?

src/game/server/NextBot/NextBotVisionInterface.cpp
```diff
@@ -32,6 +32,14 @@
 ConVar nb_blind( "nb_blind", "0", FCVAR_CHEAT, "Disable vision" );
 ConVar nb_debug_known_entities( "nb_debug_known_entities", "0", FCVAR_CHEAT, "Show the 'known entities' for the bot that is the current spectator target" );
 
+#ifdef NEO
+ConVar neo_bot_remember_incoming_threat( "neo_bot_remember_incoming_threat", "1", FCVAR_CHEAT,
+	"If true, bots will refresh their memory timers when an unseen enemy is moving towards them.", true, 0, true, 1 );
```
Rainyan, line 37: Do I understand correctly that this is effectively the bots cheating? I'm not sure how I feel about it, especially since this is a behavior directly intended to combat against the player being cheated, which could feel a bit unfair.

I could also see a problem where a shrewd human player would learn to recognize his fellow bots staring at an angle fixedly to mean there is a target there, giving them a kind of wallhack-by-proxy.

One hacky way around this could be to make it probabilistic, so the bot may or may not identify an incoming threat, who may or may not be a false positive. Sort of emulating how human players can have a hunch and hold a completely irrelevant angle, or sometimes predict the push perfectly. This would also make it moot to try and use the bot holding behavior to proxy-wallhack the enemy.

sunzenshen: Another approach could be to estimate how long it could take for an enemy to arrive at an emergence point near the bot, and add to the timer based on this travel estimate.

That said, such an approach might take more computation, and I'd like to merge in other PRs to get an idea of how much CPU budget we have left with various player machines.  If it turns out that we have more budget for calculations, it could be a good idea to consider such advanced reasoning.

Rainyan: Scheduling some kind of think timer to the future sounds like a good idea. I imagine it's not much more expensive than `NavAreaBuildPath`, plus storing the total length and dividing by class speed.

PR #1747, opened 2026-02-26
Server convar sv_neo_reject_opengl_mesa_check to block OpenGL Mesa clients

nullsystem:
## Description
New server side convar sv_neo_reject_opengl_mesa_check to reject OpenGL clients. Also turned on for competitive mode. Only Linux have OpenGL renderer and buggy due to mesa drivers. Similar to the git hash check, utilizes convar send to server for checks. Also make sure both `__cl_neo_...` convars are not changable via console.

## Toolchain

- Linux GCC Distro Native GCC 15/Arch Linux

## Linked Issues

* related #1717


Masterkatze:
Since this issue is present only on Mesa, we can check if it's really Mesa (and not NVIDIA for example) using `glGetString(GL_RENDERER)` from `gl.h`.

Masterkatze:
LGTM. Small nitpick: extra dot at the end of error string:

Screenshot_20260301_005912

Masterkatze (approved):

PR #1749, opened 2026-02-27
Bots follow NAV_MESH_CROUCH attribute

sunzenshen:
## Description
Bots crouch in response to a NAV_MESH_CROUCH attribute of a NavArea

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/neo_bot_locomotion.cpp
+++ b/src/game/server/neo/bot/neo_bot_locomotion.cpp
@@ -22,6 +22,11 @@ void CNEOBotLocomotion::Update( void )
 	// always 'crouch jump'
 	if ( IsOnGround() )
 	{
+		CNavArea* currentArea = me->GetLastKnownArea();
+		if (currentArea && (currentArea->GetAttributes() & NAV_MESH_CROUCH))
+		{
+			me->PressCrouchButton( 0.3f );
+		}
 #ifdef NEO
 		// NEO JANK resetting of crouch timer moved to NextBotPlayer::PressJumpButton
 		// so far crouch jump seems to still be working, but watch out for a regression


src/game/server/neo/bot/neo_bot_locomotion.cpp
```diff
@@ -22,6 +22,11 @@ void CNEOBotLocomotion::Update( void )
 	// always 'crouch jump'
 	if ( IsOnGround() )
 	{
+		CNavArea* currentArea = me->GetLastKnownArea();
+		if (currentArea && (currentArea->GetAttributes() & NAV_MESH_CROUCH))
```
Rainyan, line 26: Since we have the duration-based logic, I'm not sure if this is an issue in practice, but it seems the attributes system supports the concept of a "crouch jump" by applying both `NAV_MESH_CROUCH` and `NAV_MESH_JUMP` to a nav node (console commands `nav_crouch;nav_jump;` to a highlighted node):

Image

Should we check something like `IsOnGround() || ( attributes & (NAV_MESH_CROUCH|NAV_MESH_JUMP) )` to respect this even mid-jump, regardless of on-ground status? Since uncrouching too early could mess up the intended crouch jump on some geometry.

Rainyan (approved):

src/game/server/neo/bot/neo_bot_locomotion.cpp
sunzenshen: The current implementation of crouch jump is certainly a bit unconditional (e.g. the else block below), but as far as I can tell, any time a bot jumps, they always crouch jump because of the else case, and there's no apparent downside of doing so.

As for introducing a Jump action for JUMP attribute NavAreas, I suppose we should consider that at some point... given that bots struggle to get into waist-high vents. But that's kind of jankily handled with the OnStuck crouch jump behavior.

I'd prefer to address JUMP attributes in a different PR with some concrete map test areas.  I'm also a little worried that forcing jump could result in odd behavior, which might warrant more consideration of the edge cases.

As is, this PR certainly doesn't address every NavArea Attribute that is available (same for PRECISE, etc), but it also doesn't break an existing working scenario, since I don't see where the JUMP attribute is handled anywhere at the moment.

Rainyan: Yeah, I almost feel like we should have separate bitflags for "if you're entering this area from elsewhere, you should jump/crouch" type hints, rather than always doing that action when doing unrelated traversal. But offtopic for this PR.

nullsystem (approved):

PR #1753, opened 2026-03-01
Reduce dev console spam

Rainyan:
## Description
* Fix a bunch of console warnings to reduce developer console noise.
* Add some asserts for where some of the errors originated from, to catch them early
* Improve some warnings' message wording

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/shared/achievementmgr.cpp
+++ b/src/game/shared/achievementmgr.cpp
@@ -1691,7 +1691,7 @@ void CAchievementMgr::Steam_OnUserStatsReceived( UserStatsReceived_t *pUserStats
 	if ( pUserStatsReceived->m_eResult != k_EResultOK )
 	{
 #ifdef NEO
-		DevMsg( __FUNCTION__ ": failed to download stats from Steam, EResult %d\n", pUserStatsReceived->m_eResult );
+		DevMsg( "%s: failed to download stats from Steam, EResult %d\n", __FUNCTION__, pUserStatsReceived->m_eResult );
 #else
 		DevMsg( "CTFSteamStats: failed to download stats from Steam, EResult %d\n", pUserStatsReceived->m_eResult );
 #endif


src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -151,12 +151,14 @@ ConVar sv_neo_client_autorecord("sv_neo_client_autorecord", "0", FCVAR_REPLICATE
 ConVar cl_neo_client_autorecord_allow("cl_neo_client_autorecord_allow", "1", FCVAR_ARCHIVE, "Allow servers to automatically record demos on the client", true, 0, true, 1);
 #endif
 
+#ifdef GAME_DLL
 // NEO JANK (nullsystem): REASON/BUG (GH-ISSUE #1717): Linux OpenGL + mesa driver clients can wallhack weapons in thermal vision
-ConVar sv_neo_reject_opengl_mesa_check("sv_neo_reject_opengl_mesa_check", "0", FCVAR_GAMEDLL | FCVAR_REPLICATED,
+ConVar sv_neo_reject_opengl_mesa_check("sv_neo_reject_opengl_mesa_check", "0", 0,
```
sunzenshen, line 156: Sanity check nit: are we sure that FCVAR_GAMEDLL isn't implied here since above `#ifdef GAME_DLL` is defined on line 154?

Otherwise FCVAR_NONE could be used in place of 0.

AdamTadeusz: The only use i've found for FCVAR_GAMEDLL/FCVAR_CLIENTDLL is if a convar is defined only on one side (client/server) and we want it visible on the other side (server/client), i.e cl_neo_pvs_cull_roaming_observer which is defined in the server dll but can still be set by clients thanks to FCVAR_CLIENTDLL

In this instance, I would be happy with FCVAR_NONE, FCVAR_GAMEDLL is implied and will be added anyway

image

image

DESTROYGIRL:
On game start without loading a map. Debug build
image

AdamTadeusz:
i think you can have convars defined in multiple places without a problem, maybe something in the engine redefines this convar without the flag

Rainyan:
Bah, that's annoying. It's probably me who has the stale setup then if the two of you are seeing it. I'll revert that change for now in a new PR.

PR #1777, opened 2026-03-02
+use sound only for local player and new sound

DESTROYGIRL:
## Description
Get rid of HL2 sound, make it local player only

[neoAssets](https://github.com/NeotokyoRebuild/neoAssets/pull/109)<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/hl2/hl2_player.cpp
+++ b/src/game/server/hl2/hl2_player.cpp
@@ -3973,8 +3973,7 @@ void CHL2_Player::ItemPostFrame()
 	{
 		m_bPlayUseDenySound = false;
 #ifdef NEO
-		CRecipientFilter filter;
-		filter.AddRecipient( this );
+		CSingleUserRecipientFilter filter( this );
 		EmitSound( filter, entindex(), "HL2Player.UseDeny" );
 #else
 		EmitSound( "HL2Player.UseDeny" );


src/game/server/hl2/hl2_player.cpp
```diff
@@ -3972,7 +3972,13 @@ void CHL2_Player::ItemPostFrame()
 	if ( m_bPlayUseDenySound )
 	{
 		m_bPlayUseDenySound = false;
+#ifdef NEO
+		CRecipientFilter filter;
```
sunzenshen: I recently just learned of this, but there's a dedicated single person recipient filter called CSingleUserRecipientFilter

It might be more idiomatic to use that single filter kind of like this:
(so that you can remove the AddRecipient call)

```cpp
#ifdef NEO
    CSingleUserRecipientFilter filter( this );
    EmitSound( filter, entindex(), "HL2Player.UseDeny" );
#else
    EmitSound( "HL2Player.UseDeny" );
#endif
```

Another example that I was recently working on:

https://github.com/NeotokyoRebuild/neo/pull/1784/changes#diff-adf9890a25146f1b379ea1a9571f31e19c5a3e61d2548b9564875d1423a38eadR477

DESTROYGIRL: Thanks, I tried to look for this when working on it actually

PR #1780, opened 2026-03-04
Only Support bots should utilize smoke for concealment

sunzenshen:
## Description
While earlier bots could use smoke to hide from each other, feedback from players indicated that humans are less fooled by this ruse, and will fire into the smoke cloud.  This often results in a shooting fish in a barrel effect.  This PR tweaks the bot logic to only deploy and retreat into smoke if their class is Support, so they can take advantage of thermal vision.  Other classes will now prefer hard cover when retreating.  This has a side benefit of saving some CPU instructions when evaluating retreat cover.

## Toolchain
- Windows MSVC VS2022


AdamTadeusz (approved):

DESTROYGIRL (approved):
i thought i already approved this

PR #1783, opened 2026-03-05
Move bot recon superjump from TacticalMonitor to MainAction

sunzenshen:
## Description
Before, the decision for a recon jump was tied up with the single threaded execution of CNEOTacticalMonitor.  That meant that when the bots were occupied with any other behavior associated with TacticalMonitor such as retreating to cover or avoiding grenades, the bots would not consider super jumping.  By moving the implementation to CNEOBotMainAction, the Recon bots will always consider super jump no matter what scenario or tactical monitor behavior the bots are currently engaging in.

## Toolchain
- Windows MSVC VS2022


src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
```diff
@@ -359,6 +367,133 @@ Vector CNEOBotMainAction::SelectTargetPoint( const INextBot *meBot, const CBaseC
 }
 
 
+//-----------------------------------------------------------------------------------------
+void CNEOBotMainAction::ReconConsiderSuperJump( CNEOBot *me )
+{
+	CNEO_Player *pNeoMe = ToNEOPlayer(me);
```
Rainyan, line 373: Nit: Because `CNEOBot` is a subclass of `CNEO_Player`, this is equivalent to:
```cpp
CNEO_Player* pNeoMe = me;
```
i.e. could remove `pNeoMe` and replace all its uses with `me`. While `ToNEOPlayer` is meant to be fast, it's still 1 extra `->IsPlayer` virtual call per every conversion so might as well avoid it when possible.

src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
```diff
@@ -359,6 +367,133 @@ Vector CNEOBotMainAction::SelectTargetPoint( const INextBot *meBot, const CBaseC
...
+	if ( !pNeoMe || pNeoMe->GetClass() != NEO_CLASS_RECON )
+	{
+		return;
+	}
+
+	// Check that bot isn't only moving sideways which wastes aux power
+	// Also determines a direction to jump towards
+	// NEO Jank: We don't check sprint here because bots don't anticipate using sprint in a smart manner
+	if ( ( pNeoMe->m_nButtons & ( IN_FORWARD | IN_BACK ) ) == 0 )
```
Rainyan, line 382: Nit: this doesn't handle the edge case where a bot holds both forward and back keys at the same time. For a human player, this would either be a no-op (when velocity 0), or AUX burning without an actual hop (when velocity >0). Kind of a finicky mechanic, but this can be used to set up a recon sprint jump without triggering the superjump, which is sometimes useful for specific jump lineups for human players where the superjump would overshoot.

Maybe instead, could use XOR of the AND bitmask to check for "A or B but not both"?
```cpp
if (!( (pNeoMe->m_nButtons & (IN_FORWARD | IN_BACK)) ^ (IN_FORWARD | IN_BACK) ))
{
	return;
}
```
or similar.

Edit: actually, that doesn't handle the case where `!(buttons & (IN_FORWARD|IN_BACK))`...
`(buttons & (IN_FORWARD | IN_BACK)) && (buttons ^ (IN_FORWARD | IN_BACK))`? Maybe there's some nicer way to express this. Well, anyway, the compiler should be able to optimize it into a single statement even if it's multiple checks.

Tested with:
```
neo_score_limit 99;
neo_bot_quota 32;
fps_max 30;
bot_class 0; // all recons
map ntre_oilstain_ctg;
host_timescale 10;

// Then join spectator team.
```
and placing that above example if-statement at the bottom of the function (for debug purposes only; I don't think this check should be moved down) with a breakpoint and waiting to hit the `return` line, where without the if-statement we would've fired a superjump despite the simultaneous `IN_FORWARD|IN_BACK` buttons.

It seems quite rare, but does occasionally trigger (took me about 10-20 full rounds fast-forwarded) to hit the breakpoint on a few tries.

src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
```diff
@@ -359,6 +367,133 @@ Vector CNEOBotMainAction::SelectTargetPoint( const INextBot *meBot, const CBaseC
...
+
+			if (flDist >= neo_bot_recon_superjump_min_dist.GetFloat())
+			{
+				bCanJump = true;
+				break;
+			}
+			else if (flDist < 0)
+			{
+				return; // Just in case of a bad value
+			}
```
Rainyan, line 476: Nit: might be prudent to check `IsFinite`, because typically a failure here would result in a NaN value, which could be positive or negative as far as IEEE 754 is concerned.

src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
sunzenshen: Have a cherry pick ready to address these comments in a separate PR.

sunzenshen: Addressed in #1899

src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
sunzenshen: Addressed in #1899

src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
sunzenshen: Addressed in #1899

PR #1786, opened 2026-03-05
Top Hud Replace profile avatars with class icons

### LINKED ISSUE #1779: When playing on NSF and setting cl_neo_hud_team_swap_sides to 0, you can see the health of Jinrai players
### Build Info

v27.1

### Description

When using the non-classic HUD and setting cl_neo_hud_team_swap_sides to 0 when playing on NSF, you can see the health of the Jinrai players

### To Reproduce

1. Disable classic squad list
2. Set cl_neo_hud_team_swap_sides to 0
3. Play on NSF

### Expected behavior

You will see the health of NSF players

### Actual behavior

You can see the health of Jinrai players, but Jinrai players cannot see NSF health

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

https://www.youtube.com/watch?v=xkbnng4GCuk

AdamTadeusz:
## Description

NeoAssets PR https://github.com/NeotokyoRebuild/neoAssets/pull/110

I decided against highlighting the objective player, and for the font Verdana works fine. I want to still do that optimization pass which would basically involve looping through each side multiple times to draw each part of the hud in one go to save on calls made to the gpu, but I need to setup a test to make sure this will actually make drawing the hud faster, I think the hud can be merged in as is

## Toolchain

- Windows MSVC VS2022

-fixes #1779

### DIFF
--- a/src/game/client/neo/ui/neo_hud_round_state.cpp
+++ b/src/game/client/neo/ui/neo_hud_round_state.cpp
@@ -458,7 +458,7 @@ void CNEOHud_RoundState::UpdateStateForNeoHudElementDraw()
 	m_ePrevRoundStatus = NEORules()->GetRoundStatus();
 }
 
-ConVar cl_neo_squad_hud_sort_players_by_class_alive_and_star("cl_neo_squad_hud_sort_players_by_class_alive_and_star", "1", FCVAR_NONE, "whether to sort the top element by squad, then within the squad by alive status and then within the two status by class", true, 0, true, 1);
+ConVar cl_neo_squad_hud_sort_players_by_class_alive_and_star("cl_neo_squad_hud_sort_players_by_class_alive_and_star", "1", FCVAR_ARCHIVE, "whether to sort the top element by squad, then within the squad by alive status and then within the two status by class", true, 0, true, 1);
 void CNEOHud_RoundState::DrawNeoHudElement()
 {
 	CheckActiveStar();


game/neo/scripts/kb_def.lst
```diff
@@ -38,12 +38,12 @@
...
-"F7" "joinstar 3"
-"F8" "joinstar 4"
-"F9" "joinstar 5"
-"F10" "joinstar 0"
+"F5" "joinstar 0"
+"F6" "joinstar 1"
+"F7" "joinstar 2"
+"F8" "joinstar 3"
+"F9" "joinstar 4"
+"F10" "joinstar 5"
```
Rainyan, line 46: Could we keep these binds parity for config compatibility, even if it means subtracting 1 at the command callback to index the array?

Because if a user comes with their OGNT config, they may get confused when the F5-F10 binds row no longer match their cfg file.

src/game/shared/neo/neo_enums.h
```diff
@@ -28,13 +28,13 @@ static constexpr int NEO_CLASS_ENUM_COUNT = NEO_CLASS__ENUM_COUNT;
 static constexpr int NEO_CLASS_LOADOUTABLE_COUNT = NEO_CLASS__LOADOUTABLE_COUNT;
 
 enum NeoStar {
-	STAR_NONE = 0,
```
Rainyan, line 31: Why do we have to change this from parity ordering?

src/game/client/neo/ui/neo_hud_round_state.cpp
```diff
@@ -520,94 +549,130 @@ void CNEOHud_RoundState::DrawNeoHudElement()
...
+			constexpr const int ALIVE_SHIFT = 1;
+			constexpr const int STAR_SHIFT = 3;
+			COMPILE_TIME_ASSERT(MAX_PLAYERS < (1 << INDEX_SHIFT));
+			COMPILE_TIME_ASSERT(NEO_CLASS_ENUM_COUNT < (1 << CLASS_SHIFT));
+			COMPILE_TIME_ASSERT(STAR__TOTAL < (1 << STAR_SHIFT));
+			COMPILE_TIME_ASSERT(INT_MAX > ((TEAM__TOTAL - FIRST_GAME_TEAM) + 1) << (INDEX_SHIFT + CLASS_SHIFT + ALIVE_SHIFT + STAR_SHIFT));
+
+			const int playerTeam = g_PR->GetTeam(m_nPlayerList[i].first);
+			m_nPlayerList[i].second = m_nPlayerList[i].first + 
+					((playerTeam - FIRST_GAME_TEAM) << (INDEX_SHIFT + CLASS_SHIFT + ALIVE_SHIFT + STAR_SHIFT));
```
Rainyan: I'm a bit bothered by not having any idea what's going on here without whipping out a bit calculator. Have you profiled to make sure this is worth micro-optimizing? Or maybe I'm just dumb and this is obvious.

AdamTadeusz: I might be dumb for not seeing a more readable way to do this. I want to sort players first by team, then by squad, then by alive/dead, then by class and finally by index (so people can't switch classes back and forth to reorder where they are within the class they wanted to be in originally).

(Edit) would you be happy with a comment showing which bits of the second value in each entry in m_npPlayerList are set by each of these attributes?

(Edit 2) it's probably also my fault that im not using a struct but instead a std::pair, where a struct would be more readable

src/game/shared/neo/neo_enums.h
AdamTadeusz: I wanted the sorting function in neo_hud_round_state in the top hud to place players with no star at the end of the list without having to check for STAR_NONE explicitly

(Edit) I didn't see any places in code that necessarily needed STAR_NONE to be at the beginning, or at least no places where having it at the beginning made the code more readable or more performant so I thought it was fine

Rainyan: I'm mostly worried about breaking compatibility with OGNT entprop defines https://github.com/softashell/sourcemod-nt-include/blob/cc51694b8e77e580c2a4d08f7dfd8f3f67509ed4/scripting/include/neotokyo.inc#L65-L70 which could be confusing when translating stuff over. But maybe it's fine.

src/game/client/neo/ui/neo_hud_round_state.cpp
Rainyan: > would you be happy with a comment showing which bits of the second value in each entry in m_npPlayerList are set by each of these attributes?

Yeah if you don't mind adding some brief comment above it, that would be nice.

I think it's indeed the std::pair syntax combined with lots of bit math that makes it a bit mysterious to parse when quickly reading through.

game/neo/scripts/kb_def.lst
Rainyan: Actually, if we're breaking compat with the NeoStar enum anyway, maybe this is fine to just leave incompatible as it is. Since it's not like NT;RE configs are 1:1 compatible with OGNT anymore to begin with.

src/game/client/neo/ui/neo_hud_round_state.cpp
Rainyan: Thanks, much clearer now

src/game/client/neo/ui/neo_hud_round_state.cpp
```diff
@@ -429,6 +458,7 @@ void CNEOHud_RoundState::UpdateStateForNeoHudElementDraw()
 	m_ePrevRoundStatus = NEORules()->GetRoundStatus();
 }
 
+ConVar cl_neo_squad_hud_sort_players_by_class_alive_and_star("cl_neo_squad_hud_sort_players_by_class_alive_and_star", "1", FCVAR_NONE, "whether to sort the top element by squad, then within the squad by alive status and then within the two status by class", true, 0, true, 1);
```
Rainyan: Perhaps FCVAR_ARCHIVE, so this preference gets stored for the user?

Rainyan (approved):
Verified this fixing #1779

Issue #1738, opened 2026-02-24
When dying near a friendly player using the ghost, you can (briefly) see the ghost markers during the death fade out


kassibuss:
### Build Info

v26

### Description

When you die next to a teammate who is using the ghost, the ghost markers are briefly shown to you during the death fade out. Does not carry over to when you're spectating your teammates

### To Reproduce

1. Die near a friendly who is holding the ghost
2. Observe

### Expected behavior

You should not the ghost markers from your teammate

### Actual behavior

You can see the ghost markers from your teammate

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

10

PR #1791, opened 2026-03-06
Don't show OBS ghost beacons whilst LIFE_DYING

Rainyan:
## Description
Tentative fix attempt for #1738

I'm unable to reproduce the bug so this would need confirmation.

## Toolchain
- Windows MSVC VS2022

## Linked Issues

- related #1738


Agiel (approved):
Wasn't able to test this with bots, but a reasonable theory Zwia had is that this bug is caused by the spectate target not being cleared when spawning (which I also noticed when working on #1767), in which case, this fix seems appropriate.

sunzenshen (approved):

PR #1795, opened 2026-03-07
Lone bots deploy detpack ambush at ghost

sunzenshen:
## Description
Consolidated bot ambush logic into its own behavior class and added the ability for bots to plant detpacks in such scenarios.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
@@ -68,7 +68,8 @@ ActionResult< CNEOBot >	CNEOBotCtgLoneWolf::Update( CNEOBot *me, float interval 
 	{
 		if ( pDetpackWeapon && !pDetpackWeapon->m_bThisDetpackHasBeenThrown && NEORules()->m_pGhost )
 		{
-			if ( me->GetAbsOrigin().DistToSqr( NEORules()->GetGhostPos() ) < Square(200.0f) )
+			const float fDetpackDeployPrepDistSq = Square( 200.0f );
+			if ( me->GetAbsOrigin().DistToSqr( NEORules()->GetGhostPos() ) < fDetpackDeployPrepDistSq )
 			{
 				return ChangeTo( new CNEOBotDetpackDeploy( NEORules()->GetGhostPos(), new CNEOBotCtgLoneWolfAmbush() ), "Moving to plant detpack" );
 			}
--- a/src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
@@ -68,9 +68,15 @@ ActionResult< CNEOBot >	CNEOBotCtgLoneWolf::Update( CNEOBot *me, float interval 
 	{
 		if ( pDetpackWeapon && !pDetpackWeapon->m_bThisDetpackHasBeenThrown && NEORules()->m_pGhost )
 		{
-			return ChangeTo( new CNEOBotDetpackDeploy( NEORules()->GetGhostPos(), new CNEOBotCtgLoneWolfAmbush() ), "Moving to plant detpack" );
+			if ( me->GetAbsOrigin().DistToSqr( NEORules()->GetGhostPos() ) < Square(200.0f) )
+			{
+				return ChangeTo( new CNEOBotDetpackDeploy( NEORules()->GetGhostPos(), new CNEOBotCtgLoneWolfAmbush() ), "Moving to plant detpack" );
+			}
 		}
-		return ChangeTo( new CNEOBotCtgLoneWolfAmbush(), "Waiting in ambush near ghost" );
+		else
+		{
+			return ChangeTo( new CNEOBotCtgLoneWolfAmbush(), "Waiting in ambush near ghost" );
+		}
 	}
 
 	return ConsiderGhostVisualCheck( me );
--- a/src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
@@ -7,6 +7,7 @@
 #include "bot/behavior/neo_bot_ctg_lone_wolf.h"
 #include "bot/behavior/neo_bot_ctg_lone_wolf_ambush.h"
 #include "bot/behavior/neo_bot_ctg_lone_wolf_seek.h"
+#include "bot/behavior/neo_bot_detpack_deploy.h"
 #include "bot/neo_bot_path_compute.h"
 #include "neo_gamerules.h"
 #include "neo_ghost_cap_point.h"
@@ -65,6 +66,10 @@ ActionResult< CNEOBot >	CNEOBotCtgLoneWolf::Update( CNEOBot *me, float interval 
 	const CNavArea *myArea = me->GetLastKnownArea();
 	if ( ghostArea && myArea && ghostArea->IsPotentiallyVisible( myArea ) )
 	{
+		if ( pDetpackWeapon && !pDetpackWeapon->m_bThisDetpackHasBeenThrown && NEORules()->m_pGhost )
+		{
+			return ChangeTo( new CNEOBotDetpackDeploy( NEORules()->GetGhostPos(), new CNEOBotCtgLoneWolfAmbush() ), "Moving to plant detpack" );
+		}
 		return ChangeTo( new CNEOBotCtgLoneWolfAmbush(), "Waiting in ambush near ghost" );
 	}
 
--- a/src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
@@ -18,7 +18,6 @@
 //---------------------------------------------------------------------------------------------
 ActionResult< CNEOBot >	CNEOBotCtgLoneWolf::OnStart( CNEOBot *me, Action< CNEOBot > *priorAction )
 {
-	m_repathTimer.Invalidate();
 	return Continue();
 }
 
@@ -110,11 +109,10 @@ ActionResult< CNEOBot > CNEOBotCtgLoneWolf::ConsiderGhostInterception( CNEOBot *
 	const Vector& vecInterceptGoal = NEORules()->GetGhostPos();
 	if ( vecInterceptGoal != CNEO_Player::VECTOR_INVALID_WAYPOINT )
 	{
-		if ( !m_repathTimer.HasStarted() || m_repathTimer.IsElapsed() )
+		if ( !m_path.IsValid() )
 		{
 			CNEOBotPathCompute( me, m_path, vecInterceptGoal, FASTEST_ROUTE );
 			m_path.Update( me );
-			m_repathTimer.Start( RandomFloat( 0.3f, 1.0f ) );
 		}
 		else
 		{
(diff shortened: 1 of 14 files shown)


src/game/server/neo/neo_player.cpp
```diff
@@ -2560,6 +2563,8 @@ void CNEO_Player::PickupObject( CBaseEntity *pObject,
 void CNEO_Player::PlayStepSound( Vector &vecOrigin,
 	surfacedata_t *psurface, float fvol, bool force )
 {
+	// Bots in ambush behavior pay more attention to player sounds
+	CSoundEnt::InsertSound(SOUND_PLAYER, GetAbsOrigin(), 300, 0.2, this);
```
sunzenshen: Honestly not sure if 300 is too big or small, but my perception this whole time is that footstep sounds are really loud.

src/game/server/neo/neo_player.cpp
```diff
@@ -1193,6 +1193,9 @@ void CNEO_Player::PlayCloakSound(bool removeLocalPlayer)
 		// effect lasts 0.5 seconds, but allow 200-300ms leeway with GetFogObscuredRatio cache window
 		m_botThermOpticCamoDisruptedTimer.Start(0.2f);
 	}
+
+	// Cloak activation is intentionally loud; bots investigate this as a combat sound.
+	CSoundEnt::InsertSound(SOUND_COMBAT, GetAbsOrigin(), 600, 0.2, this);
```
sunzenshen: Bots already react to gunshots, so I think this classification makes sense as bots should consider enemy cloak sounds as combat sounds as well.

src/game/shared/neo/weapons/weapon_detpack.cpp
```diff
@@ -304,6 +305,10 @@ void CWeaponDetpack::TossDetpack(CBasePlayer* pPlayer)
 	{
 		Assert(false);
 	}
+
+	// Notify bots after pressing keypad since reacting immediately to press fire is unforgiving
+	// especially since the sound clip has some silence at the beginning
+	CSoundEnt::InsertSound(SOUND_COMBAT, pPlayer->GetAbsOrigin(), SOUNDENT_VOLUME_EMPTY, 0.2, pPlayer, SOUNDENT_CHANNEL_WEAPON);
```
sunzenshen: This sound triggers after the keypad initiation sequence, which to be honest might be kind of late, but I didn't want to mess with delay or think timers.  Plus there's a comedic timing if one bot tries to set a detpack, and when they are done arming they get blown up by an earlier trap.

sunzenshen: The alternative was at the start of pressing the fire button, but that timing felt like the earlier trapping bot was psychic and triggered the remote immediately at the latter's intention.  That appearance may be exacerbated by an initial silence in the detpack arming sound.

Plus you could argue that a bot is waiting for the arming sound to end, as the other player may be moving while arming, and will be closest at their arming target at the end of the sound.

src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
```diff
@@ -417,7 +386,50 @@ EventDesiredResult< CNEOBot > CNEOBotCtgLoneWolf::OnMoveToFailure( CNEOBot *me,
...
 
+//---------------------------------------------------------------------------------------------
+Vector CNEOBotCtgLoneWolf::GetNearestCapturePoint( CNEOBot *me, bool bEnemyCapPoint )
+{
+	Vector vecBestCapPoint = CNEO_Player::VECTOR_INVALID_WAYPOINT;
+	float flNearestCapDistSq = FLT_MAX;
+	
+	if ( NEORules()->m_pGhostCaps.Count() > 0 )
+	{
+		const Vector& vecStart = me->IsCarryingGhost() ? me->GetAbsOrigin() : ( m_hGhost.Get() ? m_hGhost->GetAbsOrigin() : me->GetAbsOrigin() );
```
AdamTadeusz: I don't know how expensive m_hGhost.Get() is but the ghost's absorigin should be the same as the move parent's abs origin, so m_hGhost->GetAbsOrigin can be used if IsCarryingGhost, saving us from iterating through all the weapons. At any rate it would make the statement simpler, if can get ghost return the ghost origin otherwise return my origin

AdamTadeusz: Actually the ghosts' abs origin z component is higher, so if that messes up with the pathfinding algorithm then better to leave as is. Although could check if the ghost has a moveparent and then use the moveparent abs origin which would still save us from iterating through all the weapons

AdamTadeusz: also if m_hGhost is only stored here to get its position, might be better to just call NEORules()->GetGhostPos which also has a pointer to the ghost. (technically the ghost capture points can be captured with any ghost, but doing so while m_ighosterplayer isnt updated (nobody picks up the original ghost yet) crashes the game).

AdamTadeusz (commented):
generally fine, could probably drop the ghost handle and get the ghost position from neorules

AdamTadeusz (dismissed):

AdamTadeusz (dismissed):

sunzenshen:
Fixed merge conflict by accepting all new incoming files in CMakelists.txt

src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf_seek.cpp
```diff
@@ -133,6 +146,18 @@ ActionResult< CNEOBot >	CNEOBotCtgLoneWolfSeek::Update( CNEOBot *me, float inter
...
 
+	// Periodically look for better weapons
+	if ( ( !m_scavengeTimer.HasStarted() || m_scavengeTimer.IsElapsed() ) )
+	{
+		m_scavengeTimer.Start( RandomFloat( 3.0f, 6.0f ) );
+
+		CBaseEntity *pNearbyWeapon = FindNearestPrimaryWeapon( me, true, m_pIgnoredWeapons.get() );
+		if ( pNearbyWeapon )
+		{
+			return SuspendFor( new CNEOBotSeekWeapon( pNearbyWeapon, m_pIgnoredWeapons.get() ), "Scavenging for nearby weapon" );
```
sunzenshen, line 157: In order to resolve the merge conflict with #1800,  decided to accept this PR's proposed behaviors, discarding the lone wolf file changes from the weapon swapping PR.  Then I introduced again the periodic weapon search routine into the lone wolf seek behavior. (as part of this commit: https://github.com/NeotokyoRebuild/neo/pull/1795/changes/5b94e3c0c343bbe8fc166ede1a516994a414d124)  My reasoning is that I don't want to interrupt bots waiting in ambush or planting the detpack, especially since it's not likely for new weapons to be introduced to the world while waiting in those behavior states (the implication being that the bot needs to go into combat first before new weapons are free for the taking).

sunzenshen:
I decided to pare down the scope of the changes to just the bot exploration behavior around the ghost, into this PR: #2001

The idea is to review that enemy search behavior first, without having to worry about the detpack deploying and sound detection behavior.  If PR #2001 gets merged in, then I'll rebase and clean up this PR to review the detpack logic.

One can see what was excised from PR #1795 by looking at this commit: https://github.com/NeotokyoRebuild/neo/pull/2001/changes/88169cc3c790fcdd81d418c0b842c1534d9a4342

sunzenshen:
I pared down the scope of this review by removing a number of the sound detection changes that this PR previously had, to focus on just the detpack deployment and trigger behavior.

To test, one can force the bots all be Recon class as well as reducing the inter-round timers with the following script:
```
bot_class 0;
mp_chattime 2; 
```
and setting the number of bots to 2.  One can also drop in as a player and make some large noises around an enemy detpack, if the detpack avoidance behavior in the deploying behavior causes the bots not to reach inside the blast radius.

src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
```diff
@@ -57,6 +58,10 @@ ActionResult< CNEOBot >	CNEOBotCtgLoneWolf::Update( CNEOBot *me, float interval
 	const CNavArea *myArea = me->GetLastKnownArea();
 	if ( ghostArea && myArea && ghostArea->IsPotentiallyVisible( myArea ) )
```
AdamTadeusz, line 66: I was wondering if a distance check here would be good for large open maps, but even if a bot runs with a detpack in their hands and has the ignore_enemies flag, they will still react to and shoot enemies as they appear, so not an issue, and IsPotentiallyVisible looks to be a fairly cheap check since it at least doesn't do any tracelines.

AdamTadeusz (dismissed):
some weird behavior when there's no valid ambush position, but generally great

src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
```diff
@@ -57,7 +58,17 @@ ActionResult< CNEOBot >	CNEOBotCtgLoneWolf::Update( CNEOBot *me, float interval
...
 	{
-		return ChangeTo( new CNEOBotCtgLoneWolfAmbush(), "Waiting in ambush near ghost" );
+		if ( pDetpackWeapon && !pDetpackWeapon->m_bThisDetpackHasBeenThrown && NEORules()->m_pGhost )
+		{
+			if ( me->GetAbsOrigin().DistToSqr( NEORules()->GetGhostPos() ) < Square(200.0f) )
```
AdamTadeusz: worth giving this magic number a name?

AdamTadeusz (dismissed):
I notice that bots re-path very frequently when going to an ambush position, which often causes them to run back and forth in front of the ghost, is this intentional?

src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
sunzenshen: > I notice that bots re-path very frequently when going to an ambush position, which often causes them to run back and forth in front of the ghost, is this intentional?

For some time, I've suspected that I have misinterpreted the cases where short fused repath timers should have been used based on the original SDK code.  I decided to finally address that hunch in #2065 but I'll cherry pick the fix specific to this PR behavior.

AdamTadeusz (approved):

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
```diff
@@ -85,6 +89,124 @@ ActionResult< CNEOBot >	CNEOBotTacticalMonitor::OnStart( CNEOBot *me, Action< CN
...
+	if ( bShouldDetonate )
+	{
+		return SuspendFor( new CNEOBotDetpackTrigger(), "Triggering detpack!" );
+	}
+
+	return Continue();
+}
+
+
 #ifndef NEO // NEO TODO (Adam) Monitor the remote detpack
```
DESTROYGIRL, line 210: Could remove this block now

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
```diff
@@ -311,6 +433,12 @@ ActionResult< CNEOBot >	CNEOBotTacticalMonitor::Update( CNEOBot *me, float inter
 	}
 #endif
 
+	ActionResult< CNEOBot > detpackResult = MonitorArmedDetpack( me );
+	if ( detpackResult.IsRequestingChange() )
+	{
+		return detpackResult;
+	}
+
 #if 0 // NEO TODO (Adam) detonate remote detpacks
```
DESTROYGIRL, line 442: And this

DESTROYGIRL (approved):

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
sunzenshen: Will address these in an incoming follow up detpack PR.

sunzenshen: Deleted in https://github.com/NeotokyoRebuild/neo/pull/2078.

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
sunzenshen: Deleted in https://github.com/NeotokyoRebuild/neo/pull/2078.

PR #1800, opened 2026-03-08
Bots swap for more powerful weapons

sunzenshen:
## Description
Based on a bot's weapon selection preferences, bots will scavenge weapons if they are not available to the bot's current rank.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
@@ -24,6 +24,9 @@ CNEOBotCtgLoneWolf::CNEOBotCtgLoneWolf( void )
 }
 
 //---------------------------------------------------------------------------------------------
+CNEOBotCtgLoneWolf::~CNEOBotCtgLoneWolf() = default;
+
+//---------------------------------------------------------------------------------------------
 ActionResult< CNEOBot >	CNEOBotCtgLoneWolf::OnStart( CNEOBot *me, Action< CNEOBot > *priorAction )
 {
 	m_hGhost = nullptr;
--- a/src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
@@ -20,6 +20,7 @@ CNEOBotCtgLoneWolf::CNEOBotCtgLoneWolf( void )
 	m_bHasRetreatedFromGhost = false;
 	m_vecDropThreatPos = CNEO_Player::VECTOR_INVALID_WAYPOINT;
 	m_closestCapturePoint = CNEO_Player::VECTOR_INVALID_WAYPOINT;
+	m_pIgnoredWeapons = std::make_unique<CNEOIgnoredWeaponsCache>();
 }
 
 //---------------------------------------------------------------------------------------------
@@ -33,6 +34,8 @@ ActionResult< CNEOBot >	CNEOBotCtgLoneWolf::OnStart( CNEOBot *me, Action< CNEOBo
 	m_repathTimer.Invalidate();
 	m_stalemateTimer.Invalidate();
 	m_capPointUpdateTimer.Invalidate();
+	m_scavengeTimer.Invalidate();
+	m_pIgnoredWeapons->Reset();
 	m_vecDropThreatPos = CNEO_Player::VECTOR_INVALID_WAYPOINT;
 	m_closestCapturePoint = CNEO_Player::VECTOR_INVALID_WAYPOINT;
 	m_hPursueTarget = nullptr;
@@ -59,7 +62,7 @@ ActionResult< CNEOBot >	CNEOBotCtgLoneWolf::Update( CNEOBot *me, float interval 
 		// First, ensure we have a weapon.
 		if ( !me->Weapon_GetSlot( 0 ) )
 		{
-			return SuspendFor( new CNEOBotSeekWeapon(), "Scavenging for weapon to hunt threat" );
+			return SuspendFor( new CNEOBotSeekWeapon(nullptr, m_pIgnoredWeapons.get()), "Scavenging for weapon to hunt threat" );
 		}
 
 		// We have a weapon. Investigate the last known location.
@@ -204,22 +207,19 @@ ActionResult< CNEOBot >	CNEOBotCtgLoneWolf::Update( CNEOBot *me, float interval 
 		else
 		{
 			// Enemy is closer to goal (blocking us) or gaining on us.
-
-			// If we see a weapon nearby, drop the ghost and take it
-			CBaseEntity *pNearbyWeapon = FindNearestPrimaryWeapon( me->GetAbsOrigin(), true );
-			if ( pNearbyWeapon )
+			if ( m_scavengeTimer.IsElapsed() )
 			{
-				CBaseCombatWeapon *pGhostWep = me->Weapon_GetSlot( 0 );
-				if ( pGhostWep )
-				{
-					if ( me->GetActiveWeapon() != pGhostWep )
-					{
-						me->Weapon_Switch( pGhostWep );
-						return Continue();
-					}
+				m_scavengeTimer.Start( RandomFloat( 0.5f, 1.0f ) );
 
-					me->PressDropButton( 0.1f );
-					return ChangeTo( new CNEOBotSeekWeapon(), "Dropping ghost to scavenge nearby weapon" );
+				// If we see a weapon nearby, drop the ghost and take it
+				CBaseEntity *pNearbyWeapon = FindNearestPrimaryWeapon( me, true, m_pIgnoredWeapons.get() );
+				if ( pNearbyWeapon )
+				{
+					CBaseCombatWeapon *pGhostWep = me->Weapon_GetSlot( 0 );
+					if ( pGhostWep )
+					{
+						return SuspendFor( new CNEOBotSeekWeapon( pNearbyWeapon, m_pIgnoredWeapons.get() ), "Dropping ghost to scavenge nearby weapon" );
+					}
 				}
 			}
 
--- a/src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
@@ -51,6 +51,7 @@ ActionResult< CNEOBot >	CNEOBotCtgLoneWolf::OnStart( CNEOBot *me, Action< CNEOBo
 ActionResult< CNEOBot >	CNEOBotCtgLoneWolf::Update( CNEOBot *me, float interval )
 {
 	const CKnownEntity *threat = me->GetVisionInterface()->GetPrimaryKnownThreat( true );
+	CBaseCombatWeapon *pGhostWep = me->Weapon_GetSlot( 0 );
 
 	CBaseCombatWeapon *pWeapon = me->GetActiveWeapon();
 	if ( !threat && pWeapon )
@@ -63,7 +64,7 @@ ActionResult< CNEOBot >	CNEOBotCtgLoneWolf::Update( CNEOBot *me, float interval 
 	if ( m_bPursuingDropThreat )
 	{
 		// First, ensure we have a weapon.
-		if ( !me->Weapon_GetSlot( 0 ) )
+		if ( !pGhostWep )
 		{
 			return SuspendFor( new CNEOBotSeekWeapon(nullptr, m_pIgnoredWeapons.get()), "Scavenging for weapon to hunt threat" );
 		}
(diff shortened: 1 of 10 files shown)


src/game/server/neo/bot/behavior/neo_bot_seek_weapon.cpp
```diff
@@ -1,38 +1,207 @@
...
+	}
+
+	CNEOBaseCombatWeapon *pNeoWep = ToNEOWeapon( pPrimary );
+	if ( !pNeoWep )
+	{
+		return false;
+	}
+
+	const NEO_WEP_BITS_UNDERLYING_TYPE wepBits = pNeoWep->GetNeoWepBits();
+	return ( wepBits & ( NEO_WEP_BALC | NEO_WEP_SMAC ) ) != 0;
```
AdamTadeusz: technically the smac can be dropped in nt;re, and we have third person animations for most weapons (and use some other animations instead for weapons that the vip doesnt have animations for)

src/game/server/neo/bot/behavior/neo_bot_seek_weapon.cpp
```diff
@@ -1,38 +1,207 @@
...
 	CBaseEntity *pEntity = nullptr;
-	while ( ( pEntity = gEntList.FindEntityByClassnameWithin( pEntity, "weapon_*", searchOrigin, flSearchRadius ) ) )
+	while ( ( pEntity = gEntList.FindEntityByClassnameWithin( pEntity, "weapon_*", me->GetAbsOrigin(), flSearchRadius ) ) )
 	{
+		if ( pIgnoredWeapons && pIgnoredWeapons->Has( pEntity ) )
+		{
+			continue;
+		}
+
 		CBaseCombatWeapon *pWeapon = pEntity->MyCombatWeaponPointer();
```
AdamTadeusz, line 140: I think its a safe assumption at this point that all weapons derive from CNEOBaseCombatWeapon, this way you can check wep bits for a weapon of type ghost instead of doing a string comparison further down

AdamTadeusz (commented):
looks good, in two places can replace a string comparison between a weapon name and "weapon_ghost" with GetNeoWepBits() & NEO_WEP_GHOST if the pointers are cast to cneobasecombatweapons, which at this point should be a safe assumption

AdamTadeusz (dismissed):

AdamTadeusz (dismissed):

src/game/shared/neo/weapons/weapon_neobasecombatweapon.h
```diff
@@ -260,18 +260,4 @@ class CNEOBaseCombatWeapon : public CBaseHL2MPCombatWeapon
 
 };
 
-inline CNEOBaseCombatWeapon *ToNEOWeapon( CBaseEntity *pEntity )
```
sunzenshen: Removing ToNEOWeapon as per this feedback in another review:
https://github.com/NeotokyoRebuild/neo/pull/1805#discussion_r3361863785

src/game/server/neo/bot/behavior/neo_bot_ctg_lone_wolf.cpp
```diff
@@ -204,22 +210,19 @@ ActionResult< CNEOBot >	CNEOBotCtgLoneWolf::Update( CNEOBot *me, float interval
...
-				CBaseCombatWeapon *pGhostWep = me->Weapon_GetSlot( 0 );
-				if ( pGhostWep )
+				m_scavengeTimer.Start( RandomFloat( 0.5f, 1.0f ) );
+
+				// If we see a weapon nearby, drop the ghost and take it
+				CBaseEntity *pNearbyWeapon = FindNearestPrimaryWeapon( me, true, m_pIgnoredWeapons.get() );
+				if ( pNearbyWeapon )
 				{
-					if ( me->GetActiveWeapon() != pGhostWep )
+					CBaseCombatWeapon *pGhostWep = me->Weapon_GetSlot( 0 );
```
Rainyan: Is there a case where this `pGhostWep` result could differ from the one in line 66? Assuming not, perhaps we could cache somewhere around line 65 for reuse, so we don't need the double call to `->Weapon_GetSlot` for this code path? Also because this `else` path of the `if` condition at line 66 already implies `me->Weapon_GetSlot( 0 ) == true`, could probably simplify away the `if ( pGhostWep )` check at line 222 as redundant.

src/game/server/neo/bot/behavior/neo_bot_seek_weapon.cpp
```diff
@@ -1,38 +1,195 @@
...
+	auto *pNeoWep = assert_cast<CNEOBaseCombatWeapon *>( pPrimary );
+	const NEO_WEP_BITS_UNDERLYING_TYPE wepBits = pNeoWep->GetNeoWepBits();
+	return ( wepBits & NEO_WEP_BALC ) != 0;
+}
+
+//---------------------------------------------------------------------------------------------
+int GetBotWeaponPreferenceRank( CNEOBot *me, NEO_WEP_BITS_UNDERLYING_TYPE wepBit )
+{
+	// Strip tag bits to get the primary identity bit
+	const NEO_WEP_BITS_UNDERLYING_TYPE identityBit = wepBit & ~WEP_TAG_BITS_MASK;
```
Rainyan: What's the purpose of specifically filtering the generic bit flags away from the result here, rather than letting the bitwise AND at line 43 handle it? Or are there some specific bot profiles that need this?

src/game/server/neo/bot/behavior/neo_bot_seek_weapon.cpp
```diff
@@ -1,38 +1,195 @@
...
+		// We have no primary weapon at all, anything is an upgrade
+		return true;
+	}
+
+	if ( !bHasReserveAmmo )
+	{
+		if ( pTargetWep->GetPrimaryAmmoCount() > 0 )
+		{
+			return true;
+		}
```
Rainyan: Would it make sense to `return ( pTargetWep->GetPrimaryAmmoCount() > 0 )` here rather than only return for the true case, since else we would be doing the rest of the gun logic on this function for a weapon with 0 ammo? I.e. could be doing more harm than good for the bot to loot guns that they can't shoot. Although maybe it's more realistic for the bot to naively pick up empty guns, since that's something a real player might also do? Dunno.

src/game/server/neo/bot/behavior/neo_bot_seek_weapon.cpp
```diff
@@ -1,38 +1,195 @@
...
+	}
+
+	const NEO_WEP_BITS_UNDERLYING_TYPE targetWepBits = pTargetWep->GetNeoWepBits();
```
Rainyan: Nit: purely opinion based but would personally prefer `auto targetWepBits` instead of this ridiculously long type name spelled out, for readability. I guess we should refactor that bits type to some more sensible length (`NeoWepBits`?) at some point...

src/game/server/neo/bot/behavior/neo_bot_seek_weapon.cpp
```diff
@@ -1,38 +1,195 @@
...
+
+	CBaseCombatWeapon *pPrimary = me->Weapon_GetSlot( 0 );
+	if ( pPrimary )
+	{
+		auto *pNeoPrimary = assert_cast<CNEOBaseCombatWeapon *>( pPrimary );
+		if ( pNeoPrimary->GetNeoWepBits() & NEO_WEP_BALC )
+		{
+			// can't switch these weapons
+			return nullptr;
+		}
```
Rainyan: Would `->CanBePickedUpByClass` make sense here, since we also have the case of `weapon_pz` possibly not being equippable by recons, or any such future development case of class-specific guns, rather than hard-coding this to just the BALC?

src/game/server/neo/bot/behavior/neo_bot_seek_weapon.cpp
```diff
@@ -1,38 +1,195 @@
...
+
+	int myPrefRank = BOT_WEP_PREF_RANK_EMPTY;
+	bool bHasReserveAmmo = false;
+	if ( pPrimary )
+	{
+		myPrefRank = BOT_WEP_PREF_RANK_UNPREFERRED;
+		bHasReserveAmmo = pPrimary->GetPrimaryAmmoCount() > 0;
+		auto *pMyNeoWep = assert_cast<CNEOBaseCombatWeapon *>( pPrimary );
+		myPrefRank = GetBotWeaponPreferenceRank( me, pMyNeoWep->GetNeoWepBits() );
+	}
```
Rainyan: Nit: maybe merge this `if` scope to the one at line 99, since the condition is the same?

src/game/server/neo/bot/behavior/neo_bot_seek_weapon.cpp
```diff
@@ -42,49 +199,66 @@ CBaseEntity *FindNearestPrimaryWeapon( const Vector &searchOrigin, bool bShortCi
...
+		m_hTargetWeapon = FindNearestPrimaryWeapon( me, false, m_pIgnoredWeapons );
+	}
+
+	if ( !m_hTargetWeapon )
+	{
+		return Done("No valid replacement primary found");
 	}
 
-	if ( !FindAndPathToWeapon(me) )
+	auto *pNeoPrimary = pPrimary ? assert_cast<CNEOBaseCombatWeapon *>( pPrimary ) : nullptr;
```
Rainyan: Nit: casting from type `A*` to a compatible type `B*` with the value `nullptr` is always valid, so this ternary can simplify to a regular assignment: `pNeoPrimary = assert_cast<CNEOBaseCombatWeapon *>( pPrimary );` without needing any explicit nullptr check.

The debug-mode `assert_cast` case will essentially compile into:
```cpp
Assert(fromValue == dynamic_cast<ToType*>(fromValue); // Assert(nullptr == nullptr);
return static_cast<ToType*>(fromValue); // return nullptr;
```
or similar, and this is perfectly fine even if `fromValue == nullptr` (assuming type compatibility).

src/game/server/neo/bot/behavior/neo_bot_seek_weapon.h
```diff
@@ -1,20 +1,37 @@
...
+	void Add( CBaseEntity *pWep ) { m_weapons.AddToTail( pWep ); }
+
+private:
+	CUtlVector<CHandle<CBaseEntity>> m_weapons;
+};
+
+// Returns the bot's preference rank for a weapon (-1 = not preferred, -2 = empty)
+int GetBotWeaponPreferenceRank( CNEOBot *me, NEO_WEP_BITS_UNDERLYING_TYPE wepBit );
+
+bool IsWeaponPreferenceUpgrade( CNEOBot *me, CNEOBaseCombatWeapon *pTargetWep, int myPrefRank, bool bHasReserveAmmo );
```
Rainyan: Could perhaps declare `CNEOBot *me` as `const CNEOBot *me` for these two functions, because the pointers' data is not mutated

src/game/server/neo/bot/behavior/neo_bot_seek_weapon.h
```diff
@@ -1,20 +1,37 @@
...
+
+bool IsUndroppablePrimary( CBaseCombatWeapon *pPrimary );
+
 // General utility that can be used to see if it makes sense to transition into this behavior
-CBaseEntity *FindNearestPrimaryWeapon( const Vector &searchOrigin, bool bShortCircuit = false );
+CBaseEntity *FindNearestPrimaryWeapon( CNEOBot *me, bool bAllowDropGhost = false, CNEOIgnoredWeaponsCache *pIgnoredWeapons = nullptr );
```
Rainyan: Could be declared `const CNEOBot *me`, `const CNEOIgnoredWeaponsCache *pIgnoredWeapons` for the non-mutated pointers

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
```diff
@@ -324,23 +336,27 @@ ActionResult< CNEOBot >	CNEOBotTacticalMonitor::Update( CNEOBot *me, float inter
...
 		return Continue();
 	}
 
-	if ( !m_maintainTimer.IsElapsed() )
+	// Avoid swapping weapon in the middle of a fight
+	CNEO_Player* pBotPlayer = ToNEOPlayer( me->GetEntity() );
+	if ( pBotPlayer && pBotPlayer->GetTimeSinceWeaponFired() < 3.0f )
 	{
 		return Continue();
 	}
```
Rainyan, line 348: Nit: as long as `me != nullptr`, this will dynamically dispatch as simply:

```diff
        // Avoid swapping weapon in the middle of a fight
-       CNEO_Player* pBotPlayer = ToNEOPlayer( me->GetEntity() );
-       if ( pBotPlayer && pBotPlayer->GetTimeSinceWeaponFired() < 3.0f )
+       if ( me->GetTimeSinceWeaponFired() < 3.0f )
        {
                return Continue();
        }
```
without requiring any conversion.

Or if `me` can be `nullptr`, then `if ( me && me->GetTimeSinceWeaponFired ... ` etc

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -1447,6 +1447,24 @@ bool CNEOBot::EquipRequiredWeapon(void)
...
+//-----------------------------------------------------------------------------------------------------
+void CNEOBot::DropPrimaryWeapon(void)
+{
+	CBaseCombatWeapon *pPrimary = Weapon_GetSlot( 0 );
+	if ( pPrimary )
+	{
+		if ( GetActiveWeapon() != pPrimary )
+		{
+			Weapon_Switch( pPrimary );
+		}
```
Rainyan, line 1459: Just to verify, is it expected for this path to simply switch to the gun and not actually do any drop action (presumably it will be called again for some future tick by the bot think routines which handles it)?

Rainyan (commented):
Looks good for the most part, but left some nits/questions about the code

src/game/server/neo/bot/behavior/neo_bot_seek_weapon.cpp
sunzenshen: Probably being overly cautious, the code works fine without this stripping.

src/game/server/neo/bot/behavior/neo_bot_seek_weapon.cpp
sunzenshen: For now I think just checking for any ammo is a nice middle ground since bots may still pick up a weapon that has low ammo.  I'm hoping that bots not picking up weapons when in combat helps reduces the chance that this seems like a dumb choice.

src/game/server/neo/bot/neo_bot.cpp
sunzenshen: Right, I'm waiting until the next tick to see if the primary weapon is  equipped before trying to drop it.

sunzenshen:
Waiting to see if 
https://github.com/NeotokyoRebuild/neo/pull/1984
makes a difference.

Rainyan (approved):

AdamTadeusz (approved):

PR #1802, opened 2026-03-08
Allow swim up and swim down controls to move the player up and down the ladder

AdamTadeusz:
## Description

Mostly for bots which struggle to move straight up a ladder. When using these controls the direction of the player's camera is ignored when determining which way to move when on a ladder.

## Toolchain

- Windows MSVC VS2022


sunzenshen (approved):
Tried this on a slightly tilted ladder and it feels alright as a human player.

Image

(This was on a branch with the bots trying (unsuccessfully) to use these binds, but that will need some more work to get working.  But that's out of scope for this PR.) [edit: ended up figuring out the bots in #1830]

I can't think of a reason why this would have a negative impact, so I'm all for it.

Rainyan (approved):

PR #1804, opened 2026-03-09
Separate crosshair for pistol+shotgun and hipfire + unit tests

### LINKED ISSUE #661: [FEATURE] Custom crosshair - Secondary different crosshair
### Description

Allow the ability to use a different crosshair for the secondary weapons (tachi, milso, kyla)

### Proposed Solution

* Toggle to enable/disable this feature
* If enabled, the UI will allow the player to toggle between tweaking primary or secondary crosshair settings

### Alternatives Considered

_No response_

### Additional Context

_No response_

nullsystem:
## Description
Removed hipfire convar and now part of crosshair bit-flag in serialization. Separate crosshair for secondary/pistols, shotguns and throwables/grenades, the others (SMGs, rifles, and machine guns) only for default crosshair. Knife, and snipers never gets the normal crosshair.

NeoUI added in ringbox-flags, similar to ringbox-bool but for flags rather than bools.

Serialization to/from string now saving bytes by omiting values same as their default counterpart and doing run-length encoding on empty segments.

Add further sanity checking, import limits to the defined bounds, validator also checks the bounds and utilized when `cl_neo_crosshair` set directly.

Added in unit testing, for crosshair and just-serialization functionalities.

DONE:
- [x] Compress non-default by default crosshair
- [x] Run-length encode multiple empty `;;`
- [x] NeoUI - Replace "reverse bool labels" with custom 2-size labels
- [x] Maybe more general enum for weapon types
- [x] Unit test serialization cross versions + compression, and some functions
- [x] Check longest possible serialization string

## Toolchain

- Linux GCC Distro Native Arch Linux/GCC 15

## Linked Issues

* fixes #661

### DIFF
--- a/src/tests/CMakeLists.txt
+++ b/src/tests/CMakeLists.txt
@@ -5,7 +5,7 @@
 #    $ ctest --test-dir build/[PRESET]/tests/ --output-on-failure -R [REGEX]
 # To directly run the test executables:
 #    Windows:
-#        Make sure tier0.dll is in PATH, then do:
+#        Make sure tier0.dll is in PATH ([SOURCE]\devtools\bin\), then do:
 #        $ build\windows-[debug/release]\tests\test_....
 #    Linux:
 #        $ LD_LIBRARY_PATH=lib/public/linux64/ ./build/linux-[debug/release]/tests/test_....
@@ -41,6 +41,12 @@ foreach (TEST_EXE ${TEST_EXE_LIST})
         set_property(TEST ${TEST_EXE}
             PROPERTY
                 ENVIRONMENT "LD_LIBRARY_PATH=${CMAKE_SOURCE_DIR}/lib/public/linux64/"
+        )
+    elseif (OS_WINDOWS)
+        set_property(TEST ${TEST_EXE}
+            APPEND
+            PROPERTY
+                ENVIRONMENT "PATH=${CMAKE_SOURCE_DIR}\\devtools\\bin\\"
         )
     endif ()
 


src/game/shared/neo/neo_serial.cpp
```diff
@@ -0,0 +1,222 @@
...
+	SERIALVARIANTTYPE_BOOL,
+};
+
+static void StrCatCh(char (&szMutStr)[NEO_XHAIR_SEQMAX], const char ch)
+{
+	const int iSize = V_strlen(szMutStr);
+	if (iSize < NEO_XHAIR_SEQMAX)
+	{
+		szMutStr[iSize] = ch;
+		szMutStr[iSize + 1] = '\0';
```
sunzenshen, line 30: Given zero indexing, could this write over the bounds of a NEO_XHAIR_SEQMAX sized array?

Maybe the condition here was supposed to be 

```cpp
if (iSize + < NEO_XHAIR_SEQMAX)
```
?

src/game/shared/neo/neo_crosshair.h
```diff
@@ -49,6 +54,84 @@ enum NeoHudCrosshairDynamicType
...
+	CROSSHAIR_WEP_SECONDARY,
+	CROSSHAIR_WEP_SHOTGUN,
+	CROSSHAIR_WEP_DEFAULT_HIPFIRE,
+	CROSSHAIR_WEP_SECONDARY_HIPFIRE,
+	CROSSHAIR_WEP_SHOTGUN_HIPFIRE,
+
+	CROSSHAIR_WEP__TOTAL,
+};
+
+static const ENeoCrosshairWep MAP_WEAPON_TYPE_TO_XHAIR[WEP_TYPE__TOTAL] = {
```
sunzenshen: Perhaps extern const would be more appropriate in this context?

e.g.

in neo_crosshair.h:
```cpp
extern const ENeoCrosshairWep MAP_WEAPON_TYPE_TO_XHAIR[WEP_TYPE__TOTAL];
```

and then in neo_crosshair.cpp:
```cpp
const ENeoCrosshairWep MAP_WEAPON_TYPE_TO_XHAIR[WEP_TYPE__TOTAL] = {
    CROSSHAIR_WEP_NONE,         // WEP_TYPE_NIL
    CROSSHAIR_WEP_NONE,         // WEP_TYPE_THROWABLE
    // etc
};
```

src/game/shared/neo/weapons/weapon_bits.h
```diff
@@ -70,6 +73,17 @@ enum NeoWepBits : NEO_WEP_BITS_UNDERLYING_TYPE
 #endif
 };
 
+#define DEFINE_CROSSHAIR_TYPE(_x1, _x2, _x3, weptype) weptype,
+#define DEFINE_CROSSHAIR_TYPE_ALT(_x1, _x2, _x3, _x4, weptype) weptype,
+
+static const ENeoWeaponType NEO_WEAPON_TYPE[NEO_WIDX__TOTAL] = {
```
sunzenshen: maybe this should be `extern const`?

e.g.

in the header:
```cpp
extern const ENeoWeaponType NEO_WEAPON_TYPE[NEO_WIDX__TOTAL];
```

and in cpp:

```cpp
const ENeoWeaponType NEO_WEAPON_TYPE[NEO_WIDX__TOTAL] = {
    WEP_TYPE_NIL, // 0 = invalid, set none
    // etc
};
```

src/game/shared/neo/neo_serial.cpp
nullsystem: Good catch

src/game/shared/neo/neo_serial.cpp
```diff
@@ -0,0 +1,222 @@
...
+			const char *pszCurSegment = szMutStr + iPrevSegment;
+			if (pszCurSegment[0] != '\0')
+			{
+				if (endCh == CH_XH_SEGEND) // Value
+				{
+					switch (eType)
+					{
+					case SERIALVARIANTTYPE_INT:
+					{
+						const int iInitVal = atoi(pszCurSegment);
```
sunzenshen: optional nit: maybe consider using V_atoi

src/game/shared/neo/neo_serial.cpp
```diff
@@ -0,0 +1,222 @@
...
+						var.iVal = (varMin.iVal < varMax.iVal) ? clamp(iInitVal, varMin.iVal, varMax.iVal) : iInitVal;
+					} break;
+					case SERIALVARIANTTYPE_BOOL:
+					{
+						var.bVal = (atoi(pszCurSegment) != 0);
+					} break;
+					case SERIALVARIANTTYPE_FLOAT:
+					{
+						const float flInitVal = static_cast<float>(atof(pszCurSegment));
```
sunzenshen: optional nit: maybe consider using V_atof

src/game/shared/neo/neo_serial.cpp
```diff
@@ -0,0 +1,222 @@
...
+		szMutStr[iSize + 1] = '\0';
+	}
+}
+
+static SerialVariant DeserialVariant(char (&szMutStr)[NEO_XHAIR_SEQMAX], const int iSzMutStrSize,
+		int *idx, const ESerialVariantType eType, const SerialVariant varDefault, int *piSkipIdx,
+		const SerialVariant varMin, const SerialVariant varMax)
+{
+	SerialVariant var = {};
+	V_memcpy(&var, &varDefault, sizeof(SerialVariant));
```
sunzenshen: Supposedly C++ can handle this copy itself using standard assignment, like this:

```cpp
SerialVariant var = varDefault;
```

src/game/shared/neo/neo_crosshair.cpp
```diff
@@ -167,219 +185,219 @@ void InitializeClNeoCrosshair()
...
 	crh->iCircleRad = 0;
 	crh->iCircleSegments = 0;
 	crh->eDynamicType = CROSSHAIR_DYNAMICTYPE_NONE;
-	crh->bSeparateColorDot = false;
 	crh->colorDot = COLOR_WHITE;
 	crh->colorDotOutline = COLOR_BLACK;
 	crh->colorOutline = COLOR_BLACK;
+	for (int i = 1; i < CROSSHAIR_WEP__TOTAL; ++i)
+	{
+		V_memcpy(&(xhairInfo->wep[i]), crh, sizeof(CrosshairWepInfo));
```
sunzenshen: Supposedly C++ can handle this copy itself using standard assignment, like this:

```cpp
xhairInfo->wep[i] = *crh;
```

src/game/shared/neo/neo_crosshair.cpp
```diff
@@ -167,219 +185,219 @@ void InitializeClNeoCrosshair()
...
+	}
+	if (paeHipfireOpts)
+	{
+		V_memset(*paeHipfireOpts, 0, sizeof(EHipfireOpt) * CROSSHAIR_WEP__TOTAL);
```
sunzenshen, line 248: While this one maybe not since the size of the memory being copied is more manual.

src/game/shared/neo/neo_crosshair.cpp
```diff
@@ -391,14 +439,61 @@ bool ImportCrosshair(CrosshairInfo *crh, const char *pszSequence)
...
+{
+	const int iSeqSize = V_strlen(pszSequence);
+	if (iSeqSize <= 0 || iSeqSize > NEO_XHAIR_SEQMAX)
+	{
+		return false;
+	}
+
+	char szMutSeq[NEO_XHAIR_SEQMAX];
+	V_memcpy(szMutSeq, pszSequence, sizeof(char) * iSeqSize);
+	szMutSeq[iSeqSize] = '\0';
```
sunzenshen: Maybe could use V_strncpy to handle the `\0` null termination. e.g.

```cpp
char szMutSeq[NEO_XHAIR_SEQMAX];
V_strncpy(szMutSeq, pszSequence, sizeof(szMutSeq));
```

src/game/shared/neo/neo_crosshair.cpp
```diff
@@ -436,8 +436,7 @@ bool ValidateCrosshairSerial(const char *pszSequence)
 	}
 
 	char szMutSeq[NEO_XHAIR_SEQMAX];
-	V_memcpy(szMutSeq, pszSequence, sizeof(char) * iSeqSize);
-	szMutSeq[iSeqSize] = '\0';
+	V_strcpy_safe(szMutSeq, pszSequence);
```
sunzenshen, line 444: If we're confident the source string is always terminated, this could work.
(Ideally people are sharing strings that work. And trying to extend the string to be longer didn't seem to have adverse effect with displaying the clipboard fail.)

Image

Otherwise we could be more explicit with the bounds:

```cpp
const int iSafeSize = Min(iSeqSize, (int)sizeof(szMutSeq) - 1);
V_strncpy(szMutSeq, pszSequence, iSafeSize + 1);
```

nullsystem: `V_strcpy_safe` is basically equivalent to `V_strncpy(szMutSeq, pszSequence, NEO_XHAIR_SEQMAX)` in this case and should give assurance that it handles the source + dest string properly null terminated. `iSeqSize` check before also handles longer strings issue so I think the `V_strcpy_safe` change is fine.

sunzenshen (dismissed):

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -1591,168 +1617,171 @@ void Tabs(const wchar_t **wszLabelsList, const int iLabelsSize, int *iIndex,
...
-	const int iMWheelJump = iTabWide * 0.2;
-	switch (c->eMode)
-	{
-	case MODE_PAINT:
-	{
-		vgui::surface()->SetFullscreenViewport(c->rWidgetArea.x0, c->rWidgetArea.y0, c->irWidgetWide, c->irWidgetTall);
-		vgui::surface()->PushFullscreenViewport();
+		else
+		{
+			iTabWide = (c->dPanel.wide / iLabelsSize);
```
Rainyan, line 1645: Maybe we could `Assert(iLabelsSize) > 0` or similar here, just to make sure we catch the div-by-zero UB if `iLabelsSize` ever happened to be zero by some future caller.

src/game/shared/neo/weapons/neo_weapon_types.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+// NEO NOTE (nullsystem): If this is altered and may need to alter
+// for crosshair, check MAP_WEAPON_TYPE_TO_XHAIR in neo_crosshair.h
+enum ENeoWeaponType
```
Rainyan: Perhaps we could state the enum-base explicitly here as `enum ENeoWeaponType : int`, since we often cast between int <--> enum in the code, to make it clear that these are equivalent types.

src/game/shared/neo/neo_crosshair.cpp
```diff
@@ -388,17 +441,62 @@ bool ImportCrosshair(CrosshairInfo *crh, const char *pszSequence)
...
+	ResetCrosshairToDefault(&xhairInfo, nullptr);
+
+	return ImportOrExportCrosshair(SERIALMODE_CHECK, &xhairInfo, szMutSeq, iSeqSize, NEOXHAIR_SERIAL_CURRENT);
+}
+
+bool ImportCrosshair(CrosshairInfo *xhairInfo, const char *pszSequence,
+		EHipfireOpt (*paeHipfireOpts)[CROSSHAIR_WEP__TOTAL])
+{
+	const int iSeqSize = V_strlen(pszSequence);
+	if (iSeqSize <= 0 || iSeqSize > NEO_XHAIR_SEQMAX)
```
Rainyan: Maybe the upper bound check should be `>= NEO_XHAIR_SEQMAX`, because we allocate `char szMutSeq[NEO_XHAIR_SEQMAX];`, and AFAIK `V_strcpy_safe` would reserve the last character index for the null terminator. So because `strlen` is the length without the null terminator, a string with length `NEO_XHAIR_SEQMAX` would technically be 1 over the limit (I think?)

Or, if we did indeed want to support `NEO_XHAIR_SEQMAX` non-nullterminating chars, then perhaps we should be allocating `NEO_XHAIR_SEQMAX+1` for the char array?

src/game/shared/neo/neo_crosshair.cpp
```diff
@@ -388,17 +441,62 @@ bool ImportCrosshair(CrosshairInfo *crh, const char *pszSequence)
...
+	}
+
+	if (paeHipfireOpts)
+	{
+		for (int i = CROSSHAIR_WEP_DEFAULT_HIPFIRE; i < CROSSHAIR_WEP__TOTAL; ++i)
+		{
+			const NeoCrosshairWepFlags wepMask = (1 << (i - 1));
+			const NeoCrosshairHipfireCustomFlags wepHipfireMask = (1 << (i - CROSSHAIR_WEP_DEFAULT_HIPFIRE));
+			// NeoCrosshairHipfireCustomFlags flag equiv. only sets when NeoCrosshairWepFlags flag equiv. is set
+			Assert(false == ((false == (xhairInfo->wepFlags & wepMask)) && (xhairInfo->hipfireFlags & wepHipfireMask)));
```
Rainyan: Nit: The double yoda condition is a bit difficult to parse, so could we maybe flip the outer `false` into something like:
```cpp
Assert((xhairInfo->wepFlags & wepMask) || (false == (xhairInfo->hipfireFlags & wepHipfireMask)));
```

So instead of:
If not (not A) and B
we had:
If A or not B

src/game/shared/neo/neo_crosshair.h
```diff
@@ -49,6 +54,78 @@ enum NeoHudCrosshairDynamicType
 	CROSSHAIR_DYNAMICTYPE__TOTAL,
 };
 
+enum ENeoCrosshairWep
```
Rainyan: Maybe `enum ENeoCrosshairWep : int` to make it explicit that we intend to do these typecasts

src/game/shared/neo/neo_serial.cpp
```diff
@@ -0,0 +1,231 @@
...
+};
+
+enum ESerialVariantType
+{
+	SERIALVARIANTTYPE_INT = 0,
+	SERIALVARIANTTYPE_FLOAT,
+	SERIALVARIANTTYPE_BOOL,
+};
+
+static void StrCatCh(char (&szMutStr)[NEO_XHAIR_SEQMAX], const char ch)
```
Rainyan: Nit: Perhaps it would be prudent to return `bool` of whether the concat was successful from this function? Just so it can't hide the error state of someone calling `StrCatCh` on a string where there is no room for such a concatenation on the output string.

src/game/shared/neo/neo_serial.cpp
```diff
@@ -0,0 +1,231 @@
...
+	if (ctx->eSerialMode == SERIALMODE_SERIALIZE)
+	{
+		if (eCompMode == COMPMODE_EQUALS && iVal == iCompVal)
+		{
+			StrCatCh(szMutStr, CH_XH_SEGEND);
+		}
+		else
+		{
+			char szTmp[NEO_XHAIR_SEQMAX];
+			V_sprintf_safe(szTmp, "%d%c", (iMin < iMax) ? clamp(iVal, iMin, iMax) : iVal, CH_XH_SEGEND);
```
Rainyan, line 123: Maybe `Assert(iMin <= iMax)` for the Serial funcs to catch things going weird?

src/game/shared/neo/neo_serial.cpp
```diff
@@ -0,0 +1,231 @@
...
+	int iOffset = 0;
+	const int iSzMutSeqSize = V_strlen(szMutSeq);
+
+	while ((pszToRLE = V_strstr(szMutSeq + iOffset, SZ_XH_MINSEGCOMPRESS)))
+	{
+		// NEO NOTE (nullsystem): ";;;;" is already 3 segments (after the first),
+		// so start counting from the character after that (in index 4)
+		// It's also the minimal to take use of RLE as ";;;;" -> ";3^" saving 1 byte,
+		// greater saving the greater the repeats.
+		int iLen = 3;
```
Rainyan: Maybe we should grab this value from `sizeof(SZ_XH_MINSEGCOMPRESS)-2` so it's coupled to that initial value, instead of hardcoding? Or Assert the assumptions on `SZ_XH_MINSEGCOMPRESS` contents so this breaks with a bang if that initial value is ever changed, to make sure such an error is caught

src/game/shared/neo/neo_serial.h
```diff
@@ -0,0 +1,44 @@
+#pragma once
+
+// Requires neo_crosshair.h included before this for NEO_XHAIR_SEQMAX
```
Rainyan: Is there a reason not to include `neo_crosshair.h` here instead of doing this transitive dependency, since `#pragma once` should handle the re-include case cleanly?

src/tests/CMakeLists.txt
```diff
@@ -0,0 +1,171 @@
+
+# To run all the unit tests:
+#    $ ctest --test-dir build/[PRESET]/tests/ --output-on-failure
+# To run unit tests executable matching regex:
+#    $ ctest --test-dir build/[PRESET]/tests/ --output-on-failure -R [REGEX]
+# To directly run the test executables:
+#    Windows:
+#        $ build\windows-[debug/release]\tests\test_....
+#    Linux:
+#        $ LD_LIBRARY_PATH=lib/public/linux64/ ./build/linux-[debug/release]/tests/test_....
```
Rainyan, line 11: At least on Windows, this would error with:

```
---------------------------
test_neo_crosshair.exe - System Error
---------------------------
The code execution cannot proceed because tier0.dll was not found. Reinstalling the program may fix this problem. 
---------------------------
OK   
---------------------------
```

until putting SDK 2013 tier0.dll dir to PATH. Maybe this should be documented in the comment here?

nullsystem:
@Rainyan Fixed up textured crosshair hipfire issue, code review changes, and also added a PATH envvar append for Windows ctest + PATH mention so even ctest unit test from Windows should work fine now. Commits: https://github.com/NeotokyoRebuild/neo/pull/1804/changes/f1c75bc3b9bf8e64747ce7994ac5477f42330a7f and https://github.com/NeotokyoRebuild/neo/pull/1804/changes/092c9b8c798941d546c38e0ac9f5d35912b4f78d

Rainyan (approved):

PR #1805, opened 2026-03-09
Bots apply nav pathing penalties based on weapon limits

sunzenshen:
## Description
Bots consider relative level of exposure of a NavArea when deciding paths through the NavMesh
- Bots with shotguns shy away from NavAreas that have many visibility connections
- Bots with pistol caliber weapons slightly avoid areas but to a lesser degree
- Bots with scoped/semi-auto weapons favor exposed NavAreas with long sightlines
- Generally, a minor penalty is applied for NavArea exposure for bots to avoid wide open exposed NavAreas, unless they have a scoped/battle rifle

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_ctg_enemy.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ctg_enemy.cpp
@@ -49,7 +49,8 @@ ActionResult< CNEOBot >	CNEOBotCtgEnemy::Update( CNEOBot *me, float interval )
 	// Investigate the ghost carrier's position
 	if ( m_repathTimer.IsElapsed() )
 	{
-		CNEOBotPathCompute( me, m_path, pGhostCarrier->GetAbsOrigin(), DEFAULT_ROUTE );
+		// FASTEST_ROUTE: don't waste chase time looking for cover
+		CNEOBotPathCompute( me, m_path, pGhostCarrier->GetAbsOrigin(), FASTEST_ROUTE );
 		m_repathTimer.Start( RandomFloat( 0.2f, 1.0f ) );
 	}
 	m_path.Update( me );
--- a/src/game/server/neo/bot/neo_bot_path_cost.cpp
+++ b/src/game/server/neo/bot/neo_bot_path_cost.cpp
@@ -5,17 +5,33 @@
 #include "neo_bot_locomotion.h"
 #include "nav_mesh.h"
 #include "neo_bot_path_reservation.h"
+#include "weapon_neobasecombatweapon.h"
 
 extern ConVar neo_bot_path_reservation_enable;
 
 ConVar neo_bot_path_around_friendly_cooldown("neo_bot_path_around_friendly_cooldown", "2.0", FCVAR_CHEAT,
 	"How often to check for friendly path dispersion", false, 0, false, 60);
 
-ConVar neo_bot_path_penalty_jump_multiplier("neo_bot_path_penalty_jump_multiplier", "100.0", FCVAR_CHEAT,
+ConVar neo_bot_path_penalty_jump_multiplier("neo_bot_path_penalty_jump_multiplier", "1000.0", FCVAR_CHEAT,
 	"Maximum penalty multiplier for jump height changes in pathfinding", false, 0.01f, false, 1000.0f);
 
 ConVar neo_bot_path_penalty_ladder_multiplier("neo_bot_path_penalty_ladder_multiplier", "3.0", FCVAR_CHEAT,
 	"Penalty multiplier for ladder traversal in pathfinding", true, 0.1f, true, 100.0f);
+
+ConVar neo_bot_path_penalty_exposure_base("neo_bot_path_penalty_exposure_base", "5.0", FCVAR_CHEAT,
+	"General additional penalty per visible area for bots to avoid exposed areas", true, 0.0f, false, 0.0f);
+
+ConVar neo_bot_path_penalty_exposure_pistol("neo_bot_path_penalty_exposure_pistol", "10.0", FCVAR_CHEAT,
+	"Additional penalty per visible area for bots wielding pistol caliber weapons", true, 0.0f, false, 0.0f);
+
+ConVar neo_bot_path_penalty_exposure_shotgun("neo_bot_path_penalty_exposure_shotgun", "20.0", FCVAR_CHEAT,
+	"Additional penalty per visible area for shotgun-wielding bots", true, 0.0f, false, 0.0f);
+
+ConVar neo_bot_path_penalty_exposure_inverse_base_battle_rifle("neo_bot_path_penalty_exposure_inverse_base_battle_rifle", "500.0", FCVAR_CHEAT,
+	"Base penalty for calculating inverse traversal penalty for semi-auto battle rifles", true, 1.0f, false, 0.0f);
+
+ConVar neo_bot_path_penalty_exposure_inverse_base_scoped("neo_bot_path_penalty_exposure_inverse_base_scoped", "1000.0", FCVAR_CHEAT,
+	"Base penalty for calculating inverse traversal penalty for scoped weapons", true, 1.0f, false, 0.0f);
 
 //-------------------------------------------------------------------------------------------------
 CNEOBotPathCost::CNEOBotPathCost(CNEOBot* me, RouteType routeType)
@@ -125,6 +141,50 @@ float CNEOBotPathCost::operator()(CNavArea* baseArea, CNavArea* fromArea, const 
 		{
 			cost += CNEOBotPathReservations()->GetPredictedFriendlyPathCount(area->GetID(), m_me->GetTeamNumber()) * neo_bot_path_reservation_penalty.GetFloat();
 			cost += CNEOBotPathReservations()->GetAreaAvoidPenalty(area->GetID());
+
+			// Weapon range penalties
+			CNEOBaseCombatWeapon* myWeapon = ToNEOWeapon(m_me->GetActiveWeapon());
+			const int nWeaponBits = myWeapon->GetNeoWepBits();
+			if (myWeapon && (nWeaponBits & NEO_WEP_FIREARM))
+			{
+				constexpr int nShotgunBits = NEO_WEP_AA13 | NEO_WEP_SUPA7;
+				constexpr int nBattleRifleBits = NEO_WEP_M41 | NEO_WEP_M41_S;
+				constexpr int nPistolCaliberBits = NEO_WEP_MILSO | NEO_WEP_TACHI | NEO_WEP_KYLA
+					| NEO_WEP_MPN | NEO_WEP_MPN_S | NEO_WEP_JITTE | NEO_WEP_JITTE_S | NEO_WEP_SRM | NEO_WEP_SRM_S;
+
+				const int visibleAreaCount = area->GetPotentiallyVisibleAreaCount();
+
+				if (nWeaponBits & nPistolCaliberBits)
+				{
+					// Weapons that don't have max first shot accuracy
+					const float exposurePenalty = neo_bot_path_penalty_exposure_pistol.GetFloat();
+					dist += visibleAreaCount * exposurePenalty;
+				}
+				else if (nWeaponBits & nShotgunBits)
+				{
+					// Weapons that have spread that can't hit long range targets
+					const float exposurePenalty = neo_bot_path_penalty_exposure_shotgun.GetFloat();
+					dist += visibleAreaCount * exposurePenalty;
+				}
+				else if (nWeaponBits & nBattleRifleBits)
+				{
+					// Weapons that benefit from medium sightlines that can see many NavAreas
+					const float baseline_penalty = neo_bot_path_penalty_exposure_inverse_base_battle_rifle.GetFloat();
+					dist += baseline_penalty / visibleAreaCount;
+				}
+				else if (nWeaponBits & NEO_WEP_SCOPEDWEAPON)
+				{
+					// Weapons that benefit from long sightlines that can see many NavAreas
+					const float baseline_penalty = neo_bot_path_penalty_exposure_inverse_base_scoped.GetFloat();
+					dist += baseline_penalty / visibleAreaCount;
+				}
+				else
+				{
+					// Generally avoiding exposed areas when traversing a wide open area
+					const float exposurePenalty = neo_bot_path_penalty_exposure_base.GetFloat();
+					dist += visibleAreaCount * exposurePenalty;
+				}
+			}
 
 			if (m_routeType == SAFEST_ROUTE)
 			{
--- a/src/game/server/neo/bot/neo_bot_path_cost.cpp
+++ b/src/game/server/neo/bot/neo_bot_path_cost.cpp
@@ -10,13 +10,13 @@
 extern ConVar neo_bot_path_reservation_enable;
 
 ConVar neo_bot_path_around_friendly_cooldown("neo_bot_path_around_friendly_cooldown", "2.0", FCVAR_CHEAT,
-	"How often to check for friendly path dispersion", false, 0, false, 60);
+	"How often to check for friendly path dispersion", true, 0, true, 60);
 
-ConVar neo_bot_path_penalty_jump_multiplier("neo_bot_path_penalty_jump_multiplier", "1000.0", FCVAR_CHEAT,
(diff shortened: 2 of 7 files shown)


src/game/server/neo/bot/neo_bot_path_cost.cpp
```diff
@@ -5,18 +5,34 @@
...
 #include "neo_bot_path_reservation.h"
+#include "weapon_neobasecombatweapon.h"
 
 extern ConVar neo_bot_path_reservation_enable;
 
 ConVar neo_bot_path_around_friendly_cooldown("neo_bot_path_around_friendly_cooldown", "2.0", FCVAR_CHEAT,
 	"How often to check for friendly path dispersion", false, 0, false, 60);
 
-ConVar neo_bot_path_penalty_jump_multiplier("neo_bot_path_penalty_jump_multiplier", "100.0", FCVAR_CHEAT,
+ConVar neo_bot_path_penalty_jump_multiplier("neo_bot_path_penalty_jump_multiplier", "1000.0", FCVAR_CHEAT,
```
sunzenshen: Now that NavArea pathing costs are slightly elevated, need to bump up the penalty for jump paths to match the relatively increased baseline penalty costs.

src/game/server/neo/bot/neo_bot_path_reservation.cpp
```diff
@@ -29,7 +29,7 @@ ConVar neo_bot_path_reservation_avoid_penalty_enable("neo_bot_path_reservation_a
 ConVar neo_bot_path_reservation_killed_penalty("neo_bot_path_reservation_killed_penalty", "10", FCVAR_NONE,
     "Path selection penalty added to a nav area each time a bot dies moving through that area.", true, 0, false, 0);
 
-ConVar neo_bot_path_reservation_onstuck_penalty("neo_bot_path_reservation_onstuck_penalty", "10000", FCVAR_NONE,
+ConVar neo_bot_path_reservation_onstuck_penalty("neo_bot_path_reservation_onstuck_penalty", "1000", FCVAR_NONE,
```
sunzenshen, line 32: When working on ladder climbing, I noticed that the previous penalty cost was so heavy that bots would fail to climb a challenging ladder a couple times and then never touch that ladder again for the rest of the match.  So I figured toning down this penalty was warranted.

AdamTadeusz (dismissed):

src/game/server/neo/bot/neo_bot_path_cost.cpp
```diff
@@ -126,6 +142,50 @@ float CNEOBotPathCost::operator()(CNavArea* baseArea, CNavArea* fromArea, const
 			cost += CNEOBotPathReservations()->GetPredictedFriendlyPathCount(area->GetID(), m_me->GetTeamNumber()) * neo_bot_path_reservation_penalty.GetFloat();
 			cost += CNEOBotPathReservations()->GetAreaAvoidPenalty(area->GetID());
 
+			// Weapon range penalties
+			CNEOBaseCombatWeapon* myWeapon = ToNEOWeapon(m_me->GetActiveWeapon());
+			const int nWeaponBits = myWeapon->GetNeoWepBits();
```
Rainyan: Maybe should verify `myWeapon` to be non-nullptr before dereferencing `->GetNeoWepBits`, for safety.

src/game/server/neo/bot/neo_bot_path_cost.cpp
```diff
@@ -126,6 +142,50 @@ float CNEOBotPathCost::operator()(CNavArea* baseArea, CNavArea* fromArea, const
...
+				constexpr int nPistolCaliberBits = NEO_WEP_MILSO | NEO_WEP_TACHI | NEO_WEP_KYLA
+					| NEO_WEP_MPN | NEO_WEP_MPN_S | NEO_WEP_JITTE | NEO_WEP_JITTE_S | NEO_WEP_SRM | NEO_WEP_SRM_S;
+
+				const int visibleAreaCount = area->GetPotentiallyVisibleAreaCount();
+
+				if (nWeaponBits & nPistolCaliberBits)
+				{
+					// Weapons that don't have max first shot accuracy
+					const float exposurePenalty = neo_bot_path_penalty_exposure_pistol.GetFloat();
+					dist += visibleAreaCount * exposurePenalty;
```
Rainyan: I wonder if it'd be more intuitive to increase the cost rather than the dist, since that's semantically what's happening(?) But I guess opinionated, so I don't really mind leaving this as-is.

src/game/server/neo/bot/neo_bot_path_cost.cpp
```diff
@@ -126,6 +142,50 @@ float CNEOBotPathCost::operator()(CNavArea* baseArea, CNavArea* fromArea, const
...
+			CNEOBaseCombatWeapon* myWeapon = ToNEOWeapon(m_me->GetActiveWeapon());
+			const int nWeaponBits = myWeapon->GetNeoWepBits();
+			if (myWeapon && (nWeaponBits & NEO_WEP_FIREARM))
+			{
+				constexpr int nShotgunBits = NEO_WEP_AA13 | NEO_WEP_SUPA7;
+				constexpr int nBattleRifleBits = NEO_WEP_M41 | NEO_WEP_M41_S;
+				constexpr int nPistolCaliberBits = NEO_WEP_MILSO | NEO_WEP_TACHI | NEO_WEP_KYLA
+					| NEO_WEP_MPN | NEO_WEP_MPN_S | NEO_WEP_JITTE | NEO_WEP_JITTE_S | NEO_WEP_SRM | NEO_WEP_SRM_S;
+
+				const int visibleAreaCount = area->GetPotentiallyVisibleAreaCount();
```
Rainyan: Would it make sense to early-return if there were 0 visible areas (to avoid the division by zero at lines like `dist += baseline_penalty / visibleAreaCount`, etc)? I guess it's not a big deal since float div by zero well-defined, but just wondering if we'd want to avoid the work for that case.

src/game/shared/neo/weapons/weapon_neobasecombatweapon.h
```diff
@@ -260,4 +260,14 @@ class CNEOBaseCombatWeapon : public CBaseHL2MPCombatWeapon
...
 };
 
+
+inline CNEOBaseCombatWeapon *ToNEOWeapon(CBaseCombatWeapon *pWeapon)
+{
+	if (!pWeapon)
+	{
+		return nullptr;
+	}
+	return dynamic_cast<CNEOBaseCombatWeapon*>(pWeapon);
```
Rainyan: `CNEOBaseCombatWeapon` inherits via `CBaseCombatWeapon` so I don't think we need to perform the runtime `dynamic_cast` cast here, so could replace with the compile-time `static_cast` (or `assert_cast` which is an asserted `dynamic_cast` for debug builds, but optimises to a `static_cast` for release builds, if wanting to be absolutely certain).

Likewise cast of a nullptr to a compatible type will always be valid, so the:
```cpp
if (!pWeapon)
{
	return nullptr;
}
```
can be removed: the `CBaseCombatWeapon* a = nullptr; static_cast<CNEOBaseCombatWeapon*>(a);` case will correctly handle the nullptr already.

Although after these changes we basically have a wrapped `static_cast`:
```cpp
inline CNEOBaseCombatWeapon *ToNEOWeapon(CBaseCombatWeapon *pWeapon)
{
	return static_cast<CNEOBaseCombatWeapon*>(pWeapon);
}
```
which IMO we could just use directly, and remove this helper method entirely since now we're basically obfuscating the one-liner cast operation inside the function implementation.

So essentially we could rewrite this `ToNEOWeapon` use case:
```cpp
// Weapon range penalties
CNEOBaseCombatWeapon* myWeapon = ToNEOWeapon(m_me->GetActiveWeapon());
```
as simply:
```cpp
auto* myWeapon = static_cast<CNEOBaseCombatWeapon*>(m_me->GetActiveWeapon());
```

Although perhaps somewhat opinionated whether the plain `static_cast` reads better than `ToNEOWeapon`, I guess I'm fine with either one. We do have the `ToNEOPlayer` already in this naming scheme so thematically it fits, although that one casts up from `CBaseEntity` which is less straigtforward if you're unsure of the type compatibility and warrants a multi-line implementation which would get clumsy to manually type for every such conversion. If we had a case where we need a bunch of `CBaseEntity*` -> `CNEOBaseCombatWeapon*` conversions then I could see the utility in having a helper func for that situation (`->IsBaseCombatWeapon` check as opposed to the `ToNEOPlayer`'s `->IsPlayer` check, etc). But whether there is such a use case currently, I'm not sure.

Rainyan (changes requested):

Rainyan:
@sunzenshen What was the console command to display those orange paths as shown in the screenshots above?

sunzenshen:
> @sunzenshen What was the console command to display those orange paths as shown in the screenshots above?

`nb_debug path`

src/game/server/neo/bot/behavior/neo_bot_ctg_enemy.cpp
```diff
@@ -49,7 +49,8 @@ ActionResult< CNEOBot >	CNEOBotCtgEnemy::Update( CNEOBot *me, float interval )
 	// Investigate the ghost carrier's position
 	if ( m_repathTimer.IsElapsed() )
 	{
-		CNEOBotPathCompute( me, m_path, pGhostCarrier->GetAbsOrigin(), DEFAULT_ROUTE );
+		// FASTEST_ROUTE: don't waste chase time looking for cover
+		CNEOBotPathCompute( me, m_path, pGhostCarrier->GetAbsOrigin(), FASTEST_ROUTE );
```
sunzenshen, line 53: With DEFAULT_ROUTE on this PR, bots don't chase ghost carriers fast enough as they sneak behind cover which wastes pursuit time.  I decided that it's better to risk bunching up while taking the FASTEST_ROUTE, and then let the dispatch to the Attack behavior above to use DEFAULT_ROUTE when actually attacking enemies.  So pursuers beeline as fast as possible to the ghoster until they contact the enemy, and then split up along different paths at that point.

src/game/server/neo/bot/neo_bot_path_cost.cpp
```diff
@@ -9,13 +9,28 @@
 extern ConVar neo_bot_path_reservation_enable;
 
 ConVar neo_bot_path_around_friendly_cooldown("neo_bot_path_around_friendly_cooldown", "2.0", FCVAR_CHEAT,
-	"How often to check for friendly path dispersion", false, 0, false, 60);
+	"How often to check for friendly path dispersion", true, 0, true, 60);
 
-ConVar neo_bot_path_penalty_jump_multiplier("neo_bot_path_penalty_jump_multiplier", "100.0", FCVAR_CHEAT,
-	"Maximum penalty multiplier for jump height changes in pathfinding", false, 0.01f, false, 1000.0f);
+ConVar neo_bot_path_penalty_jump_multiplier("neo_bot_path_penalty_jump_multiplier", "100000.0", FCVAR_CHEAT,
+	"Maximum penalty multiplier for jump height changes in pathfinding", true, 0.01f, false, 0.0f);
```
sunzenshen, line 15: Further increasing the jump penalty as I increased the path reservation penalty below.

src/game/server/neo/bot/neo_bot_path_cost.cpp
```diff
@@ -126,6 +141,55 @@ float CNEOBotPathCost::operator()(CNavArea* baseArea, CNavArea* fromArea, const
...
+						constexpr int nShotgunBits = NEO_WEP_AA13 | NEO_WEP_SUPA7;
+						constexpr int nBattleRifleBits = NEO_WEP_M41 | NEO_WEP_M41_S;
+						constexpr int nPistolCaliberBits = NEO_WEP_MILSO | NEO_WEP_TACHI | NEO_WEP_KYLA
+							| NEO_WEP_MPN | NEO_WEP_MPN_S | NEO_WEP_JITTE | NEO_WEP_JITTE_S | NEO_WEP_SRM | NEO_WEP_SRM_S;
+
+						if (nWeaponBits & nPistolCaliberBits)
+						{
+							// Weapons that don't have max first shot accuracy
+							const float exposurePenalty = neo_bot_path_penalty_exposure_pistol.GetFloat();
+							cost += visibleAreaCount * exposurePenalty;
```
sunzenshen, line 163: Changing the penalty addition to cost enhanced the effect of bots bunching up together behind cover, so I tweaked the other penalties to encourage bots to spread out if they are in the same room.

src/game/server/neo/bot/neo_bot_path_reservation.cpp
```diff
@@ -12,13 +12,13 @@ ConVar neo_bot_path_reservation_enable("neo_bot_path_reservation_enable", "1", F
...
 
-ConVar neo_bot_path_reservation_distance("neo_bot_path_reservation_distance", "10000", FCVAR_NONE,
-    "How far along the path to reserve, in Hammer units.", true, 0, true, 1000000);
+ConVar neo_bot_path_reservation_distance("neo_bot_path_reservation_distance", "100000", FCVAR_NONE,
+    "How far along the path to reserve, in Hammer units.", true, 0, false, 0);
 
-ConVar neo_bot_path_reservation_penalty("neo_bot_path_reservation_penalty", "100", FCVAR_NONE,
-    "Pathing cost penalty for a reserved area.", true, 0, true, 1000000);
+ConVar neo_bot_path_reservation_penalty("neo_bot_path_reservation_penalty", "10000", FCVAR_NONE,
+    "Pathing cost penalty for a reserved area.", true, 0, false, 0);
```
sunzenshen, line 21: Encourage bots to spread out in a room if they are traversing the same room as other bot teammates.  Otherwise bots bunch up together to take the same covered path.

sunzenshen (commented):
Changing the CNEOBotPathCost penalty from distance to cost for weapons enhanced the bot tendency to hide behind cover, which results in bunching up, so I further increased the path reservation penalties to encourage taking alternate paths or to dividing up a room.

This is the script I paste into the console for observation:

(Divided up into forcing TDM mode, speeding up the round resets, showing grenade paths, showing pathing, and speeding up the rounds 2x.)

```
sv_neo_gamemode_enforcement 3; sv_neo_gamemode_single 0; neo_vote_game_mode 0; mp_restartgame 1;

sv_neo_bot_grenade_debug_behavior 1; mp_chattime 2; sv_neo_preround_freeze_time 2; cl_neo_grenade_show_path 1; sv_neo_grenade_show_path 1;
sv_neo_pvs_cull_roaming_observer 0;

nb_debug_path;
host_timescale 2;
```

Rainyan (approved):
LGTM, cool to see the bots taking more nuanced routes!

src/game/server/neo/bot/neo_bot_path_cost.cpp
```diff
@@ -126,6 +141,55 @@ float CNEOBotPathCost::operator()(CNavArea* baseArea, CNavArea* fromArea, const
...
+				const int nWeaponBits = myWeapon->GetNeoWepBits();
+				if (nWeaponBits & NEO_WEP_FIREARM)
+				{
+					const int visibleAreaCount = area->GetPotentiallyVisibleAreaCount();
+					if (visibleAreaCount > 0)
+					{
+						constexpr int nShotgunBits = NEO_WEP_AA13 | NEO_WEP_SUPA7;
+						constexpr int nBattleRifleBits = NEO_WEP_M41 | NEO_WEP_M41_S;
+						constexpr int nPistolCaliberBits = NEO_WEP_MILSO | NEO_WEP_TACHI | NEO_WEP_KYLA
+							| NEO_WEP_MPN | NEO_WEP_MPN_S | NEO_WEP_JITTE | NEO_WEP_JITTE_S | NEO_WEP_SRM | NEO_WEP_SRM_S;
```
AdamTadeusz, line 157: NEO_WEP_BITS_UNDERLYING_TYPE is a long long int now, we have 31 bits left over there, if you want to define these all in one place

AdamTadeusz (approved):

src/game/server/neo/bot/neo_bot_path_cost.cpp
AdamTadeusz: tbh I wonder if maybe neowepbits should only contain attributes that can be shared between weapons, whereas weapon specific bits like NEO_WEP_AA13 should have a separate enum, but thats a conversation for another pr. Also would help with custom modded weapons in the future?

Issue #1663, opened 2026-02-10
Sniper scope zoom-in effects should be parity


Rainyan:
### Build Info

6b854555

### Description

When zooming in with a scoped weapon (SRS/ZR68L/M41L) in OGNT, there are currently two visual effects that are not parity, as described in the repro steps.

### To Reproduce

* Scope in with the SRS (or other sniper; `host_timescale` can be used to slow down the animation)

### Expected behavior

These were tested with the SRS.

* 1.) Before entering the zoomed-in scope view, expecting the viewmodel scope to be raised towards the camera for about 0.1 seconds (or 6 frames of a 60 fps recording).

* 2.) The FOV zoom-in happens instantly for the scope (after the 0.1 seconds vm move).

* 3.) The FOV zoom-out happens instantly (including for a quick-switch to a different gun whilst scoped in).

* 4.) The SRS muzzle flash is not visible from within the scope.

Example video at 1/10 speed:

https://github.com/user-attachments/assets/0675890c-b432-4733-9157-9f27bfe16e51

### Actual behavior

* 1.) The scope is entered instantly.

* 2.) The FOV zoom-in is lerped like with the un-scoped weapons.

* 3.) The FOV zoom-out is also lerped.

* 4.) The SRS muzzle flash is visible from within the scope.

https://github.com/user-attachments/assets/cd6b4e9b-1df6-44d8-8eb1-ce4749d78ecf

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

11

### Additional context and Screenshots

* Additionally, in OGNT, the hip-fire crosshair makes a brief appearance in the 0.1 seconds transition state of the vm scope moving but before we've entered the scope. We don't currently do this. But I imagine we might not want to replicate this one, since we already offer the hip-fire crosshair as an accessibility feature which should cover any use-cases for this pre-scope crosshair to line up shots etc.

* This bug report does *not* include the following bugs, which are tracked separately:
  * #1385 
  * #1644

PR #1809, opened 2026-03-09
Don't stay scoped while playing bolt animation

### LINKED ISSUE #1793: Scoping in with the srs should not be possible while the weapon is being racked, or for ~the same time as the weapon racking animation if the racking animation is cancelled
In OGNT, after firing the srs, the player is forced out of the scoped view while the bolt racking animation plays. Any attempt to scope in again while this animation plays is ignored.

In NT;RE, after firing the srs, the player is forced out of the scoped view, but they can immediately enter scoped view again while the racking animation plays.

Note: NT;RE also has a non-parity feature where the sniper can continue looking through their scope without racking the bolt after firing by continuing to hold the attack key. I would consider this issue resolved when the player can continue to do so, and is only forced out of scoping in after they release the attack key, thus begining to rack the weapon. Quick switching away from the srs should not reduce how long the player is unable to scope in.

Agiel:
## Description
We should probably move `IsAllowedToZoom` to NEOBaseCombatWeapon` and then override in the SRS class, but this is good enough for now. Basically just return false if weapon is SRS and we're playing the bolt animation.

## Toolchain

## Linked Issues

- fixes #1793

### DIFF
--- a/src/game/shared/neo/neo_player_shared.cpp
+++ b/src/game/shared/neo/neo_player_shared.cpp
@@ -58,6 +58,10 @@ bool IsAllowedToZoom(CNEOBaseCombatWeapon *pWep)
 		return false;
 	}
 
+	if (pWep->GetNeoWepBits() & NEO_WEP_SRS && pWep->GetIdealActivity() == ACT_VM_PULLBACK) {
+		return false;
+	}
+
 	// These weapons are not allowed to be zoomed in with.
 	const auto forbiddenZooms =
 		NEO_WEP_DETPACK |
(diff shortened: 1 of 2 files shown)


AdamTadeusz:
this doesn't fix quickswitching away from and back to the srs allowing immediate scope in, maybe change -fixes to -related or whatever that keyword is, or open another issue?

AdamTadeusz:
aha i see we're going for plugin quickswitching

disgusteeng

Agiel:
You're not quick switching properly, as you can tell by the bolt sound. You gotta switch while still holding the fire button. If you do you'll be able to scope immediately. That said, ognt probably has something like soonestScopeTime which is set when the bolt animation starts, while here I'm just looking at the current idea activity. Anyway, given that proper quick switching lets you zoom immediately, and that we have #1663, I think it's fine to close #1793 with this.

PR #1822, opened 2026-03-10
#1821 Everyone dying on last tick results in a tie

### LINKED ISSUE #1821: NSF always wins if everyone dies at the same time
### Build Info

V28

### Description

When dying at the same tick the round will always be awarded to NSF

### To Reproduce

Last people die at the same time
NSF Wins

### Expected behavior

Its a tie not NSF win

### Actual behavior

_No response_

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

VantoNortim:
## Description
Was tested on the latest build. NSF always won because they were set as winning team last as round winner is decided on the tick people died. Now if everyone is dead it's a tie.

## Toolchain

- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues
- fixes #1821

### DIFF
--- a/src/game/shared/neo/neo_gamerules.cpp
+++ b/src/game/shared/neo/neo_gamerules.cpp
@@ -1555,12 +1555,15 @@ void CNEORules::Think(void)
 		COMPILE_TIME_ASSERT(TEAM_JINRAI == 2 && TEAM_NSF == 3);
 		if (GetGameType() != NEO_GAME_TYPE_TDM && GetGameType() != NEO_GAME_TYPE_DM && GetGameType() != NEO_GAME_TYPE_JGR)
 		{
-			for (int team = TEAM_JINRAI; team <= TEAM_NSF; ++team)
-			{
-				if (GetGlobalTeam(team)->GetAliveMembers() == 0)
-				{
-					SetWinningTeam(GetOpposingTeam(team), NEO_VICTORY_TEAM_ELIMINATION, false, true, false, false);
-				}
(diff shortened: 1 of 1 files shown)


src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -1555,12 +1555,15 @@ void CNEORules::Think(void)
...
-			{
-				if (GetGlobalTeam(team)->GetAliveMembers() == 0)
-				{
-					SetWinningTeam(GetOpposingTeam(team), NEO_VICTORY_TEAM_ELIMINATION, false, true, false, false);
-				}
+			auto jinraiAlive = GetGlobalTeam(TEAM_JINRAI)->GetAliveMembers();
+			auto nsfAlive = GetGlobalTeam(TEAM_NSF)->GetAliveMembers();
+
+			if (jinraiAlive == 0 && nsfAlive == 0) {
+				SetWinningTeam(TEAM_SPECTATOR, NEO_VICTORY_STALEMATE, false, false, true, false);
```
Agiel, line 1562: bSwitchTeams seems to be passed in inconsistently, but I guess it doesn't matter since it's not used anyway? Something to think about if we revisit half-time later down the line 🤔

VantoNortim: Yeah I went back to look at why it's like that after opening the PR then realised it's not used. Probably should be removed for now even cause if it starts being used it will start breaking stuff lol

PR #1825, opened 2026-03-11
Use the correct team icon in the new squad hud for nsf players whos class we don't know

AdamTadeusz:


src/game/client/neo/ui/neo_hud_round_state.cpp
```diff
@@ -1085,7 +1085,7 @@ void CNEOHud_RoundState::DrawPlayer(int playerIndex, int teamIndex, const TeamLo
 			}
 		}
 
-		float textureXOffset = 0.f;
+		float textureXOffset = g_PR->GetTeam(playerIndex) == TEAM_JINRAI ? 0 : TEXTURE_WIDTH;
```
Agiel, line 1091: I would put this in the if-statement below so the logic flows better. Took me a few reads to understand what cases are covered by this one and the one above 😵‍💫.

Some ASCII art of the texture atlas would be even better 😄.

AdamTadeusz: looking at this again I think I confused myself as well, but im not making ascii art of the texture sorry xD. Anyhow the logic for the x offset can be simplified

Agiel: Doesn't have to be art, I meant more a table. But a comment would be enough, like "x axis is class and y axis is team" or whichever it is 😅. (I see you posted a screenshot which helps, but would be good to have it in the code).

AdamTadeusz: to clarify the comment i just added in that last commit was unrelated to this message

PR #1830, opened 2026-03-13
Enhance bot ladder climbing reliability

sunzenshen:
## Description
Addresses various ladder climbing failure scenarios that could cause bots not to be able to successfully climb ladders
- Leverage ladder climbing swim controls to decouple climbing with look direction
- Movement helpers to help bots align with ladder and dismount

## Toolchain
- Windows MSVC VS2022

## Dependencies
- https://github.com/NeotokyoRebuild/neo/pull/1802

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
@@ -188,7 +188,7 @@ ActionResult<CNEOBot> CNEOBotLadderClimb::Update( CNEOBot *me, float /*interval*
 	}
 
 	// Check if we're on the ladder (MOVETYPE_LADDER or locomotion says so)
-	bool onLadder = me->IsOnLadder();
+	bool onLadder = me->IsBotOnLadder();
 
 	if ( !onLadder )
 	{
@@ -201,7 +201,7 @@ ActionResult<CNEOBot> CNEOBotLadderClimb::Update( CNEOBot *me, float /*interval*
 		{
 			Vector ladderClosestPoint;
 			CalcClosestPointOnLineSegment( myPos, m_ladder->m_bottom, m_ladder->m_top, ladderClosestPoint );
-			if ( myPos.DistToSqr( ladderClosestPoint ) > Square( 100.0f ) )
+			if ( myPos.DistToSqr( ladderClosestPoint ) > Square( MAX_DEVIATION_DIST ) )
 			{
 				return Done( "Fallen too far from ladder, resetting" );
 			}
@@ -227,6 +227,7 @@ ActionResult<CNEOBot> CNEOBotLadderClimb::Update( CNEOBot *me, float /*interval*
 				if ( distToTarget < mover->GetStepHeight() * 2.0f )
 				{
 					EnterDismountPhase( me );
+					return Continue();
 				}
 				else
 				{
@@ -260,96 +261,60 @@ ActionResult<CNEOBot> CNEOBotLadderClimb::Update( CNEOBot *me, float /*interval*
 		if ( m_bGoingUp )
 		{
 			body->SetDesiredPosture( IBody::STAND );
+		}
 
-			float dismountZ = targetZ;
-
-			if ( m_pExitArea && bExitIsBehind )
+		float dismountZ = targetZ;
+		if ( m_pExitArea )
+		{
+			if ( m_bGoingUp )
 			{
-				dismountZ = MIN( m_exitAreaCenter.z, targetZ );
-			}
-
-			if ( currentZ >= dismountZ )
-			{
-				EnterDismountPhase( me );
-				return Continue();
-			}
-
-			bool bIsClimbingDown = false;
-			if ( m_pExitArea )
-			{
-				if ( currentZ < m_exitAreaCenter.z )
+				if ( bExitIsBehind )
 				{
-					me->PressMoveUpButton();
-				}
-				else
-				{
-					me->PressMoveDownButton();
-					bIsClimbingDown = true;
+					dismountZ = Min( m_exitAreaCenter.z, targetZ );
 				}
 			}
 			else
 			{
-				me->PressMoveUpButton();
+				// Allow early drop-off at intermediate floors
+				dismountZ = Max( m_exitAreaCenter.z, targetZ );
 			}
-
-			if ( !bIsClimbingDown )
-			{
-				me->PressForwardButton();
-			}
-
-			// Look at the dismount height, slightly behind the ladder
-			Vector lookTarget = m_ladder->GetPosAtHeight( dismountZ );
-			lookTarget -= m_ladder->GetNormal() * 50.0f;
-			body->AimHeadTowards( lookTarget, IBody::MANDATORY, 0.1f, nullptr, "Climbing up (looking at dismount position)" );
 		}
-		else
-		{
-			float dismountZ = targetZ;
 
+		// Check if we've reached the target exit height
+		if ( m_bGoingUp ? ( currentZ >= dismountZ ) : ( currentZ <= dismountZ ) )
+		{
+			EnterDismountPhase( me );
+			return Continue();
+		}
+
+		// Adjust vertical height based on target exit area
+		bool bIsClimbingDown = false;
+		if ( onLadder )
+		{
+			bool bShouldGoUp = m_bGoingUp;
 			if ( m_pExitArea )
 			{
-				// Allow early drop-off at intermediate floors
-				dismountZ = MAX( m_exitAreaCenter.z, targetZ );
+				bShouldGoUp = ( currentZ < m_exitAreaCenter.z );
 			}
 
-			// Check if we've reached the bottom or intermediate exit height
-			if ( currentZ <= dismountZ )
+			if ( bShouldGoUp )
 			{
-				EnterDismountPhase( me );
-				return Continue();
+				me->PressMoveUpButton();
 			}
-
-			// Adjust vertical height based on target exit area
-			bool bIsClimbingDown = false;
-			if ( m_pExitArea && onLadder )
-			{
-				if ( currentZ < m_exitAreaCenter.z )
-				{
-					me->PressMoveUpButton();
-				}
-				else
-				{
-					me->PressMoveDownButton();
-					bIsClimbingDown = true;
-				}
-			}
-			else if ( onLadder )
+			else
 			{
 				me->PressMoveDownButton();
 				bIsClimbingDown = true;
 			}
-
-			// Sanity check: don't press forward if we are actively climbing down
-			if ( !bIsClimbingDown )
-			{
-				me->PressForwardButton();
-			}
-
-			// Look at the dismount height, slightly behind the ladder
-			Vector lookTarget = m_ladder->GetPosAtHeight( dismountZ );
-			lookTarget -= m_ladder->GetNormal() * 50.0f;
-			body->AimHeadTowards( lookTarget, IBody::MANDATORY, 0.1f, nullptr, "Climbing down (looking at dismount position)" );
 		}
+
+
+		// Look at and move to the dismount height, slightly behind the ladder
+		Vector lookTarget = m_ladder->GetPosAtHeight( dismountZ );
+		lookTarget -= m_ladder->GetNormal() * 50.0f;
+		body->AimHeadTowards( lookTarget, IBody::MANDATORY, 0.1f, nullptr,
+			m_bGoingUp ? "Climbing up (looking at dismount position)" : "Climbing down (looking at dismount position)" );
+		me->PressForwardButton(0.1f);
 	}
 
 	//------------------------------------------------------------
@@ -372,9 +337,7 @@ ActionResult<CNEOBot> CNEOBotLadderClimb::Update( CNEOBot *me, float /*interval*
 		// Build look target toward exit area center with vertical bias preserved
 		if ( m_pExitArea )
 		{
-			Vector dir = m_exitAreaCenter - myPos;
-			Vector lookTarget = myPos + dir;
-
+			bool bDroppingEarly = ( myPos.z >= m_exitAreaCenter.z ); 
 			// Maintain Z-height while on the ladder in the dismount phase
 			if ( onLadder )
 			{
@@ -387,12 +350,15 @@ ActionResult<CNEOBot> CNEOBotLadderClimb::Update( CNEOBot *me, float /*interval*
 					me->PressMoveDownButton();
 				}
 			}
+			else
+			{
+				bDroppingEarly = true;
+			}
 
-			body->AimHeadTowards( lookTarget, IBody::MANDATORY, 0.1f, nullptr, "Walking to exit area" );
+			body->AimHeadTowards( m_exitAreaCenter, IBody::MANDATORY, 0.1f, nullptr, "Walking to exit area" );
 
 			// Jump to detach from ladder if exit is not straight ahead, or if we have reached the exit height
 			float dot = DotProduct( toExit, m_ladderForward );
-			bool bDroppingEarly = ( myPos.z >= m_exitAreaCenter.z ); 
 
 			if ( dot < 0.5f || bDroppingEarly )
 			{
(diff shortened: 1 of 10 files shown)


sunzenshen:
Script for commanding bots around the map to climb ladders:
```
nb_debug path;
sv_neo_bot_cmdr_enable 1;
neo_ctg_round_timelimit 99;
```

Related navmesh scenarios that demo some of the edge cases handled by this PR:
- https://github.com/NeotokyoRebuild/neoAssets/pull/116
- https://github.com/NeotokyoRebuild/neoAssets/pull/106

Also relies on PR #1802 to enhance the directional reliability of climbing ladders.

(We could wait for that PR to be merged first, and then we could rebase on top of that commit.)

src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
```diff
@@ -36,74 +100,378 @@ ActionResult<CNEOBot> CNEOBotLadderClimb::OnStart( CNEOBot *me, Action<CNEOBot>
...
+		if ( mover->IsOnGround() )
+		{
+			return Done( "Reached the ground, ending climb" );
+		}
+		
+		if ( !m_bDismountPhase && !m_bJumpedOffLadder )
+		{
+			Vector ladderClosestPoint;
+			CalcClosestPointOnLineSegment( myPos, m_ladder->m_bottom, m_ladder->m_top, ladderClosestPoint );
+			if ( myPos.DistToSqr( ladderClosestPoint ) > Square( 100.0f ) )
```
Rainyan: Is this 100 the `ALIGN_RANGE`?

src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
```diff
@@ -36,74 +100,378 @@ ActionResult<CNEOBot> CNEOBotLadderClimb::OnStart( CNEOBot *me, Action<CNEOBot>
...
+		if ( m_stuckTimer.IsElapsed() )
+		{
+			float verticalDelta = fabsf( currentZ - m_flLastZ );
+			if ( verticalDelta < STUCK_Z_TOLERANCE )
+			{
+				// No vertical progress - if we're close enough to the target, consider it done
+				float distToTarget = fabsf( currentZ - targetZ );
+				if ( distToTarget < mover->GetStepHeight() * 2.0f )
+				{
+					EnterDismountPhase( me );
```
Rainyan, line 234: Should we return `Continue` here like on line 255, or is it fine to for this logic to keep going after the `EnterDismountPhase` call?

src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
```diff
@@ -36,74 +100,378 @@ ActionResult<CNEOBot> CNEOBotLadderClimb::OnStart( CNEOBot *me, Action<CNEOBot>
...
+			if ( !bIsClimbingDown )
+			{
+				me->PressForwardButton();
+			}
+
+			// Look at the dismount height, slightly behind the ladder
+			Vector lookTarget = m_ladder->GetPosAtHeight( dismountZ );
+			lookTarget -= m_ladder->GetNormal() * 50.0f;
+			body->AimHeadTowards( lookTarget, IBody::MANDATORY, 0.1f, nullptr, "Climbing down (looking at dismount position)" );
+		}
```
Rainyan: Nit: Maybe easier to read if we merged the up/down paths, since it's mostly the same things except for some binary conditions? I did find myself scrolling back up a lot to remember if I'm looking at `m_bGoingUp` or `!m_bGoingUp` logic while reading it, but maybe I just need a bigger screen or more coffee.

I was playing around with compressing it into something like this, is about 30 loc less (would have to be verified for correctness though):

```diff
diff --git a/src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp b/src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
index b1a4714e..2c03633a 100644
--- a/src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
@@ -260,21 +260,27 @@ ActionResult<CNEOBot> CNEOBotLadderClimb::Update( CNEOBot *me, float /*interval*
 		if ( m_bGoingUp )
 		{
 			body->SetDesiredPosture( IBody::STAND );
+		}
 
-			float dismountZ = targetZ;
+		float dismountZ = targetZ;
 
-			if ( m_pExitArea && bExitIsBehind )
-			{
-				dismountZ = MIN( m_exitAreaCenter.z, targetZ );
-			}
+		if ( m_pExitArea && bExitIsBehind )
+		{
+			dismountZ = Min(
+				m_bGoingUp ? m_exitAreaCenter.z : targetZ,
+				m_bGoingUp ? targetZ : m_exitAreaCenter.z );
+		}
 
-			if ( currentZ >= dismountZ )
-			{
-				EnterDismountPhase( me );
-				return Continue();
-			}
+		if ( m_bGoingUp == ( currentZ >= dismountZ ) )
+		{
+			EnterDismountPhase( me );
+			return Continue();
+		}
 
-			bool bIsClimbingDown = false;
+		bool bIsClimbingDown = false;
+
+		if ( onLadder )
+		{
 			if ( m_pExitArea )
 			{
 				if ( currentZ < m_exitAreaCenter.z )
@@ -287,69 +293,33 @@ ActionResult<CNEOBot> CNEOBotLadderClimb::Update( CNEOBot *me, float /*interval*
 					bIsClimbingDown = true;
 				}
 			}
-			else
-			{
-				me->PressMoveUpButton();
-			}
-
-			if ( !bIsClimbingDown )
-			{
-				me->PressForwardButton();
-			}
-
-			// Look at the dismount height, slightly behind the ladder
-			Vector lookTarget = m_ladder->GetPosAtHeight( dismountZ );
-			lookTarget -= m_ladder->GetNormal() * 50.0f;
-			body->AimHeadTowards( lookTarget, IBody::MANDATORY, 0.1f, nullptr, "Climbing up (looking at dismount position)" );
 		}
 		else
 		{
-			float dismountZ = targetZ;
-
-			if ( m_pExitArea )
-			{
-				// Allow early drop-off at intermediate floors
-				dismountZ = MAX( m_exitAreaCenter.z, targetZ );
-			}
-
-			// Check if we've reached the bottom or intermediate exit height
-			if ( currentZ <= dismountZ )
+			if ( m_bGoingUp )
 			{
-				EnterDismountPhase( me );
-				return Continue();
-			}
-
-			// Adjust vertical height based on target exit area
-			bool bIsClimbingDown = false;
-			if ( m_pExitArea && onLadder )
-			{
-				if ( currentZ < m_exitAreaCenter.z )
-				{
-					me->PressMoveUpButton();
-				}
-				else
-				{
-					me->PressMoveDownButton();
-					bIsClimbingDown = true;
-				}
-			}
-			else if ( onLadder )
-			{
-				me->PressMoveDownButton();
+				me->PressMoveUpButton();
 				bIsClimbingDown = true;
 			}
-
-			// Sanity check: don't press forward if we are actively climbing down
-			if ( !bIsClimbingDown )
+			else
 			{
-				me->PressForwardButton();
+				me->PressMoveDownButton();
 			}
+		}
 
-			// Look at the dismount height, slightly behind the ladder
-			Vector lookTarget = m_ladder->GetPosAtHeight( dismountZ );
-			lookTarget -= m_ladder->GetNormal() * 50.0f;
-			body->AimHeadTowards( lookTarget, IBody::MANDATORY, 0.1f, nullptr, "Climbing down (looking at dismount position)" );
+		// Sanity check: don't press forward if we are actively climbing down
+		if ( !bIsClimbingDown )
+		{
+			me->PressForwardButton();
 		}
+
+		// Look at the dismount height, slightly behind the ladder
+		Vector lookTarget = m_ladder->GetPosAtHeight( dismountZ );
+		lookTarget -= m_ladder->GetNormal() * 50.0f;
+		body->AimHeadTowards( lookTarget, IBody::MANDATORY, 0.1f, nullptr,
+			m_bGoingUp
+			? "Climbing up (looking at dismount position)"
+			: "Climbing down (looking at dismount position)" );
 	}
 
 	//------------------------------------------------------------
```

But I guess ultimately subjective whether it's clearer to have the explicit "if...else" at the cost of some DRY, up to you.

src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
```diff
@@ -36,74 +100,378 @@ ActionResult<CNEOBot> CNEOBotLadderClimb::OnStart( CNEOBot *me, Action<CNEOBot>
...
+		if ( m_dismountTimer.IsElapsed() )
+		{
+			return Done( "Dismount walk timed out" );
+		}
+
+		// Build look target toward exit area center with vertical bias preserved
+		if ( m_pExitArea )
+		{
+			Vector dir = m_exitAreaCenter - myPos;
+			Vector lookTarget = myPos + dir;
```
Rainyan: Nit: this is the same as `lookTarget = m_exitAreaCenter`(I think?)

In which case this could simplify to:

```diff
diff --git a/src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp b/src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
index b1a4714e..ebfbd5c2 100644
--- a/src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
@@ -372,9 +372,6 @@ ActionResult<CNEOBot> CNEOBotLadderClimb::Update( CNEOBot *me, float /*interval*
                // Build look target toward exit area center with vertical bias preserved
                if ( m_pExitArea )
                {
-                       Vector dir = m_exitAreaCenter - myPos;
-                       Vector lookTarget = myPos + dir;
-
                        // Maintain Z-height while on the ladder in the dismount phase
                        if ( onLadder )
                        {
@@ -388,7 +385,7 @@ ActionResult<CNEOBot> CNEOBotLadderClimb::Update( CNEOBot *me, float /*interval*
                                }
                        }

-                       body->AimHeadTowards( lookTarget, IBody::MANDATORY, 0.1f, nullptr, "Walking to exit area" );
+                       body->AimHeadTowards( m_exitAreaCenter, IBody::MANDATORY, 0.1f, nullptr, "Walking to exit area" );

                        // Jump to detach from ladder if exit is not straight ahead, or if we have reached the exit height
                        float dot = DotProduct( toExit, m_ladderForward );
```

src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
```diff
@@ -36,74 +100,378 @@ ActionResult<CNEOBot> CNEOBotLadderClimb::OnStart( CNEOBot *me, Action<CNEOBot>
...
+					// Zero Z velocity to avoid additive momentum from climbing, reducing overshoots
+					Vector currentVel = me->GetAbsVelocity();
+					currentVel.z = 0.0f;
+					mover->Reset(); // clear velocity cache in locomotion interface
+					me->SetAbsVelocity( currentVel );
+
+					Vector jumpVelocity = toExit * 150.0f;
+					jumpVelocity.z = 150.0f;
+
+					me->ApplyAbsVelocityImpulse( jumpVelocity );
```
Rainyan: Nitpicking, but since applying an inverse velocity impulse to an axis is equivalent to zeroing it, I think this could simplify to something like below, to avoid calling `SetAbsVelocity` (it will be called as part of `ApplyAbsVelocityImpulse`). It also obviates the need for the `currentVel` temporary, since `GetAbsVelocity().z` can be retrieved inline and by reference. Although admittedly the 12 byte stack alloc is no big deal either way.

```diff
--- a/src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
@@ -402,13 +402,10 @@ ActionResult<CNEOBot> CNEOBotLadderClimb::Update( CNEOBot *me, float /*interval*
                                        me->PressJumpButton(); // mostly to trigger animation if possible

                                        // Zero Z velocity to avoid additive momentum from climbing, reducing overshoots
-                                       Vector currentVel = me->GetAbsVelocity();
-                                       currentVel.z = 0.0f;
                                        mover->Reset(); // clear velocity cache in locomotion interface
-                                       me->SetAbsVelocity( currentVel );

                                        Vector jumpVelocity = toExit * 150.0f;
-                                       jumpVelocity.z = 150.0f;
+                                       jumpVelocity.z = 150.0f - me->GetAbsVelocity().z;
```

src/game/server/neo/bot/neo_bot.h
```diff
@@ -172,6 +172,8 @@ class CNEOBot : public NextBotPlayer< CNEO_Player >, public CGameEventListener
 
 	bool IsEnemy(const CBaseEntity* them) const OVERRIDE;
 
+	bool IsOnLadder( ) const;
```
Rainyan: Not sure if this was intentional with this "override", but note that since it's not virtual, any call through a downcast pointer (`CBasePlayer*` etc) will resolve down the inheritance chain instead of up, so it'll call `CBasePlayer::IsOnLadder` and not this one:

```
>	server.dll!CBasePlayer::IsOnLadder() Line 2145	C++
 	server.dll!PlayerLocomotion::Approach(const Vector & pos, float goalWeight) Line 491	C++
 	server.dll!CNEOBotLocomotion::Approach(const Vector & pos, float goalWeight) Line 57	C++
 	server.dll!PathFollower::Update(INextBot * bot) Line 757	C++
```

Is this the intended behavior? If not, then I guess upcasting `PlayerLocomotion::m_player` to `CNEOBot*` could be an option (assuming it can never hold any other kind of `CBasePlayer*`, does feel a bit brittle though...)? Another alternative could be declaring the method virtual if we don't mind eating the slight dispatch overhead, and if we do in fact want dynamic dispatch for all of these `IsOnLadder` calls(?)

The other two places in the code base where I saw `IsOnLadder` being called were `IsLadderFreeFunctor::operator()` and `CMultiplayRules::PlayFootstepSounds`, but I don't think these two are used anywhere currently. And even if they were, I'd assume at least the gamerules one doesn't care about the `CNEOBot` implementation, unsure about `IsLadderFreeFunctor` though; presumably the `ignore` input at:

```cpp
IsLadderFreeFunctor( const CNavLadder *ladder, const CBasePlayer *ignore )
```

would be a  bot.

But maybe this is working as intended, I don't know so this is more of a question than a change request.

src/game/server/neo/bot/neo_bot_locomotion.cpp
```diff
@@ -37,7 +37,15 @@ void CNEOBotLocomotion::Update( void )
 	}
 	else
 	{
+#ifdef NEO
+		// Don't try to crouch jump if we are climbing a ladder
+		if (me->GetMoveType() != MOVETYPE_LADDER)
```
Rainyan: Should this be the new `me->IsOnLadder()`, or do we specifically just wanna check the move type here?

Rainyan (dismissed):
Looks good, the bots seem a lot smoother at navigating ladders now

src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
sunzenshen: I didn't want to #include neo_bot_ladder_approach and this is really more of a similar but independent threshold where I consider the ladder climbing is horribly off the path.  Decided to assign this as an independent constant.

src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
sunzenshen: I originally intended for the bot to immediately go into the dismount phase in the same tick, but after some testing it doesn't make a noticeable difference to return Continue() so I added that.

src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
sunzenshen: It added some side effects but in the end it was likely worth tracking down the edge cases rather than spaghetti'ing the logic into two branches.

src/game/server/neo/bot/neo_bot.h
sunzenshen: I was trying to mimic the IsOnLadder signature without overriding it, but also didn't consider the potential of downcasting. I decided it was simpler to go with IsBotOnLadder to make it clear that it should only be used for that case.

src/game/server/neo/bot/behavior/neo_bot_ladder_climb.cpp
```diff
@@ -36,74 +100,336 @@ ActionResult<CNEOBot> CNEOBotLadderClimb::OnStart( CNEOBot *me, Action<CNEOBot>
...
+		int safetyCounter = 0;
+		while ( seg && safetyCounter < 100 )
+		{
+			if ( !seg->ladder )
+			{
+				break;
+			}
+
+			seg = path->NextSegment( seg );
+			safetyCounter++;
```
Rainyan, line 134: The safety counter is a good idea, but perhaps we should Assert for not ever actually reaching it, since if that happens then there's probably something wrong.

Rainyan (dismissed):

Rainyan (approved):

AdamTadeusz (approved):

PR #1833, opened 2026-03-13
WIP: Auto-generate navmesh for brush ladders

Rainyan:
## Description
WIP, not ready for review.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #
- related #

### DIFF
--- a/src/game/server/nav.h
+++ b/src/game/server/nav.h
@@ -21,7 +21,7 @@
  */
 
 #ifdef NEO
-const float GenerationStepSize = 20.0f;			// NEO: some hallways/openings were missing connections with larger value
+constexpr float GenerationStepSize = 20.0f;			// NEO: some hallways/openings were missing connections with larger value
 #else
 const float GenerationStepSize = 25.0f;			// (30) was 20, but bots can't fit always fit
 #endif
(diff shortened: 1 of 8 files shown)


Rainyan:
So the basic case works. Here's a ntre\_dawn\_ctg ladder with `nav_generate_debug_brushladders` debug. Green line marking a good side, orange too slim a side for bots to climb, and pink a side with insufficient clearance or inside a wall. Up and down facing sides are also removed from consideration.

laddergen_debug

There's some issues still with more exotic orientations that need fixing. This ladder "works", but obviously the rotation of the generated nav node is incorrect.

orientation

Also some interesting unintended routes which I would not probably have thought of manually adding, but are totally valid. This is an alternate climb path generated from the left side of the fence ladder:

routes

AdamTadeusz:
> > I don't have a map to check, can we climb upside down ladders in source? I know of at least one source game that lets you do that https://youtu.be/bbpG3iptMLE?si=GjWWkg2AiXMrovdX&t=749
> 
> I will need to check but I think not. Would be kind of cool, although probably looks quite silly in third person without hand animations.
> 
> Up-facing ladders players can mount of course, and it's a mechanic for the ladder boosts etc. But it seems the bot ladder navigation system doesn't like that, so at least for now I'm choosing to skip those cases entirely.

I think that's perfectly valid

Rainyan:
Doing a cowboy rebase from browser since I don't have my dev setup at hand currently. But it should work fine as far as I can tell. If not, previous head 9165d79 should be known good but it doesn't have the convenient unity builds etc.

(Also pointing out there are memory bugs in this PR: particularly I could spot the `polyhedron` ptr being indirected after being freed. It's WIP code.)

PR #1840, opened 2026-03-15
Juggernaut bots can alt-fire grenades

sunzenshen:
## Description
Exposed an interface to the BALC that allows firing the alt fire grenades without having to hold down a key, the latter action being quite difficult for bots to do consistently. Also added retreat logic for Juggernaut bots when their gun overheats, now that the secondary fire expends significant heat.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
@@ -964,22 +964,7 @@ void CNEOBotMainAction::FireWeaponAtEnemy( CNEOBot *me )
 		{
 			if (myWeapon->GetNeoWepBits() & NEO_WEP_BALC)
 			{
-				auto *pBalc = static_cast<CWeaponBALC *>(myWeapon);
-				// Minimum viable firing BALC
-				// TODO: Proper heat management for higher difficulty bots
-				me->ReleaseWalkButton(); // NEO Jank: this actually cancels sprint
-
-				// NEO JANK: To simplify alt fire input management
-				// we allow the bot to bypass button-hold charge firing requirement
-				if ( (me->GetTimeSinceWeaponFired() >= pBalc->GetChargeDuration())
-					&& threatRange >= 300.0f )
-				{
-					pBalc->ShootGrenade(me);
-				}
-				else
-				{
-					me->PressFireButton(GetFireDurationByDifficulty(me));
-				}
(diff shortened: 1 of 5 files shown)


src/game/shared/neo/weapons/weapon_balc.cpp
```diff
@@ -137,59 +137,70 @@ void CWeaponBALC::PrimaryAttack(void)
...
+	m_bCharging = false;
+	m_bCharged = false;
 
-		//View kick
-		pPlayer->ViewPunchReset();
-		AddViewKick();
-	}
+	// View kick
+	pPlayer->ViewPunchReset();
+	AddViewKick();
```
DESTROYGIRL: Getting a lot of asserts from bots using the gun here, I don't mind just being lazy because we're already letting them cheat by skipping the wind up. Maybe its from firing the gun without "actually" having fired it through the route guns expect to be fired but whatever

```suggestion
	if (!pPlayer->IsBot())
	{
		pPlayer->ViewPunchReset();
		AddViewKick();
	}
```

DESTROYGIRL (changes requested):

src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
```diff
@@ -903,6 +901,33 @@ void CNEOBotMainAction::FireWeaponAtEnemy( CNEOBot *me )
...
+
+	// NEO JANK: To simplify alt fire input management
+	// we allow the bot to bypass button-hold charge firing requirement
+	// with approximation of charge fire delay
+	if ( threatRange > 300.0f && ( me->GetTimeSinceWeaponFired() > pBalc->GetChargeDuration() * 1.2f ) )
+	{
+		// Check if threat is clearly exposed for a shot
+		const CNavArea *meArea = me->GetLastKnownArea();
+		const CNavArea *targetArea = TheNavMesh->GetNearestNavArea( threat->GetEntity()->GetAbsOrigin() );
+		if ( meArea && targetArea && meArea->IsCompletelyVisible( targetArea ) )
```
sunzenshen, line 1054: Added a PVS visibility check so that bots are discouraged from launching grenades into areas that are either too elevated or there is occluding cover.

DESTROYGIRL (approved):

AdamTadeusz (approved):

PR #1841, opened 2026-03-15
Bots look out for grenades when retreating

sunzenshen:
## Description
Bot tactical monitor behaviors can prevent the other behaviors from triggering, and while that mostly has limited consequence, this resulted in bots ignoring grenades when running for a cover spot during their retreat behavior.  This PR introduces the same grenade avoidance logic from tactical monitor into the retreat to cover behavior, so that bots can visibly react to grenades when they were running for cover.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_retreat_from_grenade.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_retreat_from_grenade.cpp
@@ -226,7 +226,7 @@ ActionResult< CNEOBot >	CNEOBotRetreatFromGrenade::Update( CNEOBot *me, float in
 
 	if (!m_coverArea)
 	{
-		return SuspendFor(new CNEOBotRetreatToCover, "Reacting to contact instead");
+		return Done("Reacting to contact instead");
 	}
 
 	if ( me->GetLastKnownArea() != m_coverArea || !bIsExposed )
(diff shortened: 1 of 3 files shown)


sunzenshen:
Behavior before this PR:

https://youtu.be/AGuFlZUJt20?t=28

In the above demo, you can observe that the second grenaded bot doesn't visibly react to the incoming grenade.  This is because it is occupied with the retreat to cover behavior, which is preventing the bot from evaluating other behaviors.  In my mind so far, retreat is the only behavior that really needs an interruption to avoid grenades.

Let's look at the other behaviors launched from tactical monitor:
- CNEOBotRetreatFromGrenade would be potentially recursively redundant. I think it's fine for bots to be so preoccupied from dodging one grenade that they fail to react to another following in.  
- CNEOBotSeekWeapon is primarily a non-combat behavior, and in another PR I'm refining it to only seek for visible weapons.  I think it's believable that a player could be so fixated at rushing for a desired weapon which would blind them to a grenade threat.
- It's also believable to me that someone trying to climb a ladder with CNEOBotLadderApproach would be completely oblivious to an incoming grenade as well


CNEOBotRetreatToCover felt like the one that stuck out, because it can take some time for bots to reach their desired cover and it's the behavior most likely to experience heavy incoming fire. As a consequence, I think it makes sense to watch for grenades from this behavior at a high polling rate.

AdamTadeusz (approved):

Rainyan (approved):

PR #914, opened 2025-02-19
GH-#905 - Update to TF2 branch (source-sdk-2013 2025-02-18 update)

nullsystem:
## Description
Update NT;RE to TF2 branch:
https://www.teamfortress.com/post.php?id=238809
https://github.com/ValveSoftware/source-sdk-2013/commits/master/
https://store.steampowered.com/news/app/320/view/529838805680652975?l=english

New commits from `ValveSoftware/source-sdk-2013` from misyltoad:
* Remove sp singleplayer code.: `9ad9aeb83a20c973eb6bb6c2d56a45551ec901e8`
    * Ignored, we don't have singleplayer
* Move mp to the root directory: `44f8f18cd1fd837c9785cef85500766a6e1a16c0`
    * It's basically just no longer needing `mp` separation: `9303a26e61b6a967511cb4d723dad73921235724`
* Add Team Fortress 2 SDK: `0759e2e8e179d5352d81d0d4aaded72c1704b7a9`
    * Using cherry-pick and manually merge fixing multiple files: `7cb27ed066064386e377d80d468e89780537b0ca`

## TODO
- [x] Rough commit cherry-pick
- [x] Getting it working under cmake
- [x] Rechecking/fixing remaining highlighted in `TODO_MERGE.txt`
- [x] Linux building (64-bits) + usable with latest SDK update
- [x] Windows building (64-bits) + usable with latest SDK update
- [x] Actually seems to working more properly than initially
    - [x] Fix team/class HUD broken (unclickable, crashes)
         - Removing InstallMouseHandler fixes it, removing SetAutoDelete (dangling pointer causing crash I think)
    - [x] Fix corner HUD
        - `vgui/hud/800corner1` is replaced by `vgui/hud/8x800corner`
        - Set a fixed 8px for now
    - [x] Restore the basic bot at least (`bot_add`)
    - [x] ~~Fix 3d model animations~~ No longer only T-posing
        - tonysergi's `CHL2MPPlayerAnimState` vs 2025-TF2-SDK's `CPlayerAnimState`
        - **NOTE!!!** - I wouldn't say truely fixed, but at least much more way there than just T-posing (see thirdperson and looking at other bots)
    - [x] Fix sprinting/aux/movement
        - [x] Fix sprinting/aux drain
            - `hl2mp_player` class has changed especially around AUX and sprinting
            - `m_fIsSprinting` replaced by `m_HL2Local.m_bNewSprinting`
        - [x] Fix movement speed to NT instead of HL2 level
[...]

PR #1842, opened 2026-03-15
Fix demoui smoother buttons being unreactive

### LINKED ISSUE #1843: Demo UI smoother should work
### Build Info

v28.1-alpha

### Description

The `demoui` Smoother window which uses the legacy VGUI is currently broken due to some kind of incompatibility with NeoUI.

### To Reproduce

* Press Shift+F2 to open demoui
* Load any demo in the UI
* Pause the demo in the UI
* Press `Smooth...` in the UI
  * The Demo Smoother window should pop up
* In the Demo Smoother window, the `Edge->` or `Process->` buttons

### Expected behavior

The button opens the appropriate dropdown menu.

### Actual behavior

No dropdown menu opens.

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

11

Rainyan:
## Description
Fix the `demoui` demo smoother's dropdown buttons in the legacy VGUI not responding to mouse clicks, making the smoother unusable.

If this gets refactored in the future, then https://github.com/NeotokyoRebuild/neo/blob/36087ff754a684355bbf37d9eee9e41e0b8f88a7/src/game/client/mp3player.cpp#L889-L891 should probably also be revisited.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1843 
- related #947
- related #914

### DIFF
--- a/src/game/client/cdll_client_int.cpp
+++ b/src/game/client/cdll_client_int.cpp
@@ -151,6 +151,7 @@
 #endif
 
 #ifdef NEO
+#include <vgui_controls/Button.h>
 #include <vgui_controls/MenuButton.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


src/game/client/cdll_client_int.cpp
```diff
@@ -1943,6 +1946,106 @@ void CHLClient::DecodeUserCmdFromBuffer( bf_read& buf, int slot )
...
+		parent = child;
+
+		auto* panel = vgui::ipanel()->GetPanel(parent, "BaseUI");
+		if (!panel || !panel->IsVisible())
+		{
+			return;
+		}
+
+		childCount = vgui::ipanel()->GetChildCount(parent);
+		for (i = 0; i < childCount; ++i)
```
sunzenshen: Feels a little risky to be reusing the int i iterator in a nested loop like this.  Is it possible that these nested resets of the iterator could use different variables, like i, j, k, ...?  (unless I failed to grok the optimization reason for this reused i)

Then vgui::ipanel()->GetChild(parent, [i,j,k]) could be updated to these new names

Rainyan: Whenever we reach line 1989, it will always unconditionally break at line 2044 and therefore `i` cannot tamper with the loop above it (because it must break). The same thing happens at line 2000, where after that loop it'll unconditionally break at line 2042.

I'm reusing `i` because this will run in the `CHLClient::View_Render -> FixupDemoSmoother` which is a hot path that runs every frame (inside the demo viewer context, until we get the successful fix which occurs when the Smoother window is actually opened). For regular runs we bail out at:
```cpp
if (engine->IsPlayingDemo())
{
	FixupDemoSmoother();
}
```
so as to not do all of that vgui looping every frame for regular users.

Although I agree the nesting reads kind of poorly, so I'll see about refactoring it into something clearer.

sunzenshen: Makes sense!  I figured there was something I was missing in terms of a hot loop.

Rainyan: I did some refactoring to remove the nested loops and make it read a little better.

PR #1845, opened 2026-03-15
HLTV Changes for a smoother demo watching experience

### LINKED ISSUE #1844: As a HLTV observer I should not be able to see any crosshair when roaming and cl_observercrosshair is set to 0, and not my targets crosshair if cl_neo_crosshair_network is set to 1

AdamTadeusz:
## Description

Changes

- Don't draw target crosshair when roaming in hltv, always draw own crosshair when roaming and cl_observercrosshair is set
- Added drone movement to hltv
- Added an option to switch to a spectate target when switching to in eye or chase cam view, and..
- an option to change to either in eye or chase came view when changing the spectate target and not in one of the two
- Fixed most of the new spectator commands
- Fixed ghost beacons showing when roaming
- Fixed in-eye view offset
- Fixed camera fov getting stuck at a low value after spectating a dead player and going into free roam camera mode
- SourceTV is now not considered a spectator by neobotmanager (if there are any bots on the server and sourcetv but no players, neobotmanager will start kicking bots if neo_bot_join_after_player is set)
- Fixed the neo_fov and other missing convar value error

Things that I'm aware of that still need to be fixed (not in this PR)
- player animations being choppy or downright broken when doing a lot of skipping
- sometimes when one player shoots other players will light up like they're shooting even if they're not
- round end victory music does not play in hltv 
- hltvcamera can spectate observers, not only spectators (good?) but also dead players (probably bad)

## Toolchain

- Windows MSVC VS2022

## Linked Issues

- fixes #1844

### DIFF
--- a/src/game/shared/neo/neo_gamerules.cpp
+++ b/src/game/shared/neo/neo_gamerules.cpp
@@ -318,12 +318,12 @@ BEGIN_NETWORK_TABLE_NOBASE( CNEORules, DT_NEORules )
 	SendPropInt(SENDINFO(m_iJuggernautPlayerIndex), NumBitsForCount(MAX_PLAYERS_ARRAY_SAFE), SPROP_UNSIGNED),
 	SendPropBool(SENDINFO(m_bJuggernautItemExists)),
 	SendPropEHandle(SENDINFO(m_hJuggernaut)),
-	SendPropInt(SENDINFO(m_iLastHurt)),
-	SendPropInt(SENDINFO(m_iLastShooter)),
-	SendPropInt(SENDINFO(m_iLastEvent)),
-	SendPropInt(SENDINFO(m_iLastAttacker)),
-	SendPropInt(SENDINFO(m_iLastKiller)),
-	SendPropInt(SENDINFO(m_iLastGhoster)),
+	SendPropInt(SENDINFO(m_iLastHurt), NumBitsForCount(MAX_PLAYERS_ARRAY_SAFE), SPROP_UNSIGNED),
+	SendPropInt(SENDINFO(m_iLastShooter), NumBitsForCount(MAX_PLAYERS_ARRAY_SAFE), SPROP_UNSIGNED),
+	SendPropInt(SENDINFO(m_iLastEvent), NumBitsForCount(MAX_PLAYERS_ARRAY_SAFE), SPROP_UNSIGNED),
+	SendPropInt(SENDINFO(m_iLastAttacker), NumBitsForCount(MAX_PLAYERS_ARRAY_SAFE), SPROP_UNSIGNED),
+	SendPropInt(SENDINFO(m_iLastKiller), NumBitsForCount(MAX_PLAYERS_ARRAY_SAFE), SPROP_UNSIGNED),
+	SendPropInt(SENDINFO(m_iLastGhoster), NumBitsForCount(MAX_PLAYERS_ARRAY_SAFE), SPROP_UNSIGNED),
 #endif
 END_NETWORK_TABLE()
 


src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -4803,3 +4816,68 @@ void CNEORules::InitDefaultAIRelationships( void )
...
+	if (!pNeoPlayer || !pNeoPlayer->IsObserver())
+		return false;
+
+	return true;
+
+};
+
+CON_COMMAND_F( spec_last_hurt, "Spectate the last hurt player", FCVAR_CLIENTCMD_CAN_EXECUTE )
+{
+	if (!spectateChecks)
```
sunzenshen: Is it possible that this is just checking that the pointer to the lambda is not null, and since the lambda definition is defined above, this will always return true?

Did you mean to check the contents of spectateChecks like this?

```cpp
if(!spectateChecks())
```

src/game/client/game_controls/SpectatorGUI.cpp
```diff
@@ -986,48 +988,77 @@ CON_COMMAND_F( spec_player, "Spectate player by partial name, steamid, or userid
...
-			{
-				HLTVCamera()->SetPrimaryTarget(target->entindex());
-			}
-			else
+			ConMsg( "%s: HLTV Camera is PVS locked", __FUNCTION__ );
+			return;
+		}
+
+		// We have up to date information on all the players, just do it here
+		int fastestSpeedSquared = 0;
```
sunzenshen: Do we mind that when fastestSpeedSquared receives a float value, that the value is truncated?  Is  there a reason that this is not a float?

src/game/client/neo/ui/neo_hud_round_state.cpp
```diff
@@ -1411,12 +1419,18 @@ CON_COMMAND_F( spec_player_by_hud_position, "Spectate player by position in the
...
-		engine->ClientCmd( VarArgs("spec_player_entity_number %d", entityIndex) );
+		engine->IsHLTV() ? HLTVCamera()->SetPrimaryTarget(entityIndex) : engine->ClientCmd(VarArgs("spec_player_entity_number %d", entityIndex));
 	}
 }
 
 CON_COMMAND_F( spec_next_entity_in_hud, "Spectate next valid player to the right of the current spectate target", FCVAR_CLIENTCMD_CAN_EXECUTE )
 {
+	if (engine->IsHLTV() && HLTVCamera()->IsPVSLocked())
+	{
+		ConMsg( "%s: HLTV Camera is PVS locked", __FUNCTION__ );
```
Rainyan: Should suffix with `\n`. Same thing with some other places where this printout is used in the PR.

src/game/client/neo/ui/neo_hud_round_state.cpp
```diff
@@ -1457,12 +1477,18 @@ CON_COMMAND_F( spec_previous_entity_in_hud, "Spectate next valid player to the l
...
-		engine->ClientCmd(VarArgs("spec_player_entity_number %d", playerIndex));
+		engine->IsHLTV() ? HLTVCamera()->SetPrimaryTarget(playerIndex) : engine->ClientCmd(VarArgs("spec_player_entity_number %d", playerIndex));
 	}
 }
 
 CON_COMMAND_F( select_next_alive_player_in_hud, "Select the next alive player in the top hud", FCVAR_CLIENTCMD_CAN_EXECUTE )
 {
+	if (engine->IsHLTV() && HLTVCamera()->IsPVSLocked())
+	{
+		ConMsg( "%s: Selection is used to switch observer target in spectate_player_selected_in_hud, but HLTV Camera is PVS locked", __FUNCTION__ );
```
Rainyan: Maybe I misunderstand the error message, but how is `spectate_player_selected_in_hud` related to this happening? Same with the error message at line 1506.

src/game/client/hltvcamera.cpp
```diff
@@ -470,7 +509,11 @@ void C_HLTVCamera::CalcRoamingView(Vector& eyeOrigin, QAngle& eyeAngles, float&
 
 	eyeOrigin = m_vCamOrigin;
 	eyeAngles = m_aCamAngle;
+#ifdef NEO
+	fov = neo_fov.GetInt();
```
Rainyan: Technically the fov logic supports floats (and it's being stored as such here) so maybe `.GetFloat()`

src/game/client/hltvcamera.cpp
```diff
@@ -561,6 +608,14 @@ void C_HLTVCamera::SetMode(int iMode)
 	int iOldMode = m_nCameraMode;
 	m_nCameraMode = iMode;
 
+#ifdef NEO
+	if (cl_neo_hltvcamera_spectate_next_target_on_set_mode.GetBool() && (iMode == OBS_MODE_IN_EYE || iMode == OBS_MODE_CHASE) && !GetPrimaryTarget())
+	{
+		bAllowChangingModeWhenSettingPrimaryTarget = false;
+		SpecNextPlayer(false);
+		bAllowChangingModeWhenSettingPrimaryTarget = true;
```
Rainyan, line 616: What's the purpose of toggling `bAllowChangingModeWhenSettingPrimaryTarget` on/off here?

src/game/client/hltvcamera.cpp
```diff
@@ -571,6 +626,17 @@ void C_HLTVCamera::SetMode(int iMode)
...
+#ifdef NEO
+enum
+{
+	NEO_HLTV_ON_TARGET_MODE_NOTHING = 0,
+	NEO_HLTV_ON_TARGET_MODE_IN_EYE,
+	NEO_HLTV_ON_TARGET_MODE_CHASE,
+
+	NEO_HLTV_ON_TARGET_MODE__TOTAL = NEO_HLTV_ON_TARGET_MODE_CHASE
+};
+ConVar cl_neo_hltvcamera_default_mode_when_setting_primary_target("cl_neo_hltvcamera_default_mode_when_setting_primary_target", "1", FCVAR_ARCHIVE, "What to do if changing primary targets and not in in-eye or chase. 0 = do nothing, 1 = switch to in-eye, 2 = switch to chase", true, NEO_HLTV_ON_TARGET_MODE_NOTHING, true, NEO_HLTV_ON_TARGET_MODE__TOTAL);
```
Rainyan: Nit: maybe `0` for min value so we're not hardcoding `NEO_HLTV_ON_TARGET_MODE_NOTHING` as the min, in case the enum is ever reordered

src/game/client/hltvcamera.cpp
```diff
@@ -609,6 +693,37 @@ void C_HLTVCamera::SetPrimaryTarget( int nEntity )
...
+		case NEO_SPECTATE_EVENT_LAST_GHOSTER:
+			lastEventEntityIndex = NEORules()->GetLastGhoster();
+			break;
+		case NEO_SPECTATE_EVENT_LAST_EVENT:
+		default:
+			lastEventEntityIndex = NEORules()->GetLastEvent();
+			break;
+	}
+
+	if (lastEventEntityIndex > 0)
```
Rainyan: `>= 0` because `NEO_SPECTATE_EVENT_LAST_HURT == 0`

src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
```diff
@@ -957,7 +957,7 @@ void CNEOBaseCombatWeapon::PrimaryAttack(void)
 	}
 
 #ifdef GAME_DLL
-	pPlayer->SetLastShooter();
+	NEORules()->SetLastShooter(pPlayer->entindex());
```
Rainyan: This doesn't seem to catch Supa7 or AA13 shooting

src/game/shared/neo/neo_gamerules.h
```diff
@@ -435,6 +444,21 @@ class CNEORules : public CHL2MPRules, public CGameEventListener
...
+	void SetLastKiller(const int index) { m_iLastKiller = m_iLastEvent = index; }
+	void SetLastGhoster(const int index) { m_iLastGhoster = m_iLastEvent = index; }
+#endif // GAME_DLL
+public:
+	const int GetLastHurt() { return m_iLastHurt; }
+	const int GetLastShooter() { return m_iLastShooter; }
+	const int GetLastEvent() { return m_iLastEvent; }
+	const int GetLastAttacker() { return m_iLastAttacker; }
+	const int GetLastKiller() { return m_iLastKiller; }
+	const int GetLastGhoster() { return m_iLastGhoster; }
```
Rainyan: These could be const-qualified, so `const` to the right-hand side of the declaration since they don't mutate `CNEORules`

src/game/client/neo/ui/neo_hud_round_state.cpp
AdamTadeusz: If the player is locked from changing the spectate target then there's no point changing what player is selected in the top hud, since the only reason to change the selection is to, once the desired player is selected, jump to that player with spectate_player_selected_in_hud

src/game/client/hltvcamera.cpp
AdamTadeusz: It shouldn't happen, but I just wanted to make sure there is no infinite loop where SetMode calls SpecNextPlayer which in turn calls SetPrimaryTarget which in turn calls SetMode again

src/game/client/hltvcamera.cpp
AdamTadeusz: this would be the entity index of the last player who did that event, 0 would be the entityindex of the world. Ill change the variable name to something that makes this more clear

src/game/shared/neo/weapons/weapon_neobasecombatweapon.cpp
AdamTadeusz: moved to CNEO_Player::FireBullets

src/game/client/game_controls/SpectatorGUI.cpp
AdamTadeusz: No reason not to make this a float

src/game/client/c_baseanimating.cpp
```diff
@@ -3619,16 +3619,19 @@ int C_BaseAnimating::InternalDrawModel( int flags )
...
+			{
+				static Vector ownerOrigin;
+				ownerOrigin = pOwner->EyePosition();
+				pInfo->pLightingOrigin = &ownerOrigin;
+			}
 		}
 	}
-	else if (IsBaseCombatWeapon())
-	{
+	else if (IsBaseCombatWeapon() && !GetMoveParent())
```
DESTROYGIRL, line 3633: I assume this attempts to resolve the black supas (not sure how to replicate it in the first place tbh)

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -311,6 +318,12 @@ BEGIN_NETWORK_TABLE_NOBASE( CNEORules, DT_NEORules )
 	SendPropInt(SENDINFO(m_iJuggernautPlayerIndex), NumBitsForCount(MAX_PLAYERS_ARRAY_SAFE), SPROP_UNSIGNED),
 	SendPropBool(SENDINFO(m_bJuggernautItemExists)),
 	SendPropEHandle(SENDINFO(m_hJuggernaut)),
+	SendPropInt(SENDINFO(m_iLastHurt)),
+	SendPropInt(SENDINFO(m_iLastShooter)),
+	SendPropInt(SENDINFO(m_iLastEvent)),
+	SendPropInt(SENDINFO(m_iLastAttacker)),
+	SendPropInt(SENDINFO(m_iLastKiller)),
+	SendPropInt(SENDINFO(m_iLastGhoster)),
```
DESTROYGIRL: If we expect these networked vars to be valid player indexes then we can set the proper flags n bits for a player index

```suggestion
	SendPropInt(SENDINFO(m_iLastHurt), NumBitsForCount(MAX_PLAYERS_ARRAY_SAFE), SPROP_UNSIGNED),
	SendPropInt(SENDINFO(m_iLastShooter), NumBitsForCount(MAX_PLAYERS_ARRAY_SAFE), SPROP_UNSIGNED),
	SendPropInt(SENDINFO(m_iLastEvent), NumBitsForCount(MAX_PLAYERS_ARRAY_SAFE), SPROP_UNSIGNED),
	SendPropInt(SENDINFO(m_iLastAttacker), NumBitsForCount(MAX_PLAYERS_ARRAY_SAFE), SPROP_UNSIGNED),
	SendPropInt(SENDINFO(m_iLastKiller), NumBitsForCount(MAX_PLAYERS_ARRAY_SAFE), SPROP_UNSIGNED),
	SendPropInt(SENDINFO(m_iLastGhoster), NumBitsForCount(MAX_PLAYERS_ARRAY_SAFE), SPROP_UNSIGNED),
```

AdamTadeusz: these are indeed entity indexes of players, so this should be perfect

src/game/client/c_baseanimating.cpp
AdamTadeusz: in my case it was bright supas in the green spawn on oilstain, where they really should have had a more green tint to them, but yeah i remember seeing weapons slightly too bright or too dark sometimes

PR #1852, opened 2026-03-16
DebugDrawLine grenade trails for spectators and practicing

AdamTadeusz:
## Description

Contrary to the name of the branch, all that we do server side is make sure the grenade projectile is transmitted to all players who want to see it, those being all spectators with cl_neo_grenade_show_path set to a value greater than 0, or everyone in the case of the cheat protected convar sv_neo_grenade_show_path being set to a value greater than 0. New segments of the path are added client side in the grenade projectile client think function.

## Toolchain

- Windows MSVC VS2022

### DIFF
--- a/src/game/shared/sdk/sdk_basegrenade_projectile.cpp
+++ b/src/game/shared/sdk/sdk_basegrenade_projectile.cpp
@@ -164,7 +164,7 @@ ConVar cl_neo_grenade_show_path("cl_neo_grenade_show_path", "0", FCVAR_ARCHIVE |
 		if (m_vLastDrawPosition.DistToSqr(origin) < 0.1f)
 			return;
 
-		float r, g, b = 0;
+		float r = 0, g = 0, b = 0;
 		NEORules()->GetTeamGlowColor(GetTeamNumber(), r, g, b);
 		r *= 255;
 		g *= 255;
@@ -172,9 +172,9 @@ ConVar cl_neo_grenade_show_path("cl_neo_grenade_show_path", "0", FCVAR_ARCHIVE |
 		if (GetDamage())
 		{
 			const float timeAlive = (gpGlobals->curtime - m_flNeoCreateTime) * 0.5f;
-			r = Min(255.f, Max(0.f, r * (1 - timeAlive) + (255 * timeAlive)));
-			g = Max(0.f, g * (1 - timeAlive));
-			b = Max(0.f, b * (1 - timeAlive));
+			r = clamp(r * (1 - timeAlive) + (255.f * timeAlive), 0.f, 255.f);
+			g = clamp(g * (1 - timeAlive), 0.f, 255.f);
+			b = clamp(b * (1 - timeAlive), 0.f, 255.f);
 		}
 		DebugDrawLine(m_vLastDrawPosition, origin, r, g, b, true, showPathWhenServerEnabled ? sv_neo_grenade_show_path.GetFloat() : cl_neo_grenade_show_path.GetFloat());
 		m_vLastDrawPosition = origin;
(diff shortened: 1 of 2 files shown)


src/game/shared/sdk/sdk_basegrenade_projectile.h
```diff
@@ -63,6 +67,21 @@ class CBaseGrenadeProjectile : public CBaseGrenade
...
+
+		if ( pRecipientEntity->GetTeamNumber() == TEAM_SPECTATOR ) 
+		{
+			const char* wantsToSeeProjectilePath = engine->GetClientConVarValue(pRecipientEntity->entindex(), "cl_neo_grenade_show_path");
+			if (wantsToSeeProjectilePath && *wantsToSeeProjectilePath && (V_atof(wantsToSeeProjectilePath) != 0.f))
+			{
+				return FL_EDICT_ALWAYS;
+			}
+		}
+		return FL_EDICT_PVSCHECK;
```
sunzenshen, line 82: By any chance do we want to try executing the base class that was overridden, to see if it would handle other base behavior?

```cpp
BaseClass::ShouldTransmit(pInfo)
```

src/game/shared/sdk/sdk_basegrenade_projectile.cpp
```diff
@@ -127,6 +136,48 @@ END_NETWORK_TABLE()
...
+		if (!m_vLastDrawPosition.IsValid())
+		{
+			m_vLastDrawPosition = origin;
+			return;
+		}
+
+		if (m_vLastDrawPosition.DistToSqr(origin) < 0.1f)
+			return;
+
+		float r, g, b = 0;
```
sunzenshen: Maybe GetTeamGlowColor will override all of these correctly, but it's possible that only b is being set to 0 here, and the other two r,g variables are uninitialized at this point.

src/game/shared/sdk/sdk_basegrenade_projectile.cpp
```diff
@@ -127,6 +136,48 @@ END_NETWORK_TABLE()
...
+		NEORules()->GetTeamGlowColor(GetTeamNumber(), r, g, b);
+		r *= 255;
+		g *= 255;
+		b *= 255;
+		if (GetDamage())
+		{
+			const float timeAlive = (gpGlobals->curtime - m_flNeoCreateTime) * 0.5f;
+			r = Min(255.f, Max(0.f, r * (1 - timeAlive) + (255 * timeAlive)));
+			g = Max(0.f, g * (1 - timeAlive));
+			b = Max(0.f, b * (1 - timeAlive));
```
sunzenshen: Perhaps there could be an opportunity to use Clamp here:

```cpp
r = clamp(r * (1 - timeAlive) + (255.f * timeAlive), 0.f, 255.f);
g = clamp(g * (1 - timeAlive), 0.f, 255.f);
b = clamp(b * (1 - timeAlive), 0.f, 255.f);
```

src/game/shared/sdk/sdk_basegrenade_projectile.h
```diff
@@ -63,6 +67,21 @@ class CBaseGrenadeProjectile : public CBaseGrenade
 	// sit still until it had gotten a few updates from the server.
 	void SetupInitialTransmittedGrenadeVelocity( const Vector &velocity );
 
+	int ShouldTransmit(const CCheckTransmitInfo* pInfo) override
+	{
+		CBaseEntity *pRecipientEntity = CBaseEntity::Instance( pInfo->m_pClientEnt );
```
sunzenshen, line 72: Is it possible that CBaseEntity::Instance could return a nullptr here?

src/game/shared/sdk/sdk_basegrenade_projectile.h
AdamTadeusz: ShouldTransmit is only called when the transmit state is set to FL_EDICT_FULLCHECK (via SetTransmitState). I don't see anywhere where the grenade projectiles were set to this transmit state before, so I don't believe grenade projectiles ever called the base class ShouldTransmit before, and so doing it here would probably be a waste

sunzenshen (approved):

PR #1855, opened 2026-03-17
344 server side second auto record command

### LINKED ISSUE #1788: [FEATURE] Stop unnecessary STV demos from being recorded
### Description

Right now it seems the only way to record STV demos built-in to the game is through the default autorecord functionality, but we don't always want every demo, e.g. when running a comp server we're only interested in live match demos.

### Proposed Solution

The built-in comp feature could have a cvar to only record live matches (including pauses, overtime etc) and therefore save a lot of disk space and make it easier for serverops to actually find match demos, could even be sent to a specific folder instead of the game root.

### Alternatives Considered

_No response_

### Additional Context

_No response_

AdamTadeusz:
## Description

Adds a new convar sv_neo_server_autorecord that will record a sourcetv demo from the moment the next round starts to when the event game_end is triggered. Will not work without a sourcetv client on the server.

I've delayed when game_end is triggered to the end of the last round, so people watching the demo can see the match result and have a chance to pause the demo before they get kicked out.

## Toolchain

- Windows MSVC VS2022

## Linked Issues

- fixes #1788

### DIFF
--- a/src/game/shared/neo/neo_gamerules.h
+++ b/src/game/shared/neo/neo_gamerules.h
@@ -466,7 +466,7 @@ private:
 	Vector m_vecPreviousGhostSpawn = vec3_origin;
 	Vector m_vecPreviousJuggernautSpawn = vec3_origin;
 	bool m_bGotMatchWinner = false;
-	int m_iMatchWinnder = -1;
+	int m_iMatchWinner = TEAM_UNASSIGNED;
 #endif
 	CNetworkVar(int, m_nRoundStatus);
 	CNetworkVar(int, m_iHiddenHudElements);
--- a/src/game/shared/neo/neo_misc.cpp
+++ b/src/game/shared/neo/neo_misc.cpp
@@ -47,7 +47,7 @@ bool StartAutoRecording()
 {
 #ifdef GAME_DLL
 	CBasePlayer* sourceTV = nullptr;
-	for (int i = 0; i < gpGlobals->maxClients; i++)
+	for (int i = 1; i <= gpGlobals->maxClients; i++)
 	{
 		CBasePlayer* pPlayer = UTIL_PlayerByIndex(i);
 		if (pPlayer && pPlayer->IsHLTV())
@@ -58,7 +58,7 @@ bool StartAutoRecording()
 	}
 	if (!sourceTV)
 	{
-		ConMsg("SourceTV client not found, server side demo not recording.\n");
+		ConLog("SourceTV client not found, server side demo not recording.\n");
 		return false;
 	}
 #endif // GAME_DLL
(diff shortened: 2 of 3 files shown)


src/game/shared/neo/neo_gamerules.h
```diff
@@ -466,6 +466,7 @@ class CNEORules : public CHL2MPRules, public CGameEventListener
 	Vector m_vecPreviousGhostSpawn = vec3_origin;
 	Vector m_vecPreviousJuggernautSpawn = vec3_origin;
 	bool m_bGotMatchWinner = false;
+	int m_iMatchWinnder = -1;
```
Rainyan: Typo nit: `MatchWinner`

Also maybe use `TEAM_INVALID` instead of -1, (or perhaps `TEAM_UNASSIGNED` would be more semantically accurate?)

src/game/shared/neo/neo_misc.cpp
```diff
@@ -37,9 +43,27 @@ extern ConVar sv_neo_comp_name;
 	return LoopAroundMinMax(iValue, 0, iSize - 1);
 }
 
-#ifdef CLIENT_DLL
-void StartAutoClientRecording()
+bool StartAutoRecording()
 {
+#ifdef GAME_DLL
+	CBasePlayer* sourceTV = nullptr;
+	for (int i = 0; i < gpGlobals->maxClients; i++)
```
Rainyan: Should loop `i <= gpGlobals->maxClients` because players are in the 1-maxClients range, including `maxClients`. Also I don't think a client can ever be in index 0 so could start the loop from 1, unless there's some special case where the HLTV client can take index 0(?)

src/game/shared/neo/neo_misc.cpp
```diff
@@ -37,9 +43,27 @@ extern ConVar sv_neo_comp_name;
...
+		CBasePlayer* pPlayer = UTIL_PlayerByIndex(i);
+		if (pPlayer && pPlayer->IsHLTV())
+		{
+			sourceTV = pPlayer;
+			break;
+		}
+	}
+	if (!sourceTV)
+	{
+		ConMsg("SourceTV client not found, server side demo not recording.\n");
```
Rainyan: Maybe `ConLog`, so if this does happen there is a record of it in the server logs

AdamTadeusz:
If wanting to see the output of the console when recording a demo with stv, consider merging in https://github.com/NeotokyoRebuild/neo/pull/1845 which fixes some console spam issues

src/game/shared/neo/neo_misc.cpp
AdamTadeusz: I need to stop making this mistake aargh

Rainyan (approved):

PR #1871, opened 2026-03-20
NeoUI - Table headers + main table API + section X-axis scrolling

### LINKED ISSUE #994: [FEATURE] Add a tag/column in the server list to distinguish between different types of servers
### Description

Players might want to know what kind of gamemode, plugins, or features that a server is using before they join. Furthermore, players might want some additional context as to why a server is locked when they want to play with others.

### Proposed Solution

Add a column on the server list for tags.
competitive/casual/closed/open/ghosthopping etc.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### LINKED ISSUE #1566: Games with bots show the wrong player count in the server browser
The number goes into the negatives. Bots were being added with `neo_bot_quota 10`~ while 6-8 real players were in server

### LINKED ISSUE #1815: When lowering the resolution in windowed mode, buttons should still highlight when being moused over
Could be an issue in other window modes

https://github.com/user-attachments/assets/7c5adbbd-83ce-409d-9a98-01aa68818b13

nullsystem:
## Description

NeoUI table headers and main table API implemented, now have Begin/EndTable and NextTableRow implementation replacing the previous custom paint + button. Now can just layout the table cells with widgets, although at the moment only NeoUI::Label are properly utilized and dealt with. Other widgets are not refactored up for tables yet.

The table headers now have dragable resizing and the server browser now have tags and IP address columns, where tags are visible and IP address hidden by default. There's also right-click on the header to show/hide columns.

Sections now have X-axis scrolling, mainly used for table and header scrolling support. Tabs now have its own managed X-axis scrolling separated from section X-axis scrolling.

OTHERS:
* Fix SDR/Steam networking server bot vs player number count
* Fix border in smaller resolutions

TODOs:
* [x] Fixup tab X scrolling
* [x] Extra filtering
* [x] More columns parity to legacy browser
* [x] Un-viewport the killer/damage table

FUTURE TODOs:
* Refactor painting of widgets to utilize vgui viewports
* Possible split of painting (and colors) from NeoUI internals
* Vertical layouting that expands horizontally as oppose to the default horizontal layouting that expands vertically
* Save headers size layouts, filters, and create server options
* Moved for later PRs:
    * Change from only header to any sections to reference other section's scrolls
    * Section X-scrollbar controls options/flags
        * Modes:
            * No X-scrollbar
            * Indicator/thin X-scrollbar
            * Dragable/thick X-scrollbar
    * Using BeginTable will automatically put the section into Dragable/thick X-scrollbar mode
    
BRANCHES RELIES ON THIS:
* Auto-join NeoUI: https://github.com/nullsystem/neo/compare/NeoUITable...nullsystem:neo:GH-1298_AutoJoin
* NeoUI MP3 player/NeoMP3: https://github.com/nullsystem/neo/compare/NeoUITable...nullsystem:neo:GH-1098_NeoUIMP3

## Toolchain

- Linux GCC Distro Native GCC 15/Arch Linux

## Linked Issues
* fixes #1566
* fixes #994
* fixes #1815

### DIFF
--- a/src/game/client/neo/ui/neo_ui.cpp
+++ b/src/game/client/neo/ui/neo_ui.cpp
@@ -1851,6 +1851,7 @@ void Tabs(const wchar_t **wszLabelsList, const int iLabelsSize, int *iIndex,
 
 	if (wdgState.bInView)
 	{
+		const int iTabWideEqual = (c->dPanel.wide / iLabelsSize);
 		int iTabWide = 0;
 		int iTabOffset = 0;
 		if (pState)
@@ -1860,7 +1861,6 @@ void Tabs(const wchar_t **wszLabelsList, const int iLabelsSize, int *iIndex,
 					|| pState->iWide <= 0
 					|| pState->bInYScroll != bThisYScroll)
 			{
-				const int iTabWideEqual = (c->dPanel.wide / iLabelsSize);
 				for (int i = 0; i < iLabelsSize; ++i)
 				{
 					int iCurTabWide, iTabTall;
@@ -1880,7 +1880,7 @@ void Tabs(const wchar_t **wszLabelsList, const int iLabelsSize, int *iIndex,
 		}
 		else
 		{
-			iTabWide = (c->dPanel.wide / iLabelsSize);
+			iTabWide = iTabWideEqual;
 		}
 
 		const int iOverflowBtnX = (iTabWide > (c->dPanel.wide / iLabelsSize))
@@ -1924,6 +1924,12 @@ void Tabs(const wchar_t **wszLabelsList, const int iLabelsSize, int *iIndex,
 					.x1 = iTabOffset + iXPosTab + iTabWide,
 					.y1 = c->irWidgetTall,
 				};
+				// If in equal sizing wide, last tab, and no offset, make sure it expands
+				// fully to the right edge
+				if (i == (iLabelsSize - 1) && iTabWideEqual == iTabWide && 0 == iTabOffset)
+				{
+					tabRect.x1 = iXViewportWide;
+				}
 				const bool bHotTab = IN_BETWEEN_EQ(iXViewportStart, c->iMouseAbsX, iXViewportEnd)
 						&& IN_BETWEEN_EQ(tabRect.x0 + iXViewportStart, c->iMouseAbsX, tabRect.x1 + iXViewportStart)
 						&& IN_BETWEEN_EQ(tabRect.y0 + c->rWidgetArea.y0, c->iMouseAbsY, tabRect.y1 + c->rWidgetArea.y0);


src/game/client/neo/ui/neo_root.cpp
```diff
@@ -1459,69 +1408,195 @@ void CNeoRoot::MainLoopServerBrowser(const MainLoopParam param)
...
 							}
-							if (btn.bKeyPressed || btn.bMouseDoublePressed)
+							++iShownServers;
+							m_iUpDownDirection = 0;
+							m_iUpDownInitialServer = -1;
+
+							wchar_t wszServerName[k_cbMaxGameServerName] = {};
+							wchar_t wszPlayers[15] = {};
+							wchar_t wszPing[10] = {};
+							wchar_t wszIPAddress[64] = {};
```
Rainyan: Should we allocate the `wszIPAddress` element count with `SteamNetworkingIPAddr::k_cchMaxString`:

https://github.com/NeotokyoRebuild/neo/blob/2edf01dadf31785763a79633feaabdc8c8523557/src/public/steam/steamnetworkingtypes.h#L216-L218

or is there a case where it can be larger than this define?

src/game/client/neo/ui/neo_root_serverbrowser.cpp
```diff
@@ -331,9 +339,26 @@ void CNeoServerList::ServerResponded(HServerListRequest hRequest, int iServer)
 
 	ISteamMatchmakingServers *steamMM = steamapicontext->SteamMatchmakingServers();
 	gameserveritem_t *pServerDetails = steamMM->GetServerDetails(hRequest, iServer);
-	if (pServerDetails)
+	if (pServerDetails && m_servers.size() < 1024)
```
Rainyan: What's the reason for the `< 1024` check here? Should we `Assert` for the servers size blowing up here, if that can happen under some circumstances?

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -1171,11 +1328,52 @@ void Label(const wchar_t *wszText, const bool bNotWidget)
...
+			viewportDim.tall = c->dPanel.tall;
+		}
+		else
+		{
+			viewportDim.x = Max(c->rWidgetArea.x0, c->dPanel.x);
+			viewportDim.y = c->rWidgetArea.y0;
+			viewportDim.wide = c->irWidgetWide - iLDiff - iRDiff;
+			viewportDim.tall = c->irWidgetTall;
+		}
+		// NEO NOTE (nullsystem): Sanity check the viewport makes sense
```
Rainyan: Maybe `Assert(false)` for the else case, if we expect it to never occur?

src/game/client/neo/ui/neo_root_serverbrowser.cpp
nullsystem: There wasn't really a reason for the check, I was trying to figure out a crash bug with pulling in servers list requests and throwing stuff in but otherwise it never even gotten anywhere near 1024 items. I might have a plan to mitigate the issue properly but out of scope of this PR.

nullsystem: Bare in mind the crash was referring to when removing the gamedir = neo filter and pulling any servers in, so wasn't affected by NT;RE current servers at the moment. But nevertheless need to sort it out eventually.

src/game/client/neo/ui/neo_ui.cpp
nullsystem: It's actually expected the viewport could be "invalid" if the label were to written off the section's area due to X-axis offsets. So I don't think an assert needed here, maybe a bit more comments to clarify.

PR #1884, opened 2026-03-22
v28 cherry-pick: Fix server crash when auto recording server-side (#1882)

nullsystem:
Cherry-pick of: https://github.com/NeotokyoRebuild/neo/pull/1882

## Description
The guess from the crashdump: the server trying to use VGUI global var which may not be initialized for the server and get nullptr deref. Change how datetime is fetched to a stdlib way.

## Toolchain

### DIFF
--- a/src/game/shared/neo/neo_misc.cpp
+++ b/src/game/shared/neo/neo_misc.cpp
@@ -2,12 +2,13 @@
 
 #ifdef CLIENT_DLL
 #include "steamclientpublic.h"
+#include "vgui/ISystem.h"
 #endif // CLIENT_DLL
 #include "cbase.h"
 #include "neo_gamerules.h"
-#include "vgui/ISystem.h"
 #include "tier3.h"
 #include <filesystem.h>
+#include <ctime>
 
 extern ConVar sv_neo_comp_name;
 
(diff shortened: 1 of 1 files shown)


src/game/shared/neo/neo_misc.cpp
```diff
@@ -88,9 +89,10 @@ bool StartAutoRecording()
 
 	// Time and date
 	char timeSection[16];
-	int year, month, dayOfWeek, day, hour, minute, second;
-	g_pVGuiSystem->GetCurrentTimeAndDate(&year, &month, &dayOfWeek, &day, &hour, &minute, &second);
-	V_snprintf(timeSection, sizeof(timeSection), "%04d%02d%02d-%02d%02d", year, month, day, hour, minute);
+	const time_t timeVal = time(nullptr);
+	std::tm tmStruct{};
+	std::tm tmd = *(Plat_localtime(&timeVal, &tmStruct));
```
Rainyan, line 94: Seems like `Plat_localtime` writes the result to `tmStruct`, so the by-value assignment of `*tmStruct` to `tmd` produces a redundant copy. But I guess it's not a big deal in practice, and not worth changing for the cherry-pick.

nullsystem: I kind of just took the lines from strtools.cpp `BGetLocalFormattedDateAndTime` and kept it-ish as such. But yeah not a big deal.

PR #1887, opened 2026-03-22
Gamemode specific score / round limits

DESTROYGIRL:
## Description
`neo_score_limit` \ `neo_round_limit` remain intact for CTG to avoid breaking servers and should be removed in the future. The default server configuration should also be changed for this

Additionally, this changes the default limits for TDM and JGR.
TDM maps will only have 1 round unless there is a tie. First to 1
JGR maps will be a best of 5 rounds.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/shared/neo/neo_gamerules.cpp
+++ b/src/game/shared/neo/neo_gamerules.cpp
@@ -418,12 +418,60 @@ const NeoGameTypeSettings NEO_GAME_TYPE_SETTINGS[NEO_GAME_TYPE__TOTAL] = {
 		// convert a velocity in ft/sec and a mass in grains to an impulse in kg in/s
 #define BULLET_IMPULSE(grains, ftpersec)	((ftpersec)*12*BULLET_MASS_GRAINS_TO_KG(grains)*BULLET_IMPULSE_EXAGGERATION)
 
+extern ConVar neo_score_limit;
+extern ConVar neo_round_limit;
+extern ConVar sv_neo_ctg_score_limit;
+extern ConVar sv_neo_ctg_round_limit;
 
-ConVar neo_score_limit("neo_score_limit", "7", FCVAR_REPLICATED, "Neo score limit.", true, 0.0f, true, 99.0f);
-ConVar neo_round_limit("neo_round_limit", "0", FCVAR_REPLICATED, "Max amount of rounds, 0 for no limit.", true, 0.0f, false, 0.0f);
+static void neoScoreLimitLegacyCallback(IConVar* var, const char* pOldValue, float flOldValue)
+{
+	Warning("Using legacy neo_score_limit cvar. Use sv_neo_[gamemode]_score_limit instead!\n");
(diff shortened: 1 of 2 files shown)


src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -419,12 +418,60 @@ const NeoGameTypeSettings NEO_GAME_TYPE_SETTINGS[NEO_GAME_TYPE__TOTAL] = {
...
+{
+	Warning("Using legacy neo_score_limit cvar. Use sv_neo_[gamemode]_score_limit instead!\n");
+	sv_neo_ctg_score_limit.SetValue(neo_score_limit.GetInt());
+}
+
+static void neoRoundLimitLegacyCallback(IConVar* var, const char* pOldValue, float flOldValue)
+{
+	Warning("Using legacy neo_round_limit cvar. Use sv_neo_[gamemode]_round_limit instead!\n");
+	sv_neo_ctg_round_limit.SetValue(neo_round_limit.GetInt());
+}
```
Rainyan, line 436: This will make the legacy cvars only work for CTG, whereas parity would be supporting all game modes.

If we had something like `ConVar* NEORules::GetRoundLimitCvar() const` with similar logic to the `GetScoreLimit` switch statement but instead returning the `ConVar*` of the active game mode, then this could turn into `GetRoundLimitCvar().SetValue(...)` etc.

But if we are sunsetting the legacy cvars then I guess this is fine (and our TDM isn't exactly parity to begin with, with the respawns etc).

Rainyan (approved):

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -2592,6 +2621,56 @@ CNEORules::ReadyPlayers CNEORules::FetchReadyPlayers() const
 	return readyPlayers;
 }
 
+const int CNEORules::GetScoreLimit() const
+{
+	switch (m_nGameTypeSelected)
+	{
+	case NEO_GAME_TYPE_TDM:
+		return sv_neo_tdm_score_limit.GetInt();
+		break;
```
sunzenshen, line 2630: nit: the break;(s) are superfluous here since the returns will exit but this doesn't cause any problems either.  Just surprised this isn't caught as a warning.

src/game/shared/neo/neo_gamerules.cpp
DESTROYGIRL: I did this because people really only had these set for CTG in the original game, and the only other functional gamemode was TDM (used for like 2 maps), so I imagined it would probably be more annoying / damaging to apply it to all of them in rebuild

src/game/shared/neo/neo_gamerules.cpp
DESTROYGIRL: I wasn't really sure if this was better practice or something because I saw it being done elsewhere in the codebase

PR #1905, opened 2026-03-29
WIP: Add client-side prediction for knife hits

Rainyan:
## Description

* Predict knife damage impact effects for better responsiveness for high latency players
* Fix issues with backstab damage not working despite being within the backstab angle
* Move backstab angle to a cvar, and convert from radians to degrees for more intuitive adjustment
* Add a cvar for toggling whether the pitch component of the view angle is considered in backstab calculations

### Examples

300 ms roundtrip latency (`net fakelag 150;`) without prediction (current state):

https://github.com/user-attachments/assets/4e40d929-8542-4466-ba3d-06664f5cdc86

Same but with prediction (this PR):

https://github.com/user-attachments/assets/2b80168c-6492-4b0b-887e-d08e76e8ecb0

Note how the blood impact effect is not delayed in the latter, and renders from the correct position. The sound of the video embeds seems to be bugged, but it shouldn't be desynced in either case in reality.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1904 
- related #404

### DIFF
--- a/src/game/server/NextBot/Player/NextBotPlayer.h
+++ b/src/game/server/NextBot/Player/NextBotPlayer.h
@@ -739,6 +739,7 @@ inline void _NextBot_BuildUserCommand( CUserCmd *cmd, const QAngle &viewangles, 
 #ifdef NEO
 extern ConVar bot_mimic;
 extern ConVar bot_mimic_yaw_offset;
+extern ConVar bot_mimic_mirror;
 #endif // NEO
 template < typename PlayerType >
 inline void NextBotPlayer< PlayerType >::PhysicsSimulate( void )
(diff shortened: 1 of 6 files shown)


Rainyan:
The new `bot_mimic_mirror 1` (together with `bot_mimic 1`) can be used to make the bots look the exact opposite direction (pitch included) for testing backstab mechanics when in close proximity. I didn't nerf the "stair stabbing" mechanic whereby one can land a backstab from the vertical axis, but there is a new cvar `sv_neo_backstab_ignorez` (default 1) which can be disabled to make it possible to "dodge" a vertical backstab my matching your look direction accordingly, analogous to how one can dodge the regular backstab by using the look yaw angle and continuously facing toward the opponent. Unsure which mode we eventually want, but I'll leave changing that gameplay stuff outside the scope of this PR.

PR #1913, opened 2026-03-31
NeoUI - Full server queuing + automatic joining

### LINKED ISSUE #1298: NeoUI serverlist - Auto-join when server full
Probably it'll go like this:
* The join server functionality will check if the server is full or not
* If it's full, then instead of trying to connect, go to a separate screen where it's waiting for the server to free up and interval on next check
* Then if there's free slots automatically try to connect

nullsystem:
## Description
On entering a server, a ping-reply is now sent to the server to get the most up to date players count upon entering. Auto-join is determined if the server's full when trying to enter, and if so enters an auto-join state where it'll refresh for 15s checking count then entering when there's a free slot available. The auto-join popup bar stays on any states.

## Toolchain

## Linked Issues

- fixes #1298

### DIFF
--- a/src/game/client/neo/ui/neo_root.cpp
+++ b/src/game/client/neo/ui/neo_root.cpp
@@ -154,10 +154,19 @@ static bool NetAdrIsSDR(const servernetadr_t &netAdr)
 #endif
 }
 
-// SDR servers doesn't actually includes bots in their players count
-static int PlayersCount(const gameserveritem_t *pServer)
+enum EPlayerCountMode
 {
-	return pServer->m_nPlayers - (NetAdrIsSDR(pServer->m_NetAdr) ? 0 : pServer->m_nBotPlayers);
+	PLAYERCOUNT_ONLYPLAYER = 0,
+	PLAYERCOUNT_INCLUDEBOTS,
+};
+
+// SDR servers doesn't actually includes bots in their players count
+static int PlayersCount(const gameserveritem_t *pServer, const EPlayerCountMode eCountMode)
+{
+	const int iPlayersCount = pServer->m_nPlayers - (NetAdrIsSDR(pServer->m_NetAdr) ? 0 : pServer->m_nBotPlayers);
+	return (eCountMode == PLAYERCOUNT_INCLUDEBOTS)
+			? iPlayersCount + pServer->m_nBotPlayers
+			: iPlayersCount;
 }
 
 // Only use it rarely/cached
@@ -651,7 +660,7 @@ void CNeoRoot::OnRelayedKeyTyped(wchar_t unichar)
 // copy not ref/pointer gameserveritem_t
 void CNeoRoot::OnEnterServer(const gameserveritem_t gameServer, const char *pszServerPassword)
 {
-	const int iPlayersCount = PlayersCount(&gameServer);
+	const int iPlayersCount = PlayersCount(&gameServer, PLAYERCOUNT_INCLUDEBOTS);
 	m_serverPingAutoJoin.m_serverInfo =
 #ifdef DEBUG
 			(iPlayersCount < (gameServer.m_nMaxPlayers - cl_neo_autojoin_offset.GetInt()))
@@ -776,7 +785,6 @@ void CNeoRoot::OnMainLoop(const NeoUI::Mode eMode)
 			static constexpr float FL_AUTO_JOIN_WAIT = 15.0f;
 
 			const auto &server = m_serverPingAutoJoin.m_serverInfo;
-			const int iPlayersCount = PlayersCount(&server);
 			wchar wszText[k_cbMaxGameServerName] = {};
 
 			static constexpr const int ROWLAYOUT_REFRESH[] = { 10, 15, 35, 20, -1 };
@@ -791,7 +799,8 @@ void CNeoRoot::OnMainLoop(const NeoUI::Mode eMode)
 			g_pVGuiLocalize->ConvertANSIToUnicode(server.GetName(), wszText, sizeof(wszText));
 			NeoUI::Label(wszText);
 
-			V_swprintf_safe(wszText, L"Players: %d/%d", iPlayersCount, server.m_nMaxPlayers);
+			const int iPlayersOnlyCount = PlayersCount(&server, PLAYERCOUNT_ONLYPLAYER);
+			V_swprintf_safe(wszText, L"Players: %d/%d", iPlayersOnlyCount, server.m_nMaxPlayers);
 			NeoUI::Label(wszText);
 
 			V_swprintf_safe(wszText, L"Refresh: %ds", Max(0, static_cast<int>((m_flAutoJoinLastAttempt + FL_AUTO_JOIN_WAIT) - gpGlobals->realtime)));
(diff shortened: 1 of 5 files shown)


src/game/client/neo/ui/neo_root.cpp
```diff
@@ -684,8 +748,69 @@ void CNeoRoot::OnMainLoop(const NeoUI::Mode eMode)
...
+			else if (CNeoServerPing::PINGSTATE_NIL != m_serverPingAutoJoin.m_ePingState)
+			{
+#ifdef DEBUG
+				if ((iPlayersCount < (server.m_nMaxPlayers - cl_neo_autojoin_offset.GetInt()))
+#else
+				if ((iPlayersCount < server.m_nMaxPlayers)
+#endif
+						&& (CNeoServerPing::PINGSTATE_SUCCESS == m_serverPingAutoJoin.m_ePingState))
+				{
+					OnEnterServer(m_serverPingAutoJoin.m_serverInfo, nullptr);
```
sunzenshen: Are we only supporting autojoin for passwordless servers, or would it be worth saving the password for autojoining a password protected server?

src/game/client/neo/ui/neo_root.cpp
```diff
@@ -1836,40 +1972,38 @@ void CNeoRoot::MainLoopServerBrowser(const MainLoopParam param)
...
-
-							ConVarRef("password").SetValue("");
-							V_memset(m_wszServerPassword, 0, sizeof(m_wszServerPassword));
+							eEnterServer = ENTERSERVER_PING;
+						}
+					}
+					if (m_iSelectedServer >= 0
+							|| m_serverPingEnter.m_serverInfo.m_NetAdr.GetIP() != 0)
+					{
+						const auto &gameServerSelected = m_serverBrowser[m_iServerBrowserTab].m_filteredServers[m_iSelectedServer];
```
sunzenshen: This may be unlikely with the UI, but is it possible that only 

```cpp
m_serverPingEnter.m_serverInfo.m_NetAdr.GetIP() != 0)
```

could be true above, and a m_iSelectedServer is a weird value like -1?  The GUI seems to prevent this, but would it be worth sanity checking in the code?

src/game/client/neo/ui/neo_root.cpp
```diff
@@ -1836,40 +1972,38 @@ void CNeoRoot::MainLoopServerBrowser(const MainLoopParam param)
...
-								engine->ClientCmd_Unrestricted("disconnect");
-							}
-
-							ConVarRef("password").SetValue("");
-							V_memset(m_wszServerPassword, 0, sizeof(m_wszServerPassword));
+							eEnterServer = ENTERSERVER_PING;
+						}
+					}
+					if (m_iSelectedServer >= 0
+							|| m_serverPingEnter.m_serverInfo.m_NetAdr.GetIP() != 0)
```
sunzenshen, line 2009: What happens if m_iSelectedServer is less than 0, but the GetIP() check passes in this OR?

src/game/client/neo/ui/neo_root.cpp
```diff
@@ -684,8 +748,69 @@ void CNeoRoot::OnMainLoop(const NeoUI::Mode eMode)
...
+			g_pVGuiLocalize->ConvertANSIToUnicode(server.GetName(), wszText, sizeof(wszText));
+			NeoUI::Label(wszText);
+
+			V_swprintf_safe(wszText, L"Players: %d/%d", iPlayersCount, server.m_nMaxPlayers);
+			NeoUI::Label(wszText);
+
+			V_swprintf_safe(wszText, L"Refresh: %ds", static_cast<int>((m_flAutoJoinLastAttempt + FL_AUTO_JOIN_WAIT) - gpGlobals->realtime));
+			NeoUI::Label(wszText);
+
+			if ((m_flAutoJoinLastAttempt + FL_AUTO_JOIN_WAIT) <= gpGlobals->realtime)
```
sunzenshen, line 809: What are the chances that a frame could go long and the time ticks to less than zero?  Maybe we could get the max between 0 and the countdown?

```cpp
V_swprintf_safe(wszText, L"Refresh: %ds",
    Max(0, static_cast<int>((m_flAutoJoinLastAttempt + FL_AUTO_JOIN_WAIT) - gpGlobals->realtime)));
```

sunzenshen (dismissed):

AdamTadeusz:
Ill review the PR properly tomorrow, but having skim read it I should point out that if bots are set up to automatically vacate when a player joins, then the bot manager will maintain the number of bots such that there is always space on the server for a player to join

AdamTadeusz:
Yeah so my understanding from reading the code is that in OnEnterServer, if the number of players without bots is greater than or equal to the max number of players, the gameserveritem gets saved to the autojoin thing and we enter the autojoin state.

What this doesn't take into account is that if the server is set up to auto kick bots to allow players to join, there will always be an empty spot on the server until the server is next to full and there are no bots in the server, and if the server is not set up to auto kick bots then we can't assume that just because there are fewer real people than the max number of players, that the local player can join the server. In both cases, we should therefore be comparing the total number of players including fake ones against the max number of players.

nullsystem:
@AdamTadeusz Not sure what you mean by the first part since it does already logically assume the server kicks the bots by excluding it from the player count? Then if the server doesn't kick bots then it'll have to utilize tags to determine that which might work out.

Why would the bot count even need to be included for case where the server will kick them out anyway? It's no different to no-bots servers.

nullsystem:
And for the case about bots filling/kicking, the convar `neo_bot_quota_mode` that's not "fill" can be exposed as server tags so the client can determine what to do with the bot count. Kind of a niche case though since they're not defaults.

nullsystem:
I've thrown it as a separate PR: https://github.com/NeotokyoRebuild/neo/pull/1935 as it's an extra feature. But otherwise the current one works as intended.

AdamTadeusz:
It's not an extra feature and you don't need to pass the convar as a server tag. If the server is set to automatically kick bots to allow players to join, then as long as there are bots on the server, the returned player count including bots will always be less than the max number of players that can join the server, because neo_bot_manager will maintain the number of bots such that there is always an empty slot in the server. If the server is set up to not automatically kick bots to allow players to join, then neo_bot_manager will not do that, and in situations where player count + bot count is equal to the max number of players on the server the player will not be able to join. In both cases there's no need for a client joining the server to do any special checks, a single comparison of the total number of players including bots against the max number of players will correctly predict whether the player can join the server immediately or whether they need to join the queue. No need to read any server tags or work out what settings the server is using

AdamTadeusz:
Here is an example.
neo_bot_quota 32;maxplayers 10;neo_bot_auto_vacate 1;
image
The server has 9 players connected and not 10 because neo_bot_auto_vacate is set to 1. The number of connected players including bots is less than the max number of players, anybody trying to join the server will do so without a problem and when they do a bot will be kicked to maintain an empty slot



image
If I set neo_bot_auto_vacate to 0 then another bot joins and the server is now full, and no one can join the server.
<br/><br/>
If the auto join used the total number of players including bots, then it would predict correctly in the first case that it can join the server because 9 < 10, and in the second case it would correctly predict that the server is full and it needs to wait in the queue because 10 == 10. No need to read any server tags.

src/game/client/neo/ui/neo_root.cpp
```diff
@@ -638,6 +648,68 @@ void CNeoRoot::OnRelayedKeyTyped(wchar_t unichar)
 	OnMainLoop(NeoUI::MODE_KEYTYPED);
 }
 
+// copy not ref/pointer gameserveritem_t
+void CNeoRoot::OnEnterServer(const gameserveritem_t gameServer, const char *pszServerPassword)
+{
+	const int iPlayersCount = PlayersCount(&gameServer);
```
AdamTadeusz: iPlayerCount here should including the number of bots connected to the server

AdamTadeusz (changes requested):

AdamTadeusz (approved):

sunzenshen (approved):

PR #1930, opened 2026-04-05
Bots crouch when following pings if commander was crouched

sunzenshen:
## Description
Allow ordering bots to stay in a crouched position by having bots check the commander's crouch state when they receive a ping waypoint.  If the commander was crouched, bots will stay crouched when moving to and waiting at the ping waypoint.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
@@ -226,6 +226,7 @@ bool CNEOBotCommandFollow::FollowCommandChain(CNEOBot* me)
 		// Use sv_neo_bot_cmdr_stop_distance_sq for consistent bot collection range
 		// follow_stop_distance_sq would be confusing if player doesn't know about distance tuning
 		me->m_hLeadingPlayer = pCommander;
+		m_bCrouchWhenFollowingPing = false; // watch commander's stance
 		m_vGoalPos = CNEO_Player::VECTOR_INVALID_WAYPOINT;
 		pCommander->m_vLastPingByStar.GetForModify(me->GetStar()) = CNEO_Player::VECTOR_INVALID_WAYPOINT;
 	}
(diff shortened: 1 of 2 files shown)


sunzenshen:
Demo:

https://github.com/user-attachments/assets/f2272558-f139-4f20-bc1c-73ee7a9983e6

Testing Notes:

- Remember to press use on bots to get them to follow you first
- Stand while placing ping to have bots stand when following ping waypoints
- Crouch while placing ping to have bots crouch when following ping waypoints

Potential Controversial Points:
- This might cause an awkward situation when a commander is pinned down behind cover while crouching and if they wanted to have their followers run to a ping waypoint they would need to stand up.  Otherwise the bots would move to the waypoint while crouch walking.
- This could make bot jumps over obstacles less likely to succeed if the starting crouch interferes with the maximum jump height of the bot.
- The initial PR proactively sets m_bCrouchWhileFollowingPing to false even if it's not needed, but maybe this is clearer for developers looking at member variables with a debugger.

AdamTadeusz:
Whats the purpose of this? I can understand holding the walk key when placing a ping to have bots approach a location quietly, but I don't see the vision here

sunzenshen:
> Whats the purpose of this? I can understand holding the walk key when placing a ping to have bots approach a location quietly, but I don't see the vision here

My thinking was that there's currently no way to order a bot to crouch behind some cover and wait in ambush, or to sit crouched from a sniping position.

After thinking about this PR some more, I wonder if it might be better to have the bots run as normal between waypoints but when they are not pressing any movement buttons, then they crouch. So the pattern of movement from a crouched position would be to standup to a running movement, arrive at the waypoint, and then crouch.

sunzenshen:
> holding the walk key when placing a ping to have bots approach a location quietly

Decided to finally tackle this silent walk behavior in this PR: https://github.com/NeotokyoRebuild/neo/pull/1937

https://github.com/user-attachments/assets/fdec6e1a-35d4-4338-9a0f-988fe96a15d4

Rainyan (approved):

sunzenshen:
https://github.com/NeotokyoRebuild/neo/pull/1937 might be the less controversial proposal, so let's review and consider that one first, and then I'll rebase this PR on top of that one.

sunzenshen:
Use case addressed in #1937

PR #1937, opened 2026-04-09
Bots silent walk to ping waypoints if commander had +walk engaged

sunzenshen:
## Description
Bots check the commander's walk state to determine if they should walk while following their commander or a waypoint ping

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
@@ -205,6 +205,15 @@ bool CNEOBotCommandFollow::FollowCommandChain(CNEOBot* me)
 		{
 			me->ReleaseCrouchButton();
 		}
+
+		if (pPlayerToMirror->IsWalking())
+		{
+			me->PressSneakButton(0.5f);
+		}
+		else
+		{
+			me->ReleaseSneakButton();
+		}
 	}
 
 	// Calibrate dynamic follow distance
@@ -233,6 +242,7 @@ bool CNEOBotCommandFollow::FollowCommandChain(CNEOBot* me)
 		// Use sv_neo_bot_cmdr_stop_distance_sq for consistent bot collection range
 		// follow_stop_distance_sq would be confusing if player doesn't know about distance tuning
 		me->m_hLeadingPlayer = pCommander;
+		m_bSneakWhenFollowingPing = false; // watch commander's stance
 		m_vGoalPos = CNEO_Player::VECTOR_INVALID_WAYPOINT;
 		pCommander->m_vLastPingByStar.GetForModify(me->GetStar()) = CNEO_Player::VECTOR_INVALID_WAYPOINT;
 	}
@@ -244,6 +254,7 @@ bool CNEOBotCommandFollow::FollowCommandChain(CNEOBot* me)
 		{
 			SendUpdateToCommander( me, "On my way." );
 			me->m_hLeadingPlayer = nullptr; // Stop following and start travelling to ping
+			m_bSneakWhenFollowingPing = pCommander->IsWalking();
 			m_vGoalPos = pCommander->m_vLastPingByStar.Get(me->GetStar());
 			me->m_vLastPingByStar.GetForModify(me->GetStar()) = pCommander->m_vLastPingByStar.Get(me->GetStar());
 
@@ -256,8 +267,14 @@ bool CNEOBotCommandFollow::FollowCommandChain(CNEOBot* me)
 			else
 			{
 				me->m_hLeadingPlayer = pCommander; // fallback to following commander
+				m_bSneakWhenFollowingPing = false; // watch commander's stance
 				// continue with leader following logic below
 			}
+		}
+
+		if (m_bSneakWhenFollowingPing)
+		{
+			me->PressSneakButton(0.3f);
 		}
 
 		if (FanOutAndCover(me, m_vGoalPos))
@@ -274,6 +291,7 @@ bool CNEOBotCommandFollow::FollowCommandChain(CNEOBot* me)
 		else
 		{
 			me->m_hLeadingPlayer = pCommander; // fallback to following commander
+			m_bSneakWhenFollowingPing = false; // react to enemies at full speed
 			// continue with leader following logic below
 		}
 	}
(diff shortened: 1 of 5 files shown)


sunzenshen:
Demo:

https://github.com/user-attachments/assets/fb486665-af65-496a-a2f8-a63a2a54d10a

(Needs to be unmuted to make any sense)

Can collect bot followers in a normal match by pressing use on them.

To test in peace, the following script may be useful:

```
sv_neo_preround_freeze_time 2; mp_chattime 2;
sv_neo_bot_cmdr_enable 1;
neo_ctg_round_timelimit 99;
```

Rainyan (dismissed):
The code LGTM, one nit on the behaviour though: it seems that if you first press the use key to make the bot follow you, and only then start holding the walk key, the bot won't react to the walk key input.

Using F as the use key, and ALT as the walk key in the video:

[ex_use_then_walk.webm](https://github.com/user-attachments/assets/86e9ce8b-7dd8-4444-83e7-2038c3f6ce95)

This could be frustrating if you're standing still and wanting to be sneaky, and the stomping bot announces your hiding spot as it approaches (I think typically a human player won't hold down walk when standing still for a long time, since they're silent anyway by not moving). Maybe it would be more robust if the bot mimiced whether their commander is moving in such a way that can produce footsteps, so for example standing still would be considered as a sneaky commander?

It seems we're using `CBasePlayer::UpdateStepSound` for determining whether the player should be emitting footstep noises, which sadly is just side effects with a `void` return, so there's no useful return value information available. There's also the `virtual bool CMultiplayRules::PlayFootstepSounds(CBasePlayer*)` which we aren't using, but maybe the `UpdateStepSound` stuff could be refactored to override that one in `CNEORules` so the bool return value could be reused for checking commander footsteps audibility. Or I guess refactoring `CBasePlayer::UpdateStepSound` with `NEO` ifdefs to return boolean instead of void could be a less disruptive approach, since code that doesn't care could keep using the method just as before and simply ignore the retval.

Another place where I could see this footstep audibility stuff being useful is for un-controlled bots in the vicinity of human players, since again it could be annoying to have your sneak ruined by a clomping bot alerting enemies to your area, but possibly out of scope for this PR.

But maybe this all this refactoring business is something to revisit some other time, I'm ok with merging this one without changes.

sunzenshen:
>Maybe it would be more robust if the bot mimiced whether their commander is moving in such a way that can produce footsteps, so for example standing still would be considered as a sneaky commander?

I like this idea enough that I went ahead and made a commit to make bots base their stealthiness on the commander's current behavior.  

https://github.com/NeotokyoRebuild/neo/pull/1937/changes/168cb252890be40a3f2eaacb2a5b7ddc83d3ee96

I removed the mimicking of crouch and walk from the snake line leader to save cycles, since the behavior cascaded down from commanders to leaders to followers anyway. The main side effect is that bots will instantly crouch and walk based on their commander's current behavior, which could probably be detected by observant players.  But for now I want to keep the code simple to preserve cycles instead of worrying about how to make the behavior less predictable.

Also, at the risk of being too clever, I intentionally made the check such that even if you're standing still, by pressing the sprint key, you can urge your bot followers to follow a waypoint at a running pace.

I also bundled in a check to match the commander's current crouch behavior.  I think at this point if this PR goes in as is, we don't need #1930.  My thinking is that just matching the commander's current crouch behavior is enough as it resolves this concern I had in that discussion:

> This might cause an awkward situation when a commander is pinned down behind cover while crouching and if they wanted to have their followers run to a ping waypoint they would need to stand up. Otherwise the bots would move to the waypoint while crouch walking.

In combination with #2000, bots would move at a faster pace when they start attacking the enemy or retreating from gunfire, so I don't think we need to worry about bots getting stuck crouch walking to their destination.

Regarding:

> Another place where I could see this footstep audibility stuff being useful is for un-controlled bots in the vicinity of human players, since again it could be annoying to have your sneak ruined by a clomping bot alerting enemies to your area, but possibly out of scope for this PR.

This point I would prefer to discuss outside of this PR, because while the pointer to the commander is readily available, my impression is that it would take more calculations to figure out which friendly players are nearby to the bot.

src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
```diff
@@ -185,6 +185,34 @@ bool CNEOBotCommandFollow::FollowCommandChain(CNEOBot* me)
...
 
+	// Sneak if commander is currently quiet
+	bool commanderIsQuiet = false;
+	if (pCommander->IsWalking())
+	{
+		commanderIsQuiet = true;
+	}
+	else if ( !(pCommander->m_nButtons & (IN_FORWARD | IN_BACK | IN_MOVELEFT | IN_MOVERIGHT | IN_JUMP | IN_SPEED | IN_RUN )) )
+	{
+		// Even while standing still, press IN_SPEED / IN_RUN to urge bots to move faster
```
Rainyan, line 198: Not sure how complex this would be to implement (again I'm fine with leaving this feature out of scope), but similar to the sprint key urging the bot to run, could we also have the walk key urge the bot to sneak, regardless of the initial follow speed? Since it seems that currently, holding down `IN_WALK` for a running follower won't make them switch to sneak mode. This can happen for a commander who was not sneaky when the follow command was given, resulting in running followers, but later the commander wants to become sneaky and pass that order along to their follower who is currently beelining towards their commander from afar.

Repro steps:

* Don't be sneaky
* Press the use key to add a follower in non-sneaky mode
* Start being sneaky
* Press the walk key
* Expect the bot to be urged to also walk

Rainyan (dismissed):
Cool, seems to work nicely (and the implementation is much cleaner than my suggestion 😄)

src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
sunzenshen: It turns out that a previous assumption in  PlayerLocomotion::Approach (src/game/server/NextBot/Player/NextBotPlayerLocomotion.cpp) made it so that if a bot was already going at a running pace, PressRunButton would be continuously pressed down.  This made it so that if I ever invoked any path calculation after calling ReleaseRunButton, that keypress flag would have gotten overridden by the continue running logic.

https://github.com/NeotokyoRebuild/neo/pull/1937/changes/4ef1ef47fbd7fcbd62000796d5663539b33ebfc3#diff-758acc2bcd5d95b2cd59732802c600d322d7f59c20259c41485a3f57deb0e474R585

What worked for me was to check if the walk input flag was set, and to skip the continue running check if that was the case.

This has the side effect that if you are standing still, you need to hold the shift key to keep the bots running.  That might be unintuitive, but maybe it's also the the most reactive thing to do, to allow players to stop making noise ASAP with their bot followers.

Rainyan (approved):

AdamTadeusz (approved):

PR #1938, opened 2026-04-09
Player Water Effects

### LINKED ISSUE #1781: If you die to an enemy in water and then spectate a teammate, the footsteps of the enemy will sound as if they are still in water even if they are not
### Build Info

v27.1

### Description

Dying to a player who is in water will cause you to hear their footsteps as if theyre still in water even if they move out of the water

### To Reproduce

1. Die to an enemy who is in water
2. Spectate a teammate

### Expected behavior

The footsteps of the player you died to should change sounds when they get out of the water

### Actual behavior

You will hear the footsteps of the enemy as if they were still in water even if they aren't. Teammate who is being spectated cannot those steps

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

"When playing on maps with water we found that if you saw an enemy in water, died and then spectated a teammate, that enemy would keep max volume water steps even when they were no longer in water.

The teammate you were spectating wouldn't hear those steps, but you would. Dont know if this is relevant to the above question or a known issue."

### LINKED ISSUE #1782: Initial splash when jumping into water is louder than other splashes
### Build Info

v27.1

### Description

When jumping into water, the first splash is louder than the ones afterwards

### To Reproduce

Jump in water

### Expected behavior

all splashed have the same volume

### Actual behavior

first splash is louder

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

https://github.com/user-attachments/assets/261a5338-aeac-4d9c-92f5-b9217193b5f5

AdamTadeusz:
## Description

Network m_nWaterLevel to all players so players make the correct footstep noises when in water.
Duplicated the watersplash effect for players where the volume of the effect is halved.
Started dispatching the waterriples and watersplash effects for players client side for the local player.

## Toolchain

- Windows MSVC VS2022

## Linked Issues

- fixes #1781
- fixes #1782

### DIFF
--- a/src/game/client/fx_water.cpp
+++ b/src/game/client/fx_water.cpp
@@ -248,6 +248,139 @@ void FX_GunshotSplash( const Vector &origin, const Vector &normal, float scale )
 	C_BaseEntity::EmitSound( filter, SOUND_FROM_WORLD, ep );
 }
 
+#ifdef NEO
+void FX_PlayerSplash( const Vector &origin, const Vector &normal, float scale )
+{
+	VPROF_BUDGET( "FX_PlayerSplash", VPROF_BUDGETGROUP_PARTICLE_RENDERING );
+	
+	if ( cl_show_splashes.GetBool() == false )
+		return;
+
+	Vector	color;
+	float	luminosity;
+	
+	// Get our lighting information
+	FX_GetSplashLighting( origin + ( normal * scale ), &color, &luminosity );
+
+	float flScale = scale / 8.0f;
+
+	if ( flScale > 4.0f )
+	{
+		flScale = 4.0f;
+	}
+
+	// Setup our trail emitter
+	CSmartPtr<CTrailParticles> sparkEmitter = CTrailParticles::Create( "splash" );
+
+	if ( !sparkEmitter )
+		return;
+
+	sparkEmitter->SetSortOrigin( origin );
+	sparkEmitter->m_ParticleCollision.SetGravity( 800.0f );
+	sparkEmitter->SetFlag( bitsPARTICLE_TRAIL_VELOCITY_DAMPEN );
+	sparkEmitter->SetVelocityDampen( 2.0f );
+	sparkEmitter->GetBinding().SetBBox( origin - Vector( 32, 32, 32 ), origin + Vector( 32, 32, 32 ) );
+
+	PMaterialHandle	hMaterial = ParticleMgr()->GetPMaterial( "effects/splash2" );
+
+	TrailParticle	*tParticle;
+
+	Vector	offDir;
+	Vector	offset;
+	float	colorRamp;
+
+	//Dump out drops
+	for ( int i = 0; i < 16; i++ )
+	{
+		offset = origin;
+		offset[0] += random->RandomFloat( -8.0f, 8.0f ) * flScale;
+		offset[1] += random->RandomFloat( -8.0f, 8.0f ) * flScale;
+
+		tParticle = (TrailParticle *) sparkEmitter->AddParticle( sizeof(TrailParticle), hMaterial, offset );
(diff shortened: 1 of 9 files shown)


AdamTadeusz:
FX_PlayerSplash is a copy of FX_GunshotSplash, and FX_PlayerSlimeSplash is a copy of FX_GunshotSlimeSplash, with the sound volume halved. While lowering the sound level would make the noise drop faster over distance, im worried this may sound weird. SNDLVL_NORM sounds (from the name) like the falloff sound level we want for the effect, so I only made the splash quieter for now

src/game/shared/neo/neo_player_shared.cpp
```diff
@@ -393,3 +400,35 @@ void CNEO_Player::CheckAimButtons()
...
+		data.m_fFlags |= FX_WATER_IN_SLIME;
+	}
+	
+#ifdef GAME_DLL
+	CPASFilter filter( data.m_vOrigin );
+#else
+	CLocalPlayerFilter filter;
+#endif // GAME_DLL
+
+	float flSpeed = GetAbsVelocity().Length();
```
sunzenshen: Maybe could be LengthSqr to avoid a square root calculation.

src/game/client/fx_water.cpp
```diff
@@ -248,6 +248,139 @@ void FX_GunshotSplash( const Vector &origin, const Vector &normal, float scale )
...
+		colorRamp = random->RandomFloat( 0.75f, 1.25f );
+
+		tParticle->m_color.r = MIN( 1.0f, color[0] * colorRamp ) * 255;
+		tParticle->m_color.g = MIN( 1.0f, color[1] * colorRamp ) * 255;
+		tParticle->m_color.b = MIN( 1.0f, color[2] * colorRamp ) * 255;
+		tParticle->m_color.a = luminosity * 255;
+	}
+
+	// Setup the particle emitter
+	CSmartPtr<CSplashParticle> pSimple = CSplashParticle::Create( "splish" );
```
sunzenshen, line 324: Just to sanity check, but is this intentionally not "splash" to not conflict with an exiting name?

AdamTadeusz: The parameter is called debugName so I don't think it should matter, but CSplashParticle::Create uses splish everywhere far as I can tell

Image

Rainyan (approved):
Somewhat related to #1782: in OGNT, if you sneak (below footstep velocity) into or out of a pool of water, the waterlevel transition between water/land doesn't play the splash sound cue, whereas for NTRE it seems we currently always play it. Maybe it'd be good to implement the walking check for parity, if you think it fits this PR.

I guess the logic at `SPLASH_SPEED_THRESHOLD` would be the thing that controls this? Or rather, for the walk case, the "waterripple" wouldn't be played by OGNT I presume

sunzenshen (approved):

AdamTadeusz:
> Somewhat related to #1782: in OGNT, if you sneak (below footstep velocity) into or out of a pool of water, the waterlevel transition between water/land doesn't play the splash sound cue, whereas for NTRE it seems we currently always play it. Maybe it'd be good to implement the walking check for parity, if you think it fits this PR.
> 
> I guess the logic at `SPLASH_SPEED_THRESHOLD` would be the thing that controls this? Or rather, for the walk case, the "waterripple" wouldn't be played by OGNT I presume

I opened https://github.com/NeotokyoRebuild/neo/issues/1952 to address this in the future in another PR

PR #1941, opened 2026-04-11
Neo mp3 tweaks

AdamTadeusz:
## Description

Moved the mp3 element left and down, element now moves up when joining the auto-queue
Removed the hamburger icon, song name instead can be clicked to select a song
Song name now scrolls if name of song too long to fit inside the mp3 player, with a 2 second delay
Right align song duration label
~~Progress slider doesn't ignore the first click if not active before click~~
Hovering over buttons in the mp3 player doesn't give off button hover sounds
Removed the second scrollbar visible to the right of the song selection popup scrollbar
Switched the play and pause "icons" around when music is playing/paused
Changed pause "icon" character to pipe

## Toolchain

- Windows MSVC VS2022


AdamTadeusz:
Apparently miniaudio allows smooth volume transitions, but sadly I can't figure out how to get those to work.

src/game/client/neo/ui/neo_ui.h
```diff
@@ -453,6 +453,12 @@ struct Context
 	int *piTableVisColsWide = nullptr;
 };
 
+enum ButtonFlag_
+{
+	BUTTONFLAG_NONE = 0,
+	BUTTONFLAG_SCROLLTEXT = 1 << 0, // If the button text is too long, scroll to show the whole text over time
+};
+
```
nullsystem, line 462: Need `typedef int ButtonFlags` and use that for function parameter.

src/game/client/neo/ui/neo_ui.h
```diff
@@ -534,6 +540,7 @@ void EndOverrideFgColor();
 /*SW*/ void HeadingLabel(const wchar_t *wszText);
 /*1W*/ void Label(const wchar_t *wszText, const bool bNotWidget = false);
 /*1W*/ void Label(const wchar_t *wszText, const LabelExOpt &opt);
+/*1W*/ void Label(const wchar_t *wszText, const TextStyle textAlignment);
```
nullsystem: Redundent to `LabelExOpt` version above.

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -1733,9 +1788,9 @@ NeoUI::RetButton ButtonCheckbox(const wchar_t *wszText, const bool bVal)
 	return BaseButton(wszText, "", BASEBUTTONTYPE_CHECKBOX, bVal);
 }
 
-NeoUI::RetButton ButtonToggle(const wchar_t *wszText, const bool bVal)
+NeoUI::RetButton ButtonToggle(const wchar_t *wszText, const bool bVal, ButtonFlag_ flags, const float scrollStart)
```
nullsystem: `ButtonFlag_` -> `const ButtonFlags`

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -1439,13 +1439,22 @@ void Label(const wchar_t *wszText, const LabelExOpt &opt)
...
 }
 
+void Label(const wchar_t *wszText, const TextStyle textAlignment)
+{
+	const TextStyle bkup = c->eLabelTextStyle;
+
+	c->eLabelTextStyle = textAlignment;
+	Label(wszText);
+	c->eLabelTextStyle = bkup;
+}
```
nullsystem: Redundent to `LabelExOpt` version.

src/game/client/neo/ui/neo_ui.h
```diff
@@ -556,12 +563,12 @@ struct TabsState
 		const TabsFlags flags = TABFLAG_DEFAULT,
 		TabsState *pState = nullptr);
 /*1W*/ RetButton BaseButton(const wchar_t *wszText, const char *szTexturePath,
-		const EBaseButtonType eType, const bool bVal = false);
+		const EBaseButtonType eType, const bool bVal = false, const ButtonFlag_ flags = BUTTONFLAG_NONE, const float scrollOffset = 0.f);
```
nullsystem: Inconsistent h vs cpp function parameter naming: `scrollOffset` vs `scrollStart`. Also name style need to be `iScrollOffset`/`iScrollStart` instead.

Float styling change from `0.f` to `0.0f`.

src/game/client/neo/ui/neo_ui.h
```diff
@@ -556,12 +563,12 @@ struct TabsState
...
 /*1W*/ RetButton Button(const wchar_t *wszText);
 /*2W*/ RetButton Button(const wchar_t *wszLeftLabel, const wchar_t *wszText);
 /*1W*/ RetButton ButtonTexture(const char *szTexturePath);
 /*1W*/ RetButton ButtonCheckbox(const wchar_t *wszText, const bool bVal);
-/*1W*/ RetButton ButtonToggle(const wchar_t *wszText, const bool bVal);
+/*1W*/ RetButton ButtonToggle(const wchar_t *wszText, const bool bVal, const ButtonFlag_ flags = BUTTONFLAG_NONE, const float scrollOffset = 0.f);
```
nullsystem: Inconsistent h vs cpp function parameter naming: `scrollOffset` vs `scrollStart`. Also name style need to be `iScrollOffset`/`iScrollStart` instead.

Float styling change from `0.f` to `0.0f`.

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -1471,10 +1480,56 @@ NeoUI::RetButton BaseButton(const wchar_t *wszText, const char *szTexturePath, c
 				vgui::surface()->DrawFilledRectArray(&c->rWidgetArea, 1);
 
 				const auto *pFontI = &c->fonts[c->eFont];
-				const int x = XPosFromText(wszText, pFontI, c->eButtonTextStyle);
+				int x = XPosFromText(wszText, pFontI, c->eButtonTextStyle);
```
nullsystem: Put `const` back to x, see other comment on `fmod`.

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -1471,10 +1480,56 @@ NeoUI::RetButton BaseButton(const wchar_t *wszText, const char *szTexturePath, c
...
+					{ // Haven't started scrolling yet, but still need to only draw within the bounds of the button
+						vgui::surface()->DrawSetTextPos(x, y);
+						vgui::surface()->DrawPrintText(wszText, wszTextLen);
+					}
+					else
+					{ // Scrolling text
+						textWidth += textHeight;
+						// 1 pixel per second scroll speed if using just time, scale by size of font so text scrolls relatively fast even on large displays.
+						x -= fmod((gpGlobals->curtime - scrollStart) * textHeight, textWidth);
+						const int x2 = x + textWidth;
```
nullsystem: Keep `x` const:

`const int x2 = x - fmod((gpGlobals->curtime - scrollStart) * textHeight, textWidth) + textWidth;`

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -1471,10 +1480,56 @@ NeoUI::RetButton BaseButton(const wchar_t *wszText, const char *szTexturePath, c
...
+				int x = XPosFromText(wszText, pFontI, c->eButtonTextStyle);
 				const int y = pFontI->iYFontOffset;
-				vgui::surface()->DrawSetTextPos(c->rWidgetArea.x0 + x, c->rWidgetArea.y0 + y);
-				vgui::surface()->DrawPrintText(wszText, V_wcslen(wszText));
+				const int wszTextLen = V_wcslen(wszText);
+				
+				auto normalDraw = [](const int x, const int y, const wchar_t* wszText, const int wszTextLen) {
+					vgui::surface()->DrawSetTextPos(c->rWidgetArea.x0 + x, c->rWidgetArea.y0 + y);
+					vgui::surface()->DrawPrintText(wszText, wszTextLen);
+				};
```
nullsystem: Unnecessary splitting it out to separate function. Can re-arrange the code.

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -1471,10 +1480,56 @@ NeoUI::RetButton BaseButton(const wchar_t *wszText, const char *szTexturePath, c
...
+						vgui::surface()->DrawSetTextPos(x2, y);
+						vgui::surface()->DrawPrintText(wszText, wszTextLen);
+					
+						vgui::surface()->DrawSetTextPos(x, y);
+						vgui::surface()->DrawPrintText(wszText, wszTextLen);
+					}
+					
+					vgui::surface()->PopFullscreenViewport();
+					vgui::surface()->SetFullscreenViewport(0, 0, 0, 0);
+				}
```
nullsystem: Re-arrange: Put this entire scope under an `if (flags & BUTTONFLAG_SCROLLTEXT)`

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -1471,10 +1480,56 @@ NeoUI::RetButton BaseButton(const wchar_t *wszText, const char *szTexturePath, c
...
+
+						vgui::surface()->DrawSetTextPos(x2, y);
+						vgui::surface()->DrawPrintText(wszText, wszTextLen);
+					
+						vgui::surface()->DrawSetTextPos(x, y);
+						vgui::surface()->DrawPrintText(wszText, wszTextLen);
+					}
+					
+					vgui::surface()->PopFullscreenViewport();
+					vgui::surface()->SetFullscreenViewport(0, 0, 0, 0);
```
nullsystem: Re-arrange: Put this scope under an `if (textWidthOver > 0)`

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -1471,10 +1480,56 @@ NeoUI::RetButton BaseButton(const wchar_t *wszText, const char *szTexturePath, c
...
+				}
 			} break;
```
nullsystem, line 1530: Re-arrange: Then on end just check on a `bool` if it's drawn or not to then draw the normal text.

src/game/client/neo/ui/neo_root.cpp
```diff
@@ -1151,18 +1151,22 @@ void CNeoRoot::MainLoopRoot(const MainLoopParam param)
...
+		const int NUM_ROWS = 4;
+		g_uiCtx.dPanel.x = param.wide - flMP3Wide - g_uiCtx.iMarginX;
+		g_uiCtx.dPanel.y = param.tall - (NUM_ROWS * g_uiCtx.layout.iRowTall) - g_uiCtx.iMarginY;
+		if (m_serverPingAutoJoin.m_serverInfo.m_NetAdr.GetIP() != 0)
+			g_uiCtx.dPanel.y -= g_uiCtx.layout.iDefRowTall;
 		g_uiCtx.dPanel.wide = flMP3Wide;
 		g_uiCtx.dPanel.tall = param.tall;
 
-		NeoUI::BeginSection(NeoUI::SECTIONFLAG_PLAYBUTTONSOUNDS);
+		NeoUI::BeginSection();
```
nullsystem, line 1164: Why drop button sounds for MP3 player?

src/game/client/neo/ui/neo_root.cpp
```diff
@@ -1151,18 +1151,22 @@ void CNeoRoot::MainLoopRoot(const MainLoopParam param)
 	{
 		// Close equivalence to ~400px in 1080p (as 4:3 so 1440x1080)
 		const float flMP3Wide = 0.28f * m_flWideAs43;
-		g_uiCtx.dPanel.x = param.wide - flMP3Wide;
-		g_uiCtx.dPanel.y = param.tall - (flMP3Wide / 2) - g_uiCtx.layout.iRowTall;
+		const int NUM_ROWS = 4;
+		g_uiCtx.dPanel.x = param.wide - flMP3Wide - g_uiCtx.iMarginX;
+		g_uiCtx.dPanel.y = param.tall - (NUM_ROWS * g_uiCtx.layout.iRowTall) - g_uiCtx.iMarginY;
+		if (m_serverPingAutoJoin.m_serverInfo.m_NetAdr.GetIP() != 0)
+			g_uiCtx.dPanel.y -= g_uiCtx.layout.iDefRowTall;
```
nullsystem, line 1159: Styling: Wrap single line indent with `{ }`

src/game/client/neo/ui/neo_ui.cpp
nullsystem: Scrolling: I think it'll be worth to at least put some few seconds pause when it's at the start/restart of the scroll. There's also a bit of a jitter when it restarts the scroll currently.

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -1471,10 +1480,56 @@ NeoUI::RetButton BaseButton(const wchar_t *wszText, const char *szTexturePath, c
...
+				int textWidth = 0, textHeight = 0;
+				vgui::surface()->GetTextSize(pFontI->hdl, wszText, textWidth, textHeight);
+				const int textWidthOver = textWidth - c->dPanel.wide;
+				if (textWidthOver <= 0)
+				{
+					normalDraw(x, y, wszText, wszTextLen);
+					break;
+				}
+
+				{ // Draw with scrolling text
```
nullsystem: Add a `NEO TODO` comment in here noting this might be useful for use for plain labels. This might get refactored out as a label in the future but not necessary now.

src/game/client/neo/ui/neo_root.cpp
```diff
@@ -1238,7 +1268,7 @@ void CNeoRoot::MainLoopRoot(const MainLoopParam param)
 		}
 
 		// Play/Pause button
-		if (NeoUI::Button(mps->bPlaying ? L"\u25B6" : L"II").bPressed)
+		if (NeoUI::Button(mps->bPlaying ? L"||" : L"\u25B6").bPressed)
```
nullsystem: `||` looks a little too tall as a pause indicator than `II` in my opinion. Unfortantly the current unicode/emoji set is also unsuitable so I think stick with `II`.

src/game/client/neo/ui/neo_root.cpp
```diff
@@ -1252,28 +1282,11 @@ void CNeoRoot::MainLoopRoot(const MainLoopParam param)
...
-							.wide = g_uiCtx.dPanel.wide,
-							.tall = iPopupTall,
-						});
-			}
-		}
-
 		g_uiCtx.eButtonTextStyle = NeoUI::TEXTSTYLE_LEFT;
 		NeoUI::EndSection();
 
+		NeoUI::Dim previousDPanel = g_uiCtx.dPanel; // NEO TODO (Adam) mzync please fix this properly, I don't know how. Stops a second y scrollbar from popping up next to the below popup
```
nullsystem: This is noted as a TODO in `neo_ui.h`: `// NEO TODO (nullsystem): Popups should get its own XY offsets`.

nullsystem: Also I'm not seeing an issue here removing this currently, the other y scrollbar didn't popup and this `NeoUI::Section` should be large enough it won't?

nullsystem: Also `mzync` -> `nullsystem` if you're to refer in code.

src/game/client/neo/ui/neo_root.cpp
```diff
@@ -1175,7 +1179,33 @@ void CNeoRoot::MainLoopRoot(const MainLoopParam param)
...
-		NeoUI::Label(wszText);
+
+		static wchar_t previousWszText[128] = {};
+		static float scrollStart = gpGlobals->curtime;
+		if (Q_wcscmp(previousWszText, wszText) != 0)
+		{
+			const float SCROLL_DELAY = 2.f;
+			scrollStart = gpGlobals->curtime + SCROLL_DELAY;
+			V_swprintf_safe(previousWszText, wszText);
+		}
```
nullsystem: There's no need for string comparison, can just swap out to just checking the current song's index instead.

nullsystem (changes requested):

src/game/client/neo/ui/neo_ui.cpp
AdamTadeusz: x is used further down

src/game/client/neo/ui/neo_root.cpp
AdamTadeusz: its an mp3 player, I don't want to hear various beeps when adjusting the music that's being played

src/game/client/neo/ui/neo_root.cpp
AdamTadeusz: I was worried this would look bad if someone switched to a different font

src/game/client/neo/ui/neo_root.cpp
AdamTadeusz: You've seen the video I've posted on discord right?

src/game/client/neo/ui/neo_root.cpp
nullsystem: Only deal with it once if there's a different font.

src/game/client/neo/ui/neo_ui.cpp
nullsystem: Instead set:

```
const int iXDiff = fmod((gpGlobals->curtime - scrollStart) * textHeight, textWidth);
const int x2 = x - iXDiff + textWidth;
// ...
vgui::surface()->DrawSetTextPos(x - iXDiff, y);
```

src/game/client/neo/ui/neo_root.cpp
nullsystem: Still never managed to replicate it but nevertheless it is a known issue so.

src/game/client/neo/ui/neo_ui.cpp
AdamTadeusz: check on a bool meaning you don't want early breaks instead?

src/game/client/neo/ui/neo_ui.cpp
AdamTadeusz: There used to be jitter when I was using curtime in one place and realtime in another place, but after fixing that I'm not seeing any jitter 🤔 . There is of course the problem associated with the scroll being constant, instead of ramping up to the constant speed after the 2 second delay, but it doesn't seem to be too jarring to me

AdamTadeusz: ok looking very closely there is indeed a tiny bit of jitter during the transition

AdamTadeusz: well, adding a short delay between each scroll makes this not noticeable

src/game/client/neo/ui/neo_ui.cpp
nullsystem: Yes, early breaks not needed

src/game/client/neo/ui/neo_root.cpp
AdamTadeusz: well there's nothing stopping people from changing the fonts in clientscheme.res, figured if the pipe was already used in next/previous song may as well use it there since capital "I" could have little legs like for example in visual studio or github
image

nullsystem: Not really our concern if the player wants to steer away from the default font. The whole UI designs are around the fonts we set ourselves, not any non-default clientscheme.res and would a much bigger scope to even bother about user-customized theming. Not saying we "shouldn't" support custom theme, but I wouldn't bother about it and just focus on what we set ourselves.

nullsystem: Also the proper way to deal with this is just to have some symbols font that gives out plain media controls visuals (or even make it ourselves) we can use just for the controls instead.

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -1733,9 +1783,9 @@ NeoUI::RetButton ButtonCheckbox(const wchar_t *wszText, const bool bVal)
 	return BaseButton(wszText, "", BASEBUTTONTYPE_CHECKBOX, bVal);
 }
 
-NeoUI::RetButton ButtonToggle(const wchar_t *wszText, const bool bVal)
+NeoUI::RetButton ButtonToggle(const wchar_t *wszText, const bool bVal, ButtonFlags flags, const float flScrollStart)
```
nullsystem: `ButtonFlags flags` -> `const ButtonFlags flags`, this even differs from the header file which does have `const`

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -1473,8 +1473,58 @@ NeoUI::RetButton BaseButton(const wchar_t *wszText, const char *szTexturePath, c
...
+					vgui::surface()->GetTextSize(pFontI->hdl, wszText, textWidth, textHeight);
+					const int textWidthOver = textWidth - c->dPanel.wide;
+					if (textWidthOver > 0)
+					{ // Draw with scrolling text // NEO TODO (Adam) could be useful for plain labels too
+						vgui::surface()->SetFullscreenViewport(c->dPanel.x, c->dPanel.y, c->dPanel.wide, c->dPanel.tall);
+						vgui::surface()->PushFullscreenViewport();
+
+						{ // Scrolling text
+							textWidth += textHeight; // Separation between two draws of wszText
+							const float PAUSE_BETWEEN_SCROLLS_SECONDS = 2.0f;
```
nullsystem: I think the pause duration is too short, doesn't give times to read at non-moving state. Bump it to 5 seconds.

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -1473,8 +1473,58 @@ NeoUI::RetButton BaseButton(const wchar_t *wszText, const char *szTexturePath, c
...
+				const int wszTextLen = V_wcslen(wszText);
+
+				bool drawn = false;
+				if (flags & BUTTONFLAG_SCROLLTEXT)
+				{
+					int textWidth = 0, textHeight = 0;
+					vgui::surface()->GetTextSize(pFontI->hdl, wszText, textWidth, textHeight);
+					const int textWidthOver = textWidth - c->dPanel.wide;
+					if (textWidthOver > 0)
+					{ // Draw with scrolling text // NEO TODO (Adam) could be useful for plain labels too
```
nullsystem: NIT: Move the comments to its own line rather than at the side of `{` bracket

nullsystem (changes requested):

src/game/client/neo/ui/neo_ui.cpp
AdamTadeusz: You're not supposed to be able to read the whole text in the pause duration, just the start that's about to disappear. 5 seconds is way too long

src/game/client/neo/ui/neo_root.cpp
```diff
@@ -1151,22 +1151,28 @@ void CNeoRoot::MainLoopRoot(const MainLoopParam param)
...
 		g_uiCtx.eButtonTextStyle = NeoUI::TEXTSTYLE_LEFT;
 
 		wchar_t wszText[128] = {};
 
+		// Track Selection
 		NeoUI::SetPerRowLayout(1);
 		if (mps->songs[mps->iCurIdx].wszArtist[0])
 		{
-			V_swprintf_safe(wszText, L"%ls - %ls",
+			V_swprintf_safe(wszText, L"%ls - %ls aaaaaaaaaaaaaaaaaaaaaaaaaaaa",
```
nullsystem: Left out your a's

nullsystem (changes requested):

nullsystem (approved):
Cleanup your commit message when you squash

sunzenshen (approved):

Issue #1942, opened 2026-04-12
Selecting bots under the cursor to add to a player's squad or to ask for a weapon drop should be done client side


AdamTadeusz:
Currently adding bots to a player's squad is done server side in CNEO_Player::PlayerUse. The traceline done in this function does not benefit from lag compensation, meaning the player has to aim in front of a bot if the bot they are trying to select is running perpendicular to the player, with the amount of lead depending on the players ping, relative speed and angle etc.

Instead what we should do is override the client side PlayerUse function. When a valid bot is found via a trace client side, the client should then request that the bot be added to the local player's squad. The server can then check if this request is valid, and if it is, assign the bot to the requesting player.

The same should be done for requesting a bot to drop a weapon

PR #1943, opened 2026-04-12
Bots message their commander when leaving the squad

sunzenshen:
## Description
Bots message commander about more state changes:
- Notify when leaving the squad
- Notify when joining another player's squad
- Notify when changing squad star (alpha, bravo, etc)

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #1942 (potential collision of lines)

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_command_follow.cpp
@@ -36,7 +36,7 @@ CNEOBotCommandFollow::CNEOBotCommandFollow() : m_vGoalPos( CNEO_Player::VECTOR_I
 //---------------------------------------------------------------------------------------------
 ActionResult< CNEOBot >	CNEOBotCommandFollow::OnStart(CNEOBot *me, Action< CNEOBot > *priorAction)
 {
-	SendUpdateToCommander( me, "Joining your squad." );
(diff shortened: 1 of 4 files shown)


sunzenshen:
Demo:

https://github.com/user-attachments/assets/8dc112c5-7c36-45af-a39c-1197f708e6f6

Helpful script to prevent bots from running everywhere when testing squad disconnects:
```
sv_neo_preround_freeze_time 2; mp_chattime 2;
sv_neo_bot_cmdr_enable 1;
neo_ctg_round_timelimit 99;
```

To test bots notifying that they are joining another player's squad, one can use `bot_mimic 1` to have bots press the use key on another bot.

AdamTadeusz (approved):

AdamTadeusz:
I think this definitely could be improved a bit. Not sure if the chat is the right place for this information (though many games do this), and I'm not confident that replacing the squad name with the commander's name is a good way to differentiate the bot joining the squad the local player is in from the bot joining some other squad.

The arma series does the same thing you're doing here, it will also highlight messages that are relevant to the local player with an inverted colour scheme to separate them from chatter that's relevant to other squad members or chatter that's relevant to the whole squad
ingame_16x10
I think this could be an interesting avenue of investigation, but probably for another PR

nullsystem (approved):

PR #1944, opened 2026-04-13
Expand context hint hud element, picking up weapons with use key

### LINKED ISSUE #1942: Selecting bots under the cursor to add to a player's squad or to ask for a weapon drop should be done client side

### LINKED ISSUE #1957: All weapons should be pickup-able with the use key
In og weapons can be picked up with the use key as long as the corresponding weapon slot is empty, with the ghost being the exception and being pickupable always. It would be better if the way the ghost can be picked up worked the same for all weapons.

### LINKED ISSUE #1958: Utility weapons with more than one ammo should be duplicated into multiple weapons with one ammo each when dropped on death
A smoke grenade "weapon" has a reserve ammo pool of 2, meaning that when picking up a seemingly single smoke grenade off the ground, the player may be able to then throw two smoke grenades. It would make more sense visually if, upon death, a single smoke grenade "weapon" dropped for every ammo in the reserve pool

### LINKED ISSUE #1959: Players should see contextual hints when they can interact with objects
Expand the takeover player hint to show weapons that can be picked up, objects interacted with, bots added to squad

AdamTadeusz:
## Description

- Adds contextual hints when the player is hovering over most interactable objects
- Option to highlight the interactable object
- Allows weapon pickup with the use key. The weapon in the corresponding slot (if any) will be automatically dropped (currently even if the new weapon cannot be picked up sometimes)
- On death utility will now drop with one ammo, and a duplicate utility weapon with one ammo will drop for each extra ammo the original weapon had
- Changes how ghost startup sound is played and stopped to work using the ghost with the use key, shows the uplink state hud element if the ghost is not active but sv_neo_ctg_ghost_beacons_when_inactive is enabled, hides uplink hud element after the round is over (when ghost beacons can no longer be drawn)
- Requesting bot to drop a weapon or join squad is now done client side

## Toolchain

- Windows MSVC VS2022

## Fixes
-fixes #1959
-fixes #1957
-fixes #1942
-fixes #1958


src/game/server/neo/neo_player.cpp
```diff
@@ -2686,15 +2722,15 @@ bool CNEO_Player::BumpWeapon( CBaseCombatWeapon *pWeapon )
 		}
 	}
 
-	// We need to run this for its side-effects, even in the IsDead case below... should be refactored.
-	const bool okRet = BaseClass::BumpWeapon(pWeapon);
```
AdamTadeusz, line 2710: CHL2MP_Player::BumpWeapon calls Weapon_Equip and generally i figured we should follow the same logic as in CBasePlayer::GiveNamedItem( after the misyl fix down in CNEO_Player::GiveLoadoutWeapon(

src/game/client/neo/c_neo_player.cpp
```diff
@@ -2104,3 +2094,62 @@ const char* C_NEO_Player::GetPlayerNameWithTakeoverContext(int player_index)
...
+	return nullptr;
+}
+
+
+//-----------------------------------------------------------------------------
+// Purpose: Return true if this object can be +used by the player
+//-----------------------------------------------------------------------------
+bool C_NEO_Player::IsUseableEntity( CBaseEntity *pEntity, unsigned int requiredCaps )
+{
+#ifdef NEO
```
Rainyan: Probably don't need the NEO ifdef guards, since we're in a neo-exclusive file (and method)

src/game/client/neo/ui/neo_hud_context_hint.cpp
```diff
@@ -70,88 +71,209 @@ bool CNEOHud_ContextHint::ShouldDraw()
...
+	ClearUseEntityListEntry();
 
-	auto eObserverMode = pLocalPlayer->GetObserverMode();
-	bool bIsSpectating = (eObserverMode == OBS_MODE_CHASE || eObserverMode == OBS_MODE_IN_EYE);
+	C_NEO_Player* pLocalNeoPlayer = C_NEO_Player::GetLocalNEOPlayer();
+	if (!pLocalNeoPlayer)
+		return;
 
-	if (pObserverTargetPlayer != m_hLastSpecTarget.Get())
+	char szUppercaseKeyBinding[16]; // Assuming keybinds won't exceed 15 characters + null terminator
```
Rainyan, line 98: Maybe assert that strlen of `useKeyBinding` < sizeof this alloc, so that we get a warning if this assumption (the one in the comment message) ever fails

src/game/client/neo/ui/neo_hud_context_hint.cpp
```diff
@@ -70,88 +71,209 @@ bool CNEOHud_ContextHint::ShouldDraw()
...
+	if (useKeyBinding && useKeyBinding[0] != '\0')
+	{
+		V_strncpy(szUppercaseKeyBinding, useKeyBinding, sizeof(szUppercaseKeyBinding));
+		V_strupr(szUppercaseKeyBinding);
+	}
+	else
 	{
-		m_bHintShownForCurrentSpecTarget = false;
-		m_hLastSpecTarget = pObserverTargetPlayer;
+		const char notBoundText[] = "+use unbound\0";
```
Rainyan: C style string literals have an implicit trailing null terminator - do we need to manually insert the trailing `\0` here or could it be cleaned up? Since doing `"foobar\0"` actually results in double nullterminator, it will be `"foobar\0\0"` because of the implicit terminator + the extra one. If there is not some edge case where we must have the explicit '\0' that I didn't realise.

src/game/client/neo/ui/neo_hud_context_hint.cpp
```diff
@@ -70,88 +71,209 @@ bool CNEOHud_ContextHint::ShouldDraw()
...
+			
+			// Weapon pickup hint
+			if (pUseEntity->IsBaseCombatWeapon())
+			{
+				C_NEOBaseCombatWeapon* pNeoWeapon = static_cast<C_NEOBaseCombatWeapon*>(pUseEntity);
+				
+				// Ghost pickup hint
+				if (pNeoWeapon->GetNeoWepBits() & NEO_WEP_GHOST)
+				{
+					V_snwprintf(m_wszHintText, ARRAYSIZE(m_wszHintText), L"[%hs] pickup the Ghost", szUppercaseKeyBinding);
```
Rainyan: language nit: "pickup" -> "pick up" when used as a verb and not a noun

src/game/client/neo/ui/neo_hud_context_hint.cpp
```diff
@@ -70,88 +71,209 @@ bool CNEOHud_ContextHint::ShouldDraw()
...
+					m_flDisplayEndTime = gpGlobals->curtime + 1.f;
+				}
+				// Weapon pickup hint
+				else if (pNeoWeapon->CanBePickedUpByClass(pLocalNeoPlayer->GetClass()))
+				{
+					V_snwprintf(m_wszHintText, ARRAYSIZE(m_wszHintText), L"[%hs] pickup %hs", szUppercaseKeyBinding, pNeoWeapon->GetPrintName());
```
Rainyan: language nit: "pickup" -> "pick up"

src/game/client/neo/ui/neo_hud_context_hint.cpp
```diff
@@ -70,88 +71,209 @@ bool CNEOHud_ContextHint::ShouldDraw()
...
+					m_flDisplayEndTime = gpGlobals->curtime + 1.f;
+				}
+				else if (pLocalNeoPlayer->m_nButtons & IN_USE)
+				{
+					V_snwprintf(m_wszHintText, ARRAYSIZE(m_wszHintText), L"Cannot pickup weapon");
```
Rainyan: language nit: "pickup" -> "pick up".

Also this could probably be `V_wcscpy_safe` or similar since there are no varargs for the sprintf.

src/game/client/neo/ui/neo_hud_context_hint.cpp
```diff
@@ -163,13 +285,44 @@ void CNEOHud_ContextHint::DrawNeoHudElement()
...
+	int x = 0;
+	int y = 0;
+	const int size = iTextTall / 8;
+
+	vgui::surface()->DrawSetColor(m_TextColor);
+	for (int i = 0; i <= useEntityListLastIndex; i++)
+	{
+		if (C_BaseEntity* entity = useEntityList[i].entity.Get())
+		{
+			GetVectorInScreenSpace(entity->CollisionProp()->WorldSpaceCenter(), x, y);
```
Rainyan: Would it make sense to early-continue for off-screen elements here based on `GetVectorInScreenSpace` return value?

src/game/server/buttons.cpp
```diff
@@ -78,6 +78,11 @@ END_DATADESC()
 
 LINK_ENTITY_TO_CLASS( func_button, CBaseButton );
 
+#ifdef NEO
+IMPLEMENT_SERVERCLASS_ST( CBaseButton, DT_BaseButton )
+	SendPropInt( SENDINFO(m_spawnflags), 13, SPROP_UNSIGNED )
```
Rainyan: Should this be 14 to accommodate `SF_BUTTON_JIGGLE_ON_USE_LOCKED`?

Doing a naive strlen check in Python:
```
>>> bin(8192)
'0b10000000000000'
>>> _.split("0b")[1]
'10000000000000'
>>> len(_)
14
```

src/game/server/buttons.h
```diff
@@ -16,6 +16,11 @@ class CBaseButton : public CBaseToggle
 public:
 
 	DECLARE_CLASS( CBaseButton, CBaseToggle );
+#ifdef NEO
+	DECLARE_SERVERCLASS();
+
+	int updateTransmitState() { return SetTransmitState(FL_EDICT_PVSCHECK); }
```
Rainyan: This method seems to be unused, maybe remove?

src/game/server/neo/neo_player.cpp
```diff
@@ -2405,6 +2432,33 @@ void CNEO_Player::Weapon_DropAllOnDeath( const CTakeDamageInfo &info )
...
 		}
 
+		if (pNeoWeapon->GetSlot() == NEO_THROWABLES_WEAPON_SLOT)
+		{ // drop individual throwables with ammo counts of 1
+			int numThrowablesCreated = 0;
+			for (int i = 1; i < pNeoWeapon->m_iPrimaryAmmoCount; i++)
+			{
+				CBaseEntity* pEnt = CreateEntityByName(pNeoWeapon->GetClassname());
+				if (!pEnt)
+					break; // Assuming this will not work for all subsequent tries
```
Rainyan: Perhaps also assert to catch this error state

src/game/server/neo/neo_player.cpp
```diff
@@ -2405,6 +2432,33 @@ void CNEO_Player::Weapon_DropAllOnDeath( const CTakeDamageInfo &info )
...
+
+				CNEOBaseCombatWeapon* pNeoEnt = static_cast<CNEOBaseCombatWeapon*>(pEnt);
+				if (!pNeoEnt)
```
Rainyan: We know pEnt is non-null here, and non-null static_cast can never result in a non-null pointer, so this check can be removed since it can never happen.

If we want to verify the conversion validity then perhaps assert via dynamic cast (which would return nullptr for incorrect conversion) for debug builds? For example: `auto pNeoEnt = assert_cast<CNEOBaseCombatWeapon*>(pEnt)` which is shorthand for doing the dynamic_cast assertion for debug, but optimizes to static_cast for release builds.

src/game/server/neo/neo_player.cpp
```diff
@@ -2686,19 +2766,19 @@ bool CNEO_Player::BumpWeapon( CBaseCombatWeapon *pWeapon )
...
-	// We need to run this for its side-effects, even in the IsDead case below... should be refactored.
-	const bool okRet = BaseClass::BumpWeapon(pWeapon);
-
 	// We had some cases of dead players chilling around with visible guns.
 	// While that will be addressed in ShouldDraw, here's a preventive measure
 	// to avoid that situation from occurring altogether.
 	if (IsDead())
 		return false;
 
+	// We need to run this for its side-effects, even in the IsDead case below... should be refactored.
```
Rainyan: This comment becomes outdated if we move the okRet below the IsDead line above, so should probably remove this comment altogether.

Also maybe just `return` directly from the baseclass call so we can get rid of the okRet variable.

src/game/shared/neo/neo_juggernaut.h
```diff
@@ -39,11 +39,14 @@ class CNEO_Juggernaut : public CBaseAnimating
 
 	bool	m_bPostDeath = false;
 #endif
+	virtual int	ObjectCaps(void) { return BaseClass::ObjectCaps() | FCAP_ONOFF_USE; }
```
Rainyan: Nit: preferably mark `override` (I know you're just moving the method in the header, but figured I'd mention it since it was edited)

src/game/shared/neo/neo_player_shared.cpp
```diff
@@ -58,6 +64,17 @@ ConVar sv_neo_ghost_delay_secs("sv_neo_ghost_delay_secs", "3.3", FCVAR_NOTIFY |
...
+ConVar sv_neo_spec_replace_player_bot_enable("sv_neo_spec_replace_player_bot_enable", "1", FCVAR_REPLICATED, "Allow spectators to take over bots.", true, 0, true, 1);
+ConVar sv_neo_spec_replace_player_afk_enable("sv_neo_spec_replace_player_afk_enable", "0", FCVAR_REPLICATED, "Allow spectators to take over AFK players.", true, 0, true, 1);
+ConVar sv_neo_spec_replace_player_min_exp("sv_neo_spec_replace_player_min_exp",
+	"0", FCVAR_REPLICATED,
+	"Minimum experience allowed to takeover players ",
+	true, -999, true, 999);
+ConVar sv_neo_spec_replace_player_afk_time_sec( "sv_neo_spec_replace_player_afk_time_sec",
+	"180", FCVAR_NONE,
+	"Seconds of inactivity before a player is considered AFK for spectator takeover.",
+	true, -1, true, 999);
```
Rainyan, line 76: If we allow the `-1` minimum value, should we also add logic such that the "is AFK" logic will always return false for that special case, as a way to disable the AFK logic?

src/game/shared/neo/neo_player_shared.cpp
```diff
@@ -433,3 +450,260 @@ void CNEO_Player::Splash()
...
+		pFoundByTrace = pObject;
+#endif
+		bool bUsable = IsUseableEntity(pObject, 0);
+		while ( pObject && !bUsable && pObject->GetMoveParent() )
+		{
+			pObject = pObject->GetMoveParent();
+			bUsable = IsUseableEntity(pObject, 0);
+		}
+
+		if ( bUsable && !pObject->IsBaseCombatWeapon() )
```
Rainyan: At line 529 here, is there a risk of `pObject` being nullptr here at the `->` dereference? Maybe safer to do a `if pObject` check to avoid nullptr dereference, unless it absolutely cannot happen

src/game/client/c_buttons.cpp
```diff
@@ -0,0 +1,15 @@
...
+
+LINK_ENTITY_TO_CLASS(func_button, C_BaseButton);
+
+IMPLEMENT_CLIENTCLASS_DT( C_BaseButton, DT_BaseButton, CBaseButton )
+	RecvPropInt( RECVINFO(m_spawnflags) ),
+END_RECV_TABLE()
+
+int C_BaseButton::ObjectCaps(void) {
+	return BaseClass::ObjectCaps() | ((m_spawnflags & SF_BUTTON_USE_ACTIVATES) ? (FCAP_IMPULSE_USE | FCAP_USE_IN_RADIUS) : 0);
+};
```
Rainyan: Nit: unnecessary semicolon at end of function scope at line 15, after the ending `}`

src/game/shared/neo/neo_player_shared.cpp
AdamTadeusz: the way IsUseableEntity works right now it will never happen, but can add a check in case someone overwrites IsUseableEntity

Rainyan (dismissed):
LGTM

Masterkatze (dismissed):
LGTM. Rebased on master and tested locally.

Masterkatze (approved):

sunzenshen (approved):

Issue #1053, opened 2025-03-25
[FEATURE] VPK pack assets


brysondev:
### Description

Pack up assets into VPKs to improve performance for lower-end systems.

### Proposed Solution

Add VPK packing in github actions.

blaberry:
Some notes on this: 
I think we're better off doing this on the build machine for steam:
1. We don't want to distribute all the ognt assets directly from github
2. The default vpk.exe valve ships is broken (surprise surprise) so doing this at build time will be more annoying than just using the tools available.
3. We're already doing a lot of file-mangling on there that requires various third party utilities, adding another one isn't too much.

blaberry:
Alright, we've got the packing working for the latest development build. Seems like there's one issue that surfaces in a couple of places: Some ambient sounds don't play. So far motion vision sound and recon cloak sounds have been reported as not playing.

Without looking at the code, my assumption is that these custom ambient sounds are loaded using naive c++ fopen calls or something similar - just like how weapon scripts and loaded - instead of whatever engine method should be used to load files from either VPKs or disk. We can probably fix this for these sounds. I'll assign this to @Rainyan to either delegate to someone or diagnose/fix.

Rainyan:
> Alright, we've got the packing working for the latest development build. Seems like there's one issue that surfaces in a couple of places: Some ambient sounds don't play. So far motion vision sound and recon cloak sounds have been reported as not playing.
> 
> Without looking at the code, my assumption is that these custom ambient sounds are loaded using naive c++ fopen calls or something similar - just like how weapon scripts and loaded - instead of whatever engine method should be used to load files from either VPKs or disk. We can probably fix this for these sounds. I'll assign this to [@Rainyan](https://github.com/Rainyan) to either delegate to someone or diagnose/fix.

@blaberry Should be resolved by #1950 once merged

blaberry:
Awesome, good job! And thanks :)

Rainyan:
@blaberry Sorry about the delay, but as a heads up, the abovementioned blocker should now be cleared with #1950 merged to master, so I'll un-assign myself from this.

blaberry:
Thanks! We've started packing and signing VPKs for sounds etc in the latest development builds, was this included in v29? If not, I guess it'll start working from v30 :)

Rainyan:
> Thanks! We've started packing and signing VPKs for sounds etc in the latest development builds, was this included in v29? If not, I guess it'll start working from v30 :)

[...]

PR #1950, opened 2026-04-17
Fix missing script sound entries for VPK build

Rainyan:
## Description
Address the script sound problem at https://github.com/NeotokyoRebuild/neo/issues/1053#issuecomment-4213357081 by manually bringing in the relevant entries from OGNT scripts.

Fixes (among other things) the following sound errors:

```
PrecacheScriptSound 'NeoPlayer.GhostPing' failed, no such sound script entry
PrecacheScriptSound 'NeoPlayer.ThermOpticOn' failed, no such sound script entry
PrecacheScriptSound 'NeoPlayer.ThermOpticOff' failed, no such sound script entry
PrecacheScriptSound 'NeoPlayer.VisionOn' failed, no such sound script entry
PrecacheScriptSound 'Victory.Draw' failed, no such sound script entry
PrecacheScriptSound 'Victory.Jinrai' failed, no such sound script entry
PrecacheScriptSound 'Victory.NSF' failed, no such sound script entry
```

The added `game_sounds_player.txt` file is verbatim from OGNT definitions, except for:
* 435f88ea87 removes the unused rank up/down SFX
* 602d9698cf removes the `Victory.Jinrai, ...` entries, because we instead define them as `HUD.JinraiWin, ...` (and also call them directly in code due to needing to adjust the volume based on the jingle volume cvar so we can't rely on a hardcoded script volume)

This includes some HL2/"shared" definitions as they are in OGNT; I figured we probably want to support these for map compat, even if things like the geiger counter sounds aren't directly relevant, so they are not removed from `game_sounds_player.txt`. And some of the shared sounds we definitely want, such as the weapon pickup SFX.

Other minor changes:

* 91cdade550 Fix the missing `HL2MPTypeDeath` font nag - we don't use this HL2MP kill icon font for anything so I simply removed it entirely
* f64d1b6bc1 Refactor: Use the correct channel definitions for `HUD.GhostPickUp`
* 6744e99330 Update the `g_pChannelNames` values - these are mostly unused because the closed Valve dlls have them compiled in the binary, but might as well make our implementation match them correctly

## Steps to test

* Download a fresh development branch version of NT;RE from Steam (i.e. preferably no existing files present)
  * Steam depot buildid [22739584](https://steamdb.info/app/3172910/history/?changeid=35160216) (current latest dev build as of 2026/04/17)
  * neo\_version:
```
Build version: 20260410_8714773
Build date: 2026-04-10
Git hash: 8714773c1dd5457bc047285d515bad8803a199d7
OS: Windows
```

* Launch the game, and enter the following to dev console:
```
developer 1;
clear;
map devbox;
```

Upon loading into the map:

### Before this patch

* Console prints the above `PrecacheScriptSound ... failed, no such sound script entry` errors, and those erroring sounds don't play

### With this patch

* Build this PR, and copy the dlls + the new `game/neo/scripts/game_sounds_player.txt` and other modified assets to the Steam app path
* Launch again with the above steps
* No script sound console errors, and the relevant sounds play correctly

## Toolchain
- Windows MSVC VS2022

## Linked Issues

- related #1053


game/neo/scripts/game_sounds_player.txt
```diff
@@ -0,0 +1,317 @@
...
+	"wave"	"plats/train_use1.wav"
+}
+
+"Player.UseDeny"
+{
+	"channel"	"CHAN_ITEM"
+	"volume"	"0.4"
+	"soundlevel"  "SNDLVL_75dB"
+
+	"wave"	"common/wpn_select.wav"
```
sunzenshen, line 298: Just to sanity check, but should this be
```cpp
"wave"  "common/wpn_denyselect.wav"
```
?

game/neo/scripts/game_sounds_player.txt
```diff
@@ -0,0 +1,317 @@
+"NeoPlayer.GhostPing"
+{
+	"volume"     	"VOL_NORM"
```
sunzenshen, line 3: Is it normal/expected that there is no "channel" field like the other entries?  Is the default value expected?

game/neo/scripts/game_sounds_player.txt
```diff
@@ -0,0 +1,317 @@
...
+"Player.FallGib"
+{
+	"channel"	"CHAN_ITEM"
+	"volume"	"0.75"
+	"pitch"		"90,110"
+	"soundlevel"	"SNDLVL_NORM"
+
+	"rndwave"
+	{
+		"wave"		"physics/body/body_medium_break2.wav"
```
sunzenshen, line 240: Given there is only one option here, would it make sense to remove the rndwave and just set wave directly?

```
"Player.FallGib"
{
    "channel"    "CHAN_ITEM"
    "volume"     "0.75"
    "pitch"      "90,110"
    "soundlevel" "SNDLVL_NORM"
    "wave"       "physics/body/body_medium_break2.wav"
}
```

game/neo/scripts/game_sounds_player.txt
Rainyan: These are imported verbatim from the OGNT `scripts/game_sounds_player.txt` - I don't know. I guess I can look into where this file is used.

sunzenshen: I guess it's arguably parity.  Feel free to resolve this thread if we want to look into changes in a separate PR.

sunzenshen (approved):
Assuming parity with OGNT

game/neo/scripts/game_sounds_player.txt
Rainyan: > I guess it's arguably parity. Feel free to resolve this thread if we want to look into changes in a separate PR.

I'm swamped currently but I'll try and make time to get this resolved some time next week. I think it's reasonable to follow up on before merging this to make sure we get this right in a single PR.

AdamTadeusz (approved):

game/neo/scripts/game_sounds_player.txt
AdamTadeusz: We've already strayed away from parity using DG's keyboard click sound for useDeny

AdamTadeusz: if we want to keep using the keyboard click, it is under common/wpn_denyselect.wav

DESTROYGIRL: Knowing when a use input is successful and unsuccessful is a crucial part of gameplay ok

DESTROYGIRL: > Knowing when a use input is successful and unsuccessful is a crucial part of gameplay ok

I say this semi jokingly but having the confirmation is nice

game/neo/scripts/game_sounds_player.txt
Rainyan: It will default to `CHAN_AUTO` if unspecified, which is probably fine here. For the ghost pickup, we have a custom channel `CHAN_GHOST_PICKUP` to ensure the critical sound cue is never overridden by an overlapping sound on the same channel, but the ping sfx probably don't need such a thing.

game/neo/scripts/game_sounds_player.txt
Rainyan: Seems like a good idea to refactor, but I'll prioritize just getting this merged without introducing any more changes for now since this is quite overdue.

game/neo/scripts/game_sounds_player.txt
Rainyan: It seems the custom key click deny sound is using `HL2Player.UseDeny` instead of this `Player.UseDeny` entry, so I'll keep this as-is/parity for now. Perhaps in the future this could be refactored to be `NeoPlayer.UseDeny` so it's more obvious it isn't a base engine sound.

Rainyan:
Rebased to master and reverted the removal of `NeoPlayer.Rankup` and `NeoPlayer.RankDown` - while these entries aren't used in OGNT, them existing is technically parity, and the entries could have some use either by custom content/plugins, or by us in the future. So probably better not to remove them.

DESTROYGIRL (approved):

AdamTadeusz (approved):

PR #1951, opened 2026-04-18
MP3 player + Root menu multi-keybinding + NeoUI minor changes

nullsystem:
## Description
* Main root menu now responds to `neo_mp3` bind
* `NeoUI::ProgressDrag` fix not immediately responding on click

## Toolchain

### DIFF
--- a/src/game/client/neo/ui/neo_ui.cpp
+++ b/src/game/client/neo/ui/neo_ui.cpp
@@ -7,6 +7,7 @@
 #include <filesystem.h>
 #include <stb_image.h>
 #include <materialsystem/imaterial.h>
+#include <IGameUIFuncs.h>
 
 #include "neo_misc.h"
 
@@ -2217,7 +2218,7 @@ void Progress(const float flValue, const float flMin, const float flMax)
 
 void ProgressDrag(float *flValue, const float flMin, const float flMax)
 {
-	const auto wdgState = BeginWidget(WIDGETFLAG_MOUSE | WIDGETFLAG_MARKACTIVE);
+	auto wdgState = BeginWidget(WIDGETFLAG_MOUSE | WIDGETFLAG_MARKACTIVE);
 	if (wdgState.bInView)
 	{
 		switch (c->eMode)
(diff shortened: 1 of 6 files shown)


AdamTadeusz (dismissed):

src/game/client/neo/ui/neo_ui.cpp
```diff
@@ -3528,6 +3530,11 @@ void EndIgnoreXOffset()
 	c->bIgnoreXOffset = false;
 }
 
+bool Bind(const char *pszBindName)
+{
+	return c->eMode == MODE_KEYPRESSED && c->eCode == gameuifuncs->GetButtonCodeForBind(pszBindName);
+}
```
Rainyan: Would prefer a function-level comment on what this function does, and what does the boolean return value mean, for improved readability from code that calls this. Since a boolean `if Bind(...)` at least for me doesn't ring a bell as to what exactly is being returned by the `Bind` function without some comment, unless I read the whole implementation.

Rainyan (commented):
I'm unsure whether this is working correctly since I don't really know what the expected behavior is, but I'm seeing the following:

* When entering `neo_mp3` in the dev console, the main menu music either stops playing or resumes playing - is this intended?
* When binding `bind p "neo_mp3"` and pressing the keyboard key P in the main menu, nothing happens - is this intended?

nullsystem:
@Rainyan

* `neo_mp3` is now changed to just doing pause/play music toggle so that's expected.
* I think that was down to `gameuifuncs->GetButtonCodeForBind` not supporting multiple bindings so whatever it think is first is picked up instead. This is also noted as a TODO in `neo_root.cpp` regarding `keys.bcConsole`. Could try to do a workaround which was already present for the settings menu where all keys were checked and put into a hashtable: `htAllBinds` variable in `neo_root_settings.cpp` but hashtable updated limited to like a `InstallChangeCallback`

src/game/client/neo/ui/neo_ui.cpp
nullsystem: Reverted in next commit: https://github.com/NeotokyoRebuild/neo/pull/1951/changes/a132708f6f935d3753e3b7c1fe7b6f90c846b9c7 due to dealing with multi-keys binding now.

nullsystem:
@Rainyan I've updated how the root main menu deal with keybinds: https://github.com/NeotokyoRebuild/neo/pull/1951/changes/a132708f6f935d3753e3b7c1fe7b6f90c846b9c7
Now it's able to deal with multiple keybinds, but I had ended up limiting the console's keybind setting UI due to unable with getting the toggleconsole enforce to deal with multiple binds. It already had that issue to prior to this PR.

Rainyan (approved):
LGTM, thanks for also fixing the main menu bind

AdamTadeusz (approved):

PR #1955, opened 2026-04-28
Feet face aim yaw when leaning

AdamTadeusz:
## Description

Move feet to aim towards the aim yaw when leaning, making peeking around corners at certain angles more fair

## Toolchain

- Windows MSVC VS2022


AdamTadeusz:
There's a problem on master especially noticeable with bots at the very start of freeze time where the upper body rotation stutters when moving while feet are planted in one spot. I've included a janky fix for that since its more noticeable on this PR with the feet following the player aim when leaning.

src/game/shared/hl2mp/hl2mp_playeranimstate.cpp
```diff
@@ -870,6 +883,24 @@ float CHL2MPPlayerAnimState::GetCurrentMaxGroundSpeed()
 	return speed;
 }
 
+#ifdef NEO
+bool CHL2MPPlayerAnimState::IsLeaning(CStudioHdr *pStudioHdr)
+{
+	if (CBaseAnimating* pBaseAnimating = GetBasePlayer()->GetBaseAnimating();
+		pBaseAnimating)
```
Rainyan: Nit: the value assignment already evaluates to its result inside the `if(x = ...)`, so the `; pBaseAnimating` at the end of the condition could be removed

Rainyan (dismissed):

AdamTadeusz:
@Rainyan I've noticed there is a hitch when moving between an angle of -180 and 180 when leaning. the issue seems to be in CMultiPlayerAnimState::ConvergeYawAngles where the value flDeltaYawAbs used to calculated the scale and therefore speed of rotation is calculated before the change in angle is normalized, so flScale will jump up to a value far greater than it should be, then get clamped down to 1 resulting in a small jitter as the rotation speeds up then slows down when moving between the two angles


https://github.com/user-attachments/assets/0b1ab503-e44e-4937-a93a-1472d898e2b7

Rainyan (approved):

src/game/shared/hl2mp/hl2mp_playeranimstate.cpp
```diff
@@ -819,13 +824,21 @@ void CHL2MPPlayerAnimState::ComputePoseParam_AimYaw( CStudioHdr *pStudioHdr )
...
 	// Turn off a force aim yaw - either we have already updated or we don't need to.
 	m_bForceAimYaw = false;
+#ifdef NEO
+	// NEO JANK (Adam) this fixes the stuttering animation when players turn their camera while their feet are planted in one spot. Probably not the right way to do it
+	// NEO TODO (Adam) I'm sure this is done server side only for a reason, also this doesn't work great if the local player is in third person, find correct fix for this?
+	QAngle angle = GetBasePlayer()->GetAbsAngles();
+	angle[YAW] = m_flCurrentFeetYaw;
 
+	GetBasePlayer()->SetAbsAngles( angle );
+#else
```
AdamTadeusz: This solution, while removing the stuttering is not the right way to solve the stuttering and is the reason the PR is in draft. I don't have time right now to look for another fix, if anyone else wants to pick this up, or find another solution similar to the og solution of a single turn ever x seconds after not moving then please go ahead.

AdamTadeusz: Tournament organizers are keen on this being fixed before the upcoming ADS tournament

sunmachine: Really dumb question: have we ruled out gimbal lock? 

(Edit: I'll give this a review of the math here leading up to it, but I lack time and a system that can compile the game client for a few weeks; I'm hoping we just have to understand why this smells like a Gimbal lock and the problem will reveal itself, & why your stopgap works.)

AdamTadeusz: The stutter described here is independent of the stutter in the video above and present at any angle so I don't think gimbal lock is the problem

AdamTadeusz: I only vaguely know what gimbal lock is though

sunmachine: Great info on the symptom, thank you. If you have another video of the specific symptom I might have a better idea of what it is (e.g. state sync fighting, another symptom of a gimbal lock, etc.). What I'll do for you is to at least get static analysis going so I don't flub my read, and to recreate the math in a unit test. If I can recreate a stutter/jitter on the graph, that would narrow down the scope of the cause. 9/10 with rotational stuff it's an accumulated roll state (gimbal lock). 

What avenues have you dug into so far when testing?

AdamTadeusz: It's likely some state fighting where the upper body yaw wants to return to 0, it's actually visible on master too, especially with bots right after they re-spawn and pick a new direction to look towards, since they're frozen they don't move, so the upper body can rotate independently of the legs and you can see the issue. I'm making some comparison videos I will send soon

AdamTadeusz: Now I'm confused, I'm looking at the animations with and without the janky fix and I don't see any difference. I think the animations when the bots re-spawn look better now but that wasn't intentional.

https://github.com/user-attachments/assets/7b051cf8-da51-4138-8e2d-82884ebc2055

AdamTadeusz: theres still a bit of stutter when re-spawning, I guess ill just remove the offending code for now since it doesn't seem to be doing anything and reopen the pr for review. Thank you for your help @sunmachine XD

sunmachine: ~~No problem; I say just add another commit and I'll figure out what I can do with this. Your gut feeling seems right to me, too, and that video shows a lot.~~ Misread; good job!

~~I'm also trying to remember where I saw it (typing from a rural coffee stand drive thru window lmfao) but I saw a floating point getting compared absolutely to a 0. I'll comment on it directly soon but that might be one of the regressions.~~ Nevermind I think you ditched it. 😁

sunmachine: Also thank you for the copious video detail, you're teaching me a ton about workflow and your approach just from that detail.

Rainyan (approved):

sunzenshen (approved):

PR #1956, opened 2026-05-01
Enable free roaming in demos while paused

sm90x:
## Description

Enables moving around with roaming view in paused demos.

The two impediments to do this were 1) the movement using frametime, which I guess stops when the engine pauses the game, and 2) when paused the engine stops processing in-game inputs so pressing to move forward would do nothing.

Changed the roaming view to use realtime instead of frametime, and added a hacky way to detect movement button presses while paused using vgui.

## Toolchain
- Linux GCC 10 Sniper 3.0

### DIFF
--- a/src/game/client/hltvcamera.cpp
+++ b/src/game/client/hltvcamera.cpp
@@ -441,11 +441,7 @@ void C_HLTVCamera::CalcRoamingView(Vector& eyeOrigin, QAngle& eyeAngles, float& 
 
 		AngleVectors ( m_LastCmd.viewangles, &forward, &right, &up);  // Determine movement angles
 
-#ifdef NEO
-		if ( m_LastCmd.buttons & IN_SPEED && !bUsePausedMovement)
-#else
 		if ( m_LastCmd.buttons & IN_SPEED )
-#endif // NEO
 		{
 			factor /= 2.0f;
 		}
@@ -453,9 +449,6 @@ void C_HLTVCamera::CalcRoamingView(Vector& eyeOrigin, QAngle& eyeAngles, float& 
 		// Copy movement amounts
 		float fmove = m_LastCmd.forwardmove * factor;
 		float smove = m_LastCmd.sidemove * factor;
-#ifdef NEO
-		float umove = m_LastCmd.upmove * factor;
-#endif
 
 		const bool bDroneMove = m_LastCmd.buttons & IN_WALK;
 		if (bDroneMove)
@@ -475,11 +468,7 @@ void C_HLTVCamera::CalcRoamingView(Vector& eyeOrigin, QAngle& eyeAngles, float& 
 
 		for (int i=0 ; i<3 ; i++)       // Determine x and y parts of velocity
 			wishvel[i] = forward[i]*fmove + right[i]*smove;
-#ifdef NEO
-		wishvel[2] += umove;
-#else
 		wishvel[2] += m_LastCmd.upmove * factor;
-#endif // NEO
 
 		VectorCopy (wishvel, wishdir);   // Determine magnitude of speed of move
 		wishspeed = VectorNormalize(wishdir);
--- a/src/game/client/hltvcamera.cpp
+++ b/src/game/client/hltvcamera.cpp
@@ -26,9 +26,7 @@
 #include "neo_gamerules.h"
 #include "c_neo_player.h"
 #include "shareddefs.h"
-#include "vgui/IInput.h"
-#include "ienginevgui.h"
-#include "IGameUIFuncs.h"
+#include "input.h"
 #endif // NEO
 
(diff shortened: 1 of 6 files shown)


AdamTadeusz:
I'm impressed, there are some issues with the demo playback but I would have to double check whether that's a fault with this pr or if it is on master, I'm heading to bed so i will just jot some things down here for myself to check later.

Bots (and most likely players too) repeatedly make footstep noises while the server demo is paused. an early return in CBaseEntity::EmitSound fixes the footstep noises, would need this everywhere where there is a call to g_SoundEmitterSystem.EmitSound( at least I guess, also check if there are any other ways to emit sounds that we want to pause when a demo is paused.

Bots (and most likely players too) teleport to the exact position at the given tick when a demo is paused, instead of remaining in the interpolated position they were in just before the pause. This can be fixed in C_BaseEntity::InterpolateServerEntities() by removing the engine->IsPaused() check. The check is there for a reason, probably so when watching demos we can see exactly where any given player was at any given time, might be worth keeping this functionality and adding some kind of cvar to disable it by default? Would need further discussion.

The "drone movement" relies on the walk key being pressed. Since user input isn't dealt with when a demo is paused like you said, the player will be stuck in drone movement if they had the walk key pressed while pausing the demo, or stuck outside of drone movement if they didn't, with no way to switch between the two while paused. bDroneMove should ignore m_lastCmd.buttons and instead check if the key assigned to +walk is pressed. You might want to also deal with whether the player is using toggle or hold to walk, as that makes a difference when the demo isn't paused.

I'm not suggesting you implement these right away, more testing needs to be done still.

(Edit) also the +speed command is used in spectate to slow the spectator down, also not handled when paused

(Edit 2) also in drone movement the swim up and down commands are used to move the camera up and down when in drone movement

AdamTadeusz:
wrt interpolation, we could just always interpolate during a paused demo, and if someone wants to see the exact position they can toggle cl_interpolate themselves

AdamTadeusz (changes requested):
looking into the code further, I think you can drop all the vgui stuff. We have a class CInput that keeps track of what buttons the user is pressing even when a demo is paused. You could create a third CreateMove function in CInput that instead takes a CUserCmd created in C_HLTVCamera, like the dummy command in the CInput ::ExtraMouseSample function, populates the command and returns it to be used in place of m_LastCmd in C_HLTVCamera. this should handle all user input regardless of if the player is using a mouse and keyboard or controller or whatever, and means you need very few changes in C_HLTVCamera

sm90x:
I am probably not getting what you mean, but `C_HLTVCamera::CreateMove` and `CInput::CreateMove` don't get called when paused.

The sound looping issue is present in master.

Another issue already in master, is the menu and console key becoming unresponsive. I am not sure how to reproduce, maybe pausing and unpausing, it doesn't happen every time though.

AdamTadeusz:
@sm90x are you happy with me opening up a PR to your branch? I can implement my suggestion and you can merge it if you're happy with it

sm90x:
Sure.

AdamTadeusz:
@sm90x I opened https://github.com/sm90x/neo/pull/1

sm90x:
Looks clean now, thanks. My assumption on not processing inputs was wrong.

src/game/client/hltvcamera.cpp
```diff
@@ -443,7 +463,6 @@ void C_HLTVCamera::CalcRoamingView(Vector& eyeOrigin, QAngle& eyeAngles, float&
 				smove *= absSMove / moveMagnitude;
 			}
 		}
-#endif // NEO
```
AdamTadeusz, line 446: my apologies I made a mistake here, this code block should be surrounded by an #ifdef NEO still

AdamTadeusz (changes requested):

src/game/client/hltvcamera.cpp
```diff
@@ -614,6 +648,17 @@ void C_HLTVCamera::SetMode(int iMode)
 	m_nCameraMode = iMode;
 
 #ifdef NEO
+	// If we pause demo playback while in eye view, the observer target will be invisible when switching observer mode because we also pause simulation etc, update visibility here
+	if (iOldMode == OBS_MODE_IN_EYE && engine->IsPaused())
+	{
+		C_BaseEntity* target = ClientEntityList().GetEnt( m_iTraget1 );
+		if ( target && target->IsPlayer())
```
AdamTadeusz: After thinking about this some more, this isPlayer check here is unnecessary, could you remove it and define target inside the if statement? So it would look something like 

```
if ( C_BaseEntity* target = ClientEntityList().GetEnt( m_iTraget1 ))
{
```

src/game/client/hltvcamera.cpp
sm90x: restored

AdamTadeusz (approved):

Agiel (approved):

PR #1965, opened 2026-05-17
Shadows on water surfaces

AdamTadeusz:
## Description

neo assets pr https://github.com/NeotokyoRebuild/neoAssets/pull/120

Adds support for the $lightmapwaterfog water texture param used in games post l4d2. I wasn't able to get the scale to which the shadow is applied to the fog colour, so I made it depend on the alpha value of the water fog colour.

Keep in mind, water with this parameter needs to have a lightmap. Depending on the bsp compiler used, setting $lightmapwaterfog before compiling may be enough, or you may have to add a compiler flag like "%compileKeepLight". You can check if a water surface has a lightmap with the mat_luxels 1 command. If the surface is entirely filled in in blue, there is no lightmap. If instead you can see a square grid pattern corresponding to the lightmap resolution, the water has a lightmap.

Some example images
$lightmapwaterfog 0
image

$lightmapwaterfog 1 $fogcolor" "{52 49 44 0}
image

$lightmapwaterfog 1 $fogcolor" "{52 49 44 1}
image

## Toolchain

- Windows MSVC VS2022

### DIFF
--- a/src/materialsystem/stdshaders/include/water_ps20b.inc
+++ b/src/materialsystem/stdshaders/include/water_ps20b.inc
@@ -20,6 +20,7 @@ class water_ps20b_Static_Index
 	unsigned int m_nABOVEWATER : 2;
 	unsigned int m_nBLURRY_REFRACT : 2;
 	unsigned int m_nNORMAL_DECODE_MODE : 1;
+	unsigned int m_nLIGHTMAPWATERFOG : 2;
 #ifdef _DEBUG
 	bool m_bBASETEXTURE : 1;
 	bool m_bMULTITEXTURE : 1;
@@ -28,6 +29,7 @@ class water_ps20b_Static_Index
 	bool m_bABOVEWATER : 1;
 	bool m_bBLURRY_REFRACT : 1;
 	bool m_bNORMAL_DECODE_MODE : 1;
+	bool m_bLIGHTMAPWATERFOG : 1;
 #endif	// _DEBUG
 public:
 	void SetCONVERT_TO_SRGB( int i )
@@ -99,6 +101,15 @@ public:
 #endif	// _DEBUG
 	}
 
+	void SetLIGHTMAPWATERFOG( int i )
+	{
+		Assert( i >= 0 && i <= 1 );
+		m_nLIGHTMAPWATERFOG = i;
+#ifdef _DEBUG
+		m_bLIGHTMAPWATERFOG = true;
+#endif	// _DEBUG
+	}
+
 	water_ps20b_Static_Index(  )
 	{
 		m_nCONVERT_TO_SRGB = g_pHardwareConfig->NeedsShaderSRGBConversion();
@@ -109,6 +120,7 @@ public:
 		m_nABOVEWATER = 0;
 		m_nBLURRY_REFRACT = 0;
 		m_nNORMAL_DECODE_MODE = 0;
+		m_nLIGHTMAPWATERFOG = 0;
 #ifdef _DEBUG
 		m_bBASETEXTURE = false;
 		m_bMULTITEXTURE = false;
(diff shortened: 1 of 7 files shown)


src/materialsystem/stdshaders/water.cpp
```diff
@@ -243,6 +265,11 @@ BEGIN_VS_SHADER( Water_DX90,
 				SET_STATIC_PIXEL_SHADER_COMBO( BASETEXTURE, params[BASETEXTURE]->IsTexture() );
 				SET_STATIC_PIXEL_SHADER_COMBO( BLURRY_REFRACT, params[BLURREFRACT]->GetIntValue() );
 				SET_STATIC_PIXEL_SHADER_COMBO( NORMAL_DECODE_MODE, (int) NORMAL_DECODE_NONE );
+#ifdef NEO
+				SET_STATIC_PIXEL_SHADER_COMBO(LIGHTMAPWATERFOG, blightMapWaterFog);
+#else // Still custom NEO code
+				SET_STATIC_PIXEL_SHADER_COMBO(LIGHTMAPWATERFOG, 0);
+#endif // NEO
```
DESTROYGIRL, line 272: Was a lil confused by this until I realised it's because we can't `#ifdef` stuff in the shader code correct?
I don't think I was mindful of that when changing lightmappedgeneric. I wonder if it's the correct thing to do or we should just pull the original shader sources if you want to do a non-NEO build?

DESTROYGIRL (approved):
Looks good, didn't seem to create issues for existing materials, tested on rogue:

Image

src/materialsystem/stdshaders/water.cpp
AdamTadeusz: that was the thought process yes. And you're probably right, if anyone is building without the NEO define for whatever reason they should just pull the original shaders

src/materialsystem/stdshaders/include/water_ps20b.inc
```diff
@@ -117,18 +129,19 @@ public:
...
 #endif	// _DEBUG
 	}
 
 	int GetIndex() const
 	{
-		Assert( m_bBASETEXTURE && m_bMULTITEXTURE && m_bREFLECT && m_bREFRACT && m_bABOVEWATER && m_bBLURRY_REFRACT && m_bNORMAL_DECODE_MODE );
+		Assert( m_bBASETEXTURE && m_bMULTITEXTURE && m_bREFLECT && m_bREFRACT && m_bABOVEWATER && m_bBLURRY_REFRACT && m_bNORMAL_DECODE_MODE && m_bLIGHTMAPWATERFOG );
 		AssertMsg( !( m_nMULTITEXTURE && m_nBASETEXTURE ), "Invalid combo combination ( MULTITEXTURE && BASETEXTURE )" );
-		return ( 6 * m_nCONVERT_TO_SRGB ) + ( 12 * m_nBASETEXTURE ) + ( 24 * m_nMULTITEXTURE ) + ( 48 * m_nREFLECT ) + ( 96 * m_nREFRACT ) + ( 192 * m_nABOVEWATER ) + ( 384 * m_nBLURRY_REFRACT ) + ( 768 * m_nNORMAL_DECODE_MODE ) + 0;
+		return ( 6 * m_nCONVERT_TO_SRGB ) + ( 12 * m_nBASETEXTURE ) + ( 24 * m_nMULTITEXTURE ) + ( 48 * m_nREFLECT ) + ( 96 * m_nREFRACT ) + ( 192 * m_nABOVEWATER ) + ( 384 * m_nBLURRY_REFRACT ) + ( 768 * m_nNORMAL_DECODE_MODE ) + ( 768 * m_nLIGHTMAPWATERFOG ) + 0;
```
Masterkatze, line 140: Duplicate `768` value?

AdamTadeusz: autogenerated file

Rainyan (changes requested):
I'm seeing a problem with the underwater view, at least on Windows 11 with Intel dGPU and Vulkan (or with the implicit Intel GPU DX9->DXVK, sadly I can't test pure DX mode with my GPU).

With current master, going underwater looks like:

underwater_master

When trying with this PR, with the [associated neoAssets shaders](https://github.com/NeotokyoRebuild/neoAssets/pull/120) combined in, I'm seeing this:

underwater_pr

It looks to me like the above-water world is not correctly rendered when viewed from inside the water volume.

For both cases, this was tested with these graphics settings:

settings

This was tested in the pond in `ntre_shrine_ctg`.

GPU info:
```
GPU: Intel Arc B580
Intel Graphics Software Version: 25.32.1758.2
Graphics Driver Version: 21.0.101.8132
D3D API Version: 12.0
Shader Model Version: 6,7
Vulkan API Version: 1.4.325
```

AdamTadeusz:
@Rainyan This was a case of me putting the wrong/outdated pixel shader in neoassets, I've updated it now

Rainyan (approved):

Issue #1967, opened 2026-05-19
cl_neo_spectator_marker and similar cvars represented as parameter sets cannot be bound to a key


Sakhanist:
### Build Info

2026-03-22_c49384432cef6b50aeead1c25acfe06b89323d9b

### Description

Cvar values represented as a parameter set delimited by semicolons(;) cannot be bound to a key.

Source console uses ; as a command separator; in order to use it in a command to delimit values, one has to escape them with double quotes on both ends of the set.

In order to switch between value sets quick enough (i.e. during spectating), one should have an ability to bind the specific set of values.

When using bind, it is necessary to use double quotes when using semicolon in [command body]
bind "key" "[command]"
If semicolons are used inside the [command] body, they're interpreted as console command separators. A user cannot change this behavior by using double quotes, nor by using any other symbol inside the [command] body (unless it's implemented but undocumented by Valve)

### To Reproduce

1. Open demo / Create server, add bots: 1 for Jinrai, 1 for NSF, join spectator team;
2. Open Options > HUD > Scroll to bottom > IFF Markers > Spectator > Show squad marker = Enabled (OPTIONALLY BACKUP YOUR MARKER SETTINGS FOR SPECTATOR MODE)
3. Open dev console;
4. Input bind "[" "cl_neo_spectator_marker "0;0;0;0;0.50;0;1;1;0;32;"";
5. Press Enter;
6. Press [ (or any bound key to the command above)

### Expected behavior

The fourth position in a set represents the team marks (a 90 degree cross formed by 5 rectangles, green or blue depending on the teams if unmodded)

By having it set as 0, it should hide the mentioned marks.

### Actual behavior

The team marks are not hidden in spectator mode after pressing the bound key for the command in step 4

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

Platform-independent

### Additional context and Screenshots

There is an old workaround to it, usually applied to Sourcemod command syntax, but:

[spec_hud_cfg.txt](https://github.com/user-attachments/files/27981169/spec_hud_cfg.txt)
[vision_spec.txt](https://github.com/user-attachments/files/27981171/vision_spec.txt)
[vision_spec_raw.txt](https://github.com/user-attachments/files/27981170/vision_spec_raw.txt)

spec_hud_cfg.cfg

// default cl_neo_spectator_marker "0;0;0;1;0.50;1;1;1;1;32;"
alias raw  "exec vision_spec_raw.cfg";
[...]

PR #1968, opened 2026-05-20
Change the marker cvars delimiter char

Rainyan:
## Description
Change the `cl_neo_...marker` cvars' delimiter character from `';'` to `','`, because the engine tokenizer uses the semicolon as a command-level delimiter, and there is no escape character available. This causes issues for nested semicolons in config file binds etc, where the inner semicolon of the marker syntax is incorrectly interpreted as the end-of-command token.

I'm taking the easy route of simply changing our marker delimiter character to something better supported, because we don't have source code access to the lexer internals, so else we'd have to detour or shim the engine interface which gets ugly.

This has the one-time side effect of mauling people's marker configs for the next patch (causing those marker values to reset to defaults), but that shouldn't be too much of an issue since the config is trivially fixable for anyone with custom settings.

I'm limiting the scope of this PR to simply fixing the marker cvars, but we should probably also revisit the crosshair code syntax (and any other such serializations: `CH_XH_SEGEND`), and also change their delimiter characters to `','` or some other agreed upon character that has no conflicts.

### Steps to test

* Load into a map with bots
* Join spectator
* Bind for example:
```
bind o "cl_neo_spectator_xray_marker 4,0,0,0,0.50,1,0,1,1,32,";
bind p "cl_neo_spectator_xray_marker 4,0,0,1,0.50,1,0,1,1,32,";
```
* Use the bound keys to toggle the squad icon state.

Previously, with the `;` delimiter, this kind of a bind would fail. Likewise attempting to nest `"` quotes as a workaround to escape the semicolons would not work (but nested/escaped tokens or any kind of changes to the lexer in general are out of scope for this PR).

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #1967


AdamTadeusz:
It would be cool if people couldn't change their crosshairs with the click of a key bind, since networked crosshairs are a thing

Rainyan:
> It would be cool if people couldn't change their crosshairs with the click of a key bind, since networked crosshairs are a thing

I agree that we should prevent spam/abuse (both visual annoyance but also networked traffic spam), but maybe this could be via rate limiting, or by only networking the player's initial crosshair upon their server join (AFAIK this is what CS does to address the issue). Or perhaps something like networking once per round start could be a middle ground, where if someone changes their crosshair mid-round it's still somewhat synced.

But I think for the power users who want multiple crosshairs for whatever reason it would be nice to accommodate that use case (regardless of whether/when it's synced across clients).

AdamTadeusz (approved):

Masterkatze (approved):

PR #1978, opened 2026-05-31
Reduce instruction set baseline to just x86-64 (GCC, Linux only)

nullsystem:
## Description
There been a report of Phenom II unable to open NT;RE, so just reduce it back to just the basic x86-64 support + CMPXCHG16B.

## Toolchain

- Linux GCC Distro Native GCC 16 - Arch

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -120,9 +120,9 @@ elseif(CMAKE_BUILD_TYPE STREQUAL "Release" OR CMAKE_BUILD_TYPE STREQUAL "RelWith
 endif()
 
 add_compile_definitions(
-	# Whether to validate narrowing conversions done with narrow_cast<T>(...)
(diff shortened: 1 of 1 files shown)


Rainyan:
Do we know if this has any meaningful performance implications for the default case of more modern hardware?

nullsystem:
@Rainyan The performance change if any will be negligible `core2` doesn't give much more than the x86-64 baseline, AVX/AVX2 wasn't even compiled in the first place, and `-mtune=generic` optimize for the common hardware. Plus Windows build was already set at the x86-64/SSE2 baseline so will put Linux on equal hardware support footing. Basically there wasn't much point on setting `core2` in the first place.

Rainyan (approved):
Tested on linux pop-os 6.18.7-76061807-generic and seems fine

AdamTadeusz (approved):

PR #1986, opened 2026-06-11
Silence leaf shadow count assert spam

Rainyan:
## Description
Silence the assertion spam that would trigger for every frame on debug builds when `m_ShadowsInLeaf > 2000`, tanking dev build performance.

We previously had a clunky per-map check in place to try and silence the warning, but the map list wasn't comprehensive, and also completely hiding the issue wasn't too great either. So instead just assert once, until we can either fix the shadow counts or decide to bump the shadow count upper bound for the assertion.

## Toolchain
- Windows MSVC VS2022


AdamTadeusz (approved):

DESTROYGIRL (approved):

Rainyan:
GitHub, push to master, make no mistakes.

AdamTadeusz:
please

PR #1987, opened 2026-06-11
Bump actions versions

Masterkatze:
## Description
Fix for this warning:
`Windows Native Libraries Debug
Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4, actions/upload-artifact@v4, ilammy/msvc-dev-cmd@v1. Actions will be forced to run with Node.js 24 by default starting June 16th, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/`

`ilammy/msvc-dev-cmd@v1` is not active anymore ([link](https://github.com/ilammy/msvc-dev-cmd/pull/101)) and will be replaced by `TheMrMilchmann/setup-msvc-dev@v4` fork.

### DIFF
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -192,7 +192,7 @@ jobs:
         echo "install_dir=$install_dir" >> "$GITHUB_ENV"
         echo "preset_name=$preset_name" >> "$GITHUB_ENV"
 
-    - uses: TheMrMilchmann/setup-msvc-dev@v4
+    - uses: TheMrMilchmann/setup-msvc-dev@79dac248aac9d0059f86eae9d8b5bfab4e95e97c
       with:
         arch: x64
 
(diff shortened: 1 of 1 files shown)


Rainyan:
Should we maybe pin `TheMrMilchmann/setup-msvc-dev@v4` by the release hash instead, to minimize the risk of a supply-chain attack where `v4` gets pointed to something else? The maintainer seems legit but this doesn't seem like an action that will need updates so maybe it would be safer to do so. Just so we don't have to worry about a random supply chain compromise by whatever means.

For the official github `action` org actions I doubt there is a need to hard-pin a specific version, since intrinsically we have to trust GitHub already.

Rainyan (approved):

nullsystem (approved):

PR #1997, opened 2026-06-19
Bots loot weapons they could not select on spawn due to class restrictions

sunzenshen:
## Description
Introduces a configuration for bots to loot weapons that they could not select on spawn due to class restrictions.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #1800


game/neo/scripts/ntre_bot_profiles.txt
```diff
@@ -22,20 +22,23 @@ ntre_bot_profiles
 				corporal "jittescoped"
 				sergeant "zr68c"
 				lieutenant "supa7"
+				loot "zr68s mx mx_silenced aa13"
```
sunzenshen, line 25: I initially considered allowing the specification of relative class-exclusive weapon preferences by ranking, but having that level of control turned out to be both messy, and in my opinion not that useful.

One can see the prototype for the rank-specific preference here in this earlier commit:
https://github.com/NeotokyoRebuild/neo/pull/1997/changes/66f7eafd8b293eb8d88088c989ea11ce56d11b25

It seems easier to just assume that any class-exclusive looting preferences should apply to the LT level, mostly because when I tried to come up with initial configurations I ended up putting most preferences at the LT level anyway.

sunzenshen: Also for reference, since I needed to consider whether selections would have been redundant and already possible with #1800,  these are the weapons that each class can never select even if they were LT rank:

Recon:
- ZR suppressed
- MX
- MX suppressed
- SRS
- AA13

Assault:
- SRM suppressed
- Jitte suppressed
- ZR68 scoped
- PZ

Support:
- SRM suppressed
- Jitte suppresed
- ZR suppressed
- ZR scoped
- SRS
- AA13


For testing, the permutations are kind of tricky, but a basic test is to set all bots to Recon with `bot_class 0`, spawn as an Assault with a ZR suppressed, and throw it at a friendly recon bot.  Before with #1800 the bots would not consider the swap.

src/game/server/neo/bot/neo_bot_profile.cpp
```diff
@@ -223,6 +223,37 @@ static void SetProfileTempBotCommon(CNEOBotProfile *pProfile, KeyValues *kv)
...
+				(NEO_WEP_BITS_UNDERLYING_TYPE)0,
+				WEP_BIT_FLAG_CMP,
+				wepClassKv->GetString("loot"),
+				&pProfile->flagTemplateApplied,
+				static_cast<BotTemplateApplied_>(
+					BOT_TEMPLATE_APPLIED_WEP_PREF_RECON_LOOT << idxClass));
+
+			// Append loot preference to lieutenant level
+			const int idxLieutenant = NEO_RANK__TOTAL - 1;
+			pProfile->flagsLootWepPrefs[idxClass][idxLieutenant] |= lootFlags;
```
nullsystem, line 255: Nit: `idxLieutenant` -> `IDX_LIEUTENANT`

game/neo/scripts/ntre_bot_profiles.txt
nullsystem: Was wondering why it's all tied to LT rank but checking `GetBotWeaponPreferenceRank` function made sense. So LGTM

nullsystem (approved):

src/game/server/neo/bot/neo_bot_profile.cpp
sunzenshen: With other coding guidelines that I've worked under, we tended to reserve all caps for just macro definitions, though I have have also heard of capitalizing all constants in some other style guidelines.

In this case, I wanted to keep the naming similar to idxClass since it might be distracting to have them look different like,
[idxClass][IDX_LIEUTENANT]

game/neo/scripts/ntre_bot_profiles.txt
sunzenshen: Yeah, I thought maybe bots may want to loot intermediate rank weapons at lower ranks, but in practice I didn't feel that the benefits were worth it for the extra complication of the loot preferences config population code, especially since there's already other options provided by the spawn selection preferences.

AdamTadeusz (approved):

PR #1998, opened 2026-06-19
Fix bots not being able to see ghoster behind smoke

sunzenshen:
## Description
Fixes bots not being able to see a ghost carrier from behind smoke despite the apparent glowing beacon visible by human players.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/neo_player.h
+++ b/src/game/server/neo/neo_player.h
@@ -377,4 +377,13 @@ inline CNEO_Player *ToNEOPlayer(CBaseEntity *pEntity)
 	return assert_cast<CNEO_Player*>(pEntity);
 }
 
+inline const CNEO_Player *ToNEOPlayer(const CBaseEntity *pEntity)
+{
+	if (!pEntity || !pEntity->IsPlayer())
+	{
+		return nullptr;
+	}
+	return assert_cast<const CNEO_Player*>(pEntity);
+}
+
 #endif // NEO_PLAYER_H
--- a/src/game/server/NextBot/NextBotVisionInterface.cpp
+++ b/src/game/server/NextBot/NextBotVisionInterface.cpp
@@ -723,13 +723,22 @@ bool IVision::IsLineOfSightClear( const Vector &pos ) const
 bool IVision::IsLineOfSightClearToEntity( const CBaseEntity *subject, Vector *visibleSpot ) const
 {
 #ifdef NEO
-	// Special case for Support-class bots to see through smoke
-	bool isSupport = false;
+	bool canSeeThroughSmoke = false;
 	if (auto player = ToNEOPlayer(GetBot()->GetEntity()))
 	{
-		isSupport = (player->GetClass() == NEO_CLASS_SUPPORT);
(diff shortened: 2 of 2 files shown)


sunzenshen:
To test, 
- set bots to a non-support class using `bot_class 1`
- use `sv_neo_infinite_smoke_grenades 1`  to create a smoke screen while holding the ghost
- before this patch, bots would not see you, after the patch they will see you if you are carrying the ghost.

AdamTadeusz (approved):

src/game/server/neo/neo_player.h
```diff
@@ -374,4 +374,13 @@ inline CNEO_Player *ToNEOPlayer(CBaseEntity *pEntity)
 	return assert_cast<CNEO_Player*>(pEntity);
 }
 
+inline const CNEO_Player *ToNEOPlayer(const CBaseEntity *pEntity)
```
AdamTadeusz, line 380: We already have ToNEOPlayer in c_neo_player.h, maybe keep the declaration here and move both definitions to neo_player_shared.cpp?

sunzenshen: To be honest I looked into this, and I'm not exactly sure what you mean by the relation with c_neo_player.h

I just needed the const for `const CNEO_Player *`, so that I can get a const pointer for my use case here.

I also wasn't sure how moving both definitions to neo_player_shared.cpp actually made things clearer, because with how I understand this, it would involve defining branches around #ifdef CLIENT_DLL, which I'm not sure is actually more readable, e.g.:

In neo_player_shared.h:

```cpp
#ifdef CLIENT_DLL
class C_BaseEntity;
class C_NEO_Player;
C_NEO_Player *ToNEOPlayer(C_BaseEntity *pEntity);
#else
class CBaseEntity;
class CNEO_Player;
CNEO_Player *ToNEOPlayer(CBaseEntity *pEntity);
const CNEO_Player *ToNEOPlayer(const CBaseEntity *pEntity);
#endif
```

and in neo_player_shared.cpp:

```cpp
#ifdef CLIENT_DLL
C_NEO_Player *ToNEOPlayer(C_BaseEntity *pEntity)
{
	if (!pEntity || !pEntity->IsPlayer())
	{
		return nullptr;
	}
#if _DEBUG
	Assert(dynamic_cast<C_NEO_Player*>(pEntity));
#endif
	return static_cast<C_NEO_Player*>(pEntity);
}
#else
CNEO_Player *ToNEOPlayer(CBaseEntity *pEntity)
{
	if (!pEntity || !pEntity->IsPlayer())
	{
		return nullptr;
	}
	return assert_cast<CNEO_Player*>(pEntity);
}

const CNEO_Player *ToNEOPlayer(const CBaseEntity *pEntity)
{
	if (!pEntity || !pEntity->IsPlayer())
	{
		return nullptr;
	}
	return assert_cast<const CNEO_Player*>(pEntity);
}
#endif
```

... if it's okay, maybe  we can go with the original PR approach?

Masterkatze (approved):

PR #2002, opened 2026-06-30
Hitbox group pen

AdamTadeusz:
## Description

Any bullet that hits a player model extremity now does a second check where it hit a player while ignoring those hitbox groups, generally decreasing the ttk as for example arms no longer block shots to the chest.

Currently stops shots that would have gibbed a model from doing so if the shot ends up hitting a main hitbox behind it, I'm going to say that's fine for now? My explanation for that is uhh the limb has something to brace against so it doesn't pop off as easily.

Might need to ask the community what they think.

Shotguns are very scary now.

Tbh even an srm shreds now.

## Toolchain

- Windows MSVC VS2022

### DIFF
--- a/src/game/shared/neo/neo_player_shared.cpp
+++ b/src/game/shared/neo/neo_player_shared.cpp
@@ -755,28 +755,31 @@ bool CNEO_Player::TestHitboxes(const Ray_t& ray, unsigned int fContentsMask, tra
 		tr.surface.flags = SURF_HITBOX;
 		tr.surface.surfaceProps = physprops->GetSurfaceIndex( pBone->pszSurfaceProp() );
 		
-		switch (tr.hitgroup)
+		if (fContentsMask & CONTENTS_HITBOX)
 		{
-			case HITGROUP_LEFTARM:
-			case HITGROUP_RIGHTARM:
-			case HITGROUP_LEFTLEG:
-			case HITGROUP_RIGHTLEG:
-			case HITGROUP_GEAR:
-				trace_t secondTrace;
-				unsigned int fHitboxGroupMask = UINT_MAX;
-				fHitboxGroupMask = fHitboxGroupMask & ~(1 << HITGROUP_LEFTARM);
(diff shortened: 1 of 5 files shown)


src/game/shared/neo/neo_player_shared.cpp
```diff
@@ -707,4 +710,90 @@ CBaseEntity *CNEO_Player::FindUseEntity()
...
+	if ( TraceToStudio( physprops, ray, pStudioHdr, set, hitboxbones, fContentsMask, GetRenderOrigin(), GetModelScale(), tr ) )
+#endif // GAME_DLL
+	{
+		mstudiobbox_t *pbox = set->pHitbox( tr.hitbox );
+		mstudiobone_t *pBone = pStudioHdr->pBone(pbox->bone);
+		tr.surface.name = "**studio**";
+		tr.surface.flags = SURF_HITBOX;
+		tr.surface.surfaceProps = physprops->GetSurfaceIndex( pBone->pszSurfaceProp() );
+		
+		if (fContentsMask & CONTENTS_HITBOX)
```
AdamTadeusz: the only changes from the CBaseAnimating baseclass implementation should be in this if loop, I've skipped the ismultiplayer checks in the other player classes that otherwise scale the player's hitbox down in singleplayer

sunmachine (dismissed):
Just nit-picks so far, like taking a second pass at some indentations here and there, cleanup a TODO, blah blah.

The only thing I thought I had a bite on was HITBOX_GEAR as pondering why/how it would be used. Then I realize we don't use those hit boxes at all, do we?

Then I'm toying around on raycast perf, but even the worst case is a big "so what".

Good addition. LGTM.

AdamTadeusz:
We indeed don't use HITBOX_GEAR. I thought that it might be for things like pouches or other jiggle bones or whatever attached to the player model, but now I'm wondering if it's used for things like ballistic shields, and if so, we actually definitely shouldn't be doing this check, so I think ill remove it.

sunmachine (approved):

sunzenshen (approved):

PR #2006, opened 2026-07-04
Don't copy takeover bot's class loadout choice

sunzenshen:
## Description
Do not overwrite player's class/loadout choices when they take over a bot.

## Toolchain
- Windows MSVC VS2022


sunzenshen:
This seems to resolve the class/loadout being stuck to the bot's choices in offline play, but I'd like to try this out on a live server.

AdamTadeusz (approved):
makes sense

Agiel (approved):

PR #2007, opened 2026-07-04
Fix bots choosing empty weapon slots

### LINKED ISSUE #1619: Lieut. recon & support bots shouldn't fail to choose loadout
### Build Info

4228d58182881

### Description

When a support bot in CTG mode first reaches the lieutenant rank, for some reason they may attempt to pick a `loadoutNumber` loadout slot from `s_LoadoutWeapons` larger than the one of the PZ, which is the largest valid loadout value for supports. This leads to the bot requesting for the weapon named "weapon_empty".

Callstack:

```
 	server.dll!CNEO_Player::RequestSetLoadout(int loadoutNumber) Line 290	C++
>	server.dll!CNEO_Player::Spawn() Line 609	C++
 	server.dll!NextBotPlayer<CNEO_Player>::Spawn() Line 653	C++
 	server.dll!CNEOBot::Spawn() Line 640	C++
 	server.dll!CNEORules::StartNextRound() Line 2644	C++
 	server.dll!CNEORules::Think() Line 1205	C++
 	server.dll!CGameRules::FrameUpdatePostEntityThink() Line 614	C++
 	server.dll!CMultiplayRules::FrameUpdatePostEntityThink() Line 437	C++
 	server.dll!InvokePerFrameMethod(void(IGameSystemPerFrame::*)() f, const char * timed) Line 377	C++
 	server.dll!IGameSystem::FrameUpdatePostEntityThinkAllSystems() Line 338	C++
 	server.dll!CServerGameDLL::GameFrame(bool simulating) Line 1283	C++
 	[External Code]	
```

with locals:
```
+		this	0x0000023f7e272710 {m_bOnTarget=true m_qPrevShouldAim=ANSWER_YES (1) m_flLastShouldAimTime=1032.29993 ...}	CNEO_Player * {CNEOBot}
		classChosen	2	int
		iLoadoutClass	2	const int
		loadoutNumber	11	int
+		pEnt	{...}	CHandle<CBaseEntity>
+		pNeoWeapon	0xcccccccccccccccc {m_bWeaponIsLowered=??? m_weaponHandling={weaponID=??? spreadInfo=0xccccccccccccd41c {...} ...} ...}	CNEOBaseCombatWeapon *
+		pszWepName	0x0000023f29d69ae8 "weapon_empty"	const char *
		result	true (204)	bool
		retAssert	SPEW_DEBUGGER (0)	SpewRetval_t
```

Notice `loadoutNumber` being `11` (I also spotted the value `10` on another run).

### To Reproduce

```
// load some CTG map
map ntre_saitama_ctg;

// join spectator, and let the bots join the match

// increase the round limit so the map won't change too soon
neo_round_limit 99;

// increase match speed, and wait for a support to gain the lieutenant rank
host_timescale 10;
```

### Expected behavior

The lieutenant support picks a valid weapon loadout, and the below assertion/warning is not hit.

### Actual behavior

The lieutenant support picks an incorrect loadout value, and the assertion fails: https://github.com/NeotokyoRebuild/neo/blob/eae0d512394c078f244da019de4fdf487d4893f3/src/game/server/neo/neo_player.cpp#L3124-L3129

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

11

* I happened to stumble upon this with the commit 4228d58 during unrelated code review, so that's the build info reported, but I highly suspect this bug also exists in current master eae0d51.

* Likely not limited to CTG mode, but I haven't tested it with other game modes.

sunzenshen:
## Description
Fix bots choosing empty weapon slots after they choose a different class, where they previously selected a weapon with a selection index that would be an invalid/empty weapon slot with their new class.

## Toolchain
- Windows MSVC VS2022

## Related
- fixes: #1619

### DIFF
--- a/src/game/server/neo/bot/neo_bot.cpp
+++ b/src/game/server/neo/bot/neo_bot.cpp
@@ -601,52 +601,17 @@ CNEOBot::~CNEOBot()
 void CNEOBot::Spawn()
 {
 	// CNEOBot do m_iNeoClass a bit earlier
-	if ((m_iNextSpawnClassChoice != NEO_CLASS_RANDOM) && (m_iNeoClass != m_iNextSpawnClassChoice))
+	// so that we get correct loadout choices for the class
+	if (m_iNextSpawnClassChoice == NEO_CLASS_RANDOM) 
+	{
+		m_iNeoClass = ChooseRandomClass();
+	}
+	else if (m_iNeoClass != m_iNextSpawnClassChoice)
 	{
 		RequestSetClass(m_iNextSpawnClassChoice);
 	}
 
-	const ENeoRank eRank = static_cast<ENeoRank>(GetRank(m_iXP) - 1);
-	if (eRank == NEO_RANK_RANKLESS_DOG || (false == IN_BETWEEN_EQ(NEO_CLASS_RECON, m_iNeoClass, NEO_CLASS_VIP)))
-	{
-		m_iLoadoutWepChoice = 0;
-	}
-	else
-	{
-		const NEO_WEP_BITS_UNDERLYING_TYPE wepPrefsForCurRank = m_profile.flagsWepPrefs[m_iNeoClass][eRank];
-
-		int iChosenWeps[MAX_WEAPON_LOADOUTS] = {};
-		int iChosenWepsSize = 0;
-		for (int i = 0; i < MAX_WEAPON_LOADOUTS; ++i)
-		{
-			if (wepPrefsForCurRank & CNEOWeaponLoadout::s_LoadoutWeapons[m_iNeoClass][i].info.m_iWepBit)
-			{
-				iChosenWeps[iChosenWepsSize++] = i;
-			}
-		}
-
-		if (iChosenWepsSize == 0)
-		{
-			// Generally shouldn't happen, but if so, just pick from any under the XP limit
-			for (int i = 0; i < MAX_WEAPON_LOADOUTS; ++i)
-			{
-				if (CNEOWeaponLoadout::s_LoadoutWeapons[m_iNeoClass][i].m_iWeaponPrice > m_iXP)
-				{
-					break;
-				}
-				iChosenWeps[iChosenWepsSize++] = i;
-			}
-		}
-
-		if (iChosenWepsSize == 1)
-		{
-			m_iLoadoutWepChoice = iChosenWeps[0];
-		}
-		else
-		{
-			m_iLoadoutWepChoice = iChosenWeps[RandomInt(0, iChosenWepsSize - 1)];
-		}
-	}
+	ChooseRandomWeapon();
 
 	BaseClass::Spawn();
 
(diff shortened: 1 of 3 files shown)


sunzenshen:
Addresses an issue where bots would attempt to pick a weapon, but the selected a slot that was invalid, so they end up falling back to the default MPN choice.  This often leads to bots with higher ranks being forced to pick the rankless weapon default.

The symptoms coincide with the following error logging:

```cpp
// in IServerNetworkable *CEntityFactoryDictionary::Create( const char *pClassName )
Warning("Attempted to create unknown entity type %s!\n", pClassName );
// where pClassName was "weapon_empty"

// in bool CNEO_Player::RequestSetLoadout(int loadoutNumber)
Warning("NULL Ent in RequestSetLoadout!\n");
// where loadoutNumber ended up being an index that was outside of the class's valid weapon slot range

// in bool CNEO_Player::RequestSetLoadout(int loadoutNumber)
DevMsg("Insufficient XP for %s\n", pszWepName);
// where for example pszWepName was "weapon_pz"
```

To my understanding from looking at the debugger, it seems that bots don't update their requested loadout index before calling RequestSetLoadout, even if they had just changed their class in Spawn code.  This can result in the existing loadout index ending up being past the valid index range for a present class and rank.

My proposal is to improve the consistency of bots selecting a class and then loadout, by first making sure the class choice index is set early, and to then select another random loadout choice index immediately after,  My thinking is that this will make sure that the loadout index is always in a valid range for the bot's selected class/rank.

For my testing using the debug build for asserts, I ran a match with 31 bots with the following script:
```
mp_chattime 2; 
host_timescale 3;
```
... and let the game run in the background while waiting for any triggered asserts.  I checked that after this change, the aforementioned asserts no longer happen, and that I don't see any Support bots getting stuck with the MPN at higher ranks.

sunzenshen:
Example of the error logs:

Screenshot 2026-07-04 003624

Screenshot 2026-07-04 003559

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -651,6 +616,53 @@ void CNEOBot::Spawn()
...
+    else
+    {
+        return iChosenWeps[RandomInt(0, iChosenWepsSize - 1)];
+    }
+}
+
+void CNEOBot::ChooseRandomWeapon()
+{
+	m_iLoadoutWepChoice = ChooseRandomWeaponIndex();
+}
```
sunzenshen, line 688: Tried to make this block as similar as possible with what was originally in void CNEOBot::Spawn() , but doesn't quite show up easily in the GitHub diff.

Image

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -577,52 +577,17 @@ CNEOBot::~CNEOBot()
...
-		if (iChosenWepsSize == 1)
-		{
-			m_iLoadoutWepChoice = iChosenWeps[0];
-		}
-		else
-		{
-			m_iLoadoutWepChoice = iChosenWeps[RandomInt(0, iChosenWepsSize - 1)];
-		}
-	}
+	ChooseRandomWeapon();
```
sunzenshen, line 614: Refactored this section below as ChooseRandomWeapon()

nullsystem (approved):

AdamTadeusz (approved):

PR #2009, opened 2026-07-05
Assisted suicides also by world and indicate assists in kill feed.

Agiel:
## Description
#1851 fixed awarding assists when suiciding by your own grenade, but missed the case where the killer is worldspawn. This PR should handle that. I also added the assist indicator to the kill feed.

Note that the kill feed text will sometimes overflow the container but I haven't been able to figure out why. Pretty sure this was a problem before as well though. I restructured the space calculation and text rendering so they're closer and easier to follow but that didn't help me narrow it down. I think we can leave it until someone decides to rewrite the feed using NeoUI.

Without assist:
image
With assist:
image

## Toolchain

- Linux GCC Distro Native - CachyOS - gcc version 16.1.1 20260625 (GCC)

## Linked Issues

- fixes #
- related #

### DIFF
--- a/src/game/client/neo/ui/neo_hud_deathnotice.cpp
+++ b/src/game/client/neo/ui/neo_hud_deathnotice.cpp
@@ -384,35 +384,35 @@ void CNEOHud_DeathNotice::SetDeathNoticeItemDimensions(DeathNoticeItem* deathNot
 	}
 	else
 	{	// Player killed message
-		if (deathNoticeItem->Killer.iEntIndex != 0 && !deathNoticeItem->bSuicide)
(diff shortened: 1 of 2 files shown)


Rainyan (approved):
LGTM, there's still the case of #2010 but maybe that's out of scope for this one

Agiel:
> LGTM, there's still the case of #2010 but maybe that's out of scope for this one

Yes let's handle that separately. Another issue we might have is subtracting too many points when negative assists are enabled, but rather than fixing it I think we should discuss removing the feature altogether. I can't see any servers enabling it.

AdamTadeusz (approved):
Image

👍

PR #2013, opened 2026-07-08
Make `prop_physics_override` use `CPhysicsPropMultiplayer`

DESTROYGIRL:
## Description
This is a somewhat blind attempt to fix the issue where players will be teleported when touching `prop_physics_override` props. I couldn't replicate the issue in testing.
The reasoning behind this change is that almost all physics props in the game are, or should be `prop_physics_multiplayer`. This is because this entity class handles physics differently to avoid issues in multiplayer. `prop_physics_override` doesn't have this, and therefore I believe it's the reason we see issues with them but not other props.

I can't think of any scenarios that exist currently where a `prop_physics_override` would need to be a `CPhysicsProp` to function correctly. The game is multiplayer at it's core so I don't think this change will have any major negative impact.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/props.cpp
+++ b/src/game/server/props.cpp
@@ -2404,7 +2404,9 @@ void COrnamentProp::InputDetach( inputdata_t &inputdata )
 //=============================================================================
 LINK_ENTITY_TO_CLASS( physics_prop, CPhysicsProp );
 LINK_ENTITY_TO_CLASS( prop_physics, CPhysicsProp );	
+#ifndef NEO
 LINK_ENTITY_TO_CLASS( prop_physics_override, CPhysicsProp );	
+#endif
 
 BEGIN_DATADESC( CPhysicsProp )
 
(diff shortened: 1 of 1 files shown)


Rainyan:
Is there any demo or informal bug report/repro steps of the bug this fix is targeting anywhere?

DESTROYGIRL:
> Is there any demo or informal bug report/repro steps of the bug this fix is targeting anywhere?

There's no solid repro steps other than jump around on them a bunch. Video examples from ntre_rise_ctg before they were fixed (not mine):


https://github.com/user-attachments/assets/e4d9e2e6-2d48-4118-8e06-4accf6209f56


https://github.com/user-attachments/assets/4f8ecbb7-6ce7-46a5-a57c-49991ad1776b


If you load the original nt_rise_ctg, they should have the same behaviours I think

AdamTadeusz:
What was the map fix for this?

DESTROYGIRL:
> What was the map fix for this?

They are no longer physics props, since they never moved in the first place

AdamTadeusz (approved):
Can't replicate the issue on either version but I also don't see any issues with merging this

Masterkatze (approved):

Issue #1874, opened 2026-03-20
The damage report lies


AdamTadeusz:
Often the numbers in the damage report that shows up after dying do not reflect reality. Players that show up as killed are actually still alive, and unless we have some serious hitreg issues I don't trust the bullets hit and damage done too

AdamTadeusz:
I found some steps to reproduce the red damage numbers issue at least.
1. map ntre_oilstain_ctg
2. nb_player_stop 1
3. pick a play team
4. once the round starts, go to the enemy spawn and kill all the bots.
5. once the next round starts, go to the enemy spawn and shoot all the bots once non-fatally.
6. nb_player_stop 0
7. Let the enemy bots kill you
8. The damage report will look something like this:

<img width="491" height="414" alt="Image" src="https://github.com/user-attachments/assets/283948ee-f7b3-42eb-8d06-74c0ffec43fb" />

Note how all the damage done by the local player numbers are red. This indicates that the local player killed all these bots. However that's not what happened, we only killed all these bots in the previous round. This round we only shot them once and they are all still alive when the screenshot was taken.

nullsystem:
The client's NEO local player never really got a proper reset code path when it spawns in. Fixing that through the use of spawn counter to put a true on first tick + resets this fixes that and potentially other issues that may been caused by it.

PR #2018, opened 2026-07-12
Fix local player's players killed report not resetting each round

nullsystem:
## Description

Two fixes: First m_bFirstAliveTick never gets updated when the player respawns past the first respawn since not dying on a round will not turn this to true and the respawn resets through m_bFirstAliveTick for local player in PreThink doesn't get utilized. This may also be the cause of other non-reset issues people seen before, but this has affected the killed players report especially.

Second, move the array from C_NEO_Player into its own global array as really it's only a single instance (for the local player) and pointless for every single C_NEO_Player to hold it. For sanity check, change from it being entity index array based to UserID based to be sure which is which when its displayed and not affected by entity index changes.

Killer info/scoreboard:

- Checking for pre-round freeze in killer_info no longer needed
  - Now the c_neo_player first tick alive got fixed, can just do it
    there.
- Use MAX_PLAYERS_ARRAY_SAFE
- Move local player check

## Toolchain

## Linked Issues

* related https://github.com/NeotokyoRebuild/neo/issues/1874


src/game/client/neo/ui/neo_scoreboard.cpp
```diff
@@ -392,7 +392,15 @@ void CNEOScoreBoard::Update()
 			pPlayerInfo->iTakenDmgs = pLocalPlayer->GetAttackersScores(pNeoPlayer->entindex());
 			pPlayerInfo->iTakenHits = pLocalPlayer->GetAttackerHits(pNeoPlayer->entindex());
 			pPlayerInfo->bKilledYou = (pNeoPlayer->entindex() == g_neoKillerInfos.iEntIndex);
-			pPlayerInfo->bYouKilled = pLocalPlayer->m_rfNeoPlayerIdxsKilledByLocal[pNeoPlayer->entindex()];
+			pPlayerInfo->bYouKilled = false;
+			for (int i = 0; i < g_neoUserIDsLocalKilledSize; ++i)
```
Masterkatze: Rename `i` so it won't shadow `i` at line 334?

Masterkatze: `g_neoUserIDsLocalKilledSize` is used but `c_neo_killer_infos.h` is not included? Might break in non-unity build.

sunmachine:
Would clearing on the `round_start` event instead of the first alive tick be cleaner here, since it also covers players who don't respawn (dead/spectating) at round changeover?

EDIT: It's defined in `neo_gamerules.cpp:2954`, and I think you'd need to register it at... `neo_hud_deathnotice.cpp:208`, `CNEOHud_DeathNotice::Init`.

src/game/client/neo/ui/neo_hud_deathnotice.cpp
sunmachine, line 214: I think you can register the round_start event and get it even before the first tick and avoid the edge cases.

```suggestion
	ListenForGameEvent( "vip_death" );
	ListenForGameEvent( "round_start" ); 
```

sunmachine, line 718: At the top of `CNEOHud_DeathNotice::FireGameEvent`, we can just check for the `round_start` event right?

```suggestion
	// Not sure if this is *the best* place for it, 
	// but you can make sure it's always checked at round_start.
	auto eventName = event->GetName();
	if (!Q_stricmp(eventName, "round_start"))
	{
		NeoUserIDsLocalKilledClear();
		V_memset(&g_neoKillerInfos, 0, sizeof(CNEOKillerInfos));
		return;
	}
```

sunmachine, line 732: But if you go with my suggestion, you'll need to move this up over the `start_round` check.

src/game/client/neo/c_neo_player.cpp
```diff
@@ -1207,6 +1208,9 @@ void C_NEO_Player::PreThink( void )
 			CLocalPlayerFilter filter;
 			enginesound->SetPlayerDSP(filter, 0, true);
 
+			NeoUserIDsLocalKilledClear();
+			V_memset(&g_neoKillerInfos, 0, sizeof(CNEOKillerInfos));
+
```
sunmachine, line 1213: See the above comments, but these two lines get moved (removed here) if we're checking in `CNEOHud_DeathNotice::FireGameEvent` at `start_round` time.

sunmachine (changes requested):
This is what I mean [from my comment before](https://github.com/NeotokyoRebuild/neo/pull/2018#issuecomment-5017436529). Feel free to explain why this isn't the way, but I think it'll avoid the edge-cases caused by waiting until the first alive tick.

src/game/client/neo/c_neo_player.cpp
```diff
@@ -479,6 +479,7 @@ C_NEO_Player::C_NEO_Player()
 	m_flTocFactor = 0.15f;
 
 	memset(m_szNeoNameWDupeIdx, 0, sizeof(m_szNeoNameWDupeIdx));
+	if (IsLocalPlayer()) NeoUserIDsLocalKilledClear();
```
sunmachine, line 482: This and [:1632](https://github.com/NeotokyoRebuild/neo/pull/2018/changes/BASE..3cbdb97b40b6098ae2d42c020c385338f37a79d5#diff-9b923d4958beed1904e7fc3522efcaa2512cf4fa36df20d6b619a2231105016bR1632) become redundant if you use `round_start`.

src/game/client/neo/c_neo_player.cpp
```diff
@@ -1625,11 +1629,10 @@ void C_NEO_Player::Spawn( void )
 		m_rfAttackersAccumlator.GetForModify(i) = 0.0f;
 		m_rfAttackersHits.GetForModify(i) = 0;
 	}
-	V_memset(m_rfNeoPlayerIdxsKilledByLocal, 0, sizeof(m_rfNeoPlayerIdxsKilledByLocal));
+	if (IsLocalPlayer()) NeoUserIDsLocalKilledClear();
```
sunmachine, line 1632: Same note. Redundant if using the `round_start` suggestion.

nullsystem:
@sunmachine I think first alive tick also covers when players are allowed to respawn within the round (DM/TDM/JGR) which round_start won't, also the context of the report is for when the player's alive since the latest spawning in so first tick makes more sense.

nullsystem:
@Masterkatze Changes based on code review: https://github.com/NeotokyoRebuild/neo/pull/2018/changes/71703f81a9a5778798d583e30ae7a22a10a0d4d8

sunmachine (approved):

Masterkatze (approved):

sunzenshen (approved):

PR #2019, opened 2026-07-14
Launch Linux and Windows through a launcher, with SDK dependency

sunmachine:
Co-authored-by: @DESTROYGIRL

## Description

NT;RE ships its own engine binaries on its own appid; only the HL2/platform assets come from Source SDK Base 2013 MP, mounted through the `|appid_243750|` search paths in `gameinfo.txt`. Valve's stock launcher assumes the opposite layout and runs the game from the SDK install, which is why the SDK asset mounts silently fail on Linux. This PR makes `launcher_main` a real standalone mod launcher: the game runs from its own install with its own `bin/`, and the SDK is nothing more than an asset dependency. 

Much of its functionality can be found in Valve's `hl2.sh` script (the old `ntre.sh` script), except now ported over directly into the launcher. 

Details:
- Linux launches in-process like Windows: own `bin/linux64` `LD_LIBRARY_PATH`, dlopen our `launcher.so`, no `hl2.sh` or Valve bootstraps.
- Works around the engine opening `|appid_243750|` mounts with the SDK name partly lowercased (silent ENOENT on case-sensitive filesystems) via  self-repairing symlink.
- Missing or broken SDK install shows an error dialog instead of crashing in `MXR_LoadAllSoundMixers`.
- No `-game` in launch options; derived from the binary name (`neo_linux64` -> `neo`).
- `-wait_for_debugger` now works on Linux.
- CMake: `NEO_MOD_APPID` defaults to 3172910; opt-in `NEO_INSTALL_LAUNCHER` packages the launcher per the `NEO_INSTALL_*` conventions. No CI changes.
- The earlier `SDK_EXEC_DIR` fix is removed, superseded by this layout.
- Dev setups are unaffected; `sourcemods/` and explicit `-game` launches don't use this launcher.

Verified on Linux via the Steam Play button (`-steam` only): full bot match with all SDK content mounted, error dialog when the SDK files are missing, stale symlink repaired automatically. In-game `path` dump in the comments.

> [!WARNING]
> The Windows flow is code-complete but untested; needs a run on a real Windows install, including with the SDK absent.

Depot follow-ups (not this PR): launcher binaries at the install root, launch options pointing at them with `-steam` only, drop the renamed Valve bootstraps, preserve `bin/` filename casing (engine dlopens `GameUI.so` etc. by exact name).

### Build (branch: linux-build-fix)

- From `src/`, in a Steam Runtime 3.0 (sniper) environment:
  - `cmake --preset linux-release`
  - `cmake --build build/linux-release --target launcher_main`
- Output: `src/build/linux-release/launcher_main/neo_linux64`

### Overlay onto Steam (local testing only)

- Set NT;RE to the `sdk-dependency-testing` branch, "-steam only (dev testing)" launch option.
  - You don't need to do a compatibility override for Sniper, it's already configured depot-side.
- Copy `neo_linux64` into `~/.local/share/Steam/steamapps/common/NEOTOKYOREBUILD/` (next to `neo/`).
- NOTE: This step can be skipped if you're using the new `Launch game without script (dev testing)` option at Play-time.
  - Replace `ntre.sh` there with this shim, then `chmod +x neo_linux64 ntre.sh`:
  ```bash
  #!/bin/bash
  exec "$(cd "${0%/*}" && pwd)/neo_linux64" "$@"
  ```
- Press Play. Optional: -condebug in launch options writes neo/console.log.

Screenshot 2026-07-18 at 12 55 51

## Depot Changes 

- `ntre.sh` isn't required anymore.
- The depot will need to execute from `neo_linux64` as the shim's functionality was ported into the executable. 

## Toolchain
- Windows MSVC VS2022
- Linux GCC 10 Sniper 3.0

## Linked Issues
- fixes #
- related #

## TODO

- [x] Test Windows.
- ~~Update CI so that we can switch NEO_BUILD_LAUNCHER to ON by default.~~
  - To be done in a follow-up PR.


src/CMakeLists.txt
```diff
@@ -51,6 +51,8 @@ option(NEO_USE_SEPARATE_BUILD_INFO "Use separate build info on Linux" OFF)
 option(NEO_ENABLE_CPACK "Enable CPack" OFF)
 option(NEO_USE_MEM_DEBUG "Enable USE_MEM_DEBUG defines" OFF)
 option(NEO_GENERATE_GAMEDATA "Generate SourceMod gamedata" ${NEO_DEDICATED})
+option(NEO_BUILD_LAUNCHER "Build the Steam mod launcher" OFF) # NOTE Linux only
```
Masterkatze: Why `OFF` by default?

sunmachine:
Marking as a draft for now; this proved the problem and POC for the fix (with the patch POC in discord), but it needs to be tidied up before proper review.

sunmachine:
Fix-proof-of-concept is here: https://discord.com/channels/1235346473827434517/1526783378718855268/1526787211507535954

DESTROYGIRL:
Windows build @ https://github.com/sunmachine/neotokyo-rebuild/pull/1

src/CMakeLists.txt
sunmachine: Off for now until CI is modified. 

See updated PR under # TODO

> - [ ] Update CI so that we can switch NEO_BUILD_LAUNCHER to ON by default. @Masterkatze

src/launcher_main/main.cpp
DESTROYGIRL, line 1: Please wrap all code changes in `#ifdef NEO` blocks and leave the original SDK code in the `#else` or `#ifndef NEO` where applicable

src/launcher_main/CMakeLists.txt
```diff
@@ -0,0 +1,77 @@
...
+            $<TARGET_FILE:launcher_main>
+            ${CMAKE_SOURCE_DIR}/../game/
+        COMMAND ${CMAKE_COMMAND} -E make_directory
+            ${CMAKE_SOURCE_DIR}/../game/bin/linux64
+        COMMAND ${CMAKE_COMMAND} -E copy_if_different
+            ${CMAKE_SOURCE_DIR}/lib/public/linux64/libsteam_api.so
+            ${CMAKE_SOURCE_DIR}/../game/bin/linux64/
+        VERBATIM
+    )
+endif()
```
DESTROYGIRL, line 90: These strictly use linux paths and names, I didn't change these for the windows build because I wasn't totally clear on the intended functionality xd
Otherwise it'd just be `game/bin/linux64` to `game/bin/x64` and `libsteam_api.so` to `steam_api.dll` I believe

Also- is copying the steamapi still desired? It seems to have been intentionally excluded in the changes to `src/CMakeLists.txt`

src/launcher_main/main.cpp
DESTROYGIRL, line 253: Up for debate, it was my personal feelings that we shouldn't be enforcing the rules around how the SDK is installed before the gameinfo loading stage (though, this might be unavoidable given how your fix to the bug is implemented), but we seem to have lost a bit of UX here between the base behaviour and our changes:

Original behaviour (sdk-dependancy-testing branch unmodified with `-steam -game neo`. Prompts user that SDK is missing, brings up steam install dialog:

Image

New behaviour:

Image

Can the steam install dialog be restored?

DESTROYGIRL (changes requested):
Works and compiles without warning on windows. Tested playing on a v30 dedicated server, no issues noticed.

Tested:
- Running on Windows works
- Running on Windows without SDK 2013 already installed works
- Running on Linux works, though I noticed some console errors regarding shaders, unsure if its expected. A more seasoned linux player can verify
- Running on Linux without SDK 2013 already installed works (check comments)

DESTROYGIRL:
A TODO on this is to create a workflow for github actions that produces a new archive containing the launcher as part of the builds
(maybe this can be handled in a different PR...?)

src/launcher_main/main.cpp
DESTROYGIRL: Tested on windows, exact same scenario

sunmachine: Great that the error appeared! Still looking into restoring the previous UX. 💪 For some reason it never triggered for me... and I think I know why and it's stupid.

sunmachine: Check ✅ [`62a126c`](https://github.com/NeotokyoRebuild/neo/pull/2019/commits/62a126cd856fe679f80bc1a65f2635d8cb5963c1)

sunmachine: Restored! [`b91a9b9`](https://github.com/NeotokyoRebuild/neo/pull/2019/commits/b91a9b91d760b9147fc39887f37ba8110242bf85)

src/launcher_main/CMakeLists.txt
sunmachine: I adjusted this by guarding the copies per-OS. 

- The launcher copy now runs on both platforms. 
- The steam_api copy is Linux-only, and that's on purpose: the repo only carries a runtime steam_api for Linux. 
  - On Linux the linker needs the real `libsteam_api.so` at build time, so it's checked in under `lib/public/linux64` and we can stage it into `game/bin/linux64` for from-source builds. 
  - On Windows the linker only needs the import stub (`lib/public/x64/steam_api64.lib`). The actual `steam_api64.dll` isn't in the repo at all, so there's nothing to copy; it ships with the engine binaries in `bin/x64`.
  
As for copying the steam_api, and is it desired? Yup! 

The exclusion from the install rule in `src/CMakeLists.txt` is intentional for the same reason: the packaged install overlays onto a depot that already has steam_api next to the engine binaries, so only the launcher itself is staged. The POST_BUILD copy exists just for the dev layout, where a from-source `game/` tree has no `bin/` directory until we create one.

sunmachine: Oh, [`c79f7a3`](https://github.com/NeotokyoRebuild/neo/pull/2019/changes/c79f7a35e84408cac9d6fd935cba224652482f17)

DESTROYGIRL (dismissed):
Working on linux / windows with SDK install prompt as well

src/launcher_main/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
+# Steam mod launcher. Runs the game from the mod's own install directory
+# using the mod's shipped engine binaries; the Source SDK Base 2013 MP
+# install is only an asset dependency, mounted through gameinfo.txt.
+if(OS_LINUX)
+find_package(SDL2 REQUIRED)
```
Masterkatze: No indentation here and in other `if` branches.

src/launcher_main/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
...
+            ${CMAKE_SOURCE_DIR}/../game/
+        VERBATIM
+    )
+    # The repo only carries the Linux steam_api runtime; on Windows
+    # lib/public/x64 has the import lib, and the launcher loads
+    # bin/x64/steam_api64.dll from a Source SDK Base / depot install.
+    if(OS_LINUX)
+        add_custom_command(TARGET launcher_main POST_BUILD
+            COMMAND ${CMAKE_COMMAND} -E make_directory
+                ${CMAKE_SOURCE_DIR}/../game/bin/linux64
```
Masterkatze: Missing `/`?

sunmachine:
Marking TODOs as satisfied.

src/CMakeLists.txt
```diff
@@ -211,7 +217,6 @@ if(OS_WINDOWS)
         _ALLOW_RUNTIME_LIBRARY_MISMATCH
         _ALLOW_ITERATOR_DEBUG_LEVEL_MISMATCH
         _ALLOW_MSC_VER_MISMATCH
-        fopen=dont_use_fopen
```
Masterkatze, line 214: Add this define to `tier1` target.

DESTROYGIRL: My bad!

src/launcher_main/CMakeLists.txt
sunmachine: Conforming to style 👍

src/launcher_main/CMakeLists.txt
sunmachine: Conforming to best practice 👍

src/CMakeLists.txt
sunmachine: Done

DESTROYGIRL (dismissed):

Masterkatze (dismissed):
LGTM

Masterkatze (approved):

DESTROYGIRL (approved):

PR #2000, opened 2026-06-27
Bots can attack toward an optional coordinate

sunzenshen:
## Description
Allows inputting an optional vector position into the bot Attack behavior which then serves as a goal direction to look for cover towards while engaging the enemy.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #1806 which served as a prototype for selecting cover to advance toward the primary known threat

PR #2020, opened 2026-07-14
Bots avoid grenade-related hazard areas

sunzenshen:
## Description
Bots avoid hazard areas:
- areas that are in line of sight of a smoke cloud
- areas that are in line of sight of a grenade trajectory

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/NextBot/NextBotVisionInterface.cpp
+++ b/src/game/server/NextBot/NextBotVisionInterface.cpp
@@ -21,6 +21,8 @@
 #include "neo_player.h"
 #include "neo_smokelineofsightblocker.h"
 #include "bot/neo_bot.h"
+#include "bot/neo_bot_path_reservation.h"
+#include "nav_mesh.h"
 #endif
 
 #include "tier0/vprof.h"
@@ -794,6 +796,33 @@ bool IVision::IsLineOfSightClearToEntity( const CBaseEntity *subject, Vector *vi
 	UTIL_TraceLine( GetBot()->GetBodyInterface()->GetEyePosition(), firstPosition, MASK_BLOCKLOS_AND_NPCS|CONTENTS_IGNORE_NODRAW_OPAQUE, &filter, &result );
 	if ( result.DidHit() )
 	{
+		// Check if first center traceline hit any blocking smoke
+		if (neo_bot_path_reservation_enable.GetBool()
+			&& result.m_pEnt
+			&& FStrEq(result.m_pEnt->GetClassname(), SMOKELINEOFSIGHTBLOCKER_ENTITYNAME))
+		{
+			auto pSmoke = static_cast<CNEOSmokeLineOfSightBlocker*>(result.m_pEnt);
+			// Assume that threat could be hiding inside navarea where smoke blocked sight
+			if (neoBot)
+			{
+				int team = neoBot->GetTeamNumber();
+
+				CNavArea *smokeArea = TheNavMesh->GetNearestNavArea(result.endpos);
+				if (smokeArea)
+				{
+					// Register smoke sightlines for team to avoid as hazardous
+					CNEOBotPathReservations()->AddSmokeHazard(smokeArea->GetID(), pSmoke->GetNextThink(), team);
+				}
+
+				CNavArea *myArea = neoBot->GetLastKnownArea();
+				if (myArea)
+				{
+					// Also register the area I saw the smoke from as hazardous
+					CNEOBotPathReservations()->AddSmokeHazard(myArea->GetID(), pSmoke->GetNextThink(), team, false);
+				}
+			}
+		}
+
 		UTIL_TraceLine( GetBot()->GetBodyInterface()->GetEyePosition(), secondPosition, MASK_BLOCKLOS_AND_NPCS|CONTENTS_IGNORE_NODRAW_OPAQUE, &filter, &result );
 
 		if ( result.DidHit() )
--- a/src/game/server/neo/bot/behavior/neo_bot_attack.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_attack.cpp
@@ -66,6 +66,7 @@ public:
 		m_threatArea = threat->GetLastKnownArea();
 		m_goalArea = goalArea ? goalArea : m_threatArea; // prioritize movement towards input goal area or threat
 		m_myDistToGoalSq = m_goalArea ? ( m_goalArea->GetCenter() - m_me->GetAbsOrigin() ).LengthSqr() : 0;
+		m_onStuckPenalty = neo_bot_path_reservation_onstuck_penalty.GetFloat();
 	}
 
 	bool IsBetterCandidate( CNavArea *area, float avoidPenalty ) const
@@ -118,6 +119,20 @@ public:
 		if ( area == m_myArea )
 		{
 			return true; // skip our starting area
+		}
+
+		// Skip areas that are hazardous or where bots get stuck
+		if ( neo_bot_path_reservation_enable.GetBool() )
+		{
+			int navAreaId = area->GetID();
+			if (CNEOBotPathReservations()->IsAreaHazardous(navAreaId, m_me))
+			{
+				return true;
+			}
+			if (CNEOBotPathReservations()->GetAreaAvoidPenalty(navAreaId) >= m_onStuckPenalty)
+			{
+				return true;
+			}
 		}
 
 		float avoidPenalty = neo_bot_path_reservation_enable.GetBool()
@@ -215,6 +230,7 @@ public:
 	const CNavArea *m_myArea;     // reference point of myself
 	const CNavArea *m_threatArea; // reference point of the threat
 	float m_myDistToGoalSq;       // the bot's current distance to the threat
+	float m_onStuckPenalty;       // cache onstuck penalty
 };
 
 
@@ -466,11 +482,6 @@ QueryResultType	CNEOBotAttack::ShouldRetreat( const INextBot *me ) const
 	if (!m_bSawEnemySinceLastPathCompute)
 	{
 		return ANSWER_UNDEFINED;
-	}
-
-	if ( m_goalArea && m_attackCoverArea )
-	{
-		return ANSWER_NO;
 	}
 
 	return ANSWER_UNDEFINED;
--- a/src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.cpp
@@ -9,6 +9,8 @@
 #include "weapon_grenade.h"
 #include "weapon_smokegrenade.h"
 
+ConVar sv_neo_bot_grenade_polite_frag("sv_neo_bot_grenade_polite_frag", "1", FCVAR_NONE, "Force bots to consider teammates when throwing frag grenades.", true, 0, true, 1);
+ConVar sv_neo_bot_grenade_polite_smoke("sv_neo_bot_grenade_polite_smoke", "1", FCVAR_NONE, "Force bots to consider teammates when throwing smoke grenades.", true, 0, true, 1);
 ConVar sv_neo_bot_grenade_use_frag("sv_neo_bot_grenade_use_frag", "1", FCVAR_NONE, "Allow bots to use frag grenades", true, 0, true, 1);
 ConVar sv_neo_bot_grenade_use_smoke("sv_neo_bot_grenade_use_smoke", "1", FCVAR_NONE, "Allow bots to use smoke grenades", true, 0, true, 1);
 ConVar sv_neo_bot_grenade_throw_cooldown("sv_neo_bot_grenade_throw_cooldown", "10", FCVAR_NONE, "Cooldown in seconds between grenade throws for bots");
@@ -78,29 +80,32 @@ Action< CNEOBot > *CNEOBotGrenadeDispatch::ChooseGrenadeThrowBehavior( const CNE
 	// Should I toss a smoke grenade?
 	if ( pSmokeGrenade && (me->GetClass() == NEO_CLASS_SUPPORT) )
 	{
-		for ( int i = 1; i <= gpGlobals->maxClients; i++ )
+		if ( sv_neo_bot_grenade_polite_smoke.GetBool() )
 		{
-			CNEO_Player *pPlayer = ToNEOPlayer( UTIL_PlayerByIndex( i ) );
-			if ( !pPlayer || !pPlayer->IsAlive() || pPlayer == me )
+			for ( int i = 1; i <= gpGlobals->maxClients; i++ )
 			{
-				continue;
-			}
-
-			if ( !me->IsEnemy( pPlayer ) )
-			{
-				if ( !pPlayer->IsBot() && pPlayer->GetClass() != NEO_CLASS_SUPPORT )
+				CNEO_Player *pPlayer = ToNEOPlayer( UTIL_PlayerByIndex( i ) );
+				if ( !pPlayer || !pPlayer->IsAlive() || pPlayer == me )
 				{
-					// Avoid blocking the vision of a friendly human
-					// (Bots benefit from concealment without the disorientation)
-					return nullptr;
+					continue;
 				}
-			}
-			else
-			{
-				if ( pPlayer->GetClass() == NEO_CLASS_SUPPORT )
+
+				if ( !me->IsEnemy( pPlayer ) )
 				{
-					// Avoid giving an enemy with thermal vision a free smoke screen
-					return nullptr;
+					if ( !pPlayer->IsBot() && pPlayer->GetClass() != NEO_CLASS_SUPPORT )
+					{
+						// Avoid blocking the vision of a friendly human
+						// (Bots benefit from concealment without the disorientation)
+						return nullptr;
+					}
+				}
+				else
+				{
+					if ( pPlayer->GetClass() == NEO_CLASS_SUPPORT )
+					{
+						// Avoid giving an enemy with thermal vision a free smoke screen
+						return nullptr;
+					}
 				}
 			}
 		}
@@ -111,9 +116,12 @@ Action< CNEOBot > *CNEOBotGrenadeDispatch::ChooseGrenadeThrowBehavior( const CNE
 	// Should I toss a frag grenade? 
 	if ( pFragGrenade )
 	{
-		if ( !CNEOBotGrenadeThrowFrag::IsFragSafe( me, threat->GetLastKnownPosition() ) )
+		if ( sv_neo_bot_grenade_polite_frag.GetBool() )
 		{
-			return nullptr;
+			if ( !CNEOBotGrenadeThrowFrag::IsFragSafe( me, threat->GetLastKnownPosition() ) )
+			{
+				return nullptr;
+			}
 		}
 
 		return new CNEOBotGrenadeThrowFrag( pFragGrenade, threat );
--- a/src/game/server/neo/bot/behavior/neo_bot_grenade_throw_frag.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_grenade_throw_frag.cpp
@@ -1,11 +1,13 @@
 #include "cbase.h"
 #include "bot/neo_bot.h"
 #include "bot/neo_bot_path_compute.h"
+#include "bot/neo_bot_path_reservation.h"
 #include "bot/behavior/neo_bot_grenade_throw_frag.h"
 #include "neo_gamerules.h"
 #include "neo_player.h"
 #include "weapon_neobasecombatweapon.h"
 
+#include "nav_mesh.h"
 #include "nav_pathfind.h"
 
 extern ConVar sv_neo_grenade_blast_radius;
@@ -104,6 +106,15 @@ CNEOBotGrenadeThrow::ThrowTargetResult CNEOBotGrenadeThrowFrag::UpdateGrenadeTar
 	if ( !IsFragSafe( me, m_vecTarget ) )
 	{
 		return THROW_TARGET_CANCEL; // risk of friendly fire
+	}
+
+	// Register explosive hazard at the target position
+	// so the throwing bot's team can avoid the trajectory and landing areas
+	CNavArea *area = TheNavMesh->GetNearestNavArea(m_vecTarget);
+	if (area)
+	{
+		int team = me->GetTeamNumber();
+		CNEOBotPathReservations()->AddFragHazard(area->GetID(), gpGlobals->curtime + sv_neo_grenade_fuse_timer.GetFloat(), team);
 	}
 
 	return THROW_TARGET_READY;
--- a/src/game/server/neo/bot/behavior/neo_bot_retreat_from_grenade.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_retreat_from_grenade.cpp
@@ -5,6 +5,8 @@
 #include "bot/behavior/neo_bot_retreat_from_grenade.h"
 #include "bot/behavior/neo_bot_retreat_to_cover.h"
 #include "bot/neo_bot_path_compute.h"
+#include "bot/neo_bot_path_reservation.h"
+#include "nav_mesh.h"
 #include "sdk/sdk_basegrenade_projectile.h"
 
 // memdbgon must be the last include file in a .cpp file!!!
@@ -72,6 +74,7 @@ public:
 	{
 		m_me = me;
 		m_grenade = grenade;
+		m_onStuckPenalty = neo_bot_path_reservation_onstuck_penalty.GetFloat();
 		m_pGrenadeStats = dynamic_cast<CBaseGrenadeProjectile *>( grenade );
 		m_safeRadiusSqr = m_pGrenadeStats ? Square(m_pGrenadeStats->m_DmgRadius * sv_neo_bot_grenade_frag_safety_range_multiplier.GetFloat()) : 0.0f;
 
@@ -89,6 +92,20 @@ public:
 		{
 			// edge case: we just want to bail to let Update exit this behavior
 			return false;  // can't search if there's no threat source
+		}
+
+		// Skip areas that are hazardous or where bots get stuck
+		if ( neo_bot_path_reservation_enable.GetBool() )
+		{
+			int navAreaId = area->GetID();
+			if (CNEOBotPathReservations()->IsAreaHazardous(navAreaId, m_me))
+			{
+				return true;
+			}
+			if (CNEOBotPathReservations()->GetAreaAvoidPenalty(navAreaId) >= m_onStuckPenalty)
+			{
+				return true;
+			}
 		}
 
 		if ( m_pGrenadeStats )
@@ -138,6 +155,7 @@ public:
 	CNEOBot *m_me;
 	CBaseEntity *m_grenade;
 	CBaseGrenadeProjectile *m_pGrenadeStats;
+	float m_onStuckPenalty;
 	float m_safeRadiusSqr;
 	CUtlVector< CNavArea * > m_coverAreaVector;
 };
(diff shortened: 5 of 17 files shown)


src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
```diff
@@ -195,16 +197,14 @@ EventDesiredResult<CNEOBot> CNEOBotMainAction::OnKilled( CNEOBot *me, const CTak
 	// Intended to add some variance to pathing for similar starting scenarios
 	if ( const CNavArea *navArea = me->GetLastKnownArea() )
 	{
-		CNEOBotPathReservations()->IncrementAreaAvoidPenalty( navArea->GetID(), neo_bot_path_reservation_killed_penalty.GetFloat() );
-	}
-	else
-	{
-		// Fallback if GetLastKnownArea is null, try finding nearest nav area
-		CNavArea *nearestArea = TheNavMesh->GetNearestNavArea( me->GetAbsOrigin() );
```
sunzenshen, line 203: I decided that if for some reason the situation was bugged enough that we don't know the GetLastKnownArea of a bot, we shouldn't bother trying to search for GetNearestNavArea.  GetNearestNavArea is a bit more expensive and, if the situation is glitched, might not even return an appropriate navarea.

src/game/server/neo/bot/neo_bot_path_cost.cpp
```diff
@@ -55,158 +55,164 @@ float CNEOBotPathCost::operator()(CNavArea* baseArea, CNavArea* fromArea, const
...
+
+	if ( CNEOBotPathReservations()->IsAreaHazardous(area->GetID(), m_me) )
 	{
-		if (!m_me->GetLocomotionInterface()->IsAreaTraversable(area))
+		if ( (m_routeType != RETREAT_ROUTE) || (m_routeType != FASTEST_ROUTE) )
 		{
-			return -1.0f;
+			return -1.0f; // attempt to route around hazards
 		}
+	}
```
sunzenshen, line 73: The diff on this file is kind of ugly, but this hazard avoidance marking is the new change. A return value of -1.0 is considered an impassable navarea.  The assumption is that this impassability applies only when a bot is not in a hazard area, and if a bot was in a hazard area, then they will check that in tactical monitor and try to exit the hazard area.

src/game/server/neo/bot/neo_bot_path_cost.cpp
```diff
@@ -55,158 +55,164 @@ float CNEOBotPathCost::operator()(CNavArea* baseArea, CNavArea* fromArea, const
 		// first area in path, no cost
 		return 0.0f;
 	}
-	else
```
sunzenshen, line 58: The rest of this file was getting rid of one level of indentation inherited from copying CSimpleBotPathCost, by getting rid of this else.

sunzenshen (commented):
The goal of this PR was to have bots avoid areas that were visible from smoked areas, if they do not have thermal vision.  Before this PR, bots tended to disregard smoke and would fall victim to either Support bots or humans that know to spray into the smoke.  This PR introduces a time-expiring hazard area marking lookup for smoke and general hazardous areas.

Along with support for smoke, I added hazard area marking for grenades and death locations.  The thinking for bot grenade trajectories is that they are so erratic that it's safer to assume that any place along the trajectory towards the target and beyond is a dangerous area.  For marking death areas of friendly bots, it's mostly to have bots pause towards running into a fatal funnel if the room is narrow enough to be blocked by the bot death locations.

sunzenshen:
Demo example of bots moving out of a room with smoke coverage, into one where there is no exposed smoke:

https://github.com/user-attachments/assets/4f38e815-998e-46ad-92e5-29e008c129e3

src/game/server/neo/bot/behavior/neo_bot_attack.cpp
```diff
@@ -79,10 +80,18 @@ class CSearchForAttackCover : public ISearchSurroundingAreasFunctor
...
+		// Skip areas that are hazardous or where bots get stuck
+		if ( neo_bot_path_reservation_enable.GetBool() )
 		{
-			return true; // skip areas that have had navigation hiccups
+			int navAreaId = area->GetID();
+			if (CNEOBotPathReservations()->IsAreaHazardous(navAreaId, m_me))
+			{
+				return true;
+			}
+			if (CNEOBotPathReservations()->GetAreaAvoidPenalty(navAreaId) >= m_onStuckPenalty)
```
sunzenshen, line 91: I forgot I added avoidance penalties for areas where a bot died, so I made the threshold under which a bot got stuck in a location, allowing to minor penalty for death areas to be used as candidates if under the OnStuck threshold.

src/game/server/neo/bot/behavior/neo_bot_attack.cpp
```diff
@@ -452,11 +462,6 @@ QueryResultType	CNEOBotAttack::ShouldRetreat( const INextBot *me ) const
 		return ANSWER_UNDEFINED;
 	}
 
-	if ( m_goalArea && m_attackCoverArea )
-	{
-		return ANSWER_NO;
```
sunzenshen, line 455: While preventing the bots from retreating when they are rushing for a cover area was okay for proving out #2000, it made the bots a bit suicidal when their chosen advancing cover was too exposed, so I think it's better to retreat when there's any incoming fire.

This also encourages more grenade throw attempts which is helpful for viewing the smoke and grenade avoidance behavior.

src/game/server/neo/bot/behavior/neo_bot_retreat_from_grenade.cpp
```diff
@@ -247,6 +278,9 @@ ActionResult< CNEOBot >	CNEOBotRetreatFromGrenade::Update( CNEOBot *me, float in
 //---------------------------------------------------------------------------------------------
 EventDesiredResult< CNEOBot > CNEOBotRetreatFromGrenade::OnStuck( CNEOBot *me )
 {
+	m_coverArea = FindCoverArea( me );
+	CNEOBotPathCompute(me, m_path, m_coverArea->GetCenter(), FASTEST_ROUTE);
+	CNEOBotPathReservations()->IncrementAreaAvoidPenalty(me->GetLastKnownArea()->GetID(), neo_bot_path_reservation_onstuck_penalty.GetFloat());
```
sunzenshen: If a bot gets stuck, sometimes the chosen cover area is the culprit, so look for another random cover area to reset.

sunzenshen (commented):
I forgot that I also incremented OnStuck area avoidance penalties for bot death areas, not just for bots getting stuck so here's another set of adjustments when testing the allowance of bot death areas as candidates for where to escape a hazard.

src/game/server/neo/bot/behavior/neo_bot_ctg_escort.cpp
```diff
@@ -196,7 +196,7 @@ ActionResult< CNEOBot >	CNEOBotCtgEscort::Update( CNEOBot *me, float interval )
 			m_chasePath.Invalidate();
 
 			CNEOBotPathCompute( me, m_path, vecMoveTarget, SAFEST_ROUTE );
-			m_repathTimer.Start( RandomFloat( 1.0f, 2.0f ) );
+			m_repathTimer.Start( RandomFloat( 5.0f,	10.0f ) );
```
sunzenshen, line 199: Excessively short repath timers can cause bots to indecisively fail to move towards a position, so I increased these timers to allow some movement towards a goal.

src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.cpp
```diff
@@ -9,6 +9,8 @@
 #include "weapon_grenade.h"
 #include "weapon_smokegrenade.h"
 
+ConVar sv_neo_bot_grenade_polite_frag("sv_neo_bot_grenade_polite_frag", "1", FCVAR_NONE, "Force bots to consider teammates when throwing frag grenades.", true, 0, true, 1);
+ConVar sv_neo_bot_grenade_polite_smoke("sv_neo_bot_grenade_polite_smoke", "1", FCVAR_NONE, "Force bots to consider teammates when throwing smoke grenades.", true, 0, true, 1);
```
sunzenshen, line 13: It can be rare to see a smoke deployment with the default bot deployment decisions, so I added a ConVar to induce bots to throw grenades more readily. This helps observe the hazard avoidance behavior more frequently.

sunzenshen: The existing logic below should be mostly the same despite the indentation changes that caused many diff line changes.

src/game/server/neo/bot/behavior/neo_bot_grenade_throw_frag.cpp
```diff
@@ -106,6 +108,15 @@ CNEOBotGrenadeThrow::ThrowTargetResult CNEOBotGrenadeThrowFrag::UpdateGrenadeTar
 		return THROW_TARGET_CANCEL; // risk of friendly fire
 	}
 
+	// Register explosive hazard at the target position
+	// so the throwing bot's team can avoid the trajectory and landing areas
+	CNavArea *area = TheNavMesh->GetNearestNavArea(m_vecTarget);
+	if (area)
+	{
+		int team = me->GetTeamNumber();
+		CNEOBotPathReservations()->AddFragHazard(area->GetID(), gpGlobals->curtime + sv_neo_grenade_fuse_timer.GetFloat(), team);
```
sunzenshen, line 117: Expanding the hazard area around grenades by adjacent Potentially Visible Set is intended as a shorthand for avoid the area along the line through the target, as well as avoid the radiating circle around the idea detonation zone, with some leeway for the trajectory to be off as it usually is.

src/game/server/neo/bot/behavior/neo_bot_retreat_from_hazard_area.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+    virtual bool operator()(CNavArea *baseArea, CNavArea *priorArea, float travelDistanceSoFar)
+    {
+        int id = baseArea->GetID();
+        float candidateHazardTime = CNEOBotPathReservations()->GetAreaHazardousTime(id, m_me);
+        if (candidateHazardTime > 0)
+        {
+            if ( m_me->IsDebugging( NEXTBOT_PATH ) )
+            {
+                baseArea->DrawFilled(255, 0, 0, MIN(255, candidateHazardTime), candidateHazardTime);
```
sunzenshen, line 27: if `nb_debug path` is enabled, then the areas that are considered hazardous will be highlighted as red floor tiles:

Image

One existing feature of `nb_debug path` is that arrows are drawn from the point a bot tried to make a full path and where the goal was, when the path ended up being blocked.  With this PR, this also happens when bots try to path to a position that is marked blocked by the hazard areas.  

I tweaked the default path computation API to allow for partial paths that are as close as possible to the goal without crossing the artificially blocked areas.  This side effect should only affect the debug overlay, while encouraging bots to hide out of view of a smoke cloud.

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
```diff
@@ -287,6 +290,21 @@ ActionResult< CNEOBot >	CNEOBotTacticalMonitor::Update( CNEOBot *me, float inter
...
+	// Might be annoying if bots ignored following or waypoints (e.g. smoke sightlines avoidance)
+	if ( !me->m_hCommandingPlayer.Get() && m_hazardCheckTimer.IsElapsed() )
+	{
+		m_hazardCheckTimer.Start( 0.5f );
+		CNavArea *myArea = me->GetLastKnownArea();
+		if ( myArea )
+		{
+			if ( CNEOBotPathReservations()->IsAreaHazardous( myArea->GetID(), me ) )
+			{
+				return SuspendFor( new CNEOBotRetreatFromHazardArea( ), "Avoiding hazard area" );
```
sunzenshen, line 303: Allow bot commanders to micromanage where they will go instead of having bots cower from smoke clouds.

src/game/server/neo/bot/neo_bot_path_compute.cpp
```diff
@@ -39,20 +39,25 @@ static void CNEOBotReservePath(CNEOBot* me, PathFollower& path)
...
 		CNEOBotReservePath(bot, path);
 		return true;
 	}
 
-	CNEOBotPathCost cost_without_reservations(bot, route);
+	CNEOBotPathCost cost_without_reservations(bot, FASTEST_ROUTE);
 	cost_without_reservations.m_bIgnoreReservations = true;
-	if (path.Compute(bot, goal, cost_without_reservations, maxPathLength, includeGoalIfPathFails, requireGoalArea) && path.IsValid())
+	path.Compute(bot, goal, cost_without_reservations, maxPathLength, includeGoalIfPathFails, requireGoalArea);
+	if (path.IsValid())
```
sunzenshen, line 60: By allowing path.Compute to return false, we can allow bots to create partial paths closer to their goal areas, even if the path is blocked by a banned area.  I use this to allow bots to wander around the edge of a hazard area until a smoke cloud dissipates.

src/game/server/neo/bot/neo_bot_path_compute.h
```diff
@@ -14,7 +14,7 @@ bool CNEOBotPathCompute
 	const Vector& goal,
 	RouteType route,
 	float maxPathLength = PATH_NO_LENGTH_LIMIT,
-	bool includeGoalIfPathFails = PATH_TRUNCATE_INCOMPLETE_PATH,
+	bool includeGoalIfPathFails = true, // return incomplete path if full path not found by default
```
sunzenshen, line 17: By default, allow bots to plan partial routes to a goal even if the goal is blocked off from the bot's navigable area.

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -2823,6 +2824,10 @@ void CNEORules::StartNextRound()
 	m_flNeoNextRoundStartTime = 0;
 	m_flGhostLastHeld = 0;
 
+#ifdef GAME_DLL
+	CNEOBotPathReservations()->ClearRound();
```
sunzenshen, line 2828: If sv_neo_preround_freeze_time and mp_chattime are set too low, the round reset won't leave enough time for smoke hazard areas to reset.  It's also probably cleaner to clear out the bot path reservations as well.

sunzenshen (commented):
Finally figured out why bots completely stop when blocked by a hazard area, and implemented partial pathing up to hazard area boundaries.

Opening up this PR for review again.

Here's the script I used to set up observation of bot behavior:

```

mp_chattime 2; 
cl_neo_grenade_show_path 1; sv_neo_grenade_show_path 1;

sv_neo_bot_grenade_polite_smoke 0;
sv_neo_bot_grenade_polite_frag 0;

nb_debug path;
```

src/game/server/NextBot/NextBotVisionInterface.cpp
```diff
@@ -794,6 +796,33 @@ bool IVision::IsLineOfSightClearToEntity( const CBaseEntity *subject, Vector *vi
 	UTIL_TraceLine( GetBot()->GetBodyInterface()->GetEyePosition(), firstPosition, MASK_BLOCKLOS_AND_NPCS|CONTENTS_IGNORE_NODRAW_OPAQUE, &filter, &result );
 	if ( result.DidHit() )
 	{
+		// Check if first center traceline hit any blocking smoke
+		if (neo_bot_path_reservation_enable.GetBool()
+			&& result.m_pEnt
+			&& FStrEq(result.m_pEnt->GetClassname(), SMOKELINEOFSIGHTBLOCKER_ENTITYNAME))
```
AdamTadeusz, line 802: I'm wondering what the reasoning behind the existence of FStrEq is, is it better to inline the pointer comparison and save yourself a function call if the two pointers are the same, and also more readable than V_stricmp == 0? In the case where the two pointers aren't the same you're comparing the two pointers twice.

It seems to me like it would be better to remove the pointer comparison from FStrEq and replace all calls to V_stricmp with FStrEq where we don't care which string has a higher numerical value.

Anyway, not suggesting we change that here, just something to ponder.

src/game/server/NextBot/NextBotVisionInterface.cpp
```diff
@@ -794,6 +796,33 @@ bool IVision::IsLineOfSightClearToEntity( const CBaseEntity *subject, Vector *vi
...
+		{
+			auto pSmoke = static_cast<CNEOSmokeLineOfSightBlocker*>(result.m_pEnt);
```
AdamTadeusz, line 804: We're assuming here that, if GetClassname returns exactly SMOKELINEOFSIGHTBLOCKER_ENTITYNAME, this static_cast is guaranteed to return a valid pointer to a CNEOSmokeLineOfSightBlocker. I think that's a safe assumption to make, Looking at IClassMap it seems to me that it's not possible for two entities to have the same class name (not sure about the case where NO_ENTITY_PREDICTION is defined), I'm just wondering how expensive a dynamic_cast is compared to a string comparison, there's 28 characters in SMOKELINEOFSIGHTBLOCKER_ENTITYNAME after all.

Again I have no idea which is more performant and I'm happy with this solution, another thing to ponder I guess, maybe someone with more knowledge can chime in.

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
```diff
@@ -287,6 +290,21 @@ ActionResult< CNEOBot >	CNEOBotTacticalMonitor::Update( CNEOBot *me, float inter
 		return SuspendFor( new CNEOBotPathClearBreakable( breakable ), "Clearing breakable in path" );
 	}
 
+	// Don't want to interfere with human squad leader's control just to ignore hazards
+	// Might be annoying if bots ignored following or waypoints (e.g. smoke sightlines avoidance)
+	if ( !me->m_hCommandingPlayer.Get() && m_hazardCheckTimer.IsElapsed() )
+	{
+		m_hazardCheckTimer.Start( 0.5f );
+		CNavArea *myArea = me->GetLastKnownArea();
+		if ( myArea )
```
AdamTadeusz, line 364: I like the coding template, or whatever you call it I can't remember, where you put the above two lines into one line, but I specifically forbid you from doing that in this pr now

AdamTadeusz (dismissed):

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
sunmachine: Why? Intentionally naïve here to make sure I understand.

AdamTadeusz: I was joking, but from what I've seen FCC likes to add things people suggest even if its more subjective rather than the right thing to do I guess? I just wanted to stress that this suggestion of mine shouldn't be taken as law, I'm happy with how it's written right now.

sunzenshen:
Needed to rebase a merge conflict from https://github.com/NeotokyoRebuild/neo/pull/2065

In reviewing the conflicts, I realized that I needed to clean up the repath timer still in the new hazard avoidance class to match the #2065 convention, and also fixed a potential null dereference with a bot's chosen safe area  (we hadn't seen it because for the most part bots have been able to find safe areas easily in the maps we tested).

AdamTadeusz (dismissed):

sunzenshen:
Needed to rebase

AdamTadeusz (approved):

src/game/server/neo/bot/neo_bot_path_reservation.cpp
```diff
@@ -342,3 +352,192 @@ float CNEOBotPathReservationSystem::GetAreaAvoidPenalty(unsigned int navAreaID)
...
+		blank.hazardExpireTime = 0.0f;
+		blank.smokeExpireTime = expireTime;
+		index = m_HazardAreas[teamID].Insert(navAreaID, blank);
+	}
+    else
+    {
+        HazardInfo &existing = m_HazardAreas[teamID][index];
+        // Optimizing simplification: assume new time is later to skip comparisons
+        // May also work for resetting an area to be non-hazardous early
+        existing.smokeExpireTime = expireTime;
```
sunmachine, line 476: When the entry already carries this expiry, nothing downstream changes, but we still fall through to `ForAllCompletelyVisibleAreas` and re-mark every visible area. Early return here to skip the walk on repeat calls, while leaving the first registration and any later smoke with a different expiry untouched.

```suggestion
		if (existing.smokeExpireTime == expireTime)
		{
			// already registered with this expiry, skip the visibility walk
			return; 
		}
		existing.smokeExpireTime = expireTime;
```

One thing worth knowing before you just take the suggestion, though. 

The early return also skips the walk when the propagating call arrives after a non-propagating one has already stored the same expiry for that tile. In your solution, that ordering happens only when the bot's own tile is also the smoke tile (which is harmless), but if a future caller relies on propagation after a plain registration, it would need to be reworked.

src/game/server/NextBot/NextBotVisionInterface.cpp
```diff
@@ -794,6 +796,33 @@ bool IVision::IsLineOfSightClearToEntity( const CBaseEntity *subject, Vector *vi
...
+			// Assume that threat could be hiding inside navarea where smoke blocked sight
+			if (neoBot)
+			{
+				int team = neoBot->GetTeamNumber();
+
+				CNavArea *smokeArea = TheNavMesh->GetNearestNavArea(result.endpos);
+				if (smokeArea)
+				{
+					// Register smoke sightlines for team to avoid as hazardous
+					CNEOBotPathReservations()->AddSmokeHazard(smokeArea->GetID(), pSmoke->GetNextThink(), team);
```
sunmachine, line 814: See the suggestion I provided in this review. This runs inside the per-enemy vision query, so every bot re-registers the same smoke several times a second for as long as the cloud is up, and each call re-walks the smoke area's visible set. The trace that gets us here is the expensive part of the query today. This walk can end up costing more than the trace once a few bots share a sightline. Again, see the suggestion I tossed in `AddSmokeHazard` for a cheap early-out. Not sure if it's the best solve for it but it's definitely better.

sunmachine (changes requested):

src/game/server/neo/bot/neo_bot_path_cost.cpp
```diff
@@ -55,158 +55,167 @@ float CNEOBotPathCost::operator()(CNavArea* baseArea, CNavArea* fromArea, const
...
+
+	if (!m_me->GetLocomotionInterface()->IsAreaTraversable(area))
+	{
+		return -1.0f;
+	}
+
+	if ( CNEOBotPathReservations()->IsAreaHazardous(area->GetID(), m_me) )
 	{
-		if (!m_me->GetLocomotionInterface()->IsAreaTraversable(area))
+		if ( m_routeType == DEFAULT_ROUTE )
```
sunmachine, line 66: This wall applies to commanded bots too. `CNEOBotCommandFollow` paths with `DEFAULT_ROUTE` and requires `COMPLETE_PATH`, so a hazard area between a bot and its commander ends the follow behavior and clears the commander. Suggest exempting bots with `m_hCommandingPlayer` set here, to match the exemption in the tactical monitor.

src/game/server/neo/bot/behavior/neo_bot_tactical_monitor.cpp
```diff
@@ -352,6 +355,21 @@ ActionResult< CNEOBot >	CNEOBotTacticalMonitor::Update( CNEOBot *me, float inter
 		return SuspendFor( new CNEOBotPathClearBreakable( breakable ), "Clearing breakable in path" );
 	}
 
+	// Don't want to interfere with human squad leader's control just to ignore hazards
+	// Might be annoying if bots ignored following or waypoints (e.g. smoke sightlines avoidance)
+	if ( !me->m_hCommandingPlayer.Get() && m_hazardCheckTimer.IsElapsed() )
```
sunmachine, line 360: This is half of the exemption. The other half is in `CNEOBotPathCost`, which still returns -1 for hazard areas regardless of commander, so commanded bots don't cower but they also can't be led through smoke. See the comment on `neo_bot_path_cost.cpp:66`.

src/game/server/neo/bot/neo_bot_path_reservation.cpp
sunzenshen: My understanding as that as you pointed out, trying to optimize here needs to solve a propagation issue.  The one scenario I had in mind was that teammates during the same tick might traceline a threat around the same area, and because it's the same tick, we can't exactly infer that we are originating from the same origin point as a previous friendly marked.  These potentially visible sets are relatively short lists and I'm not engaging in recursion either.

src/game/server/NextBot/NextBotVisionInterface.cpp
sunzenshen: My understanding is that this is actually not always the same area.  While there are some redundant hits,  a smoke cloud can cover multiple navareas, and we're just getting the nearest navarea at the traceline end point.  There's probably an optimization opportunity here but this allows different areas of a smoke cloud to be processed when relevant based on the bot's relative position.

src/game/server/neo/bot/neo_bot_path_cost.cpp
sunzenshen: This is a good finding.  Because commanding bots is a relatively less used feature, I feel like we can address this in a different PR.

sunmachine (approved):

src/game/server/NextBot/NextBotVisionInterface.cpp
sunmachine: Agreed, and I don't think the early-out touches that. Different smoke tiles are different keys, so they each still get walked. Let's continue on the `AddSmokeHazard` thread since both comments end up at the same question. Writing that up there...

src/game/server/neo/bot/neo_bot_path_reservation.cpp
sunmachine: Oh I almost forgot to post this:

I think we're closer than it looks. A few questions to check that.

Different parts of the cloud: if two traces end on different smoke tiles, those are different keys in `m_HazardAreas`. Does the early-out ever fire in that case? As far as I can tell it can't, so the multi-tile behaviour you're describing is preserved.

The redundant hits you mentioned: what does the walk take as input? Just (team, smoke tile, expiry), not the bot's origin. So when a second trace ends on the same tile with the same expiry, what does re-walking it add that the first walk didn't? Origin only feeds the separate non-propagating call on the bot's own tile, which the early-out doesn't block.

Now the part I got wrong earlier, and the one that I think actually threatens your multi-tile goal: what does the walk write for the tiles it visits? A plain entry with the same expiry. A blocker is 280 units across on a 20-unit nav step, so a cloud sits on several tiles, and Y's walk plain-marks its neighbour X. When a later trace lands on X and asks to propagate, what does X's entry look like? Under the early-out as I wrote it, X is skipped, and the tiles visible from X but not from Y never get marked. That's the opposite of "different areas processed when relevant," and it isn't limited to a bot standing in the smoke tile as I first said.

So: does keying the early-out on "was this tile propagated with this expiry" rather than "was this tile registered with this expiry" fix that? One extra float in `HazardInfo`, set only in the `propagatePVS` branch. Same-tile repeats still dedupe exactly, different tiles still walk, and a plain mark never suppresses a later propagating call.

sunzenshen: Yeah, let me think about this, but I hear you on ideating on this.  Just decided it's probably not needed to get this in tonight.  I think I might try to work something out in a local commit on top of this branch, see if I can address the feedback and decide if I want to merge and then immediately cherry pick the changes, or just go for another round of reviews.

sunzenshen:
Addressed @sunmachine 's code review feedback in https://github.com/NeotokyoRebuild/neo/pull/2128

PR #2021, opened 2026-07-16
Bots that cannot capture ghost will patrol around area

sunzenshen:
## Description
Bots that cannot capture the ghost will transition into the area patrolling behavior.
This is intended to address situations when the ghost is moved into an awkward area where the bots cannot pick up the ghost.
This prevents the bots from entering a cycle where they fail to pick up the ghost and then immediately retry their attempt.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_ctg_seek.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ctg_seek.cpp
@@ -89,7 +89,7 @@ ActionResult< CNEOBot > CNEOBotCtgSeek::Update( CNEOBot *me, float interval )
 				}
 				else
 				{
-					return Done("Capture target was not a ghost");
+					return SuspendFor(new CNEOBotCtgLoneWolf, "Capture target is blocked by some other entity, searching around the nearest areas");
 				}
 			}
 		}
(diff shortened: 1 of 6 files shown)


sunzenshen:
The easiest way to reproduce this kind of capture failure scenario is to noclip to pick up the ghost quickly to throw it into an awkward pickup area.  One such pickup area is in ntre_oilstain_ctg where the alleyway bike in the corner can block bots if the ghost somehow ends up behind it in the corner:

Screenshot 2026-07-15 224647

Screenshot 2026-07-15 224659

Before, bots could enter a cycle state where they drop their weapon and try to pick up the ghost, fail, and then retry their attempt.

With this PR, when bots fail to capture the ghost, they enter the area patrolling ctgLoneWolfSeek behavior:

Screenshot 2026-07-15 223757

When the ghost is picked up again by either a lucky bot or a more sophisticated human player, the behavior stack is reset in such a way that bots can reevaluate whether than can try capturing the ghost again:
(such as noclipping behind the impassable barrier and dropping the ghost back into a playable area)

Screenshot 2026-07-15 223822

Use `noclip` to pass through walls, and `nb_debug behavior` to display the current behavior state of the bots.

src/game/server/neo/bot/behavior/neo_bot_ctg_seek.cpp
```diff
@@ -89,7 +89,7 @@ ActionResult< CNEOBot > CNEOBotCtgSeek::Update( CNEOBot *me, float interval )
 				}
 				else
 				{
-					return Done("Capture target was not a ghost");
+					return SuspendFor(new CNEOBotCtgLoneWolf, "Capture target is blocked by some other entity, searching around the nearest areas");
```
sunzenshen, line 92: This case can be reached if the ghost is behind a bike, as the line of sight check hits the bike before the ghost.

nullsystem (approved):

AdamTadeusz (approved):

PR #2026, opened 2026-07-18
Add NEO_INSTALL_ENGINE_BUNDLE, support for Linux and Windows

sunmachine:
## Description

Following the `NEO_INSTALL_LIBRARIES`/`NEO_INSTALL_RESOURCES` patterns.
Stages everything to run without the SDK installed as a dependency, created from the 2013 Source MP SDK.

- `bin/`, `hl2/`, `platform/`: engine and base content, exact filename casing preserved
- Valve's stock launch chain, renamed to `ntre.*`
- a `gameinfo.txt` with plain relative search paths instead of `|appid_243750|` tokens

For local staging and testing. The payload comes entirely from the SDK install, so nothing needs to be compiled first. This PR should update the CI steps as well before leaving draft.

```
cmake --preset linux-release -DNEO_INSTALL_ENGINE_BUNDLE=ON
cmake --install build/linux-release --prefix <dir>
```

Overlay the output onto the existing game content. That's the depot.

**Depot notes:** Launch options must pass `-game neo`. `-steam only` won't work and that's intentional at this point. The upload pipeline must preserve filename casing.

### Cause

The Linux launch is broken by two case-sensitivity bugs: the engine lowercases the SDK folder name when resolving `|appid_243750|` mounts (so base content silently fails to mount), and the depot shipped lowercased binary names like `gameui.so` that the engine dlopens by exact name. Relative paths skip the buggy (or exactly as designed? we'll never know) lookup entirely. Same packaging model as the original NEOTOKYO.

## Toolchain

- Linux GCC 10 Sniper 3.0

## TODO

**Windows is completely untested at time of writing. Halp!**

- [x] Linux verification: launches and plays from the Play button on a real install, including with SDK Base renamed away. In-game `path` table resolves every mount inside the install dir.
- [ ] Windows verification: confirm the staged bootstraps (`hl2.exe` to `ntre.exe`, `hl2_win64.exe` to `ntre64.exe`) match what the depot ships, then a Play-button run.
- [ ] Update CI steps.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -47,6 +47,12 @@ else ()
 endif ()
 option(NEO_INSTALL_LIBRARIES "Install game libraries" OFF)
 option(NEO_INSTALL_RESOURCES "Install game resources" OFF)
+option(NEO_INSTALL_ENGINE_BUNDLE "Install the engine, base content, and launch chain for a self-contained depot" OFF)
+if(OS_WINDOWS)
+    set(NEO_SDK_BASE_DIR "C:/Program Files (x86)/Steam/steamapps/common/Source SDK Base 2013 Multiplayer" CACHE PATH "Source SDK Base 2013 Multiplayer install directory")
+else()
+    set(NEO_SDK_BASE_DIR "$ENV{HOME}/.local/share/Steam/steamapps/common/Source SDK Base 2013 Multiplayer" CACHE PATH "Source SDK Base 2013 Multiplayer install directory")
+endif()
 option(NEO_USE_SEPARATE_BUILD_INFO "Use separate build info on Linux" OFF)
 option(NEO_ENABLE_CPACK "Enable CPack" OFF)
 option(NEO_USE_MEM_DEBUG "Enable USE_MEM_DEBUG defines" OFF)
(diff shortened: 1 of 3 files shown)


sunmachine:
@DESTROYGIRL Give the windows side a spin!

src/CMakeLists.txt
```diff
@@ -47,6 +47,12 @@ else ()
 endif ()
 option(NEO_INSTALL_LIBRARIES "Install game libraries" OFF)
 option(NEO_INSTALL_RESOURCES "Install game resources" OFF)
+option(NEO_INSTALL_ENGINE_BUNDLE "Install the engine, base content, and launch chain for a self-contained depot" OFF)
+if(OS_WINDOWS)
+    set(NEO_SDK_BASE_DIR "C:/Program Files (x86)/Steam/steamapps/common/Source SDK Base 2013 Multiplayer" CACHE PATH "Source SDK Base 2013 Multiplayer install directory")
+else()
```
sunmachine, line 53: There's no switch statement in CMAKE right? Google-fu failed.

sunmachine:
LOL NOPE NOT THIS! Look at #2019 instead!

PR #2028, opened 2026-07-23
Damage reporting fixup: Remove m_rfAttacker arrays from the client, just rely on UserMessage

### LINKED ISSUE #1874: The damage report lies

nullsystem:
## Description
The 1st commit in this branch has demoed when the server sends over the damage hit values via UserMessage, this is the reliable and actual number the server have. So if we're using UserMessage, might as well have the scoreboard take it from that instead (which just transmits once on death like the killer infos) and save having to transmit those arrays through the network tables constantly.

Also enabled damage report for deathmatch/respawnables, properly clearing index other held on spawning. `CNEORules::PlayerKilled` also changed to only deal with the scores server side: there wasn't a point in modifying it client side when those XP scores are enforced and transmitted by the server anyway and also `FetchAssists` server side only now. Renamed `c_neo_killer_infos.h/cpp` to `c_neo_killer_damage_infos.h/cpp` and move message fetch and printout to there.

Minor refactor on scoreboard info update handling, zero the scoreboard info array before updating.

Removed unused code paths.

## Toolchain

## Linked Issues

- fixes #1874


src/game/client/neo/c_neo_killer_damage_infos.cpp
```diff
@@ -114,38 +134,35 @@ static void __MsgFunc_KillerDamageInfo(bf_read &msg)
...
+		if (pDmgReport->iDealtDmgs > 0 && pDmgReport->iDealtHits > 0)
+		{
+			V_sprintf_safe(infoLine, "Damage dealt to %s [%s]: %d in %d hits\n",
+					   dmgerName, dmgerClass,
+					   pDmgReport->iDealtDmgs, pDmgReport->iDealtHits);
 			ConMsg("%s", infoLine);
-
-			totals += attackerInfo;
+			totals.iTakenDmgs += pDmgReport->iDealtDmgs;
+			totals.iTakenHits += pDmgReport->iDealtHits;
```
sunmachine: At a glance this looks like a mistaken accumulation but I'm wondering if my glancing read is too fast. 😅 

```suggestion
			totals.iDealtDmgs += pDmgReport->iDealtDmgs;
			totals.iDealtHits += pDmgReport->iDealtHits;
```

src/game/client/neo/c_neo_killer_damage_infos.cpp
```diff
@@ -114,38 +134,35 @@ static void __MsgFunc_KillerDamageInfo(bf_read &msg)
...
 		}
+		if (pDmgReport->iTakenDmgs > 0 && pDmgReport->iTakenHits > 0)
+		{
+			V_sprintf_safe(infoLine, "Damage taken from %s [%s]: %d in %d hits\n",
+					   dmgerName, dmgerClass,
+					   pDmgReport->iTakenDmgs, pDmgReport->iTakenHits);
+			ConMsg("%s", infoLine);
+			totals.iDealtDmgs += pDmgReport->iTakenDmgs;
+			totals.iDealtHits += pDmgReport->iTakenHits;
```
sunmachine: Same issue here, right (assuming my understanding is correct)?

```suggestion
			totals.iTakenDmgs += pDmgReport->iTakenDmgs;
			totals.iTakenHits += pDmgReport->iTakenHits;
```

sunmachine (changes requested):
Quick read-through before dinner with family. Will review some more later this evening.

src/game/client/neo/c_neo_killer_damage_infos.cpp
nullsystem: Yeah that's a mistake, fixed in commit: 27fa03a6978271b61587234798e88b3186ff7a73

src/game/client/neo/c_neo_killer_damage_infos.cpp
```diff
@@ -94,15 +107,22 @@ static void __MsgFunc_KillerDamageInfo(bf_read &msg)
...
-		{
-			continue;
-		}
 
-		auto* neoAttacker = assert_cast<C_NEO_Player*>(UTIL_PlayerByIndex(pIdx));
+	NeoDamageReportClear();
+	AttackersTotals totals = {};
+
+	// Read (server side) per-player damage stats
+	const int iAtkSize = Min(msg.ReadShort(), MAX_PLAYERS);
```
sunmachine: Is anything bounding the message itself?

With the send props change/removal, each record written here is 12 bytes against `MAX_USER_MSG_DATA`'s `255`, which runs out somewhere around twenty or so attackers.

Which to me probably doesn't matter, right? You'd have to have a crazy situation and even then an overflow isn't going to crash, you're just gonna' drop reports. 

EDIT: Yea I can't even imagine a scenario where the whole enemy team in a 16 versus 16 game, plus your teammates' friendly fire, makes this something concerning at all.

sunmachine (approved):

src/game/client/neo/c_neo_killer_damage_infos.cpp
nullsystem: It would be very much an edge case, but nevertheless can be improved to handle it also so updated it with this commit: 9ae5aa5853142b365f4d1680f838812a67799d29

This have both compacting from shorts to unsigned bytes if max dmgs/hits <= 255 which will be the case for CTG but can go over for JGR/respawning gamemodes and a backup just incase secondary UserMessage to handle the remaining players when needed.

At least with this, it now took 30 players damage report in CTG with a killer player using a weapon with long enough name (EX: supa7) to cause the secondary UserMessage to be used for the 30th player.

nullsystem:
Also fixed it up for bot takeover: 5e93ecceb59b0afd4d64adc8676db326a460a525

src/game/client/neo/c_neo_player.cpp
```diff
@@ -1996,6 +1996,7 @@ void __MsgFunc_CSpectatorTakeoverPlayer(bf_read &msg)
 		// Save for later in C_NEO_Player::OnDataChanged
 		pSpectatorTakingOver->m_hSpectatorTakeoverPlayerTarget = pPlayerTakeoverTarget;
 		pSpectatorTakingOver->m_bCopyOverTakeoverPlayerDetails = true;
+		NeoAllKDReportsClear();
```
sunmachine: Only if they're a local player, right?

```suggestion
		if (pSpectatorTakingOver->IsLocalPlayer())
		{
			NeoAllKDReportsClear();
		}
```

sunmachine (changes requested):

src/game/client/neo/c_neo_player.cpp
nullsystem: I somewhat thought it was sent to a single player but seems not.

Fixed here: 1feb702ad2c549942f86d4e0c8a62b22f74d4355

src/game/server/neo/neo_player.cpp
```diff
@@ -2390,18 +2392,70 @@ void CNEO_Player::StartShowDmgStats(const CTakeDamageInfo *info)
...
+
 		for (int i = 0; i < iAtkSize; ++i)
 		{
+			if (i >= iAtkFirstSize)
+			{
+				MessageEnd();
+				UserMessageBegin(filter, "KillerDamageInfoExtra");
+				WRITE_BYTE(static_cast<char>(iAtkSize - iAtkFirstSize));
+				WRITE_BYTE(flags);
+			}
```
sunmachine: The two-message split looks right! I walked the byte accounting above and it lines up.

One thing I want to check. Line 2432 writes `iAtkSize - iAtkFirstSize`, the count of all remaining records. That's only an accurate header if this branch runs exactly once, at the boundary. What holds it to one entry?

`i >= iAtkFirstSize` is still true on the next iteration. Tracing `iAtkSize` at 25, and `iAtkFirstSize` at 20, I get a new `UserMessageBegin` at `i` at 20, 21, 22, 23, 24. Five continuation messages, each claiming 5 records and carrying 1.

Edit: Oh, I think `==` gets you what you want.

sunmachine (changes requested):

src/game/server/neo/neo_player.cpp
nullsystem: Whoops, good catch. Fixed here: 9296589319c66be605dc1174d9d910e215508680

src/game/client/neo/c_neo_player.cpp
sunmachine: Yea it's a bit insideous

src/game/server/neo/neo_player.cpp
sunmachine: I get nailed by these level and edge problems all the time. Why is it easier when reviewing someone elses code but not my own 🙃

sunmachine (approved):

DESTROYGIRL (approved):

sunzenshen (approved):

PR #2032, opened 2026-07-25
Bots won't ping and only see own pings in non-teamplay modes

### LINKED ISSUE #2031: Bots place pings in non-teamplay gamemodes
It's not needed. The ping sound is also audible for players on the same "team", which it shouldnt be

DESTROYGIRL:
## Description
In DM, change behaviours so that pings are blocked entirely, not just prevented from drawing. Brings them back for the local player. Stops bots from pinging.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #2031

### DIFF
--- a/src/game/client/neo/ui/neo_hud_player_ping.cpp
+++ b/src/game/client/neo/ui/neo_hud_player_ping.cpp
@@ -117,16 +117,16 @@ void CNEOHud_PlayerPing::FireGameEvent(IGameEvent* event)
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


sunzenshen (approved):
Seems to work.  Tested with:
```
sv_neo_gamemode_enforcement 3; sv_neo_gamemode_single 3; neo_vote_game_mode 3; mp_restartgame 1;
```

nullsystem (approved):

PR #2035, opened 2026-07-26
Add generic Linux dev setup & bootstrap

sunmachine:
## Description

Optional developer tooling for Linux + Clang. It does not change how NT;RE builds.

- `tools/ntre-dev-setup.sh` — sets up working C++ IntelliSense: a pinned `clangd`, a `compile_commands.json` symlinked at the repo root, and a `.clangd` carrying the flags the SteamRT toolchain needs.
- `.devcontainer/` — a dev container on the same sniper SDK image the CI runners use, which runs that script on first open.

Why the script is needed: a unity build's `compile_commands.json` lists only the generated `unity_*.cxx` blobs, so roughly 1300 translation units get no compile flags and indexing is dead across most of `src/game`. The script works around that with a separate, build free CMake directory configured with the unity options off, written to the git-ignored `src/CMakeUserPresets.json`. Your real build directory keeps whatever settings you gave it, and nothing is ever compiled from the IntelliSense one.

Editor config for VS Code and Zed is written on request. Everything generated is git-ignored, and the existing Qt Creator and CLI build workflows are untouched.

Also removes two Source SDK leftovers: the `buildallprojects` gitattributes entry and `src/.vscode/tasks.json`, which both refer to a build script this repo no longer has.

> [!NOTE]
> **Rebased on #2052**
>
> Rebased onto @nullsystem's 6d33c0ea03e804fb0034b9ecd0121bc301a18910, which adds the `NEO_UNITY_BUILD_CLIENT_SERVER` and `NEO_UNITY_BUILD_OTHERS` options. The IntelliSense preset sets both to `OFF` for itself alone. Following @Masterkatze's review, this branch no longer touches shared build defaults or injects CMake code into the project. That commit appears in the diff below until #2052 merges. This also gives me another day to hammer on Zed a bit.

### Trying it out

1. `git clone` this repo and check out this branch.
2. Open the folder in VS Code with the [Dev Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) extension installed.
3. Accept the *"Reopen in Container"* prompt, or run *Dev Containers: Reopen in Container* from the command palette.
4. Wait for the first build; it pulls the sniper SDK image and runs the setup script, so expect a few minutes. Later opens are seconds.
5. Open any `.cpp` file under `src/` and confirm go-to-definition and completion work.

Requires Docker or Podman on the host. The `README.md` covers running the script outside a container.

## Toolchain
- Linux GCC 10 Sniper 3.0

## Linked Issues
- fixes #
- related #

## Zed Example

https://github.com/user-attachments/assets/be93f365-8467-4a7f-b5e1-63a494d7b327

https://github.com/user-attachments/assets/4a241b90-c4de-4345-8a48-1023fe82a3d9


tools/ntre-dev-setup.sh
```diff
@@ -0,0 +1,424 @@
...
+
+# ------------------------------------------- 2. unity-build override for CMake
+# Injected via CMAKE_PROJECT_<name>_INCLUDE, so no tracked CMakeLists.txt changes
+# and upstream merges stay clean.
+if write_if_absent "${IDE_DIR}/cmake/ide-overrides.cmake" <<'EOF'
+# Loaded via CMAKE_PROJECT_neo_INCLUDE by the IntelliSense preset only; clears
+# UNITY_BUILD so the database gets one entry per translation unit (see the
+# header of tools/ntre-dev-setup.sh). The real build keeps its unity speedup.
+
+function(_neo_ide_disable_unity dir)
```
Masterkatze: This thing is kinda clever, but I think it would be better to create a CMake option `NEO_USE_UNITY_BUILD` instead and use its value to initialize the `UNITY_BUILD` property of all unity build targets. Set it to `ON` only for the CI build and `OFF` otherwise (which is what we need for `Linux Debug (IntelliSense index only)` preset and also the developers since the build is incremental). Apart from this - LGTM.

sunmachine: Happy to make the change, thank you for the review!

sunmachine:
Note to self: I still need to toy around with Zed more to make sure this is solid. 

Note to self 2: Get more details on @Agiel 's specific Zed workflows to make sure I covered his concerns adequately.

tools/ntre-dev-setup.sh
sunmachine: ~~Done in `9f2a52c1` and `9b173e9a`. One note on the "ON only for CI" part: rather than adding `-DNEO_USE_UNITY_BUILD=ON` to the workflow, the option defaults to `${NEO_CI_BUILD}` (same pattern as `NEO_GENERATE_GAMEDATA` defaulting to `NEO_DEDICATED`), which covers both the Linux and Windows CI jobs with no workflow change. The IntelliSense preset sets it to `OFF` explicitly, and the injected override is gone. Happy to switch to an explicit workflow flag instead if you prefer it discoverable there.~~

D'oh, almost. Gotta' do another pass.

sunmachine: Updated + rebased atop of @nullsystem's work.

sunmachine:
#2052 is merged in, ~~will rebase in the morning~~ rebase complete.

sunmachine:
Just verified Zed fully. It worked about as well as they described from their new (Jan 2026) devcontainer support. 

@Agiel Hopes this helps to stop from switching IDEs so much. 🤞 Check out the videos above.

sunmachine:
Working with @Agiel to address some updates and feedback.

PR #2038, opened 2026-07-28
Apparatus Soundscapes

DESTROYGIRL:
Soundscapes for https://github.com/NeotokyoRebuild/neoAssets/pull/126

### DIFF
--- a/game/neo/scripts/soundscapes_ntre_apparatus_dm.txt
+++ b/game/neo/scripts/soundscapes_ntre_apparatus_dm.txt
@@ -1,0 +1,68 @@
+"apparatus.outside"
+
+{
+	"dsp"	"1"
+	"playlooping"
+	{
+		"volume"	"0.8"
+		"pitch"		"100"
+		"wave"		"ocean_rain_lp.wav"
+	}
+}
+
+"apparatus.inside"
+
+{
+	"dsp"	"1"
+	"playlooping"
+	{
+		"volume"	"0.8"
+		"pitch"		"100"
+		"wave"		"ocean_rain_inside_lp.wav"
+	}
+
+	"playlooping"
+	{
+		"volume"	"0.2"
+		"pitch"		"100"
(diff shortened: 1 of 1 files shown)


game/neo/scripts/soundscapes_ntre_apparatus_dm.txt
```diff
@@ -0,0 +1,68 @@
...
+		"wave"		"ocean_rain_inside_lp.wav"
+	}
+
+	"playlooping"
+	{
+		"volume"	"0.6"
+		"pitch"		"100"
+		"wave"		"ambient/atmosphere/underground_hall_loop1.wav"
+	}
+}
```
Rainyan, line 68: Nit on sound file locations: should we maybe have a dedicated folder for the sounds? As they kinda stand out being in the root path:

sounds_location

Admittedly OGNT also does this for some sound assets, but maybe it'd be neater to organize for example by map name into a folder to keep things tidy

Also curious on sound origin - are these ambient rain sounds cleared in terms of copyright for being included?

Rainyan (dismissed):

game/neo/scripts/soundscapes_ntre_apparatus_dm.txt
DESTROYGIRL: I was considering a specific folder (having sounds in the root folder is bad practice imo but its too late to change that with all the og sounds...), I put them there to display that they aren't incredibly specific and can be used for other maps if desired. Unlike many of the sounds in `rogue` or `tutorial`. But, I guess it'd be better to set a good example for the future if it's the preference

The ocean rain sounds were built on a large mixed bowl of sounds I've recorded myself and ones I have from royalty free libraries which permit use in projects like these, though which clips and where specifically I couldn't tell you as I started playing with this map concept in 2024. The rain sfx clip itself I did not record.

Rainyan (approved):

AdamTadeusz (approved):

PR #2039, opened 2026-07-30
Jump 2

AdamTadeusz:
## Description

+jump2 command that does a normal jump without doing a superjump. Personally I bind mwheelup to +jump2 and mwheeldown to +jump

Decided to extend m_nButtons to 64 bits so the PR is a bit involved

## Toolchain

- Windows MSVC VS2022

## Linked Issues

- fixes #

### DIFF
--- a/src/game/server/episodic/vehicle_jeep_episodic.h
+++ b/src/game/server/episodic/vehicle_jeep_episodic.h
@@ -66,7 +66,7 @@ public:
 	void			AddPropToCargoHold( CPhysicsProp *pProp );
 
 	virtual CBaseEntity *OnFailedPhysGunPickup( Vector vPhysgunPos );
-	virtual void	DriveVehicle( float flFrameTime, CUserCmd *ucmd, int iButtonsDown, int iButtonsReleased );
+	virtual void	DriveVehicle( float flFrameTime, CUserCmd *ucmd, int iButtonsDown, int iButtonsReleased ); // NEO TODO (Adam) change to int64 when building this
 	virtual int DrawDebugTextOverlays( void );
 
 	DECLARE_DATADESC();
--- a/src/game/server/episodic/vehicle_jeep_episodic.h
+++ b/src/game/server/episodic/vehicle_jeep_episodic.h
@@ -66,7 +66,7 @@ public:
 	void			AddPropToCargoHold( CPhysicsProp *pProp );
 
 	virtual CBaseEntity *OnFailedPhysGunPickup( Vector vPhysgunPos );
-	virtual void	DriveVehicle( float flFrameTime, CUserCmd *ucmd, int iButtonsDown, int iButtonsReleased ); // NEO TODO (Adam) change to int64 when building this
+	virtual void	DriveVehicle( float flFrameTime, CUserCmd *ucmd, int iButtonsDown, int iButtonsReleased );
 	virtual int DrawDebugTextOverlays( void );
 
 	DECLARE_DATADESC();
--- a/src/game/server/hl2/vehicle_cannon.cpp
+++ b/src/game/server/hl2/vehicle_cannon.cpp
@@ -122,7 +122,7 @@ public:
 	virtual	bool CreateVPhysics( void );
 	virtual void UpdateOnRemove( void );
 
-	void	DriveCannon( int iDriverButtons, int iButtonsPressed ); // NEO TODO (Adam) change iDriverButtons and iButtonsPressed to int64 when building this
+	void	DriveCannon( int iDriverButtons, int iButtonsPressed );
 	void	ResetUseKey( CBasePlayer *pPlayer );
 	void	InitCannonSpeeds( void );
 	void	RunCraneMovement( float flTime );
@@ -380,6 +380,9 @@ void CPropCannon::SetupMove( CBasePlayer *player, CUserCmd *ucmd, IMoveHelper *p
 //-----------------------------------------------------------------------------
 void CPropCannon::DriveCannon( int iDriverButtons, int iButtonsPressed )
 {
+#ifdef NEO
+	Assert(false); // NEO TODO (Adam) change iDriverButtons and iButtonsPressed to int64 when building this
+#endif // NEO
 	bool bWasExtending = m_bExtending;
 
 	// Handle rotation of the crane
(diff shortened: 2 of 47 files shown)


AdamTadeusz:
~~There are a lot of places where IN_JUMP is used like CommentarySystem, fourwheelvehiclephysics etc. I'm not going to touch those that we don't use right now~~

(Edit) only if its not built?

src/game/server/episodic/vehicle_jeep_episodic.h
```diff
@@ -66,7 +66,7 @@ class CPropJeepEpisodic : public CPropJeep
 	void			AddPropToCargoHold( CPhysicsProp *pProp );
 
 	virtual CBaseEntity *OnFailedPhysGunPickup( Vector vPhysgunPos );
-	virtual void	DriveVehicle( float flFrameTime, CUserCmd *ucmd, int iButtonsDown, int iButtonsReleased );
+	virtual void	DriveVehicle( float flFrameTime, CUserCmd *ucmd, int iButtonsDown, int iButtonsReleased ); // NEO TODO (Adam) change to int64 when building this
```
Rainyan, line 69: Nit: maybe instead of commenting this, add a NEO-ifdef guarded `Assert(false)` in the impl, so whoever touches this in the future cannot miss the comment without hitting an assert?

src/game/server/hl2/vehicle_cannon.cpp
```diff
@@ -122,7 +122,7 @@ class CPropCannon : public CBaseProp, public IDrivableVehicle
 	virtual	bool CreateVPhysics( void );
 	virtual void UpdateOnRemove( void );
 
-	void	DriveCannon( int iDriverButtons, int iButtonsPressed );
+	void	DriveCannon( int iDriverButtons, int iButtonsPressed ); // NEO TODO (Adam) change iDriverButtons and iButtonsPressed to int64 when building this
```
Rainyan, line 125: Nit: would maybe prefer a hard `Assert(false)` in the implementation so we don't rely on the comment being seen but rather fail loudly

src/game/server/hl2/vehicle_jeep.cpp
```diff
@@ -1322,7 +1322,7 @@ void CPropJeep::SetupMove( CBasePlayer *player, CUserCmd *ucmd, IMoveHelper *pHe
 //-----------------------------------------------------------------------------
 void CPropJeep::DriveVehicle( float flFrameTime, CUserCmd *ucmd, int iButtonsDown, int iButtonsReleased )
 {
-	int iButtons = ucmd->buttons;
+	int iButtons = ucmd->buttons; // NEO TODO (Adam) change to int64 when building this
```
Rainyan, line 1325: Maybe ifdef NEO `Assert(false)` here so this cannot work without the coder noticing & fixing this in the future

src/game/server/hl2/weapon_physcannon.cpp
```diff
@@ -3308,7 +3308,7 @@ void CWeaponPhysCannon::ItemPostFrame()
 	}
 
 	// NOTE: Attack2 will be considered to be pressed until the first item is picked up.
-	int nAttack2Mask = pOwner->m_nButtons & (~m_nAttack2Debounce);
+	int nAttack2Mask = pOwner->m_nButtons & (~m_nAttack2Debounce); // NEO TODO (Adam) Change to int64 when building this?
```
Rainyan, line 3311: `Assert(false)` so whoever hits this has to deal with it?

src/game/server/sdk/sdk_vehicle_jeep.cpp
```diff
@@ -1324,7 +1324,7 @@ void CPropJeep::SetupMove( CBasePlayer *player, CUserCmd *ucmd, IMoveHelper *pHe
 //-----------------------------------------------------------------------------
 void CPropJeep::DriveVehicle( float flFrameTime, CUserCmd *ucmd, int iButtonsDown, int iButtonsReleased )
 {
-	int iButtons = ucmd->buttons;
+	int iButtons = ucmd->buttons; // NEO TODO (Adam) change to int64 when building this
```
Rainyan, line 1327: Ditto on maybe doing `Assert(false)` or similar here to force a fix for whoever tries to use it

src/tier1/bitbuf.cpp
```diff
@@ -1066,7 +1066,7 @@ int32 bf_read::ReadSignedVarInt32()
 
 int64 bf_read::ReadSignedVarInt64()
 {
-	uint32 value = ReadVarInt64();
+	uint32 value = ReadVarInt64(); // NEO TODO (Adam) https://github.com/ValveSoftware/source-sdk-2013/issues/322
```
Rainyan, line 1069: Would it make sense to fix this retval type 32->64 here as part of this PR, while we're at it modifying this stuff anyway?

Rainyan (dismissed):

src/tier1/bitbuf.cpp
AdamTadeusz: I did a little test, writing the button state twice, once using writelonglong and once using writesignedvarint64

```c++
	if ( to->buttons != from->buttons )
	{
		buf->WriteOneBit( 1 );
#ifdef NEO
	  	buf->WriteLongLong( to->buttons );
	  	buf->WriteSignedVarInt64( to->buttons );
#else
	  	buf->WriteUBitLong( to->buttons, 32 );
#endif // NEO
 	}
```

```c++
	// read buttons
	if ( buf->ReadOneBit() )
	{
#ifdef NEO
		move->buttons = buf->ReadLongLong();
		int64 buttons2 = buf->ReadSignedVarInt64();
		Assert(move->buttons == buttons2);
#else
		move->buttons = buf->ReadUBitLong( 32 );
#endif // NEO
	}
```

using jump2 triggers the Assert before the fix and doesn't trigger the Assert after the fix

AdamTadeusz: of course this isn't an exhaustive test, ReadSignedVarInt64 exists separate to writeLongLong for a reason, probably handles a sign bit, but i'm also confident the fix is fine

AdamTadeusz: I'll mention the issue in a github comment for posterity https://github.com/ValveSoftware/source-sdk-2013/issues/322

Rainyan (approved):

PR #2041, opened 2026-07-31
Fix Vision Prediction, add vision off sound and separate sound channel for vision sounds. Also fix ghost pickup sound not playing on its own special channel

### LINKED ISSUE #2042: toggling vision mode should be smooth at high ping

### LINKED ISSUE #2043: The Ghost pickup sound should play on its own channel so it doesn't get overriden by any other sound
CHAN_GHOST_PICKUP doesn't work the way its used now since EmitSound doesn't care about the channel value in the param passed to it. Only the channel in the soundscripts matters, and custom channel enums aren't recognized by the engine when parsing the sound scripts.

AdamTadeusz:
## Description

-----> EPILEPSY WARNING <------

Makes toggling the vision mode far smoother at high ping.

Recommend testing through the following:

bind mwheeldown +vision
test how quickly the vision mode can be triggered on a listen server and pay attention to the sound
then
net_fakelag 200
and do the same

also notice how individual inputs at high ping also don't cause the screen to flicker multiple times

## Toolchain

- Windows MSVC VS2022

## Linked Issues

- fixes #2042
- fixes #2043


AdamTadeusz:
I added the tournament priority label since this also fixed the ghost pickup sound not playing on its own channel. If need be I can move this fix to a separate PR

AdamTadeusz:
also this stops C_NEO_Player::PreDataUpdate returning early before calling the base class that does some important things so thats also important for the tourney

src/public/soundflags.h
```diff
@@ -30,7 +30,13 @@ enum
 	CHAN_VOICE_BASE	= 8,		// allocate channel for network voice data
 	CHAN_USER_BASE	= (CHAN_VOICE_BASE+128)		// Anything >= this number is allocated to game code.
 #ifdef NEO
-	,CHAN_GHOST_PICKUP	= CHAN_USER_BASE,
+	// NEO NOTE (Adam) These cannot be used inside of sound scripts, nor does using them in code do much (EmitSound ignores the channel set in the EmitSound_t param passed to it).
+	// We could add a flag like SND_CHANGE_PITCH and SND_CHANGE_VOL, but the reason SND_CHANGE_CHANNEL doesn't exist is because it cannot be changed for already playing sounds.
+	// Still might be useful for sounds that aren't playing yet if theres some disclaimer attached to the flag I guess. If you add something like that please update this comment.
+	// For now add any channels you may need here and in SoundParametersInternal.cpp, then use the integer equivalent in the sound scripts.
+	// If any of the enums here change, remember to update all the soundscripts.
```
sunmachine, line 37: Preface: no change required.

This is the first PR where I'll adventure gently into the topic of minimizing chances for future human error. Speaking generally on my personal approach, when I hit a problem like this where the sanest path is to just comment to future devs to update some specific non-static sites, is to also suggest a way to mark or wrangle this tech debt and DX friction. Compounding micro-disciplines.  

So on this: at worst for this case, I see this being a single and brief "god damnit" to the next dev working in here. I don't think we're gonna' iterate in this area too much. My opining aside, this is good given my read of it. Thank you for the clear detail on this.

src/game/shared/neo/neo_player_shared.cpp
```diff
@@ -728,4 +729,80 @@ CBaseEntity *CNEO_Player::FindUseEntity()
...
+	}
+
+	m_flVisionLastTime = gpGlobals->curtime;
+	m_bInVision = !m_bInVision;
+
+#ifdef CLIENT_DLL
+	if (prediction->IsFirstTimePredicted())
+	{
+		// NEO TODO (Adam) Is this still needed?
+		DevMsg("Playing sound at :%f\n", gpGlobals->curtime);
```
sunmachine: Nit-pick: No. Delete this. 💣 

Rationale: useful for active work, useless otherwise. If anything it's just more noise. If the hope is that this is meant to help passively catch stuff, that never works over time and there are better methods of providing that kind of observability.

game/neo/scripts/game_sounds_player.txt
```diff
@@ -25,13 +25,22 @@
...
+	"channel"	"137" // CHAN_VISION
 	"volume"     	"VOL_NORM"
 	"soundlevel"	"SNDLVL_75dB"
 
 	"wave"	"player/vision_on.wav"
 }
 
+"NeoPlayer.VisionOff"
+{
+	"channel"	"137" // CHAN_VISION
```
sunmachine, line 37: [`neo_gamerules.cpp:1555`](https://github.com/AdamTadeusz/neo/blob/370_fixVisionPrediction/src/game/shared/neo/neo_gamerules.cpp#L1555) may be dead code because of this.

game/neo/scripts/game_sounds_ui.txt
```diff
@@ -92,7 +92,7 @@
 
 "HUD.GhostPickUp"
 {
-	"channel"		"CHAN_ITEM"
+	"channel"		"136" // CHAN_GHOST_PICKUP
```
sunmachine, line 95: [`weapon_ghost.cpp:221`](https://github.com/AdamTadeusz/neo/blob/370_fixVisionPrediction/src/game/shared/neo/weapons/weapon_ghost.cpp#L221) may be dead code because of this.

src/public/SoundParametersInternal.cpp
```diff
@@ -41,6 +41,7 @@ static SoundChannels g_pChannelNames[] =
 	{ CHAN_VOICE_BASE, "CHAN_VOICE_BASE" },
 	{ CHAN_USER_BASE, "CHAN_USER_BASE" },
 	{ CHAN_GHOST_PICKUP, "CHAN_GHOST_PICKUP" },
+	{ CHAN_VISION, "CHAN_VISION" },
```
sunmachine, line 44: I really wish the SDK issues like this were easier to manage. 😭

No change required, only noting here because of the 👃 that's not your fault, and I do hope we can make this whole chain more sane somehow w/o requiring Valve.

sunmachine (commented):
Leaving these as comments for now. Reading more tonight. I feel I'm still really trying to get a handle on this ahead of asking y'all questions as a point of self-exercise. The static stuff was easy to figure out at least.

The other thing I'm trying to figure out, too, is whether this should be two PRs? I can see a world where we need to do a revert for one of the things changed in this, but we'd need to pull out perfectly fine refactoring (so far as I can tell 👀) as a result.

sunmachine (dismissed):
Solid PR. Good refactor. 

I have a broader question about supposedly dead imports my compiler is warning me about, but that's not only affecting this PR, and its for chat (e.g. how worried are we about transitive dependencies from unused imports)?

sunmachine (approved):

src/game/server/neo/neo_message.cpp
```diff
@@ -87,7 +87,7 @@ void CNEO_Message::InputShowMessage(inputdata_t& inputData)
 		filter.AddAllPlayers();
 
 		EmitSound_t ep;
-		ep.m_nChannel = CHAN_BODY;
+		ep.m_nChannel = CHAN_BODY; // NEO TODO (Adam) This doesn't change the channel this sound is played on, set correct channel in sound script
```
DESTROYGIRL, line 90: Comment is irrelevant here cuz the entity plays whatever sound its set to play, including raw sounds detached from a sound script. Not sure if there's any other way to handle it

DESTROYGIRL (approved):

PR #2049, opened 2026-08-04
Carry over values from the damage report to show in the next round during freeze and pause

AdamTadeusz:
## Description

Show damage report in freeze and pause

## Toolchain

- Windows MSVC VS2022

## Linked Issues

- fixes #

### DIFF
--- a/src/game/client/neo/c_neo_player.cpp
+++ b/src/game/client/neo/c_neo_player.cpp
@@ -472,8 +472,13 @@ C_NEO_Player::C_NEO_Player()
 	m_flTocFactor = 0.15f;
 
 	memset(m_szNeoNameWDupeIdx, 0, sizeof(m_szNeoNameWDupeIdx));
-	ClearLocalPlayerDmgReports();
 	m_szNameDupePos = 0;
+
+	if (IsLocalPlayer())
+	{
+		ListenForGameEvent( "game_freezetime_end" );
+		ClearLocalPlayerDmgReports();
+	}
 }
 
 C_NEO_Player::~C_NEO_Player()
@@ -1186,8 +1191,6 @@ void C_NEO_Player::PreThink( void )
 			// so it could arrive too late.
 			CLocalPlayerFilter filter;
 			enginesound->SetPlayerDSP(filter, 0, true);
-
-			ClearLocalPlayerDmgReports();
 
 			// Reset the cache of other players crosshair data on spawning in
 			if (CHudCrosshair *crosshair = GET_HUDELEMENT(CHudCrosshair))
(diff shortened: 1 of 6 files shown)


src/game/client/neo/c_neo_player.cpp
```diff
@@ -1187,8 +1192,6 @@ void C_NEO_Player::PreThink( void )
 			CLocalPlayerFilter filter;
 			enginesound->SetPlayerDSP(filter, 0, true);
 
-			ClearLocalPlayerDmgReports();
```
nullsystem, line 1190: Major regression: This broken the damage/kill report and now it kept values cross rounds.

nullsystem (changes requested):

nullsystem (changes requested):
No, this change also carries over to the next round without freeze time and pause.

AdamTadeusz:
I can't reproduce the issue. I don't see how its possible to have a new round without a pre round freeze, even with sv_neo_preround_freeze_time 0 StartNextRound calls SetRoundStatus(NeoRoundStatus::PreRoundFreeze);

nullsystem:
2026-08-05T181027_screen

1) I can reproduce the bug in your branch where the killer indicator haven't been cleared
2) Deathmatch/respawning KD report haven't been considered which cannot be relying on freeze/pause to clean up between spawning anyway

@AdamTadeusz

src/game/client/neo/c_neo_player.cpp
```diff
@@ -472,8 +472,13 @@ C_NEO_Player::C_NEO_Player()
...
 
 	memset(m_szNeoNameWDupeIdx, 0, sizeof(m_szNeoNameWDupeIdx));
-	ClearLocalPlayerDmgReports();
 	m_szNameDupePos = 0;
+
+	if (IsLocalPlayer())
+	{
+		ListenForGameEvent( "game_freezetime_end" );
+		ClearLocalPlayerDmgReports();
+	}
```
sunmachine, line 481: `IsLocalPlayer()` is always false here. 

The static local-player pointer it compares against is only assigned later in `C_BasePlayer::PostDataUpdate`. 

So, the listener never registers! `FireGameEvent` never runs, and with the per-respawn clear gone from `PreThink`, nothing clears the report between rounds. I think that's the cross-round regression @nullsystem is hitting.

The base class registers it's own event unconditionally in the constructor; doing the same here and checking for local player at event time works. See my other suggestion deeper in the file, too

```suggestion
      ListenForGameEvent( "game_freezetime_end" );
```

src/game/client/neo/c_neo_player.cpp
```diff
@@ -2148,3 +2156,12 @@ void C_NEO_Player::ClearLocalPlayerDmgReports()
...
 	}
 }
+
+void C_NEO_Player::FireGameEvent(IGameEvent* event)
+{
+	auto eventName = event->GetName();
+	if (!Q_stricmp(eventName, "game_freezetime_end"))
+	{
+		NeoAllKDReportsClear();
+	}
```
sunmachine, line 2166: Chain to the base handler (since it handles `base_player_teleported`), and do the local player check here where it's meaningful

```suggestion
	BaseClass::FireGameEvent(event);
	if (IsLocalPlayer() && !Q_stricmp(event->GetName(), "game_freezetime_end"))
	{
		NeoAllKDReportsClear();
	}
```

sunmachine (changes requested):

PR #2050, opened 2026-08-04
Try to fix MP3 exit again

nullsystem:
## Description
Add a Join call in deinit

## Toolchain

## Linked Issues

* fixes #1925


Rainyan (approved):
Can't trigger the bug reliably but this seems reasonable

AdamTadeusz (approved):

PR #2052, opened 2026-08-05
Add toggles for cmake unity build

nullsystem:
## Description
One for client/server and one for vgui2/tier1/mathlib as may want to turn it off only for working on just client/server library side. Also fixup compile/linker errors since the changes and add NEO_UNITY defines to fixup for those ifdef changes.

## Toolchain

- Windows MSVC VS2022

## Linked Issues

- fixes #
- related #

### DIFF
--- a/src/game/client/viewpostprocess.cpp
+++ b/src/game/client/viewpostprocess.cpp
@@ -87,7 +87,7 @@ ConVar mat_tonemap_algorithm( "mat_tonemap_algorithm", "1", FCVAR_CHEAT, "0 = Or
 ConVar mat_tonemap_percent_target( "mat_tonemap_percent_target", "60.0", FCVAR_CHEAT );
 ConVar mat_tonemap_percent_bright_pixels( "mat_tonemap_percent_bright_pixels", "2.0", FCVAR_CHEAT );
 ConVar mat_tonemap_min_avglum( "mat_tonemap_min_avglum", "3.0", FCVAR_CHEAT );
-#ifdef NEO // Unity build
+#ifdef NEO_UNITY // Unity build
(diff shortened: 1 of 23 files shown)


Rainyan (approved):
LGTM, tested on Win11 with MSVC 2022 and toggling the new options in CMakeLists.txt and then running "Project" -> "Delete Cache and Reconfigure" and "Build" -> "Rebuild All" to force the non-unity rebuild.

src/game/client/viewpostprocess.cpp
```diff
@@ -87,7 +87,7 @@ ConVar mat_tonemap_algorithm( "mat_tonemap_algorithm", "1", FCVAR_CHEAT, "0 = Or
 ConVar mat_tonemap_percent_target( "mat_tonemap_percent_target", "60.0", FCVAR_CHEAT );
 ConVar mat_tonemap_percent_bright_pixels( "mat_tonemap_percent_bright_pixels", "2.0", FCVAR_CHEAT );
 ConVar mat_tonemap_min_avglum( "mat_tonemap_min_avglum", "3.0", FCVAR_CHEAT );
-#ifdef NEO // Unity build
+#ifdef NEO_UNITY // Unity build
```
AdamTadeusz, line 90: Is this comment still needed?

AdamTadeusz (approved):

Masterkatze (approved):
LGTM

src/game/client/viewpostprocess.cpp
nullsystem: Don't see an issue leaving it there + easily grep for it

PR #2054, opened 2026-08-07
Toggles for cap zones

DESTROYGIRL:
## Description
This will be required for saitama in the upcoming tournament.
Also cleaned up a bunch of ancient code

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
@@ -429,7 +429,7 @@ Vector CNEOBotCtgCarrier::GetNearestCapPoint( const CNEOBot *me ) const
 		float flNearestCapDistSq = FLT_MAX;
 		for( int i=0; i<NEORules()->m_pGhostCaps.Count(); ++i )
 		{
-			CNEOGhostCapturePoint *pCapPoint = assert_cast<CNEOGhostCapturePoint*>( UTIL_EntityByIndex( NEORules()->m_pGhostCaps[i] ) );
+			CNEOGhostCapturePoint *pCapPoint = dynamic_cast<CNEOGhostCapturePoint*>( UTIL_EntityByIndex( NEORules()->m_pGhostCaps[i] ) );
 			if (!pCapPoint || !pCapPoint->GetActive())
 			{
 				continue;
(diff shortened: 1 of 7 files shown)


src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
```diff
@@ -429,8 +429,8 @@ Vector CNEOBotCtgCarrier::GetNearestCapPoint( const CNEOBot *me ) const
 		float flNearestCapDistSq = FLT_MAX;
 		for( int i=0; i<NEORules()->m_pGhostCaps.Count(); ++i )
 		{
-			CNEOGhostCapturePoint *pCapPoint = dynamic_cast<CNEOGhostCapturePoint*>( UTIL_EntityByIndex( NEORules()->m_pGhostCaps[i] ) );
-			if (!pCapPoint)
+			CNEOGhostCapturePoint *pCapPoint = assert_cast<CNEOGhostCapturePoint*>( UTIL_EntityByIndex( NEORules()->m_pGhostCaps[i] ) );
```
Rainyan: I don't have time for a full review right now so apologies for the random drive-by comment, but AFAIK, since we currently cache the entity indices as:

https://github.com/NeotokyoRebuild/neo/blob/de226650729d9514fb5ffb4384cbe623beb1deb2/src/game/shared/neo/neo_gamerules.cpp#L3233 

this means that if for whatever reason the cached `CNEOGhostCapturePoint` entity represented by that entindex number is killed (either by our native logic or by a SourceMod plugin etc) after it was pushed to `m_pGhostCaps`, there's a >0 chance of that freed entidex number being recycled to represent something else.

This could cause the `UTIL_EntityByIndex( NEORules()->m_pGhostCaps[i] ) );` to return a valid pointer to some completely unrelated entity which happened to get the recycled entindex, but very likely is not compatible with the conversion of `static_cast<CNEOGhostCapturePoint*>` (which is what `assert_cast<T*>` optimizes into in Release builds). This would trigger undefined behaviour and likely crash the server or do something unexpected. So sadly I don't think we can safely remove the `dynamic_cast` RTTI checks here nor in `neo_bot_ctg_escort.cpp` file without refactoring the `m_pGhostCaps` data type to something like CHandles. Or at least I think so(?)

DESTROYGIRL: I see, well it's unrelated to the functionality in this PR and I just saw it as a cleanup opportunity, so I'll revert. Should I also do it for the line in `neo_bot_ctg_lone_wolf.cpp` as well since it's the same setup?

Rainyan: Yeah I think it would be better to switch that one to `dynamic_cast` as well

Agiel (approved):

AdamTadeusz (approved):

PR #2055, opened 2026-08-08
Auto-fix lowercased engine binaries at launch

sunmachine:
## Description

> [!WARNING]
> 
> This just "fixes" poisoned installs. That's why it's hacky. The real fix will require verifying the depot files' names at deploytime fixing the filenames in the depot.

The depot ships 7 `bin/` binaries fully lowercased, but the engine `dlopen`s them by exact mixed-case name (`GameUI.so` etc.) and Steam never repairs casing on disk. Linux installs created since the casing change fail at startup with "Engine Error: Could not load: GameUI"; older installs keep working, and verify/reinstall can't fix it. Follow-up to the depot notes in #2019.

- Launcher renames the seven binaries to canonical casing at startup, before anything is `dlopen`ed. Newest file wins if both casings exist; failures warn, never block launch.
- Unit tests cover the four disk states. Linux-only; no CI changes.

Verified on Linux: `ctest` green in sniper, and a real install with both `GameUI.so` lowercased (error dialog on the stock launcher) reaches the menu with this launcher.

## Testing on Linux

From `src/` in a sniper environment:

```
cmake --preset linux-release
cmake --build build/linux-release --target launcher_main test_neo_case_heal
ctest --test-dir build/linux-release/tests/ -R test_neo_case_heal
```

From the install root:

```
mv bin/GameUI.so bin/gameui.so
mv bin/linux64/GameUI.so bin/linux64/gameui.so
```

Launch stock: "Engine Error: Could not load: GameUI". Then:

```
cp <repo>/src/build/linux-release/launcher_main/neo_linux64 .
```

Launch again: reaches the menu, only `GameUI.so` left on disk.

## Depot Changes

- Re-stage `bin/` with canonical filename casing (the seven files); ship together with this launcher so new installs stop being born broken.

## Toolchain
- Linux GCC 10 Sniper 3.0

### DIFF
--- a/src/launcher_main/CMakeLists.txt
+++ b/src/launcher_main/CMakeLists.txt
@@ -11,6 +11,10 @@ if(OS_WINDOWS)
     list(APPEND LAUNCHER_SOURCES launcher_main_mod_neo.rc)
 endif()
 
+if(OS_LINUX)
+    list(APPEND LAUNCHER_SOURCES neo_case_heal.cpp neo_case_heal.h)
+endif()
+
 add_executable(launcher_main
     ${LAUNCHER_SOURCES}
 )
--- a/src/launcher_main/main.cpp
+++ b/src/launcher_main/main.cpp
@@ -704,6 +704,10 @@ static void WaitForDebuggerConnect( int argc, char *argv[], int time )
 #endif // !LINUX
 
 #ifdef NEO
+#if defined( LINUX )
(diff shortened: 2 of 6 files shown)


DESTROYGIRL (approved):
Working, tested on a previously functional install by deleting `bin` and verifying files in steam, which downloads them with the incorrect cases

AdamTadeusz (approved):

PR #2061, opened 2026-08-10
Change cap zone distance to account for the cap radius

DESTROYGIRL:
## Description
The distance from the cap zone now shows the distance from the cap radius rather than the origin. Currently in gameplay, players are not given any information about how large the radius is, which can lead to frustration from missed or unexpected caps. I think this would be a minimal but effective change that reduces cases like this.

This is a diversion from parity, so feel free to debate this change

## Toolchain
- Windows MSVC VS2022


Agiel (approved):
Good idea.

AdamTadeusz (approved):

DESTROYGIRL:
I'm gonna let this sit for a day just in case

PR #2065, opened 2026-08-15
Avoid recomputing long/static paths on short timers

sunzenshen:
## Description
Only follow short timer recompute pattern with paths that are short and without additional path requirements.
Because there are now complex criteria in the path costs, triggering an early recompute was causing bots to jitter between different alternative paths, which caused indecisive repathing in competing directions at times.
By limiting the recompute on short timers in the case where paths are intended to be short or straightforward, we can avoid these path jitter events.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
@@ -356,7 +356,6 @@ ActionResult< CNEOBot >	CNEOBotCtgCarrier::OnStart( CNEOBot *me, Action< CNEOBot
 {
 	m_chasePath.Invalidate();
 	m_aloneTimer.Invalidate();
-	m_repathTimer.Invalidate();
 	m_path.SetMinLookAheadDistance( me->GetDesiredPathLookAheadRange() );
 
 	m_teammates.RemoveAll();
@@ -542,12 +541,11 @@ void CNEOBotCtgCarrier::UpdateFollowPath( CNEOBot *me, const CUtlVector<CNEO_Pla
 			{
 				m_chasePath.Invalidate();
 
-				if ( !m_path.IsValid() || !m_repathTimer.HasStarted() || m_repathTimer.IsElapsed() )
+				if ( !m_path.IsValid() )
 				{
 					CNEOBotPathCompute( me, m_path, vecGoalPos, FASTEST_ROUTE );
-					m_repathTimer.Start( RandomFloat( 0.3f, 0.5f ) );
 				}
-				
+                
 				m_path.Update( me );
 				return;
 			}
@@ -633,10 +631,9 @@ void CNEOBotCtgCarrier::UpdateFollowPath( CNEOBot *me, const CUtlVector<CNEO_Pla
 
 	if ( bFoundGoal )
 	{
-		if ( !m_path.IsValid() || !m_repathTimer.HasStarted() || m_repathTimer.IsElapsed() )
+		if ( !m_path.IsValid() )
 		{
 			CNEOBotPathCompute( me, m_path, vecGoalPos, SAFEST_ROUTE );
-			m_repathTimer.Start( RandomFloat( 0.5f, 1.0f ) );
 		}
 		m_path.Update( me );
 	}
@@ -656,7 +653,6 @@ ActionResult< CNEOBot > CNEOBotCtgCarrier::OnResume( CNEOBot *me, Action< CNEOBo
 	// Re-evaluate nearest cap point on resume (in case we moved significantly while interrupted)
 	m_closestCapturePoint = GetNearestCapPoint( me );
 
-	m_repathTimer.Invalidate();
 	UpdateFollowPath( me, m_teammates );
 	return Continue();
 }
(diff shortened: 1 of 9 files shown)


AdamTadeusz (approved):

src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
```diff
@@ -542,12 +541,11 @@ void CNEOBotCtgCarrier::UpdateFollowPath( CNEOBot *me, const CUtlVector<CNEO_Pla
...
 				m_chasePath.Invalidate();
 
-				if ( !m_path.IsValid() || !m_repathTimer.HasStarted() || m_repathTimer.IsElapsed() )
+				if ( !m_path.IsValid() )
 				{
 					CNEOBotPathCompute( me, m_path, vecGoalPos, FASTEST_ROUTE );
-					m_repathTimer.Start( RandomFloat( 0.3f, 0.5f ) );
 				}
-				
+
```
sunmachine, line 548: nit-pick, spaces over project's tab conventions. wondering if it's worth setting up full-project linting and formatting so this stuff is just automatic...

src/game/server/neo/bot/behavior/neo_bot_retreat_to_cover.cpp
```diff
@@ -309,10 +309,8 @@ ActionResult< CNEOBot >	CNEOBotRetreatToCover::Update( CNEOBot *me, float interv
 
 		m_waitInCoverTimer.Reset();
 
-		if ( m_repathTimer.IsElapsed() )
+		if ( !m_path.IsValid() )
 		{
-			m_repathTimer.Start( RandomFloat( 0.3f, 0.5f ) );
-
 			CNEOBotPathCompute( me, m_path, m_coverArea->GetCenter(), RETREAT_ROUTE );
 		}
```
sunmachine, line 315: Just to make sure I get this, we now re-consider an invalid path each tick until it finds something, so a series of concurrent failures could get nasty on A*, right? 

So to that end, I don't have a real measure of this being a problem right now. I honestly don't have a good view yet (see: need more review time to dig on that end) to validate how much that can happen; map-specific / implementation-specific scope.

src/game/server/neo/bot/behavior/neo_bot_retreat_from_grenade.cpp
sunmachine, line 224: When this call picks a different cover area, what makes the bot stop following the path to the old one?

src/game/server/neo/bot/behavior/neo_bot_retreat_to_cover.cpp
sunmachine, line 276: Same question here as with the retreat_from_grenade behavior, except I think this one is comparatively less obvious to me: when this picks a different cover area, what makes the bot stop following the path to the old one?

sunmachine (commented):
Hoping to get more reading in on this, but so far the approach seems solid. These behavior structures are pretty fun to read.

src/game/server/neo/bot/behavior/neo_bot_retreat_from_grenade.cpp
sunzenshen: Changing the old path doesn't happen at this point yet.  By this point we have just used a functor to iterate through all potential cover areas and gotten a candidate NavArea to use as a destination.  The path is finally computed below with CNEOBotPathCompute, using the center of the chosen NavArea as a destination.

src/game/server/neo/bot/behavior/neo_bot_retreat_to_cover.cpp
sunzenshen: Changing the old path doesn't happen at this point yet. By this point we have just used a functor to iterate through all potential cover areas and gotten a candidate NavArea to use as a destination. The path is finally computed below with CNEOBotPathCompute, using the center of the chosen NavArea as a destination.

src/game/server/neo/bot/behavior/neo_bot_retreat_to_cover.cpp
sunzenshen: From what I've seen, the path tends to be invalid only when:
- It has not been initiated so it has an invalidated setting at the start
- When we invalidate it manually
- When a path fails to be created

In my experience, it's actually really rare that CNEOBotPathCompute returns an invalid path because the default fallback is to create a partial path.  I made CNEOBotPathCompute a wrapper on the HL2 DM example code to have several fallbacks, with some optional settings if one really wanted to have a full path or nothing option.

sunzenshen (commented):
Appreciate the thoughtful questions!

src/game/server/neo/bot/behavior/neo_bot_retreat_to_cover.cpp
sunmachine: Resolving + pluggin' in fix suggestion

sunmachine (approved):

src/game/server/neo/bot/behavior/neo_bot_retreat_to_cover.cpp
sunzenshen: Sun talked out this scenario out with me offline, and it turns out that there is indeed a subtle bug where bots will not immediately move to the next cover area until the repath timer expires.  The funny thing is that hl2mp_bot_retreat_to_cover.cpp (which this implementation was initially based on) also has this same bug of not invalidating the timer to force the next path recomputation.

I think this bug is outside the scope of this PR but sun will be pushing out a fix for this timer repath bug.

src/game/server/neo/bot/behavior/neo_bot_retreat_from_grenade.cpp
sunzenshen: Also has a similar timer delay as with:

> Sun talked out this scenario out with me offline, and it turns out that there is indeed a subtle bug where bots will not immediately move to the next cover area until the repath timer expires.  The funny thing is that hl2mp_bot_retreat_to_cover.cpp (which this implementation was initially based on) also has this same bug of not invalidating the timer to force the next path recomputation.
> 
> I think this bug is outside the scope of this PR but sun will be pushing out a fix for this timer repath bug.

PR #2066, opened 2026-08-15
Differ killer/damage info console printout to single ConMsg

nullsystem:
## Description
Probably the workaround of ConMsg printing at random, instead of directly printing lines with ConMsg, the have a buffer string array to fill up then just print it all at once.

## Toolchain

### DIFF
--- a/src/game/client/neo/c_neo_killer_damage_infos.cpp
+++ b/src/game/client/neo/c_neo_killer_damage_infos.cpp
@@ -15,6 +15,8 @@
 
 static inline AttackersTotals gTotals;
 static const char *BORDER = "==========================\n";
+static char gPrintBuffer[4096];
+static char gConcatBuffer[512];
 
 void NeoUserIDsLocalKilledClear()
 {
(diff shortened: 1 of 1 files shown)


Agiel (approved):

AdamTadeusz:
image

Doesn't fix the issue of damage take and dealt lines being intermixxed

AdamTadeusz:
image

Looks like they are sorted by the recipient/dealer

nullsystem:
@AdamTadeusz ? Your screenshot doesn't prove anything. The format is currently similar to the OGNT plugin for the damage dealt to/taken from.

AdamTadeusz:
Ok could you explain what issue this pr is trying to fix then, because i thought you were fixing this
image

AdamTadeusz:
Also what kind of statement is "Your screenshot doesn't prove anything"

I'm trying to do my due diligence here and make sure good PRs get merged

Agiel:
There's a bug where the lines are printed in random order, which this is trying to fix. I think what he means is that the order in your screenshot isn't random.

edit: I do agree it would be better to separate dealt and taken, but that's probably out of scope?

nullsystem:
@AdamTadeusz 

outoforderprint

This issue, where you can see the Damage dealt is above the border when it should be inside.

AdamTadeusz (approved):

Issue #2068, opened 2026-08-15
Shooting shouldn't cancel out aim punch.


Agiel:
### Build Info

v32.1

### Description

This is technically parity, but it has a very detrimental effect and so should be changed. Especially since we disable viewkick in comp, which would otherwise mask it.

### To Reproduce

1. Set `sv_suppress_viewpunch 1` to make it more noticable, and `sv_neo_wep_dmg_scale 0.1` to give yourself more time.
2. Have an enemy shoot at you (if testing with bots enable fakelag).
3. Notice the aim punch.
4. Start firing yourself.

### Expected behavior

You should still see the aim punch.

### Actual behavior

No visible aim punch.

### Additional context and Screenshots

This happens because, in order to not have view kick be cumulative, we call ViewPunchReset before applying the new kick after a shot. This also cancels out any aim punch. If your ping is higher than the time from getting shot until your own shot, by the time your client knows you were hit the viewpunch has already been canceled out, so you never see it. Since you don't see it, you can't compensate for it, and your shots are likely to miss if you're aiming for heads (although they might look like hits on your end).

PR #2069, opened 2026-08-15
Don't reset aimpunch after a shot if viewkick is disabled.

Agiel:
## Description
This adds a check to avoid calling ViewPunchReset after a shot if viewkick is disabled. Ideally we'd separate viewkick and aimpunch so they can be controlled individually (and make viewkick visual only?). But this workaround will work for comp where viewkick usually is disabled.

## Toolchain
- Cachy OS - gcc version 16.1.1 20260728

## Linked Issues

- related #2068

### DIFF
--- a/src/game/shared/neo/weapons/weapon_aa13.cpp
+++ b/src/game/shared/neo/weapons/weapon_aa13.cpp
@@ -120,7 +120,6 @@ void CWeaponAA13::PrimaryAttack(void)
 		pPlayer->SetSuitUpdate("!HEV_AMO0", FALSE, 0);
 	}
 
-	pPlayer->ViewPunchReset();
 	AddViewKick();
 }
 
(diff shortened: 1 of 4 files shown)


bauxiteDYS:
Well, I guess we'll see how this looks but I think it will be absolutely horrific, and imo this has no place in any shooter, especially if it looks anything like the spaghetti code SDK implementation. I feel the game would be far better without it and that's what I would want to test first, if it seems the game actually needs this "aim punch" (camera movement when being shot) then it would be a good idea to implement an actual good version of it, but that isn't gonna happen because it's a terrible mechanic.

bauxiteDYS:
Ok, we tested it and it's not as terrible as I thought, i still hate it but I think this is better than normal.

AdamTadeusz:
a quick search for ViewPunchReset() reveals a few locations not handled by this PR
image

Agiel:
Not the first time I forgot about the shotguns...
I move the ViewPunchReset call into AddViewKick because that's really where it belongs.

AdamTadeusz (approved):

nullsystem (approved):

PR #2072, opened 2026-08-16
Bot visible when aiming on target

sunzenshen:
## Description
If the bot is aiming at a target, consider it visible.  Intended to address scenarios where two bots are clearly running straight towards each other and are not firing.

## Toolchain
- Windows MSVC VS2022


AdamTadeusz (approved):
This seems to me like a workaround for the hundred other bot detection convars not having the right value, but i'll allow it.

DESTROYGIRL (approved):

PR #2073, opened 2026-08-17
When there are no cap zones for their team, bots search for enemy

sunzenshen:
## Description
Some CTG maps have no cap zones for one team, so as a placeholder that team should just search the map for the enemy, and try to recover the ghost if the enemy collects the ghost.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
@@ -154,30 +154,11 @@ ActionResult< CNEOBot >	CNEOBotSeekAndDestroy::OnStart( CNEOBot *me, Action< CNE
 //---------------------------------------------------------------------------------------------
 ActionResult< CNEOBot >	CNEOBotSeekAndDestroy::Update( CNEOBot *me, float interval )
 {
-	ActionResult< CNEOBot > result = UpdateCommon( me, interval );
-	if ( result.IsRequestingChange() || result.IsDone() )
-		return result;
-
 	// Check for Game Type Specific behaviors and suspend for them
-	if (NEORules()->GetGameType() == NEO_GAME_TYPE_CTG)
+	if ( NEORules()->GetRemainingPreRoundFreezeTime( true ) > 0.0f )
 	{
-		// Check if enemy has the ghost
-		if (NEORules()->GhostExists())
+		if (NEORules()->GetGameType() == NEO_GAME_TYPE_CTG)
 		{
-			int iGhosterPlayer = NEORules()->GetGhosterPlayer();
-			if (iGhosterPlayer > 0 && iGhosterPlayer <= gpGlobals->maxClients)
-			{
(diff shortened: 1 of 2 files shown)


src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
```diff
@@ -160,7 +162,43 @@ ActionResult< CNEOBot >	CNEOBotSeekAndDestroy::Update( CNEOBot *me, float interv
...
+				if (pGhostCarrier && pGhostCarrier != me && pGhostCarrier->GetTeamNumber() != me->GetTeamNumber())
+				{
+					return SuspendFor(new CNEOBotCtgEnemy, "Stopping the ghost carrier!");
+				}
+			}
+		}
+
+		if (!m_ctgCheckTimer.HasStarted() || m_ctgCheckTimer.IsElapsed())
+		{
+			m_ctgCheckTimer.Start(mp_chattime.GetFloat());
```
AdamTadeusz: Is the idea to make sure this check runs at least once before freeze-time ends? perhaps the freeze time length convar would be a better choice here, mp_chattime only affects how long until the map is changed after the winner of the match is determined. Also we're on the server, could probably wrap this in a function and call from gamerules whenever the round status changes to live, again assuming you just care about this running once each round before freeze time ends.

AdamTadeusz: Then again DG added some functionality where caps can be disabled with map logic, but again can probably recompute again whenever that logic runs instead of on a timer

DESTROYGIRL: Bots in their current state seem to react to caps being toggled mid game fine on master so I think it'll be OK.
I agree a different variable to chattime would probably be more suitable here

sunzenshen: Decided that it wasn't that big of a performance drop if bots constantly check the cap zones only during freezetime, especially if we skip the seek and destroy logic during freezetime.  Rationale is that complicating the class with yet another timer/event is more mental overhead than benefit in performance.

Some example test criteria:
- `nb_debug behavior` to check current bot state machine
- checked that seek and destroy still works on a TDM map like bullet'
- checked that bots can still be collected as followers during freezetime as that behavior dispatches through CNEOBotScenarioMonitor
- checked that JGR mode behavior still activates
- checked that CTG mode behavior still activates 

Also checked that forcing TDM still works: 
`sv_neo_gamemode_enforcement 3; sv_neo_gamemode_single 0; neo_vote_game_mode 0; mp_restartgame 1;`

On isolation specifically:
- Checked that spawning in the house one can command a bot to be a follower to verify that CNEOBotScenarioMonitor is not blocked
- Checked that outside spawn bots do try to pick up the ghost if they somehow survive all the way up the hill
- Checked that house spawn bots use the basic seek and destroy behavior

AdamTadeusz (changes requested):
missing #include "neo_ghost_cap_point.h"

AdamTadeusz (approved):

DESTROYGIRL (approved):

PR #2077, opened 2026-08-17
Add new damage modifier for explosives

### LINKED ISSUE #2076: Explosives should be affected by the damage multiplier
### Build Info

v32.2

### Description

Turns out explosive damage isn't run through the damage modifier.

### To Reproduce

Load up rise and start throwing nades at a wall (refer to [spreadsheet](https://docs.google.com/spreadsheets/d/13gRkWqTxDGI5t-CGDOwqo2QWV3d5O2Ww-UylDrjMpD0/edit?usp=sharing))

### Expected behavior

Damage values in NT;RE should match the "Normalized" column.

### Actual behavior

The values are too low

The "Multiplied" column multiplies by the damage modifier 1.485 and gets us close to the OGNT values. Would need a much larger sample size to know for sure but I don't see any reason why OGNT would use a separate value for explosives.

Agiel:
## Description
Adds a new damage modifier for explosives `sv_neo_exp_dmg_modifier` with the same default value as the weapon modifier (1.485, which is parity).

I made it a separate modifier because I can see scenarios where we want to reduce bullet damage without affecting the grenades. The modifier is applied pretty high up the chain, because `CGameRules::RadiusDamage` takes the damage info as a const, and we want the correct damage in the `WeaponHit` event.

Note that explosives can still do less damage than they would in OGNT if there's a prop between the player and the explosion. This is intentional (by Valve) and handled in `CGameRules::RadiusDamage`. I deemed it outside the scope of this issue.

## Toolchain
- Cachy OS - gcc version 16.2.1 20260810

## Linked Issues
- fixes #2076

### DIFF
--- a/src/game/shared/neo/neo_gamerules.cpp
+++ b/src/game/shared/neo/neo_gamerules.cpp
@@ -48,7 +48,9 @@ ConVar sv_neo_warmup_round_time("sv_neo_warmup_round_time", "45", FCVAR_REPLICAT
 ConVar sv_neo_preround_freeze_time("sv_neo_preround_freeze_time", "15", FCVAR_REPLICATED, "The pre-round freeze time, in seconds.", true, 0.0, false, 0);
 ConVar sv_neo_latespawn_max_time("sv_neo_latespawn_max_time", "15", FCVAR_REPLICATED, "How many seconds late are players still allowed to spawn.", true, 0.0, false, 0);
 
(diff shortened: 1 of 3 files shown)


AdamTadeusz (approved):
on devbox tested with a recon, throwing a grenade near straight down

Throw Position
bind k "setpos 0 0 180; setang 90 0 0"
Taking Damage Position
bind l "setpos 256 0 180; setang 90 0 0"

left with 88HP in both og and this branch, on master it was more like 92

nullsystem (approved):

sunmachine (approved):
This is clean. You have three approvals @Agiel 😂

PR #2085, opened 2026-08-20
Native spectator overlay - Team name/logo/win indicator

nullsystem:
## Description
ConVar set team top area infos with ability in live game to update match scores on match end. Team logo can be set with absolute/relative path PNG/JPG files. Also add new main menu settings UI that'll show up if overlay's enabled to edit those team name/logo settings also.

Also fixed up bAvatarNeedUpdate.

* DONE: Separate PR: NeoUI setting UI to toggle + setup team name + logos
* DONE: Check 1440p + 4k

## Toolchain

### DIFF
--- a/src/game/client/neo/ui/neo_hud_spectator_overlay.cpp
+++ b/src/game/client/neo/ui/neo_hud_spectator_overlay.cpp
@@ -913,12 +913,9 @@ void CNEOHud_SpectatorOverlay::DrawNeoHudElement()
 			{
 				vgui::surface()->DrawSetColor(rsColor);
 				vgui::surface()->DrawFilledRect(x, y, x + iLogoWH, y + iLogoWH);
-				if (pRoundStateHUD)
-				{
-					vgui::surface()->DrawSetColor(COLOR_FADED_DARK);
-					vgui::surface()->DrawSetTexture(pRoundStateHUD->m_teamLogoColors[TEAM_JINRAI].logo);
-					vgui::surface()->DrawTexturedRect(x, y, x + iLogoWH, y + iLogoWH);
-				}
+				vgui::surface()->DrawSetColor(COLOR_FADED_DARK);
+				vgui::surface()->DrawSetTexture(pRoundStateHUD->m_teamLogoColors[TEAM_JINRAI].logo);
+				vgui::surface()->DrawTexturedRect(x, y, x + iLogoWH, y + iLogoWH);
 			}
 
 			// NEO TODO (nullsystem): Deal with very long team name that needs multiline?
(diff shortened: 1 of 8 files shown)


Agiel (dismissed):

AdamTadeusz:
line 257 in vgui_avatarimage.cpp I'm getting a signed/unsigned mismatch warning which is treated as an error
image

nullsystem:
@AdamTadeusz Does changing uint32 -> int fixes that for you? I haven't seen the error from the compiler about this here.

AdamTadeusz:
> @AdamTadeusz Does changing uint32 -> int fixes that for you? I haven't seen the error from the compiler about this here.

Yes

nullsystem:
@AdamTadeusz Added in that change in this branch

src/game/client/neo/ui/neo_hud_spectator_overlay.cpp
```diff
@@ -760,12 +837,187 @@ void CNEOHud_SpectatorOverlay::DrawNeoHudElement()
...
+						y,
+						x + iLogoWH,
+						y + iLogoWH);
+				}
+			}
+			else
+			{
+				vgui::surface()->DrawSetColor(rsColor);
+				vgui::surface()->DrawFilledRect(x, y, x + iLogoWH, y + iLogoWH);
+				if (pRoundStateHUD)
```
AdamTadeusz: already checked if pRoundStateHUD on line 842

src/game/client/neo/ui/neo_hud_spectator_overlay.cpp
```diff
@@ -760,12 +837,187 @@ void CNEOHud_SpectatorOverlay::DrawNeoHudElement()
...
```
AdamTadeusz: ditto on pRoundStateHUD

src/game/client/neo/ui/neo_hud_spectator_overlay.cpp
```diff
@@ -760,12 +837,187 @@ void CNEOHud_SpectatorOverlay::DrawNeoHudElement()
...
+					y += iIndWinWH + iYIndGap;
+				}
+			}
+
+			x -= (iGapW + iLogoWH);
+			const int y = iYNameLogoStart;
+
+			// Team logo
+			const char *pszJinLogo = cl_neo_hud_spectator_overlay_jinrai_logo.GetString();
+			if (pszJinLogo[0])
```
AdamTadeusz, line 894: I recall seeing a coding practice(?) where we first check for pszJinLogo and then pszJinLogo[0], should that apply here? (and further down)

src/game/client/neo/ui/neo_hud_spectator_overlay.cpp
```diff
@@ -126,6 +179,30 @@ void CNEOHud_SpectatorOverlay::FireGameEvent(IGameEvent *event)
 			pCard->iLastHP = 0;
 		}
 	}
+	else if (0 == V_strcmp(type, "game_end"))
+	{
+		// NEO NOTE (nullsystem): Match winner auto update only works on live
```
AdamTadeusz, line 184: I think this is likely to cause some confusion tbh, we don't want to spoil the outcome of a match by setting matches won before the replay ends, but also spectators might wonder why the winner wasn't updated like during a live match, and thus may question if said team really won the match or if there's some other invisible mechanic

AdamTadeusz (dismissed):
Left some nits and questions

src/game/client/neo/ui/neo_hud_spectator_overlay.cpp
nullsystem: `ConVar::GetString` already checks the pointer itself, so should already guarantee a pointer from convar.

src/game/client/neo/ui/neo_hud_spectator_overlay.cpp
nullsystem: I think for now just clarify this mechanic to the people casting + streaming and see how it goes.

Agiel (approved):

AdamTadeusz (approved):

PR #2086, opened 2026-08-21
Invalidate stale cover paths when retreat destination changes

sunmachine:
## Retreat to cover...
image

#### Courtesy of Don Claudio

Screenshot 2026-08-20 at 22 46 24

## Description
Follow-up to #2065. 

When `FindCoverArea` picks a different cover area mid-retreat, the old path stayed valid and kept steering the bot toward the abandoned, exposed cover: 
- until the 1.0s repath timer elapsed in the grenade retreat, 
- and until the old path completed on its own in retreat-to-cover (where the timer was removed and `!m_path.IsValid()` is the only recompute trigger).

Invalidate the path (and repath timer, where one exists) whenever the chosen cover area changes, matching the existing `OnStuck`/`OnMoveToFailure` handling. No change when the re-pick returns the same area.

## Toolchain
- Linux GCC 10 Sniper 3.0

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_retreat_from_grenade.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_retreat_from_grenade.cpp
@@ -221,7 +221,15 @@ ActionResult< CNEOBot >	CNEOBotRetreatFromGrenade::Update( CNEOBot *me, float in
 	// track projectile and relation to escape destination every update
 	if ( !m_coverArea || ( grenadeArea && grenadeArea->IsPotentiallyVisible( m_coverArea ) ) )
 	{
+		CNavArea *pPrevCoverArea = m_coverArea;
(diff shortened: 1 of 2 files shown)


src/game/server/neo/bot/behavior/neo_bot_retreat_from_grenade.cpp
sunzenshen, line 245: Note for other reviewers, this would be when the path is recalculated when the path or timer are invalidated.

src/game/server/neo/bot/behavior/neo_bot_retreat_to_cover.cpp
sunzenshen, line 321: Note for other reviewers, this would be when the path is recalculated when the path is invalidated.

sunzenshen (approved):

AdamTadeusz (approved):

PR #2088, opened 2026-08-22
Hint controls for buttons

DESTROYGIRL:
## Description
Set custom hint messages or disable contextual hints for buttons in maps. Hints are enabled by default if the KV is not set for compatibility with legacy maps

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #1944

### DIFF
--- a/game/bin/rebuild.fgd
+++ b/game/bin/rebuild.fgd
@@ -467,3 +467,239 @@
 [
 	radius(float) : "Radius" : 12 : ""
 ]
+
+@SolidClass base(Targetname, Parentname, Origin, RenderFields,DamageFilter, Button) = func_button : 
+	"A brush entity that's designed to be used for a player-useable button. When used by the player, it moves to a pressed position."
+[
+	movedir(angle) : "Move Direction (Pitch Yaw Roll)" : "0 0 0" : "Specifies the direction of motion to move when the button is used."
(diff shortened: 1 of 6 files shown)


Rainyan (approved):
Tested with `ent_fire !picker addoutput "usehinttext Pay respects"` while aiming at a button, and LGTM

game/bin/rebuild.fgd
```diff
@@ -467,3 +467,239 @@
 [
 	radius(float) : "Radius" : 12 : ""
 ]
+
+@SolidClass base(Targetname, Parentname, Origin, RenderFields,DamageFilter, Button) = func_button :
```
AdamTadeusz, line 471: RenderFields, DamageFilter?

AdamTadeusz (approved):

game/bin/rebuild.fgd
DESTROYGIRL: These are from the base FGD for render FX flags, filters n such. Our FGD is very ugly because for us to extend base FGD definitions we have to copy the entire definition into our own. This wouldn't be needed if we only supported Hammer++ because that has @ExtendClass, but alas

AdamTadeusz: I meant the missing space before DamageFilter

DESTROYGIRL: Idk blame source developer number 6 for that mistake 22 years ago

PR #2090, opened 2026-08-23
Tweak dead avatar color

nullsystem:
## Description
Contrast from -64 to 64, brightness from 1.25 to 0.5, swapped order of operation now contrast then brightness changing.

## Toolchain


nullsystem:
Replaced by https://github.com/NeotokyoRebuild/neo/pull/2094/

PR #2092, opened 2026-08-24
NavMesh Place name support

AdamTadeusz:
## Description

NeoAssets PR https://github.com/NeotokyoRebuild/neoAssets/pull/132

Adds a hud element that displays the name of the closest nav area's place name, if any. Intended for incorporating call-outs from callout maps into the navigation meshes.

Also adds a command for displaying all nearby place names

https://github.com/user-attachments/assets/3c863d21-25bd-4889-a8cd-1da4a47a5c4a

## Toolchain

- Windows MSVC VS2022


AdamTadeusz:
I'm not happy with how client side the place names are loaded within the same function that reads the nav file, ideally that function should take a pointer to a CUtlVector and be moved to a separate file, with the original file calling it then going through all the place names and configuring how they look etc. but I'm tired.

AdamTadeusz:
Ok fixed some last minute issues and included a video showing the feature

src/game/client/neo/c_neo_point_world_text.cpp
```diff
@@ -0,0 +1,340 @@
...
+}
+
+PointWorldText::~PointWorldText()
+{
+}
+
+void PointWorldText::SetText( const char* pszText )
+{
+	m_nTextLength = V_strlen( pszText );
+	V_strncpy( m_szText, pszText, sizeof(m_szText) );
```
sunmachine:
```suggestion
	V_strncpy( m_szText, pszText, sizeof(m_szText) );
	m_nTextLength = V_strlen( m_szText );
```

Measure after the copy.

`m_szText` is lengthed 18 via `MAX_PLACE_NAME_LENGTH` (or 17 + a NUL). 
Recording the length of the untruncated source means CalcTextTotalSize and DrawModel read past the copied text. 💥

There's another chat to have about how to remove the character limit if it gets cumbersome... but c'mon, not like we're gonna' bake in the script to Shrek in here.

sunmachine (changes requested):

src/game/client/neo/ui/neo_hud_place_name.cpp
```diff
@@ -0,0 +1,524 @@
...
+	{
+		case TEXTALIGN_LEFT:
+		default:
+			textXOffset = 0;
+			break;
+		case TEXTALIGN_CENTER:
+		case TEXTALIGN_RIGHT:
+			int textWidth = 0, textHeight = 0;
+			vgui::surface()->GetTextSize(textFont, m_szPlaceName, textWidth, textHeight);
+			textXOffset = textXAlignment == TEXTALIGN_CENTER ? (wide / 2) - (textWidth / 2) : wide - textWidth;
```
sunmachine, line 117: _Not a blocker_.

- `wide` here is the private member, and it is never assigned. 
- `ApplySchemeSettings` declares a shadow `int wide`, but it  
- With the textXAlignment 2 in HudLayout.res this evaluates to 0. 
  - textWidth, DrawSetTextPos gets a negative x, 
  - vgui clips the label entirely. 
  - Centered alignment shows half of it at the top-left. 
  - only left alignment draws. 
  - The world callouts (DrawPlaceNames) do not go through here, so `+showPlaceNames` is unaffected.

The eventual fix is to just use GetWide() here (the panel is sized to the screen in ApplySchemeSettings) and drop the wide member. 

textXpos/textYpos from the .res (-4, 4) are also read but never applied. adding them at the DrawSetTextPos on line 132 gives the label its inset.

src/game/server/nav_area.cpp
```diff
@@ -1265,6 +1269,12 @@ bool CNavArea::SplitEdit( bool splitAlongX, float splitEdge, CNavArea **outAlpha
 	SplitNotification notify( this, alpha, beta );
 	TheNavMesh->ForAllLadders( notify );
 
+#ifdef NEO
+	// If the old area had a place name, the new areas will inherit it
+	alpha->m_place = m_place;
+	beta->m_place = m_place;
```
sunmachine:
```suggestion
	alpha->SetPlace( m_place );
	beta->SetPlace( m_place );
```

sunmachine (changes requested):

src/game/server/neo/neo_player.cpp
```diff
@@ -1192,6 +1193,22 @@ void CNEO_Player::PreThink(void)
...
+		{
+			if (Q_strcmp(m_szLastPlaceName.Get(), placeName))
+			{
+				Q_strncpy(m_szLastPlaceName.GetForModify(), placeName, MAX_PLACE_NAME_LENGTH);
+			}
+		}
+		else
+		{
+			Q_strncpy(m_szLastPlaceName.GetForModify(), "", MAX_PLACE_NAME_LENGTH);
+		}
```
sunmachine:
```suggestion
		// O(1) containment lookup only. Off-mesh (jumping, on props) keep the last known name.
		if (CNavArea *area = TheNavMesh->GetNavArea(GetAbsOrigin()))
		{
			const char *placeName = TheNavMesh->PlaceToName(area->GetPlace());
			if (!placeName)
			{
				placeName = "";
			}
			if (Q_strcmp(m_szLastPlaceName.Get(), placeName))
			{
				Q_strncpy(m_szLastPlaceName.GetForModify(), placeName, MAX_PLACE_NAME_LENGTH);
			}
		}
```

Double check this, seemed to work on my side but I didn't test too hard past gameplay.

src/game/client/neo/c_neo_point_world_text.cpp
```diff
@@ -0,0 +1,340 @@
...
+				behind)
+			{
+				return;
+			}
+		}
+	}
+
+	CMatRenderContextPtr pRenderContext( g_pMaterialSystem );
+	pRenderContext->Bind( pDebugText );
+	pDebugText->IncrementReferenceCount();
```
sunmachine:
```suggestion
```

Nit pick, no decrement, can crash the game if held down for some billion seconds or something.

src/game/client/neo/c_neo_point_world_text.h
```diff
@@ -0,0 +1,60 @@
+///////////////////////////////////////////////
```
sunmachine, line 3: nit

```suggestion
#pragma once

///////////////////////////////////////////////
```

sunmachine (commented):
nitpick

sunmachine (changes requested):

src/game/server/neo/neo_player.cpp
AdamTadeusz: I'm actually not happy with the range of GetNearestNavArea, I would rather nothing show up for the current nav area if the nearest nav area is far away, I would replace GetPlace here with GetNearestNavArea with some custom parameters, but for now ill change the logic here to make sure GetForModify doesn't keep getting called when there's no nearest nav area

AdamTadeusz: Jumping on props etc should not cause any issues that this suggestion would fix, noclipping outside of tetsu top spawn or at an angle above the sofa shows GetPlace is very lenient

image

image

src/game/client/neo/ui/neo_hud_place_name.cpp
AdamTadeusz: I kept wide and made sure to set its value in ApplySchemeSettings, also reading textXPos and textYPos in DrawSetTextPos again

src/game/server/neo/neo_player.cpp
sunmachine: Awesome! 💪 Thanks for sanity checking

sunmachine (dismissed):
BAM

src/game/client/neo/c_neo_point_world_text.cpp
```diff
@@ -0,0 +1,339 @@
+#include "c_neo_point_world_text.h"
+#include "view_scene.h"
+#include "view.h"
```
sunzenshen, line 3: nit: by any chance do we want to add 

```cpp
#include "tier0/memdbgon.h"
```

after these includes?

src/game/server/neo/neo_player.cpp
```diff
@@ -1192,6 +1193,19 @@ void CNEO_Player::PreThink(void)
 			SuperJump();
 		}
 	}
+
+	if (TheNavMesh)
+	{
+		const char* placeName = TheNavMesh->PlaceToName(TheNavMesh->GetPlace(GetAbsOrigin()));
```
sunzenshen: Could probably just use GetLastKnownArea() instead of neeting to query the NavMesh for GetPlace.

Something like:

```cpp
const CNavArea* pArea = GetLastKnownArea();
const char* placeName = pArea ? TheNavMesh->PlaceToName(pArea->GetPlace()) : NULL;
```

src/game/client/neo/ui/neo_hud_place_name.cpp
```diff
@@ -0,0 +1,524 @@
...
+
+	PrecacheMaterial( TEXT_MATERIAL );
+	m_Font.Init(TEXT_MATERIAL, TEXTURE_GROUP_PRECACHED, true);
+
+	placeNameRadiusChangeCallback();
+}
+
+CNEOHud_PlaceName::~CNEOHud_PlaceName()
+{
+	if (g_PlaceName == this)
```
sunzenshen, line 61: Just before, should we also call 
```cpp
IGameSystem::Remove(this);
```
?

src/game/server/nav_mesh.h
```diff
@@ -1128,9 +1135,19 @@ class CNavMesh : public CGameEventListener
 	//----------------------------------------------------------------------------------
 	// Place directory
 	//
+#ifdef NEO
+	struct PlaceNameAndCount
+	{
+		char name[MAX_PLACE_NAME_LENGTH];
+		int count;
+		Vector averageCenter;
```
sunzenshen: Would it be easier/safer to also set vec3_origin here?

```cpp
Vector averageCenter = vec3_origin;
```

src/game/client/neo/ui/neo_hud_place_name.cpp
```diff
@@ -0,0 +1,524 @@
...
+static ConCommand endshowplacenames("-showPlaceNames", [](const CCommand& args)->void {shouldDrawPlaceNames = false; });
+
+static float animationAlpha = 0.f;
+void CNEOHud_PlaceName::DrawPlaceNames()
+{
+	constexpr int ANIMATION_SPEED = 2;
+	animationAlpha =   shouldDrawPlaceNames	? min(1.0f, animationAlpha + (gpGlobals->frametime * ANIMATION_SPEED))
+											: max(0.0f, animationAlpha - (gpGlobals->frametime * ANIMATION_SPEED));
+
+	for (PlaceNameCallout place : places)
```
sunzenshen: Maybe this could be a reference to avoid a copy?

```cpp
for (PlaceNameCallout &place : places)
```

src/game/client/neo/ui/neo_hud_place_name.cpp
```diff
@@ -0,0 +1,524 @@
...
+
+		places.RemoveAll();
+
+		// read each entry
+		char placeName[256];
+		unsigned short len;
+		for( int i=0; i<placeCount; ++i )
+		{
+			len = fileBuffer.GetUnsignedShort();
+			fileBuffer.Get( placeName, MIN( sizeof( placeName ), len ) );
```
sunzenshen, line 355: At this point, might be a good idea to initalize all the placeName(s) to either have their placeName or to cap off the string with a null terminator like:

```cpp
placeName[ MIN( sizeof( placeName ) - 1, len ) ] = '\0';
```

src/game/client/neo/ui/neo_hud_place_name.cpp
```diff
@@ -0,0 +1,524 @@
...
+			}
+		}
+
+		// Place data
+		unsigned short entry = fileBuffer.GetUnsignedShort();
+		if (entry > 0 && entry <= places.Count())
+		{
+			entry -= 1;
+			Vector newNavCenter = (nwCorner + seCorner) / 2.f;
+			places[entry].origin = ((places[entry].origin * places[entry].navAreaCount) + newNavCenter) / ++places[entry].navAreaCount;
```
sunzenshen: embedding the ++ preincrement might be kind of hard to reason about so it might be easier to evaluate that up front and change this line in response:

```cpp
places[entry].navAreaCount++;
places[entry].origin = ((places[entry].origin * (places[entry].navAreaCount - 1)) + newNavCenter) / places[entry].navAreaCount;
```

also, might be undefined behavior when the preincrement is evaluated as an operand of `/`

src/game/server/nav_file.cpp
sunmachine, line 179: This is an issue that can crop up if you're editing, you can reproduce it by naming two navs "Thinktank Maintenence Bay 01" and "Thinktank Maintenence Bay 02" as an example.

PlaceDirectory::Load keys the directory by name, so any two place entries whose stored names match (which in-game editing produces whenever two callouts share their first 17 characters, since names are truncated to 17 chars) collapse into one slot on reload, and silently shift every later area onto the wrong callout once saved.

```suggestion
		placeName[ MIN( sizeof( placeName ) - 1, len ) ] = '\0';

		// Every directory entry gets its own slot, in file order, so the
		// runtime index always equals the on-disk index that each area stores.
		Place place = TheNavMesh->LoadPlace( placeName );

		if ( V_strlen( placeName ) >= MAX_PLACE_NAME_LENGTH )
		{
			Warning( "NavMesh place \"%s\" is longer than %d characters, stored as \"%s\"\n",
			placeName, MAX_PLACE_NAME_LENGTH - 1, TheNavMesh->PlaceToName( place ) );
		}
#else
		Place place = TheNavMesh->NameToPlace( placeName );
		if (place == UNDEFINED_PLACE)
		{
			Warning( "Warning: NavMesh place %s is undefined?\n", placeName );
		}
#endif // NEO
		AddPlace( place );
```

### Reproduction

**Setup:** listen server (the nav edit commands require the listen-server host), any shipped map with no callouts yet. `ntre_bullet_tdm` has the smallest nav.

```
1.  map ntre_bullet_tdm
2.  sv_cheats 1
3.  nav_edit 1
4.  nav_toggle_place_mode // edit mode: Place Painting
5.  nav_use_place "Thinktank Maintenence Bay 01"
        console: Current place set to new place 'Thinktank Maintenence Bay 01'
6.  aim at an area, nav_toggle_place_painting, then nav_toggle_place_painting again
        console: Current place set to new place 'Thinktank Maintenence Bay 02'
8.  aim at a DIFFERENT area, nav_toggle_place_painting, then again
        on-screen: Area #<id> Thinktank Mainten  // <- same text as step 6
9.  nav_use_place "Catwalk"
10. aim at a THIRD area, nav_toggle_place_painting, then again
        on-screen: Area #<id> Catwalk
11. nav_place_list
        Map uses 3 place names:
            Thinktank Mainten
            Thinktank Mainten  // <- duplicate: two places, one text
            Catwalk
12. nav_save
13. map ntre_bullet_tdm // full reload: server AND client re-read the file
14. sv_cheats 1; nav_place_list
        Map uses 2 place names: //<- one place collapsed on load
15. walk onto each of the three areas; watch the HUD place line while holding +showplacenames
```

Note: `nav_load` alone is not enough at step 13. The client only re-parses the nav file at level init, so a `map` reload is needed for both sides to read the saved file.

**Expected at step 15.** Before the reload, all three areas showed the same text on the HUD line and the on-screen edit text. After the reload, the 3D label over each area (client, reads the file 1:1) still shows what was painted, but the HUD line (server, shifted directory) disagrees on at least one. Which one depends on the area IDs chosen, since the saved directory is ordered by first encounter in area-ID order:

| area IDs in order | HUD on Bay 01 area | HUD on Bay 02 area | HUD on Catwalk area |
|---|---|---|---|
| Bay01 < Bay02 < Catwalk | `Thinktank Mainten` | **`Catwalk`** | **blank** |
| Bay01 < Catwalk < Bay02 | `Thinktank Mainten` | **blank** | `Catwalk` |
| Catwalk < Bay01 < Bay02 | `Thinktank Mainten` | **blank** | `Catwalk` |

The 3D labels read `Thinktank Mainten`, `Thinktank Mainten`, `Catwalk` in every case. A HUD line that names a different callout than the label floating over the same spot, or is blank under a label, is the reproduction. A `nav_save` at this point writes the wrong assignments to disk, and a further reload produces no warning.

**The console prints nothing that flags a problem at any step:** no truncation message at 5/7, no duplicate warning at 12, no warning at 13. The doubled line at step 11 and the truncated on-screen text are the only tells.

**Where it happens on load** (`src/game/server/nav_file.cpp`, `PlaceDirectory::Load`):

1. `:172` `LoadPlace(placeName)` appends the second `Thinktank Mainten` as its own slot.
2. `:174` `NameToPlace(placeName)` returns the *first* slot (`FStrEq` is `V_stricmp`).
3. `:112-113` `AddPlace` sees `IsKnown` and returns without appending. Directory is one short; `IndexToPlace` (`:119-133`) shifts every later area.

Image

Image

sunmachine (changes requested):
Nav-editing bug. I really hate in-game nav editing and review. Fuck this process. I'm going to make this completely obsolete mark my words.

src/game/server/nav_file.cpp
AdamTadeusz: I think the intermediate Catwalk callout is unnecessary
image

AdamTadeusz:
@sunmachine seems to me like clamping the place name before use solves this issue right?

AdamTadeusz:
A long input is one thing, an empty is another. I figured this would also come in handy instead of relying on nav_place_pick ing a nav area without a place to allow clearing places with the paint function

sunmachine:
> @sunmachine seems to me like clamping the place name before use solves this issue right?

I think so! But I am broadly not keen on tying user-facing strings as IDs either (it's not a problem now but it's a localization smell eventually, assuming that's a thing we want to do one day). Good practice, not necessary for this "fix", just calling the smell as I saw it crop up as an actual editing bug. 

Feel free to dismiss the review or update with the fix as you prescribed!

AdamTadeusz:
> > @sunmachine seems to me like clamping the place name before use solves this issue right?
> 
> I think so! But I am broadly not keen on tying user-facing strings as IDs either (it's not a problem now but it's a localization smell eventually, assuming that's a thing we want to do one day). Good practice, not necessary for this "fix", just calling the smell as I saw it crop up as an actual editing bug.
> 
> Feel free to dismiss the review or update with the fix as you prescribed!

Yep 👍 it's in [448c909](https://github.com/NeotokyoRebuild/neo/pull/2092/commits/448c90924a371ba3d1cc66b34ff9d2d737ca5495)

sunmachine:
> > > @sunmachine seems to me like clamping the place name before use solves this issue right?
> 
> > 
> 
> > I think so! But I am broadly not keen on tying user-facing strings as IDs either (it's not a problem now but it's a localization smell eventually, assuming that's a thing we want to do one day). Good practice, not necessary for this "fix", just calling the smell as I saw it crop up as an actual editing bug.
> 
> > 
> 
> > Feel free to dismiss the review or update with the fix as you prescribed!
> 
> 
> 
> Yep 👍 it's in [448c909](https://github.com/NeotokyoRebuild/neo/pull/2092/commits/448c90924a371ba3d1cc66b34ff9d2d737ca5495)

Ohhh my bad I had no idea! The UI didn't annotate the commit in the first place like I expected it would, and didn't see the change. 🤪

sunmachine (approved):

sunzenshen (approved):

PR #2093, opened 2026-08-24
Native spectator overlay - Team name/logo presets

nullsystem:
## Description
Add a simple preset setter set by overlay/presets.txt so in the overlay settings UI can just select a team rather than typing out the name and finding the logo file.

## Toolchain

### DIFF
--- a/game/neo/overlay/presets.txt
+++ b/game/neo/overlay/presets.txt
@@ -1,0 +1,23 @@
+// overlay/presets.txt
+//
+// Team name and logo presets for the overlay HUD
+//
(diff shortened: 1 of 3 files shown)


Agiel (approved):
Works great. Here's the preset for GG26 to test. (TGR Woods' logo doesn't work because it's a gif, but that's expected).
[presets.zip](https://github.com/user-attachments/files/31435753/presets.zip)

sunzenshen (approved):
Set `cl_neo_hud_spectator_overlay_enabled  1` and LGTM:

Image

Image

Image

Image

Image

PR #2095, opened 2026-08-25
Fix not showing grenade/det icon on death HUD

nullsystem:
## Description
Just relying on the previous bools given. Overlay stay on using by index.

## Toolchain


Agiel (approved):

sunzenshen (approved):
LGTM:

Image

Image

PR #2099, opened 2026-08-28
Fix bot_mimic not copying weapon swap to slot 1

sunzenshen:
## Description
Fixes bots not copying weapon swap to slot 1 when `bot_mimic` is enabled.
As a vestige of HL2DM, bots were copying weapon swaps based on entity ID and not by slot.
This commit involves retrieving the slot number of the player's weapon and mimicking the swap to that slot.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/NextBot/Player/NextBotPlayer.h
+++ b/src/game/server/NextBot/Player/NextBotPlayer.h
@@ -814,22 +814,30 @@ inline void NextBotPlayer< PlayerType >::PhysicsSimulate( void )
 				cmd = *pPlayerMimicked->GetLastUserCommand();
 				cmd.viewangles[YAW] += bot_mimic_yaw_offset.GetFloat();
 
-				if ( cmd.weaponselect != 0 ) // Mimicked player swapped weapon
+				if (cmd.weaponselect != 0) // mimicked player swapped weapon
 				{
-					// Instead of trying to select the exact weapon the mimicked player just equipped...
-					CBaseCombatWeapon *pSrcWeapon = dynamic_cast<CBaseCombatWeapon *>(
-						CBaseEntity::Instance( cmd.weaponselect ) );
+					CBaseEntity *pSrcEnt = CBaseEntity::Instance(cmd.weaponselect);
+					cmd.weaponselect = 0; // default to not swapping weapon
 
-					if ( pSrcWeapon )
+					if (pSrcEnt && pSrcEnt->IsBaseCombatWeapon())
 					{
-						// ... get the slot number of the mimicked player's weapon choice
-						CBaseCombatWeapon *pBotWeapon = pThisBot->Weapon_GetSlot( pSrcWeapon->GetSlot() );
-						// and swap to the same slot weapon if available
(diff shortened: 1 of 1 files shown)


sunzenshen:
Demo of fix:

https://github.com/user-attachments/assets/c625563b-fe34-45d4-920a-eaaf4867845e

Notice that if Recon does not have a utility, it merely skips that weapon swap.

src/game/server/NextBot/Player/NextBotPlayer.h
```diff
@@ -814,6 +814,25 @@ inline void NextBotPlayer< PlayerType >::PhysicsSimulate( void )
 				cmd = *pPlayerMimicked->GetLastUserCommand();
 				cmd.viewangles[YAW] += bot_mimic_yaw_offset.GetFloat();
 
+				if ( cmd.weaponselect != 0 ) // Mimicked player swapped weapon
+				{
+					// Instead of trying to select the exact weapon the mimicked player just equipped...
+					CBaseCombatWeapon *pSrcWeapon = dynamic_cast<CBaseCombatWeapon *>(
+						CBaseEntity::Instance( cmd.weaponselect ) );
```
Rainyan: Could perhaps check via `->IsBaseCombatWeapon()`, to avoid the RTTI overhead?

Rainyan (dismissed):
LGTM, should make it less awkward to test weapon switch stuff.

Few nits:

* Going by weapon slot will be agnostic to which of the weapons it is for slots than can have multiple guns at once (i.e. the grenades). So for example you couldn't force an assault bot who happens to also hold a smoke grenade in their inventory to specifically choose the smoke over the frag grenade:

smoke_ne_nade

One option could be sorting the found slot items by best matching NeoBits?

* Weapon switch via `use` (eg. `use weapon_smokegrenade`) is not reflected by the mimic. It's somewhat common at least for power users to have specific binds for the exact grenade types because it's faster than using the slot to cycle through. So users with these kinds of quick equip nade binds would not benefit. But maybe somewhat out of scope.

src/game/server/NextBot/Player/NextBotPlayer.h
sunzenshen: > Weapon switch via use (eg. use weapon_smokegrenade) is not reflected by the mimic. It's somewhat common at least for power users to have specific binds for the exact grenade types because it's faster than using the slot to cycle through. So users with these kinds of quick equip nade binds would not benefit. But maybe somewhat out of scope.

Seems like with the 2nd commit, after addressing the specific weapon choice before falling back to a slot, seems to work okay with the utility swap hotkeys?  Here is me pressing the default hjk binds in this second demo video:

https://github.com/user-attachments/assets/42a3fb3f-1765-4af1-82f2-5031873bf869

Rainyan: Nice, seems to work fine now.

> seems to work okay with the utility swap hotkeys?

Ah, I forgot we had specific slot commands available for these instead of the `use weapon_...` syntax.

Rainyan (approved):

AdamTadeusz (approved):

PR #2101, opened 2026-08-28
Refactor avatar handling completely for overlay + scoreboard + killer info

nullsystem:
## Description
Throw away CAvatarImage and just it ourselves

Testing: Just go through a lot of demos + reloading NT;RE and doing it again to be sure it loads up properly.

## Toolchain


Agiel (dismissed):

AdamTadeusz:
Ideally we would have a separate CAvatarImage storage that is updated as players join and leave, that other elements can reference instead of each requesting their own copy of each player's avatar, but this seems like a fine workaround for now

src/game/client/neo/ui/neo_hud_spectator_overlay.cpp
```diff
@@ -392,17 +408,6 @@ void CNEOHud_SpectatorOverlay::UpdateStateForNeoHudElementDraw()
 			}
 		}
 	}
-
-	// NEO NOTE (nullsystem): This is so the spectator don't need to tab press
-	// on the scoreboard to update the avatar. Generally only happens in the
-	// beginning of match/demo
-	//
-	// NEO TODO (nullsystem): Move avatar handling off to something separate from
-	// the scoreboard but shared between scoreboard, overlay, and killer info
```
AdamTadeusz, line 401: Keep this NEO TODO?

AdamTadeusz (dismissed):

sunmachine:
~~I am verifying a possible crash case. :joy: I am working slow though; will get it up hopefully soon @nullsystem~~

Edit: Review is up :+1: two nits, higher-res-monitor issue with icons, and image list fix.

AdamTadeusz:
Oh yeah m_pImageList has m_bDeleteImagesWhenDone set to false, when are the new CAvatarImage s deleted?

src/game/client/neo/ui/neo_hud_spectator_overlay.cpp
```diff
@@ -217,12 +224,17 @@ void CNEOHud_SpectatorOverlay::ApplySchemeSettings(vgui::IScheme* pScheme)
 
 	SetFgColor(COLOR_TRANSPARENT);
 	SetBgColor(COLOR_TRANSPARENT);
+
+	if (m_pImageList)
+	{
+		delete m_pImageList;
+	}
+	m_pImageList = new vgui::ImageList(false);
+	m_mapAvatarsToImageList.RemoveAll();
```
sunmachine: We regenerate the image list but we don't safely clear the iCardsSize.

```suggestion
	m_mapAvatarsToImageList.RemoveAll();
	m_iCardsSize = 0;
```

sunmachine (changes requested):

src/game/client/neo/ui/neo_hud_spectator_overlay.cpp
```diff
@@ -431,13 +436,26 @@ void CNEOHud_SpectatorOverlay::DrawPlayerCard(const SpectatorPlayerCard &card,
...
-		card.pAvatar->Paint();
+		CAvatarImage *pAvatarImg = nullptr;
+		if (card.avatar.i184Idx >= 0)
+		{
+			pAvatarImg = (CAvatarImage *)(m_pImageList->GetImage(card.avatar.i184Idx));
+		}
+		if ((!pAvatarImg || !pAvatarImg->IsValid()) && card.avatar.i64Idx >= 0)
+		{
+			pAvatarImg = (CAvatarImage *)(m_pImageList->GetImage(card.avatar.i64Idx));
+		}
```
sunmachine: This isn't a blocker but it's kinda' neat. I was worried it was a crash path at first because... I misread shit. Instead, you get stuck with lower-res icons at larger resolutions (1440p/4k). You won't notice this on 1080p since it picks the 64.

When the 184 image isn't ready/valid yet we swap to the 64, and only paint that. Where does the 184 image get another chance to load after this? 🤷 I don't think it does.

I'm reading how `CAvatarImage` re-polls Steam when `m_bLoadPending` is set (looking at vgui_avatarimage.cpp's `Paint()` func). 
- Failed attempts call `m_fNextLoadTime = curtime + 1.0f` there, 
- and the other path through `OnPersonaStateChanged` calls `LoadAvatarImage()`.
- if Steam answers inside that 1s window the attempt is skipped and m_bLoadPending just stays set.

So the 64 stays set and the rest of the session. The scoreboard doesn't have this since it's picking by `iAvatarSize`.

I thought I was just being old and blind but I kept seeing a slightly blurrier icon. Posting a fix below.

sunmachine: See above for context.

The choice no longer depends on validity, so the image we want is the image that keeps getting painted, and `Paint()` is what eventually makes it valid. That is the whole fix. No more circular selection 

```suggestion
		// Pick the image by draw size, same as the scoreboard. Always paint the
		// one we picked: CAvatarImage only retries loading from Paint(), so an
		// image that is never painted never becomes valid.
		CAvatarImage *pAvatarImg = nullptr;
		if (card.avatar.i184Idx >= 0 && iAvatarSize > 64)
		{
			pAvatarImg = (CAvatarImage *)(m_pImageList->GetImage(card.avatar.i184Idx));
		}
		else if (card.avatar.i64Idx >= 0)
		{
			pAvatarImg = (CAvatarImage *)(m_pImageList->GetImage(card.avatar.i64Idx));
		}
```

sunmachine (dismissed):

src/game/client/neo/ui/neo_hud_spectator_overlay.cpp
```diff
@@ -102,6 +100,7 @@ CNEOHud_SpectatorOverlay::CNEOHud_SpectatorOverlay(const char* pElementName, vgu
 	g_pNeoHudSpecOverlay = this;
 	SetAutoDelete(true);
 	m_iHideHudElementNumber = NEO_HUD_ELEMENT_SPECTATOR_OVERLAY;
+	m_pImageList = new vgui::ImageList(false);
```
sunmachine: nit-pick, but it looks like ImageList cleans up for us if we set it to `true` and prevents orphaning the images and their persona callbacks. At first I was worried this was just a style convention to not bother raising but this actually seems useful. 

```suggestion
	m_pImageList = new vgui::ImageList(true);
```

src/game/client/neo/ui/neo_hud_spectator_overlay.cpp
```diff
@@ -217,12 +224,17 @@ void CNEOHud_SpectatorOverlay::ApplySchemeSettings(vgui::IScheme* pScheme)
 
 	SetFgColor(COLOR_TRANSPARENT);
 	SetBgColor(COLOR_TRANSPARENT);
+
+	if (m_pImageList)
+	{
+		delete m_pImageList;
+	}
+	m_pImageList = new vgui::ImageList(false);
```
sunmachine: Same as ln 103

```suggestion
	m_pImageList = new vgui::ImageList(true);
```

sunmachine (commented):
non-blocking nit-picks

src/game/client/neo/ui/neo_hud_spectator_overlay.cpp
nullsystem: CAvatarImage implementation itself actually have a global variable where it caches and reuses textures so I think probably not that necessary to have separate handling now.

src/game/client/neo/ui/neo_hud_spectator_overlay.cpp
nullsystem: I would have thought `SetAvatarSteamID` would have set it straight away since `SetAvatarSize` which flips `m_bSetDesiredSize` on so it calls `LoadAvatarImage` inside of `SetAvatarSteamID`.

nullsystem: Alright I've decided to just have a single `CAvatarImage` rather than by size since the implementation never actually sticks to a 32/64/184 size anyway. I'll also need to change how the scoreboard and killer info handles this also again separately.

nullsystem:
Going to work on a "NeoAvatar" CAvatarImage replacement and extend this PR to cover redoing scoreboard + killer HUD.

nullsystem:
@Agiel @AdamTadeusz @sunmachine Just to note the PR is completely re-written so re-review it fully.

AdamTadeusz:
Ok so I'm watching back a demo from yesterday and in the spectator overlay I can only see the profile pictures of players im friends with on steam (and my own profile picture)
Zrzut ekranu 2026-08-30 144541
I removed you from my friends list, restarted steam and the game, and now your profile picture also doesn't load
Zrzut ekranu 2026-08-30 144906

(Edit) This is on windows

src/game/client/neo/ui/neo_avatar.cpp
```diff
@@ -0,0 +1,165 @@
...
+	{
+		iAvatar = steamapicontext->SteamFriends()->GetSmallFriendAvatar(m_SteamID);
+	}
+	else if (iAvatarWH <= 64)
+	{
+		iAvatar = steamapicontext->SteamFriends()->GetMediumFriendAvatar(m_SteamID);
+	}
+	else
+	{
+		iAvatar = steamapicontext->SteamFriends()->GetLargeFriendAvatar(m_SteamID);
```
AdamTadeusz: 🤔

nullsystem: I think it's more maybe to due to missing a call to `RequestUserInformation`. Weirdly it fetches everyone (friend or not friend) fine on Linux as is.

nullsystem: @AdamTadeusz I've updated the branch 120a77c530c494675eb389d1ec69968377a6da55 check it if `RequestUserInformation` helped

AdamTadeusz:
Great!

image

AdamTadeusz (dismissed):
I think steamapicontext is deprecated but lgtm

Agiel (dismissed):
Verified that it's working on Linux also 🐧

nullsystem:
@AdamTadeusz a670556429649fc6befbe5a7736519ab126a1aca removed steamapicontext now, mainly so lines aren't so long/redudent anyway

Agiel (approved):

AdamTadeusz (approved):

sunmachine:
> @Agiel @AdamTadeusz @sunmachine Just to note the PR is completely re-written so re-review it fully.

:pinched_fingers: Good shit

sunmachine (approved):

PR #2104, opened 2026-08-30
Separate ping button from IN_ATTACK3

### LINKED ISSUE #2040: ping placement should have its own button rather than relying on IN_ATTACK3

sunzenshen:
## Description
Refactor the ping button to use its own button, instead of overloading the use of IN_ATTACK3

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #2040


sunzenshen:
Drafted, to see how #2039 turns out, as that expands the button range from 32 bits to 64 bits.

PR #2105, opened 2026-08-31
Bound debug overlay draw counts to avoid resource crashes

### LINKED ISSUE #1539: nb_debug * crashes the game at high bot numbers
everytime I try to run the command nb_debug * the game will crash shortly after in the weirdest places
image
image
image

To reproduce:
click on create server
set max players and bot quota to a value greater than or equal to 15
join spectator
run the commands "sv_cheats 1;nb_debug *"
the exceptions occurs usually around round start

a smaller bot quota is required to trigger the exception provided host_timescale is increased accordingly

sunzenshen:
## Description
In order to prevent the game from crashing when nb_debug is displaying debug overlays for a large number of bots on a large map, bound the number of draw counts that can appear at a time via a ConVar.  This has the side effect of causing flickering of alternating debug overlay draws between bots, but the justification is that it's better to see periodically the debug overlays per bot, rather than to not have the options because the game would crash trying to render all debug overlays.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1539

### DIFF
--- a/src/game/client/cdll_client_int.cpp
+++ b/src/game/client/cdll_client_int.cpp
@@ -154,6 +154,7 @@
 #include <vgui_controls/Button.h>
 #include <vgui_controls/MenuButton.h>
 #include "neo_mp3player.h"
+#include "neo/neo_debugoverlay_budget.h"
 #endif
 
 extern vgui::IInputInternal *g_InputInternal;
@@ -991,8 +992,15 @@ int CHLClient::Init( CreateInterfaceFn appSystemFactory, CreateInterfaceFn physi
 		return false;
 	if ( (render = (IVRenderView *)appSystemFactory( VENGINE_RENDERVIEW_INTERFACE_VERSION, NULL )) == NULL )
 		return false;
+#ifdef NEO
+	// Route every debug overlay through the budgeting proxy (neo/neo_debugoverlay_budget.*)
+	// so a runaway per-tick visualiser such as nb_debug can't exhaust the renderer.
+	if ( (debugoverlay = NEO_InstallDebugOverlayBudget( (IVDebugOverlay *)appSystemFactory( VDEBUG_OVERLAY_INTERFACE_VERSION, NULL ) )) == NULL )
+		return false;
+#else
 	if ( (debugoverlay = (IVDebugOverlay *)appSystemFactory( VDEBUG_OVERLAY_INTERFACE_VERSION, NULL )) == NULL )
 		return false;
+#endif
 	if ( (datacache = (IDataCache*)appSystemFactory(DATACACHE_INTERFACE_VERSION, NULL )) == NULL )
 		return false;
 	if ( !mdlcache )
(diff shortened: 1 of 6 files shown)


src/game/shared/neo/neo_debugoverlay_budget.h
```diff
@@ -0,0 +1,21 @@
...
+//
+// Wrapping the interface covers every overlay emitter in the game from a single
+// place.
+
+class IVDebugOverlay;
+
+// Wrap the engine's debug overlay interface and return the pointer game code
+// should use. A NULL input (a dedicated server has no debug overlay) returns
+// NULL, so existing "if ( debugoverlay )" checks keep working.
+IVDebugOverlay *NEO_InstallDebugOverlayBudget( IVDebugOverlay *pReal );
```
sunzenshen, line 21: Under the hood, use a proxy pattern to drop draws when the flow of overlay draws is starting to exceed the ConVar defined budget neo_debugoverlay_budget.

src/game/client/cdll_client_int.cpp
```diff
@@ -991,8 +992,15 @@ int CHLClient::Init( CreateInterfaceFn appSystemFactory, CreateInterfaceFn physi
 		return false;
 	if ( (render = (IVRenderView *)appSystemFactory( VENGINE_RENDERVIEW_INTERFACE_VERSION, NULL )) == NULL )
 		return false;
+#ifdef NEO
+	// Route every debug overlay through the budgeting proxy (neo/neo_debugoverlay_budget.*)
+	// so a runaway per-tick visualiser such as nb_debug can't exhaust the renderer.
+	if ( (debugoverlay = NEO_InstallDebugOverlayBudget( (IVDebugOverlay *)appSystemFactory( VDEBUG_OVERLAY_INTERFACE_VERSION, NULL ) )) == NULL )
+		return false;
+#else
 	if ( (debugoverlay = (IVDebugOverlay *)appSystemFactory( VDEBUG_OVERLAY_INTERFACE_VERSION, NULL )) == NULL )
```
sunzenshen, line 1001: My pitch for this approach is that we only need to wrap debugoverlay at a few initialization sites, and we can handle custom changes in the new proxy class.

sunzenshen (commented):
Stress test demo of 31 bots on a wide open map, while changing between nb_debug modes:

https://streamable.com/ypau0r

The side effect is that debug draws are rejected near the limit, which causes flickering, but I'd rather have some visibility into something like paths than to not have the option because of a hard crash.

Rainyan (dismissed):
LGTM, tested with launch arg `-maxplayers 32` and cmd `neo_bot_quota 32; nb_debug PATH;` as in the `ntre_engage_ctg` example case and can verify this fixes the crash.

AdamTadeusz:
Getting this error on windows
image
casting NUM_BUCKETS to a float on line 187 is one way to solve the problem.

AdamTadeusz (changes requested):
^

sunzenshen:
Needed to rebase.

Rainyan (approved):

AdamTadeusz (approved):

PR #2109, opened 2026-09-02
Weapon API IsWeaponReloadable()

sunzenshen:
## Description
Instead of manually checking for specific weapon bits, expose an interface to check if weapon is reloadable.

## Toolchain
- Windows MSVC VS2022


Rainyan (dismissed):
For completeness, should probably also cover `CNEOBaseProjectile` and the `weapon_ghost`, which also subclass `CNEOBaseCombatWeapon`.

As for `weapon_smac` it spawns with only 1 mag, so it typically can't be reloaded (both because there's only 1 VIP and it uses a unique `AMMO_SMAC` ammo type, but also because by default you can't loot ammo from weapons). But technically it does support reloading if you modify its weapon script (despite missing the reload animation, or at least it's not hooked up). So it probably is indeed correct for the SMAC to return `true` as it is currently.

Rainyan (approved):

DESTROYGIRL (approved):

PR #2110, opened 2026-09-02
More !IsTeamplay fixes for DM

sunzenshen:
## Description
* Fixed the client deathmatch scorer scanning invalid player slot 0.
* Fixed deathmatch timelimit handling so stopped matches correctly enter overtime status.
* Fixed deathmatch round startup when all players choose the same cosmetic team.
* Disabled team-based match-point, sudden-death, and do-or-die checks in free-for-all modes.
* Fixed deathmatch assists being incorrectly treated as team damage.
* Fixed `FetchAssists` from counting the victim as their own assister and from missing valid assisters.
* Fixed friendly-fire, mirrored damage, auto-kick tracking, and spawn immunity incorrectly applying between deathmatch opponents.
* Fixed bots incorrectly seeing same-team deathmatch opponents through thermoptic camo.
* Fixed bot vantage-point searches from excluding opponents who share the bot's cosmetic team.
* Fixed bots unnecessarily avoiding collisions with deathmatch opponents.
* Fixed smoke-grenade behavior from treating same-team deathmatch opponents as allies.
* Optimized bot vantage-point searches by collecting the enemy list once instead of rescanning players for every candidate nav area.
* Fixed bots repeatedly retrying failed seek-and-destroy pathfinding every tick by adding a one-second retry backoff after failed searches.
* Added regression coverage confirming deathmatch scoring and match flow, including ties, and single-team cosmetic selections.
* Added CTG regression coverage for shared teamplay logic, including round transitions, assists, damage handling, and bot behavior.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/neo_bot.cpp
+++ b/src/game/server/neo/bot/neo_bot.cpp
@@ -1310,6 +1310,8 @@ public:
 	{
 		m_vantageArea = NULL;
 
+		m_enemies.EnsureCapacity(gpGlobals->maxClients - 1);
+
 		for (int i = 1; i <= gpGlobals->maxClients; ++i)
 		{
 			CNEO_Player* enemy = ToNEOPlayer(UTIL_PlayerByIndex(i));
@@ -1330,7 +1332,7 @@ public:
 
 		for (int i = 0; i < m_enemies.Count(); ++i)
 		{
-			CNavArea* enemyArea = (CNavArea*)m_enemies[i]->GetLastKnownArea();
+			CNavArea* enemyArea = m_enemies[i]->GetLastKnownArea();
 			if (enemyArea->IsCompletelyVisible(area))
 			{
 				// nearby area from which we can see an enemy
--- a/src/game/server/neo/bot/neo_bot.cpp
+++ b/src/game/server/neo/bot/neo_bot.cpp
@@ -1306,28 +1306,34 @@ void CNEOBot::UpdateLookingAroundForEnemies(void)
 class CFindVantagePoint : public ISearchSurroundingAreasFunctor
 {
 public:
-	CFindVantagePoint(int enemyTeamIndex)
+	CFindVantagePoint(const CNEOBot* me)
 	{
-		m_enemyTeamIndex = enemyTeamIndex;
 		m_vantageArea = NULL;
+
+		for (int i = 1; i <= gpGlobals->maxClients; ++i)
+		{
+			CNEO_Player* enemy = ToNEOPlayer(UTIL_PlayerByIndex(i));
+
+			if (!enemy || !me->IsEnemy(enemy))
+				continue;
+
+			if (!enemy->IsAlive() || !enemy->GetLastKnownArea())
+				continue;
+
+			m_enemies.AddToTail(enemy);
+		}
 	}
 
 	virtual bool operator() (CNavArea* baseArea, CNavArea* priorArea, float travelDistanceSoFar)
 	{
 		CNavArea* area = (CNavArea*)baseArea;
 
-		CTeam* enemyTeam = GetGlobalTeam(m_enemyTeamIndex);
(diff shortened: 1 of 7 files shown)


sunzenshen:
ntre_apparatus_dm still doesn't have a navmesh, so a rough one can be generated with `nav_generate`

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -1306,28 +1306,34 @@ void CNEOBot::UpdateLookingAroundForEnemies(void)
 class CFindVantagePoint : public ISearchSurroundingAreasFunctor
 {
 public:
-	CFindVantagePoint(int enemyTeamIndex)
+	CFindVantagePoint(const CNEOBot* me)
 	{
-		m_enemyTeamIndex = enemyTeamIndex;
 		m_vantageArea = NULL;
+
+		for (int i = 1; i <= gpGlobals->maxClients; ++i)
```
Rainyan, line 1315: Nit: could call `m_enemies.EnsureCapacity(gpGlobals->maxClients - 1);` before entering the loop to avoid heap reallocations whenever `AddToTail` is called. Assuming `gpGlobals->maxClients - 1` is the maximum possible enemy count(?)

src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -1306,28 +1306,34 @@ void CNEOBot::UpdateLookingAroundForEnemies(void)
...
-		for (int i = 0; i < enemyTeam->GetNumPlayers(); ++i)
+		for (int i = 0; i < m_enemies.Count(); ++i)
 		{
-			CNEO_Player* enemy = (CNEO_Player*)enemyTeam->GetPlayer(i);
-
-			if (!enemy->IsAlive() || !enemy->GetLastKnownArea())
-				continue;
-
-			CNavArea* enemyArea = (CNavArea*)enemy->GetLastKnownArea();
+			CNavArea* enemyArea = (CNavArea*)m_enemies[i]->GetLastKnownArea();
```
Rainyan: Nit: superfluous C cast, `GetLastKnownArea` already returns `CNavArea*`

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -964,27 +964,31 @@ void CNEORules::GetDMHighestScorers(
...
-		if (!g_PR->IsConnected(i))
+		if (!g_PR->IsConnected(i) || g_PR->GetTeam(i) < FIRST_GAME_TEAM)
 		{
 			continue;
 		}
 		iXP = g_PR->GetXP(i);
 #endif
 
-		if (iXP == *iHighestXP)
+		if (*iHighestPlayersTotal == 0 || iXP > *iHighestXP)
```
Rainyan: Maybe I'm missing something, but isn't `iHighestPlayersTotal` unconditionally set to value 0 at line 946, turning `*iHighestPlayersTotal == 0` into an unconditional `true` for this condition?

In the same sense, the assignment at line 982 seems redundant(?)

Rainyan (dismissed):

src/game/shared/neo/neo_gamerules.cpp
sunzenshen: This function is used for listing the players that are tied for first place.  This line was attempting to fix an edge case where ALL players have negative score, so the comparison with iHighestXP is never true.  But actually it's simpler to just disregard that edge case, because if everyone has a negative score, no one is really winning.  (Also, if everyone is somehow having a terrible time having negative scores, they can just vote to change the map.)

The assignment at line 982 is fixing a bug where if we are iterating through the player list, and we found someone with a higher score than the running threshold, then we will still list all the other players we have counted since then.  Reassigning the number of iHighestPlayersTotal is necessary to count only the players that have the new highest score threshold.

Rainyan (approved):

DESTROYGIRL (approved):
Thanks, surprised I didn't catch any of these in testing

PR #2111, opened 2026-09-03
Bots seek enemy after capturing ghost

sunzenshen:
## Description
After a team captures the objective in CTG mode, they can get into a buggy state where they are trying to pursue the objective, but there are no relevant objectives between round end and the next round start.  This behavior end for the related CTG behaviors allows the bots to look for enemies between that time period, which can help bots score some interim points.

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
+++ b/src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
@@ -381,6 +381,11 @@ ActionResult< CNEOBot >	CNEOBotCtgCarrier::Update( CNEOBot *me, float interval )
 		return Done( "No longer carrying the ghost" );
 	}
 	
+	if ( NEORules()->IsRoundOver() )
+	{
+		return Done( "Round Over: no CTG objectives to reach" );
+	}
+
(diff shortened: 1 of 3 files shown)


src/game/server/neo/bot/behavior/neo_bot_ctg_carrier.cpp
```diff
@@ -381,6 +381,11 @@ ActionResult< CNEOBot >	CNEOBotCtgCarrier::Update( CNEOBot *me, float interval )
 		return Done( "No longer carrying the ghost" );
 	}
 	
+	if ( NEORules()->IsRoundOver() )
+	{
+		return Done( "Round Over: no CTG objectives to reach" );
```
sunzenshen, line 386: Note: it is intentional that CNEOBotCtgEnemy does not have this behavior state closer.  My thinking is that normal players would know to run to the goal that the ghost carrier was closest to at the time of the round cap, so it's not that unreasonable for bots to keep pursuing the ghost carrier after the end of the round.

AdamTadeusz (approved):

sunmachine (approved):

PR #2113, opened 2026-09-03
Fix bot thrashing when attack destination has no upcoming cover

sunzenshen:
## Description
Bots could enter a behavior transition trashing state where they attempt to plot an attack path towards a destination but then bounce in and out of the attack behavior transition, because there was no cover between the bot and that location.  This would especially happen if the bot's destination was in view.  This change causes the bots to fall back to chasing the enemy if there is no suitable cover candidate.

## Toolchain
- Windows MSVC VS2022


AdamTadeusz (approved):

sunmachine:
This one's gonna' take me a bit more time, likely tomorrow AM. :sob:

sunmachine (approved):

PR #2117, opened 2026-09-06
Allow pinning ghost to a specific spawn point

sunzenshen:
## Description
Allow specifying a spawn point index to force the ghost to spawn at that location.

## Toolchain
- Windows MSVC VS2022


AdamTadeusz:
with sv_neo_ghost_spawn_bias enabled, there's no way to predict what ghost spawn has what index, maybe a disclaimer in the new convar's description is warranted that this is the case?

AdamTadeusz (approved):

Rainyan (approved):

PR #2118, opened 2026-09-06
Fix uninitialized x64 handle in CUtlHandleTable::CreateHandle

sunzenshen:
## Description

Fully zero initialize x64 handle that was being partially memcpy'd by a 32-bit source.

Aims to fix this assert:

image

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/public/tier1/utlhandletable.h
+++ b/src/public/tier1/utlhandletable.h
@@ -290,12 +290,11 @@ UtlHandle_t CUtlHandleTable<T, HandleBits>::CreateHandle( unsigned int nSerial, 
 	// but that'd widen the object from 4 bytes of data to 4+4 padding, which would need profiling.
 	static_assert(sizeof(h) == 4);
 	static_assert(sizeof(UtlHandle_t) >= sizeof(h));
-	UtlHandle_t ret;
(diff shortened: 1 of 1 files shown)


AdamTadeusz:
Is there a reliable way of replicating this assert?

sunmachine:
Same, I can't figure it out. 😭

sunzenshen:
Oddly enough it only started showing up on startup for me a couple weeks ago, but you'd think the root cause commit would be earlier than that.

Rainyan:
What does the call stack look like when entering that function? And if you can reproduce the assert fail, what are the input values for the function for that failure case? Also would be curious to know what the `ret` and `ref` values are for the assertion failure.

Also, was this in debug build or release build with asserts enabled?

sunzenshen:
> What does the call stack look like when entering that function? And if you can reproduce the assert fail, what are the input values for the function for that failure case? Also would be curious to know what the `ret` and `ref` values are for the assertion failure.
> 
> Also, was this in debug build or release build with asserts enabled?

This was in the debug build, but the irony is that after refreshing my CMake cache, I don't see this assert anymore.  I'll watch out to see if I encounter this in a different branch.

PR #502, opened 2024-07-11
Fix dying + disconnect causing Ghost/weps to noclip fall, fix ghost spawn frozen in air

nullsystem:
## Description

* On dying player ghost drop: Just use a "proper" function to drop weapons
* On disconnected/kicked player ghost drop: Use teleport/workaround so it doesn't fall through
* Ghost spawning: Give it gravity so it doesn't just stays there on given spawn origin

## Toolchain

- Linux GCC Distro Native Arch/GCC 14

## Linked Issues

* fixes #423
* fixes #433

PR #2120, opened 2026-09-08
Fix bogus ghost angles on ghoster disconnect

Rainyan:
## Description
Fix a problem with the ghost being set to uninitialized QAngles if the ghoster disconnects mid-round.

81834151 fixes the bad angles case, but I decided to follow it up with 9c95ac29 to simplify the code path.

Also should be careful not to regress #502 with this; I briefly tested the 3 cases of:

* spawned/respawned ghost not floating in the air
* ghost detaching to world properly when the ghoster:
  * dies
  * disconnects/is kicked

## Steps to reproduce

* For c6ff0807 (HEAD before this PR), set breakpoint at line: https://github.com/NeotokyoRebuild/neo/blob/c6ff080778e1d19dda672b6f551b9df91de41d08/src/game/shared/neo/neo_gamerules.cpp#L4414
* Launch a CTG map with bots
* Wait for any bot to pick up the ghost
* Use `neo_bot_kick all;` to kick the bot currently holding the ghost
* Observe `QAngle angles` local variable value as the breakpoint is hit

### What happens

* The `QAngle angles` local variable's value is uninitialized (or NaN if `VECTOR_PARANOIA`)
* The `Teleport` call trashes ghost angles with the bad `angles` value
* Furthermore, after continuing execution after the breakpoint, `IsEntityQAngleReasonable` may or may not fail (depending on what the `angles` ended up being), at: https://github.com/NeotokyoRebuild/neo/blob/c6ff080778e1d19dda672b6f551b9df91de41d08/src/game/server/baseentity.cpp#L6819-L6827

### What should happen

* Ghost angles are not trashed when the ghoster disconnects

## Toolchain
- Windows MSVC VS2022

### DIFF
--- a/src/game/shared/neo/neo_gamerules.cpp
+++ b/src/game/shared/neo/neo_gamerules.cpp
@@ -4408,13 +4408,8 @@ void CNEORules::ClientDisconnected(edict_t* pClient)
 		auto ghost = GetNeoWepWithBits(pNeoPlayer, NEO_WEP_GHOST);
 		if (ghost)
 		{
-			// NEO JANK (nullsystem): Teleport so that disconnected player don't noclips the ghost?
-			Vector stillVec{0.0f, 0.0f, 0.0f};
-			ghost->Drop(stillVec);
+			ghost->Drop(vec3_origin);
 			pNeoPlayer->Weapon_Detach(ghost);
-			Vector origin = ghost->GetAbsOrigin();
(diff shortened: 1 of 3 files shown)


AdamTadeusz (approved):

sunzenshen (approved):
Kicked the ghost carrier from the game and the ghost seemed to settle fine on the ground. LGTM

PR #2121, opened 2026-09-09
Cheat protect color blind settings

nullsystem:
## Description
It gave competitive advantage

## Toolchain


AdamTadeusz:
Idk I think we should test first whether it actually does, not let panic get the best of us yk.

Agiel (approved):

AdamTadeusz (approved):

PR #2122, opened 2026-09-10
Bots retreat from excessively close enemies

sunzenshen:
## Description
Bots tend to struggle with aiming at extremely close enemies, so having bots retreat if they aren't advancing up to a cover area helps their survivability and targeting.

## Toolchain
- Windows MSVC VS2022


sunzenshen:
To test, enable `buddha` and move closer to the enemy until they back away from you.  `sv_neo_bot_attack_too_close_retreat_ratio` to adjust the ratio applied to the optimal engagement distance constants for weapons.

PR #2123, opened 2026-09-10
Stricter bot superjump path checks

sunzenshen:
## Description
Additional checks for bots to reduce superjump attempts when it would be dangerous, such as fall risks

## Toolchain
- Windows MSVC VS2022


sunzenshen:
Aims to reduce the number of spawn location falls on the map ntre_skyline_ctg, where the rooftop spawn team would often have Recon bots try to superjump across the gap and fall.

PR #2125, opened 2026-09-10
Add scaling for background map pan and lerp

Rainyan:
## Description
Convert `cl_neo_background_pan` from boolean option into a float scale.

Also add an accompanying `cl_neo_background_lerp` cvar for controlling the lerp to be used for `cl_neo_background_pan`, to enable gentler panning movement.

When using the cvar defaults, there are no changes introduced to the panning behaviour:

* The `cl_neo_background_pan 1` default value remains at the same `CAMERA_MOVEMENT_MULTIPIER` panning strength (used as multiplier of `CAMERA_MOVEMENT_MULTIPIER * cl_neo_background_pan`) for client config compatibility.
* The `cl_neo_background_lerp` also defaults to `1` (which "lerps" at `A + (B-A)*cl_neo_background_lerp`, i.e. it is an uninterpolated plain assignment `A=B` as was the case previously).

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #1546

### DIFF
--- a/src/game/client/view.cpp
+++ b/src/game/client/view.cpp
@@ -780,7 +780,7 @@ void CViewRender::SetUpViews()
 	}
 #endif
 #if defined NEO
-	// float compare because background pan can be scalar
+	// float compare because background pan can be scalar, and .GetBool() casts via int which would round down for values 0-1
 	else if (engine->IsLevelMainMenuBackground() && cl_neo_background_pan.GetFloat() != 0)
 	{
 		if (pPlayer)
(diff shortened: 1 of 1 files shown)


Agiel:
Is it possible to allow negative values so I can finally make the background pan in the correct direction? 👀

Rainyan:
> Is it possible to allow negative values so I can finally make the background pan in the correct direction? 👀

Added support for negative values in 860d531 to flip the pan direction. Unless we want X/Y axes control separated, but maybe this is good enough?

PR #2126, opened 2026-09-11
Fix wrong FL_ONGROUND flag on player spawn

### LINKED ISSUE #1969: A recon should not be able to super jump in water
If a map has a spawn point above a large body of water, recons can super jump after swimming up past the surface. Seems to require having been a spectator or unassigned beforehand, as this doesn't work after death

Rainyan:
## Description
Fix a problem with `FL_ONGROUND` flag being set unconditionally server side for CNEO_Player::Spawn when calling into its base class. This would not always hold true, the most notable issue being #1969 . Another weird edge case would be spawning late into a round, and being able to recon-superjump if you do it before landing on the ground for the first time from a mid-air spawnpoint.

Also refactoring some old related code to be shorter and nicer.

## Test map (OGNT compatible)

[dev_swimspawn.zip](https://github.com/user-attachments/files/32080962/dev_swimspawn.zip)

Should be able to superjump (parity):
* On dry land
* On feet-deep water

Should not be able to superjump:
* When fully submerged
* On waist-deep water
* On the surface of the water
* When "walking" on the bottom of the deep water pool

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1969

### DIFF
--- a/src/game/shared/neo/neo_player_shared.cpp
+++ b/src/game/shared/neo/neo_player_shared.cpp
@@ -930,6 +930,8 @@ void CNEO_Player::FixupOnGroundFlag()
 	// Assert that the client side baseclass doesn't do this (and therefore we need do nothing for it)
 	Assert((GetFlags() & FL_ONGROUND) == IsServer());
 #ifdef GAME_DLL
+	if (GetTeamNumber() < FIRST_GAME_TEAM) // if it's not a player, don't bother
+		return;
 	const Vector& start = GetAbsOrigin();
 	const Vector& maxs = GetPlayerMaxs();
 	const Vector& mins = GetPlayerMins();
--- a/src/game/shared/neo/neo_player_shared.cpp
+++ b/src/game/shared/neo/neo_player_shared.cpp
@@ -903,21 +903,9 @@ bool CNEO_Player::IsAllowedToSuperJump()
 	if (IsAirborne())
 		return false;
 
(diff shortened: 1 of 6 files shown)


src/game/shared/neo/neo_player_shared.cpp
```diff
@@ -884,3 +884,64 @@ bool CNEO_Player::TestHitboxes(const Ray_t& ray, unsigned int fContentsMask, tra
...
+	// Assert that the client side baseclass doesn't do this (and therefore we need do nothing for it)
+	Assert((GetFlags() & FL_ONGROUND) == IsServer());
+#ifdef GAME_DLL
+	if (GetTeamNumber() < FIRST_GAME_TEAM) // if it's not a player, don't bother
+		return;
+	const Vector& start = GetAbsOrigin();
+	const Vector& maxs = GetPlayerMaxs();
+	const Vector& mins = GetPlayerMins();
+	float playerHeight = maxs.z - mins.z;
+	Vector end(start.x, start.y, start.z - playerHeight);
```
AdamTadeusz: I'm wondering what the rationale is behind using playerHeight for the trace length, seems a bit much. In CGameMovement::CategorizePosition the offset used is 2.0f hu.

Rainyan: Not sure why I was thinking about the bounding box fit at all, this should be now fixed.

PR #2127, opened 2026-09-11
Bots hold angles where out of view threat could emerge

sunzenshen:
## Description
When a bot knows of a primary threat position (usually through gunfire) that happens to be out of view, this proposal allows bots to model out the fastest area a known threat could appear from and hold that angle.  Threats still expire after 3 seconds by default, but this is an improvement over bots choosing a random look direction when a threat is not in line of sight.

## Toolchain
- Windows MSVC VS2022


src/game/server/neo/bot/behavior/neo_bot_grenade_throw.cpp
```diff
@@ -50,59 +50,7 @@ CNEOBotGrenadeThrow::CNEOBotGrenadeThrow( CNEOBaseCombatWeapon *pWeapon, const C
...
-			}
-
-			if ( area == familiarArea )
-			{
-				return vec3_invalid;
-			}
-		}
-	}
-
-	return vec3_invalid;
```
sunzenshen: Moved to neo_bot_path_compute.cpp

src/game/server/neo/bot/behavior/neo_bot_grenade_throw.cpp
```diff
@@ -50,59 +50,7 @@ CNEOBotGrenadeThrow::CNEOBotGrenadeThrow( CNEOBaseCombatWeapon *pWeapon, const C
...
-	}
-
-	ShortestPathCost cost;
-	const Vector& vecGoal = obscuredPos;
-	if ( NavAreaBuildPath( familiarArea, obscuredArea, &vecGoal, cost ) )
-	{
-		// search backwards from obscured position to find the first point visible to me
-		for ( CNavArea *area = obscuredArea; area; area = area->GetParent() )
-		{
-			// DEBUG: Draw emergence path
```
sunzenshen: Replaced with sv_neo_bot_debug_emergence_area, though sv_neo_bot_grenade_debug_behavior can still be used to see which positions bots end up targeting for throws.

PR #2131, opened 2026-09-12
Replace .start with .forcelive

### LINKED ISSUE #1994: Players cannot force a competitive match to go live/force a live match to stop
The OG competitive plugin has a !forcelive command, which players can use to start a comp match even if there are not enough players on either/both teams. The command can also be used to stop a match by typing it during a live game twice (though maybe from an UX perspective it'd be better to split it into two commands). The feature does not exist in NT;RE.

nullsystem:
## Description
Force live just simply forces live game regardless of readyups or thresholds. .start removed as it's never really used and redundent to this chat command.

## Toolchain

## Linked Issues

- fixes #1994


Rainyan (approved):
It would be good to have the option to also forcibly end the match as described in the issue, but this is definitely nicer in terms of usability to the previous over-threshold exclusive logic.

Issue #1705, opened 2026-02-18
OS mouse cursor graphical glitch


Rainyan:
### Build Info

v25.0-alpha

### Description

*(Submitting this from a Discord user report; I can't personally reproduce this on my machine.)*

When the user loads into a level, and then moves their cursor on top of the team selection UI element, the cursor graphics show a glitch/corruption where the cursor becomes a black box.

### To Reproduce

* Use Windows
* Set "Inverted" style for the "Mouse pointer style" option
* Set size 2 (second slider position from the left) for the pointer's "Size" option

<img width="1029" height="353" alt="Image" src="https://github.com/user-attachments/assets/919069e9-9df9-40ff-a14a-00b03c3a3d1b" />

* Launch the game
* Load into a level
* Hover the cursor over the team selection UI element

### Expected behavior

The cursor is rendered as usual

### Actual behavior

Corruption occurs:

https://github.com/user-attachments/assets/4bd89d36-db49-4d74-84a2-7f69ede44459

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

11

### Machine's GPU

AMD

### GPU's driver

26.1.1

PR #2134, opened 2026-09-13
Default-disable SW cursor for Linux

### LINKED ISSUE #2114: Software cursor is unaligned at non-native screen resolutions on Linux
### Build Info

5210510a29797bbb569e3f341739db51c3d82641

### Description

On Linux, when using the fullscreen video mode, and changing the game resolution to something lower than the OS desktop screen resolution, the software cursor (as default-enabled by #1704) becomes unaligned from the actual mouse position. This is somewhat akin to the cursor offset bug with OGNT, although unsure whether it is related technically.

For the reasoning on defaulting to software cursor to begin with, please see #1704

### To Reproduce

# Prerequisites

* Use Linux
  * The bug is probably not distro-specific, but it is verified occurring with:
    * Fedora Linux 44
    * SteamOS 3.8
* Use Wayland
  * Unsure if this is required but X11 is unverified
* The desktop environment probably doesn't matter, but the bug is verified on COSMIC 1.6 and KDE Plasma 6.7
* Bug verified with screens of native resolutions 1920x1080 and 2560x1440
  * Likely the exact resolution doesn't matter either

# Steps to repro

* Launch game with screen native resolution
* Use the fullscreen video mode*
* Have the software cursor enabled (as is default-enabled by #1704)
* Lower the in-game resolution via the main menu below the native resolution
* Inspect the main menu software cursor accuracy when pointing at menu elements

\* see additional context below

### Expected behavior

* The software cursor should accurately point to the correct elements after the game resolution change

### Actual behavior

* The software cursor becomes incorrectly offset from the real mouse cursor position after the game resolution change

### Operating System

- [ ] Windows
- [x] Linux

### Version/Distro

Fedora and SteamOS (probably distro unrelated)

Personally, on my Fedora COSMIC Spin, I can only repro this bug with the "fullscreen" in-game video mode, whereas "windowed" and "borderless" work correctly for me.

\* But for Sexu, who originally reported this bug to me, apparently the bug happens for all video modes, "windowed" and "borderless" modes included.

A naive workaround could be disabling the software cursor by default for Linux.

Yet another workaround could be reverting #1704 with the consequences of regressing #1705

Although I guess ideally we should figure out what causes this incorrect cursor offset.

Sexu was also reporting poor/laggy software cursor draw performance, but I'm leaving that one out of scope for this issue.

Rainyan:
## Description
Workaround for bug:
* #2114 

This will change `cl_software_cursor` from bool to per-OS int bitflags, with the current default bitflags being Windows only.

The video menu options for the SW cursor are also updated to toggle the per-platform bitflag. This way, users can have their per-platform cursor preference carry over across Linux/Windows when synced via Steam etc, since only the bit for their current platform is changed.

Worth noting that this doesn't fix #2114 but just works around it, to avoid users having a bad time with their cursor not aligning properly.

### Changes introduced

* Default appearance of the SW cursor on Linux changes to disabled, to alleviate #2114 . The Windows OS software cursor default appearance remains unchanged (enabled), so that we don't reintroduce #1705 .

* Linux will have the video options menu field for the software cursor hidden, to avoid users running into #2114
  * Power users can still access the `cl_software_cursor` bitflags directly from the console to override this

## Toolchain
- Windows MSVC VS2022
- Fedora Linux 44 (COSMIC) + GCC 16.2.1 20260819 (Red Hat 16.2.1-2)

## Linked Issues
- related #2114
- related #1705


AdamTadeusz:
software cursor looks so ass I wish we would disable it on windows too

Rainyan:
> software cursor looks so ass I wish we would disable it on windows too

In b7a12bd, I've updated the `cl_software_cursor` bit flags to now include a `EnabledForWindowsInvertedMouseOnly` option, and changed the default SW cursor preference to again be disabled for all platforms. This way, we can target just the users who are on Windows + have the problematic inverse-colored OS cursor for the software cursor hack, without needing to default all Windows users to the SW cursor.

This also bumps the config defaults for major version 35 per 1c22010, so ideally the next release version should bump the major version from 34 -> 35 accordingly.

PR #2136, opened 2026-09-14
Optimize bot grenade emergence point targeting

sunzenshen:
## Description
Replaces expensive route modeling for emergence route with simpler distance-based PVS search based on previously existing SDK example CFindClosestPotentiallyVisibleAreaToPos.  Also applied grenade throw cooldown timers uniformly to avoid recalculating grenade throw scenario immediately after a dispatcher rejection.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #1651


src/game/server/neo/bot/neo_bot.cpp
```diff
@@ -2035,6 +2063,24 @@ bool CNEOBot::IsLineOfFireClear(const Vector& where, const LineOfFireFlags flags
 }
 
 
+//-----------------------------------------------------------------------------------------------------
+// Return true if a thrown object (like a grenade) has a completely unobstructed physical line to the target.
+// Unlike IsLineOfFireClear, this requires !trace.DidHit() and will NOT treat breakable entities (like glass) as clear.
+bool CNEOBot::IsThrowLineClear(const Vector& from, const Vector& to) const
+{
+	trace_t trace;
+	NextBotTraceFilterIgnoreActors filter(NULL, COLLISION_GROUP_NONE);
```
Rainyan, line 2072: Question: would it be more appropriate to use `COLLISION_GROUP_PROJECTILE`, since that seems to be what the NT frag/smoke/det as setting their collisiongroup as? Or perhaps I'm misunderstanding how this trace filter works. Or perhaps the TraceLine mask already takes care of that(?)

I could be completely off so this is more of a question than a change suggestion per se.

Rainyan (approved):

PR #2137, opened 2026-09-14
Match snapshots, restore ghost index + spawn hdl, output to chat message

nullsystem:
# **NOTE: REQUIRES `sv_neo_comp 1` to test**

## Description
* Added `sv_neo_restore_round_snapshot` to restore a round to a snapshot. This will also immediately reset the round to the snapshot round.
* Add ability to save and restore ghost index chosen for each round
* Add ability to save and restore spawn handle chosen for each round but this feature is limited to snapshots, won't be done for session restore as it won't be do-able for crashed server restore
* Output logs to chat message
* `sv_neo_restore_xp` support steamid3

## Toolchain

## Linked Issues

- fixes #
- related #


src/game/server/neo/neo_gamerules_restore.cpp
```diff
@@ -271,44 +460,71 @@ void MatchSessionBackup()
...
+	// NEO NOTE (nullsystem): Just fall back to unused index 0 if rounds somewhat
+	// manages to get to 64+ or some invalid index for some reason.
+	// From then on it won't crash the snapshotter as it just starts writing to
+	// snapshot index-0, but round numbers are indexed-1 so they won't get used anyway.
+	const int iRoundNumber = NEORules()->roundNumber();
+	MatchSnapshot *pSnapshot =
+			(iRoundNumber < 0 || iRoundNumber >= SNAPSHOTS_TOTAL) ?
+					&gSnapshots[0] : &gSnapshots[iRoundNumber];
+	giSnapshotsMin = Min(giSnapshotsMin, iRoundNumber);
+	giSnapshotsMax = Max(giSnapshotsMax, iRoundNumber);
```
sunmachine: Clamp min/max to the slot actually written, or skip the snapshot entirely when out of range.

With the round number pushed to 64, `sv_neo_restore_round_snapshot 64` was accepted by the range check and the game segfaulted on the read one slot past the array. 😭 

Shouldn't ever happen obviously, optional and nit-picky.

```suggestion
	const bool bSnapshotInRange = (iRoundNumber >= 0 && iRoundNumber < SNAPSHOTS_TOTAL);
	MatchSnapshot *pSnapshot = bSnapshotInRange ? &gSnapshots[iRoundNumber] : &gSnapshots[0];
	if (bSnapshotInRange)
	{
		giSnapshotsMin = Min(giSnapshotsMin, iRoundNumber);
		giSnapshotsMax = Max(giSnapshotsMax, iRoundNumber);
	}
```

src/game/shared/neo/neo_gamerules.cpp
```diff
@@ -2924,6 +2936,11 @@ void CNEORules::StartNextRound()
 			continue;
 		}
 
+		if (pPlayer->m_iNextRestore.flags & NEXT_ROUND_PLAYER_RESTORE_FLAG_WEAPON)
+		{
+			pPlayer->m_iLoadoutWepChoice = pPlayer->m_iNextRestore.iLoadoutWepChoice;
+		}
```
sunmachine, line 2955: The loadout choice is applied before Spawn() (`neo_gamerules.cpp` line 2950), but XP is restored after it (line 2965). Spawn calls `RequestSetLoadout`, which downgrades to loadout 0 if current XP is insufficient. Move the XP and death restore above Spawn().

```suggestion
		// Restore XP and deaths before Spawn() so RequestSetLoadout validates
		// the restored weapon choice against the restored XP
		if (pPlayer->m_iNextRestore.flags & NEXT_ROUND_PLAYER_RESTORE_FLAG_XP)
		{
			pPlayer->m_iXP.Set(pPlayer->m_iNextRestore.iXP);
		}
		if (pPlayer->m_iNextRestore.flags & NEXT_ROUND_PLAYER_RESTORE_FLAG_DEATH)
		{
			pPlayer->ResetDeathCount();
			pPlayer->IncrementDeathCount(pPlayer->m_iNextRestore.iDeaths);
		}
		if (pPlayer->m_iNextRestore.flags & NEXT_ROUND_PLAYER_RESTORE_FLAG_WEAPON)
		{
			pPlayer->m_iLoadoutWepChoice = pPlayer->m_iNextRestore.iLoadoutWepChoice;
		}
```

Important: The post-Spawn XP/death block at lines 2963 to 2971 becomes redundant (but I think harmless). It's closing lines are outside the diff hunk, so GitHub won't accept a suggestion that deletes it, so I'm asking you here, with an imaginary Bernie Sanders meme, to delete those lines.

src/game/server/neo/neo_gamerules_restore.cpp
```diff
@@ -176,6 +362,7 @@ CON_COMMAND(sv_neo_restore_session, "Restore the previous session")
 	{
 		RestoreSetRoundNumber(kvRounds->GetInt("number"), __func__);
 		RestoreSetRoundsWon(kvRounds->GetInt("jinrai"), kvRounds->GetInt("nsf"), __func__);
+		RestoreSetGhostSpawnIdx(kvRounds->GetInt("ghost"), __func__);
```
sunmachine: No code change, just comment. Feel free to modify. I wouldn't be shocked if I misunderstood something here.

```suggestion
		// NEO NOTE: ghost spawn index is not restored across sessions. 
		// m_ghostSpawns is only shuffled on round 1, so after a server restart
		// into round N > 1 the saved index points at a different spawn than the 
		// one originally used. Snapshot restore only.
		RestoreSetGhostSpawnIdx(kvRounds->GetInt("ghost"), __func__);
```

src/game/server/neo/neo_gamerules_restore.cpp
```diff
@@ -128,28 +217,125 @@ static void RestoreSetXPDeath(CNEO_Player *pNeoPlayer, const int iXP, const int
...
+}
+
+CON_COMMAND(sv_neo_restore_round_snapshot, "Restore the current match's recorded round snapshot")
+{
+	if (2 != args.ArgC())
+	{
+		ErrorToWarningAndTalk("Usage: %s <round number>", __func__);
+		return;
+	}
+
```
sunmachine, line 271: > [!NOTE]
> I DO NOT LIKE THIS SUGGESTED FIX. IT'S NARROW AND FAST TO IMPLEMENT, AND WORKS.
> 
> The approach that's better that I didn't have time to write out is to arm the next-round flags unconditionally when called from this command, so they override the warmup wipe and the round increment, and treat Pause explicitly. You either apply the restore when the pause ends or lift the pause and reset. Keep a refusal only for Idle with too few players, where there is genuinely nothing to reset. Move `ClearSnapshots()` out of `bFromStarting` and leave it to map change, `RestartGame`, and a full match reset. 

> Every `RestoreSet*` helper only arms the next-round flags when in ready-up or `IsRoundOn` (freeze, live, post-round). Run during Pause, Warmup, Countdown, or plain Idle, the spawn restore errors out (line 230), the ghost flag is dropped (line 145), and the round number is set immediately then incremented by `StartNextRound`, landing on N+1. From Warmup the scores are zeroed and snapshots wiped as well. The command should reject unless `IsRoundOn()`.

I did an A/B to check, restoring what I'll call "snapshot 1" during `Warmup` landed on round 2 with the snapshot store wiped (range reported "2 to 2"). The control run I did during a live round correctly landed on round 1.

```suggestion
	}

	if (false == NEORules()->InReadyUpState() && false == NEORules()->IsRoundOn())
	{
		ErrorToWarningAndTalk("%s: error: Cannot restore snapshot while paused, in warmup, or in countdown", __func__);
		return;
	}
```

sunmachine (commented):
Give this a read; not marking as "request changes" because not all of the changes need to be made / are even desirable.

src/game/server/neo/neo_gamerules_restore.cpp
nullsystem: Warmup/Countdown blocked since the snapshots will be cleared on new match anyway. Pause changed so it doesn't immediately go to snapshot round and instead waits for pause to do so.

src/game/server/neo/neo_gamerules_restore.cpp
nullsystem: I've changed both ghost and player spawn matching to use entindex instead. It seems in quick initial testing it may even be suitable (as in the entindex are stable even cross server sessions) for session restore also so applied for both snapshots and session restore.

nullsystem:
@sunmachine Updated for code review changes: https://github.com/NeotokyoRebuild/neo/pull/2137/changes/576a0437a2072a051dca8376ab5e168d30baa54a

sunmachine (approved):

PR #2139, opened 2026-09-15
Fix players being able to warp through walls when touching some physics objects

DESTROYGIRL:
## Description
This fix applies a check at the tail end of what could be a larger chain of issues with override physics props (my theory, anyway). Basically just looks at the velocities vs the proposed new position to see if the move was really valid

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #2013


DESTROYGIRL:
Test map here: 
[pen_test3.zip](https://github.com/user-attachments/files/32263925/pen_test3.zip)

Best bet at repro is rubbing up against and running into the 3 chairs sitting next to each other in a circle until it happens. `developer 1` must be enabled to see the message in the fix

Rainyan (approved):

sunmachine (approved):
Clean!

PR #2140, opened 2026-09-17
Fix incorrect chat colors

### LINKED ISSUE #1400: Text in chat messages are sometimes the wrong colour
### Build Info

20250921_e55c822

### Description

Sometimes a player's status/name in the chat message will be printed in the orange spectator colour despite being assigned to a team.

Image

### To Reproduce

1. Join a team.
2. Type a message, ideally when dead.

### Expected behavior

Status/player should correspond to the colour of the team the player has been assigned to.

### Actual behavior

_No response_

### Operating System

- [x] Windows
- [ ] Linux

### Version/Distro

10

### Machine's CPU

_No response_

### Machine's GPU

_No response_

### GPU's driver

_No response_

### Build's compiler

_No response_

_No response_

Rainyan:
## Description
Use `C_PlayerResource` to find the latest player metadata when preparing to print chat messages.

Trying to fetch the client's knowledge about a a player may fail if that player is out of their PVS, because that would mean the data could be stale or perhaps never even networked to that player in the first place, until they first meet each other.

Also refactoring the `COLOR_CUSTOM` path away since we don't currently use it, and skipping profanity checks for the `"Console"` hardcoded server name since it is redundant.

Example case of the bug (build 279847474c56d3c3c71deabcdf2cc16209bfe798):

[ads_20260913-2216_ntre_hebi_ctg_b2_zwia_168097191.zip](https://github.com/user-attachments/files/32347758/ads_20260913-2216_ntre_hebi_ctg_b2_zwia_168097191.zip)

marisa's glhf message at this demo's beginning should be printed with Jinrai colors, and not in spectator orange.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- fixes #1400


sunmachine (approved):

src/game/client/hud_basechat.cpp
```diff
@@ -1593,38 +1593,20 @@ void CBaseHudChatLine::InsertAndColorizeText( wchar_t *buf, int clientIndex )
...
 				range.color = pChat->GetTextColorForClient(COLOR_CUSTOM, clientIndex);
 				m_textRanges.AddToTail(range);
 			}
 			else
 			{
-				Assert(false);
 				range.color = pChat->GetTextColorForClient(COLOR_NORMAL, clientIndex);
 				m_textRanges.AddToTail(range);
 			}
 		}
```
sunmachine: `Colorize` only inserts text covered by a range, and this was the only range covering `[0, m_iNameStart)`. With it compiled out, the `[Jinrai] ` / `[NSF] ` / `*DEAD*` prefix from the team and dead chat formats is no longer rendered at all. All-chat while alive has no prefix, so it looks fine there.

The stale color in the demo came from reading the team off the player entity. `COLOR_PLAYERNAME` goes through `GetClientColor`, which already reads it from `g_PR`, so the prefix can keep its range and just use that:

```suggestion
		range.color = pChat->GetTextColorForClient(COLOR_PLAYERNAME, clientIndex);
		m_textRanges.AddToTail(range);
```

With that, `COLOR_CUSTOM` / `SetCustomColor` really are unused and can go.

sunmachine, line 1609: Oops, almost missed this. Team name text e.g. `[Jinrai]` ain't drawing.

Image
Image

`Colorize` only draws text covered by a range, and this was the only one covering `[0, m_iNameStart)`, so the `[Jinrai]` / `*DEAD*` prefix no longer renders in team and dead chat. `COLOR_PLAYERNAME` already reads the team from `g_PR`, so the prefix can keep it's range and use that:

```suggestion
		range.color = pChat->GetTextColorForClient(COLOR_PLAYERNAME, clientIndex);
		m_textRanges.AddToTail(range);
```

With that, `COLOR_CUSTOM` / `SetCustomColor` really are unused and can go!

sunmachine (changes requested):
Found an oops 😭

PR #2141, opened 2026-09-17
Bots don't avoid smoke grenades with a known thrower

sunzenshen:
## Description
Adds smoke identification code to the bot grenade avoidance behavior.
Before bots retreated from smoke grenades while they were in flight, even in circumstances where the nature of the grenade should be obvious.  
Now checks whether the smoke grenade was thrown by a teammate (implied that the teammate communicated the intent), or whether the thrower was identified by the bot recently.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #1545


sunmachine (approved):
Straight-forward.

Rainyan (approved):

PR #2142, opened 2026-09-17
Stricter bot superjump path checks

sunzenshen:
## Description
Additional checks for bots to reduce superjump attempts when it would be dangerous, such as fall risks

## Toolchain
- Windows MSVC VS2022


src/game/server/neo/bot/behavior/neo_bot_behavior.cpp
```diff
@@ -457,6 +477,29 @@ void CNEOBotMainAction::ReconConsiderSuperJump( CNEOBot *me )
...
+		trace_t tr;
+		UTIL_TraceLine(vecProbeStart, vecProbeEnd, MASK_SOLID_BRUSHONLY, me, COLLISION_GROUP_NONE, &tr);
+
+		if (!tr.DidHit() || (vecFeet.z - tr.endpos.z) > flMaxDrop)
+		{
+			// Potential hole in ground detected
+			return;
+		}
+	}
+
```
sunmachine, line 502: Could this trace exactly `flMaxDrop` and test `!tr.DidHit()` only? Since the probes start at feet height (launch dir has z zeroed), the measured drop is the same as the trace distance, so I think the buffer and the z delta comparison never change the result: a hit between 250 and 314 units down is rejected by `> flMaxDrop`, and a 250 unit trace would reject it by missing.

If so, the comment could read: "Abort if the floor at any sample point ahead is missing or more than flMaxDrop below the bot's feet (a pit, a gap, or a long fall)."

PR #1724, opened 2026-02-23
Bots investigate gunfire sounds while searching

sunzenshen:
## Description
Bots investigate gunfire sounds while searching if those sounds are closer than the bot's current movement goal.

## Toolchain
- Windows MSVC VS2022

PR #2143, opened 2026-09-17
Improve bot tracking of recent combat sounds

sunzenshen:
## Description
Bot trackable combat sounds on average were only existing for about 200 milliseconds before disappearing, so most combat sounds were disappearing by the time bots were considering nearest combat sounds every 250ms.  The proposed fix is to add combat sounds to a temporary buffer object every tick, and let the bots query this buffer object at their leisure.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #1724


src/game/server/neo/bot/behavior/neo_bot_seek_and_destroy.cpp
```diff
@@ -20,106 +20,114 @@ extern ConVar neo_bot_defense_must_defend_time;
...
+		m_soundSearchTimer.Start( 0.25f );
 
-		if (distSqr >= flClosestDistSqr)
+		Vector vFight;
+		if ( !NEOMemorySoundCombat::FindNearestFight( me->GetAbsOrigin(), me->GetTeamNumber(), me->entindex(), vFight )
+			|| !BotInSoundPAS( me, vFight ) )
 		{
-			continue;
+			return false;
 		}
```
sunmachine, line 84: After arriving at a fight, the bot re-targets the same spot on the next search. 

`OnMoveToSuccess` restores the objective goal, then 0.25s later the commit is gone (bot is inside arrive range), the fight is still in memory, and the only "already there" test compares `vGoalBefore` to `vFight`, not the bot's position. Standing at the fight makes `flVia` equal `flDirect`, so the detour ratio passes and the bot walks back to where it is. 

_This repeats until the memory expires, with a pathfind each cycle._

```suggestion
		if ( !NEOMemorySoundCombat::FindNearestFight( me->GetAbsOrigin(), me->GetTeamNumber(), me->entindex(), vFight )
			|| !BotInSoundPAS( me, vFight )
			|| me->GetAbsOrigin().DistToSqr( vFight ) <= flArriveSqr )
		{
			return false; // no fight, out of PAS, or already there
		}
```

sunmachine (changes requested):
Fizzixit

sunmachine (approved):

PR #2144, opened 2026-09-17
Optimize bot grenade emergence point targeting

sunzenshen:
## Description
Replaces expensive route modeling for emergence route with simpler distance-based PVS search based on previously existing SDK example CFindClosestPotentiallyVisibleAreaToPos.  Also applied grenade throw cooldown timers uniformly to avoid recalculating grenade throw scenario immediately after a dispatcher rejection.

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #1651


src/game/server/neo/bot/behavior/neo_bot_grenade_dispatch.cpp
```diff
@@ -124,6 +124,12 @@ Action< CNEOBot > *CNEOBotGrenadeDispatch::ChooseGrenadeThrowBehavior( const CNE
 			}
 		}
 
+		// Don't commit to a throw when there's nowhere to throw from yet
+		if ( me->FindVisibleThrowPointNear( threat->GetLastKnownPosition() ) == vec3_invalid )
+		{
+			return nullptr;
+		}
```
sunmachine: Does this gate make the vantage-area path in `CNEOBotGrenadeThrowFrag` unreachable? 

Could this check be dropped entirely? `CNEOBotGrenadeThrow::OnStart` already returns Done when `FindVantageArea` finds nothing, and that search covers the bot's own area first, so it also handles the "throwable from here" case. With the cooldown now starting before dispatch, a Done from OnStart no longer retries immediately, which I think was the reason for this gate.

If the frag action is only created when `FindVisibleThrowPointNear` already finds a clear point from where the bot stands, the first targeting pass should find the same point, pass the looser line-of-fire check, and return READY. The WAIT branch (`FindVantageArea`, pathing there, the blocked-vantage fallback) would then only run in edge cases, where before a bot without a direct throw line would reposition and throw.

If dropping the repositioning is intended, might be worth noting in the description and removing the WAIT path. If not, maybe the gate could be something cheaper than "can throw from here right now" (like, removed?)?

sunmachine (approved):

PR #2145, opened 2026-09-17
Bots follow commander into smoke and smoke hazard optimization

sunzenshen:
## Description
Stops bots from retreating from smoke if they are following a commander, and smoke hazard propagation optimization

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #2020


sunmachine (approved):
LGTM

PR #2146, opened 2026-09-17
Bot path reservation dispersal improvements

sunzenshen:
## Description
Primarily addresses a myriad of issues with the team-specific lookup and penalization of congested routes:
- fixed bots using the same serial prefix as a reservation index (the reservation system was treating all teammates as the same individual)
- clear reservations on death
- penalize excessively long detours

## Toolchain
- Windows MSVC VS2022

## Linked Issues
- related #1535


src/game/server/neo/bot/neo_bot_path_cost.cpp
```diff
@@ -146,11 +148,29 @@ float CNEOBotPathCost::operator()(CNavArea* baseArea, CNavArea* fromArea, const
 
 	// ------------------------------------------------------------------------------------------------
 	// New path reservation related cost adjustments
-	if ( !m_bIgnoreReservations && (m_routeType != FASTEST_ROUTE) )
+	if ( NEORules()->IsTeamplay() && !m_bIgnoreReservations && (m_routeType != FASTEST_ROUTE) )
```
sunmachine: The new `IsTeamplay()` condition wraps the whole reservation block, including `GetAreaAvoidPenalty`, which has nothing to do with teams. 

So in a DM, a bot that gets stuck still records the penalty. The increment has no teamplay check. But, the path cost now ignores it, and the next path goes through the same area. 

On master, the avoid penalty applied in every mode. The gate is also redundant for the friendly and enemy counts, which already return 0 outside teamplay.

```C++
if ( NEORules()->IsTeamplay() && !m_bIgnoreReservations && (m_routeType != FASTEST_ROUTE) )
{
	// ... truncated!
	cost += CNEOBotPathReservations()->GetAreaAvoidPenalty(area->GetID());
}
```

Fix: no need for `IsTeamplay()` I think. Behavior is unchanged.

```suggestion
	if ( !m_bIgnoreReservations && (m_routeType != FASTEST_ROUTE) )
```

sunmachine (changes requested):

src/game/server/neo/bot/neo_bot_path_cost.cpp
sunzenshen: > The new `IsTeamplay()` condition wraps the whole reservation block, including `GetAreaAvoidPenalty`, which has nothing to do with teams.
> 
> So in a DM, a bot that gets stuck still records the penalty. The increment has no teamplay check. But, the path cost now ignores it, and the next path goes through the same area.
> 
> On master, the avoid penalty applied in every mode. The gate is also redundant for the friendly and enemy counts, which already return 0 outside teamplay.
> 
> ```c++
> if ( NEORules()->IsTeamplay() && !m_bIgnoreReservations && (m_routeType != FASTEST_ROUTE) )
> {
> 	// ... truncated!
> 	cost += CNEOBotPathReservations()->GetAreaAvoidPenalty(area->GetID());
> }
> ```
> 
> Fix: no need for `IsTeamplay()` I think. Behavior is unchanged.

Agree with the general Area Avoid Penalty, but we should only have bots avoid other teammate paths if there's actually a team game mode.  There's a quirk with free for all modes with the game where you still pick a team, but mostly for aesthetics.

sunmachine: Oh I'm getting sleepy. We don't have any real free-for-all maps right? It's all TDM. 🤪

sunmachine (approved):
I thought I found a bug but I didn't find a bug I found DM.

src/game/server/neo/bot/neo_bot_path_cost.cpp
sunzenshen: There's ntre_apparatus_dm, but that map doesn't have navmeshes yet.

sunmachine: Ah look at that!

sunmachine (approved):

